PR #737, opened 2018-02-14
defaults: vm UUID from vm name

fromanirh:
Let's initialize the vm UUID from the VM name, using UUID V5 (see
RFC4122). This helps achieving stable UUIDs across reboots.


fromanirh:
@rmohr ok, then we can close this PR, will make a new one with the offline controller changes.

PR #751, opened 2018-02-21
k8s-1.9.3 provider, fully dockerized

rmohr:
Add a provide which runs completely in docker containers and is completely ephemeral. It only cashes docker images in a local registry.

See the included README.md for more details.

Advantages:
 * A fresh cluster comes almost as fast up as an already fully provisioned cluster
 * Reproducible rpm-versions and reproducible k8s version, since everhtying is already prepulled
 * We can destroy the cluster before/after every test-run
 * We can share the local docker registry between different cluster versions (k8s/openshift)
 * The VMs inside the containers are built on vagrant-boxes.

Note: When starting the provider the first time, it will download about 3 gigs. That is the same size, like a fully provisioned vagrant-master takes after it has downloaded and installed everything.


rmohr:
> As a follow up, figure out a way to get the magic that's going on in your rmohr/cli and rmohr/k8s-1.9.3 containers upstream in a kubevirt project repo. Preferably before this provider is used by Jenkins.

Agreed. For reference, here is the code for creating the images right now: https://github.com/rmohr/qemu-dockerized

cynepco3hahue:
@rmohr Can we change default CI provider to this one?

cynepco3hahue:
@rmohr Or not, `virt-launcer` container failed to run with 
```
./virt-launcher: /lib64/libvirt.so.0: version `LIBVIRT_3.7.0' not found (required by ./virt-launcher)
./virt-launcher: /lib64/libvirt.so.0: version `LIBVIRT_3.1.0' not found (required by ./virt-launcher)
./virt-launcher: /lib64/libvirt.so.0: version `LIBVIRT_3.0.0' not found (required by ./virt-launcher)
./virt-launcher: /lib64/libvirt.so.0: version `LIBVIRT_3.4.0' not found (required by ./virt-launcher)
```

rmohr:
@cynepco3hahue this provider builds the docker images on your machine. Could you check the installed version of libvirt packages inside your `kubevirt/libvirt-3.7.0` container?

Here are mine:

```bash
$ docker run --rm -it kubevirt/libvirt:3.7.0 dnf list installed | grep libvirt
libvirt-client.x86_64                      3.7.0-3.fc27                @updates 
libvirt-daemon.x86_64                      3.7.0-3.fc27                @updates 
[...]
libvirt-libs.x86_64                        3.7.0-3.fc27                @updates
```
If it looks different in your case, could you try `docker pull kubevirt/libvirt-3.7.0` to see if it pulls in an image?

cynepco3hahue:
@rmohr the only images that I can see
docker images | grep libvirt
kubevirt/libvirt-kubevirt                                 latest              499138ecb5e9        6 months ago        1.26 GB
docker.io/kubevirt/libvirt                                latest              f1c6aec1619e        10 months ago       1.12 GB

So maybe building failed and I missed it, will recheck it.

PR #755, opened 2018-02-22
Use correct libvirt image for virt-launcher

rmohr:
@cynepco3hahue that should fix the issue you experienced with #751. It seems like with the switch to the decentralized libvirt pods we lost the tag in the Dockerfile. 


cynepco3hahue:
Fix helped to me, functional tests now running fine

fabiand:
Merging according to the previous comment.

PR #756, opened 2018-02-22
Enable initializerregistration for our providers

rmohr:
* k8s-1.9.3 provider uses now an image with the feature enabled
 * after a vagrant recreate the vagrant provider has the feature enabled

This is a prerequisite for #652.

### DIFF
--- a/cluster/k8s-1.9.3/provider.sh
+++ b/cluster/k8s-1.9.3/provider.sh
@@ -10,7 +10,7 @@ function up() {
     VAGRANT_NUM_NODES=${VAGRANT_NUM_NODES-0}
     # Add one, 0 here means no node at all, but in the kubevirt repo it means master-only
     VAGRANT_NUM_NODES=$((VAGRANT_NUM_NODES + 1))
-    docker run --privileged --rm -v /var/run/docker.sock:/var/run/docker.sock rmohr/cli:latest run --nodes ${VAGRANT_NUM_NODES} --tls-port 127.0.0.1:8443 --ssh-port 127.0.0.1:2201 --background --registry-port 127.0.0.1:5000 --prefix kubevirt --registry-volume kubevirt_registry --base rmohr/kubeadm-1.9.3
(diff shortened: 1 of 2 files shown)


cluster/k8s-1.9.3/provider.sh
```diff
@@ -10,7 +10,7 @@ function up() {
     VAGRANT_NUM_NODES=${VAGRANT_NUM_NODES-0}
     # Add one, 0 here means no node at all, but in the kubevirt repo it means master-only
     VAGRANT_NUM_NODES=$((VAGRANT_NUM_NODES + 1))
-    docker run --privileged --rm -v /var/run/docker.sock:/var/run/docker.sock rmohr/cli:latest run --nodes ${VAGRANT_NUM_NODES} --tls-port 127.0.0.1:8443 --ssh-port 127.0.0.1:2201 --background --registry-port 127.0.0.1:5000 --prefix kubevirt --registry-volume kubevirt_registry --base rmohr/kubeadm-1.9.3
+    docker run --privileged --rm -v /var/run/docker.sock:/var/run/docker.sock rmohr/cli:latest run --nodes ${VAGRANT_NUM_NODES} --tls-port 127.0.0.1:8443 --ssh-port 127.0.0.1:2201 --background --registry-port 127.0.0.1:5000 --prefix kubevirt --registry-volume kubevirt_registry --base "rmohr/kubeadm-1.9.3@sha256:5f9394037a62ee0d5fb8e6005cf248f0b7601e1d0a67c0af79ff8d2f8e9541bb"
```
cynepco3hahue: Why do we need image hash here?

rmohr: Since it is the only way to make sure that docker really pulls the right image. If we would just update the "latest" tag, people would not have the right image, if a `rmohr/kubeadm-1.9.3:latest` already exists on their machine. This way we can override all confusions.

rmohr: With the upcoming switch to bazel, I intend to only use hashes everywhere, to enforce the reproducibility.

PR #759, opened 2018-02-22
Make sure that kubeadm picks up pod CIDR for flannel

rmohr:
If the pod network CIDR is not picket up, flannel can't start successfully. @cynepco3hahue merge #756 too fast.


rmohr:
Confirmed that flannel comes up properly on both providers.

PR #768, opened 2018-02-25
Add VM templates for windows and linux

yanirq:
Added VM templates for RHEL 7.4 and Windows Server 2012 R2

Signed-off-by: Yanir Quinn yquinn@redhat.com

### DIFF
--- a/cluster/vm-template-rhel7.yaml
+++ b/cluster/vm-template-rhel7.yaml
@@ -1,0 +1,56 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: rhel7-vm-template
+  annotations:
+    description: "OCP KubeVirt Red Hat Enterprise Linux 7.4 VM template"
+    tags: "kubevirt,ocp,template,linux"
+  labels:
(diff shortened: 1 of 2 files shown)


yanirq:
> It's not clear to me that these belong in the manifest's directory. They're more than "examples" in the sense that they're presumably functional, but they're certainly not critical to the core operation of KubeVirt. Perhaps these belong somewhere under the cluster directory?

I placed them currently both under dev and release, if we are not ready to place them under release yet then they should be at least under manifests/dev

fabiand:
To me they are also closer to a VM definition (like the manifests in `cluster/`) than definitions for our controllers.

To me `cluster/` looks like a good place for now.

rmohr:
@fabiand ok then lets treat them right now like examples.

@yanirq could you move them back to `cluster/`?

yanirq:
> @yanirq could you move them back to cluster/?
@rmohr  done

fabiand:
Added a tracker to add functional tests with #772

rmohr:
> Added a tracker to add functional tests with #772

thx

PR #774, opened 2018-02-27
OpenShift deployment fixes

cynepco3hahue:
- user privileged policy for ISCSI container
- use alpha version instead of latest one

### DIFF
--- a/cluster/vagrant-openshift/setup_master.sh
+++ b/cluster/vagrant-openshift/setup_master.sh
@@ -56,7 +56,7 @@ openshift_clock_enabled=true
 openshift_master_identity_providers=[{'name': 'allow_all_auth', 'login': 'true', 'challenge': 'true', 'kind': 'AllowAllPasswordIdentityProvider'}]
 openshift_disable_check=memory_availability,disk_availability,docker_storage,package_availability,docker_image_availability
 openshift_repos_enable_testing=True
-openshift_image_tag=latest
+openshift_image_tag=v3.9.0-alpha.4
 containerized=true
 enable_excluders=false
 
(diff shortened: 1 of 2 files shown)


cluster/vagrant-openshift/setup_master.sh
```diff
@@ -56,7 +56,7 @@ openshift_clock_enabled=true
 openshift_master_identity_providers=[{'name': 'allow_all_auth', 'login': 'true', 'challenge': 'true', 'kind': 'AllowAllPasswordIdentityProvider'}]
 openshift_disable_check=memory_availability,disk_availability,docker_storage,package_availability,docker_image_availability
 openshift_repos_enable_testing=True
-openshift_image_tag=latest
+openshift_image_tag=v3.9.0-alpha.4
```
rmohr: Is that already enough to make it work again? In my case I also had to add the nfs storage role to a machine. Oterhwise it failed on the service broker deployment.

cynepco3hahue: I deployed and it passed just fine, so look so.

cynepco3hahue:
And I have only 3 failures on it
```
Summarizing 3 Failures:

[Timeout...] Storage Starting a VM with Alpine PVC [It] should be successfully started 
/root/go/src/kubevirt.io/kubevirt/tests/storage_test.go:115

[Timeout...] Storage Starting a VM with Alpine PVC [It] should be successfully started and stopped multiple times 
/root/go/src/kubevirt.io/kubevirt/tests/storage_test.go:136

[Fail] Networking VirtualMachine with nodeNetwork definition given [It] should be able to reach the internet 
/root/go/src/kubevirt.io/kubevirt/tests/vm_networking_test.go:139

Ran 48 of 50 Specs in 1775.807 seconds
FAIL! -- 45 Passed | 3 Failed | 0 Pending | 2 Skipped --- FAIL: TestTests (1775.81s)
```

PR #778, opened 2018-02-28
Pin k8s generator versions in the docker builder

### LINKED ISSUE #777: Use a predictable version for the k8s code generators
Recent travis builds are pulling in newer code generators than we have locally. We need to ping the generators to a specific version.

rmohr:
The generators need to be pinned to avoid different output between
developer machines and CI.

Fixes #777 


rmohr:
Travis passes and functional tests could still compile.

PR #790, opened 2018-03-06
Pin kubevirtci client to a specific version

rmohr:
In order to allow updating clients and images without interrupting
anyone, use shas for pulling the client and the images.


rmohr:
There will come some updates for `rmohr/cli` and the vm images. Therefore lets make the version expclicit to not run into strange side-effects.

PR #793, opened 2018-03-07
client-python: update deploy.sh script

lukas-bednar:
client-python should be deployed only in case when API was changed.

Fixes kubevirt/client-python#7

Signed-off-by: Lukas Bednar <lbednar@redhat.com.com>

### DIFF
--- a/hack/gen-client-python/deploy.sh
+++ b/hack/gen-client-python/deploy.sh
@@ -24,7 +24,12 @@ git config --global user.email "travis@travis-ci.org"
 git config --global user.name "Travis CI"
 
 # Push only in case something got changed in code.
-if git status --porcelain | grep 'kubevirt/' | grep --quiet "^ [AM]"; then
+# Ignore api_client.py and configuration.py because it contains version,
+# which is getting updated regardless of changes in API.
+if git status --porcelain |
+    grep 'kubevirt/' |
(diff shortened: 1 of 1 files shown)


hack/gen-client-python/deploy.sh
```diff
@@ -24,7 +24,13 @@ git config --global user.email "travis@travis-ci.org"
 git config --global user.name "Travis CI"
 
 # Push only in case something got changed in code.
-if git status --porcelain | grep 'kubevirt/' | grep --quiet "^ [AM]"; then
+# Ignore api_client.py and configuration.py because it contains version,
```
lukas-bednar: it solves my problem, but I think it could be done better.

I was thinking about computing hash from openapi.json and use it as a version for sdk. 

I am open to other suggestions as well.

stu-gott: Ignoring the files entirely isn't ideal, but I agree that I can't think of a trivial way to account for changes minus the changing User Agent/SDKPackage Version.

rmohr: Something like:

```bash
if [ "$(git diff --numstat -- api_client.py | cut -f 1)" != "1" ] && [  -n "$(git diff --numstat -- api_client.py | cut -f 1)"  ] ; then
  echo "we need an update"
fi
```

should help you. First ignore the two files on the overall status check and then check if more than one line changed.

PR #799, opened 2018-03-08
Fix offlinevm selector

petrkotas:
This PR removes the not necessary label selector from example offline virtual machine. It is confusing for the users.

@fabiand @rmohr PRM

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/799)


petrkotas:
@rmohr For sure. Already did.

Issue #789, opened 2018-03-04
Windows 2016 VM failed to get IP from DHCP server


cynepco3hahue:
### Description
VM with Windows 2016 failed to get IP from DHCP server
### VM details
```
apiVersion: kubevirt.io/v1alpha1
kind: VirtualMachine
metadata:
  name: windows2016
spec:
  terminationGracePeriodSeconds: 0
  domain:
    cpu:
      cores: 2
    features:
      acpi: {}
      apic: {}
      hyperv:
        relaxed: {}
        vapic: {}
        spinlocks:
          spinlocks: 8191
    clock:
      utc: {}
      timer:
        hpet:
          present: false
        pit:
          tickPolicy: delay
        rtc:
          tickPolicy: catchup
        hyperv: {}
    firmware:
      uuid: 5d307ca9-b3ef-428c-8861-06e72d69f223
    resources:
      requests:
        memory: 4096M
    devices:
      disks:
      - name: server2016
        volumeName: server2016
        disk:
          dev: vda
  volumes:
    - name: server2016
      persistentVolumeClaim:
        claimName: disk-windows
```
### VM libvirt xml
```
# virsh -r dumpxml 1
<domain type='qemu' id='1'>
  <name>default_windows2016</name>
  <uuid>5d307ca9-b3ef-428c-8861-06e72d69f223</uuid>
  <metadata>
    <kubevirt xmlns="http://kubevirt.io">
      <uid>e8232248-1f90-11e8-9eca-525400c99dab</uid>
      <graceperiod>
        <deletionGracePeriodSeconds>0</deletionGracePeriodSeconds>
      </graceperiod>
    </kubevirt>
  </metadata>
  <memory unit='KiB'>4000768</memory>
  <currentMemory unit='KiB'>4000000</currentMemory>
  <vcpu placement='static'>2</vcpu>
  <resource>
    <partition>/machine</partition>
  </resource>
  <sysinfo type='smbios'>
    <system>
      <entry name='uuid'>5d307ca9-b3ef-428c-8861-06e72d69f223</entry>
    </system>
  </sysinfo>
  <os>
    <type arch='x86_64' machine='pc-q35-2.10'>hvm</type>
    <boot dev='hd'/>
  </os>
  <features>
    <acpi/>
    <apic/>
    <hyperv>
      <relaxed state='on'/>
      <vapic state='on'/>
      <spinlocks state='on' retries='8191'/>
    </hyperv>
  </features>
  <cpu>
[...]

PR #801, opened 2018-03-09
Fix DHCP server options to allow Windows accept offered IP

vladikr:
Server-ID Option 54 and Domain-Name-Server Option 6
should have length 4 instead of 16

### DIFF
--- a/pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
+++ b/pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
@@ -46,12 +46,12 @@ func SingleClientDHCPServer(
 	handler := &DHCPHandler{
 		clientIP:      clientIP,
 		clientMAC:     clientMAC,
-		serverIP:      serverIP,
+		serverIP:      serverIP.To4(),
 		leaseDuration: infiniteLease,
 		options: dhcp.Options{
 			dhcp.OptionSubnetMask:       []byte(clientMask),
 			dhcp.OptionRouter:           []byte(routerIP),
-			dhcp.OptionDomainNameServer: []byte(dnsIP),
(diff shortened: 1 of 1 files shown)


vladikr:
Should resolve #789

pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
```diff
@@ -46,12 +46,12 @@ func SingleClientDHCPServer(
 	handler := &DHCPHandler{
 		clientIP:      clientIP,
 		clientMAC:     clientMAC,
-		serverIP:      serverIP,
+		serverIP:      serverIP.To4(),
 		leaseDuration: infiniteLease,
 		options: dhcp.Options{
 			dhcp.OptionSubnetMask:       []byte(clientMask),
 			dhcp.OptionRouter:           []byte(routerIP),
```
rmohr: And here we don't need a`routerIP.To4()`?

rmohr: > @vladikr what about routerIP and clientMask?

Ok, I see in the data you shared in #789 that the subnet mask seems to be serialized to four byte, let's check that routerIP is ok too.

vladikr: > let's check that routerIP is ok too.

Yea, Default-Gateway Option 3, length 4: 10.244.0.1
OptionRouter                 OptionCode = 3

rmohr: Great!

fabiand:
@cynepco3hahue can you please check if this fix is really fiing your issue?

PR #808, opened 2018-03-12
WIP RFC api: let user select volume mode

### LINKED ISSUE #816: PVC disks fail if they are block PVs
In 1.9 Kube supports block PVs.

If such a PV is used with KubeVirt, then it will fail with

```
      message: 'server error. command Launcher.Sync failed: virError(Code=38, Domain=18,
        Message=''Cannot access storage file ''/var/run/kubevirt-private/vm-disks/pvc1/disk.img''
        (as uid:107, gid:107): No such file or directory'')'
```

fromanirh:
Reviewers: This PR is by no means complete. I'm using it as vessel to discuss few design topics. Feel free to suggest a better venue for this discussion.

Topics for discussion:
- per https://kubernetes.io/docs/reference/feature-gates/ , the whole blockvolume support is behind a feature gate which is disabled per default in 1.9. Do we want/need instead to assume this feature is enabled in kubevirt installations?
- per https://schd.ws/hosted_files/kccncna17/8e/Mitsuhiro_Tanino_Block_Volume_KC_CNC_NA17.pdf looks like that the only block volume supported in 1.9 is FC. ISCSI (and more) is planned for 1.10.
Does this make this feature less desirable for the short term
- API wise, do which is the best way to expose this feature? I choose the simplest, minimal change approach as described below, but I'm not sure this is what we want.

+++

Starting with version 1.9, k8s gained the option to expose the volumes
as raw block device, instead of always creating filesystem on them.
We want to expose this feature for consumption to VMs running into PODs.

This patch implements the minimal API change, adding one optional
parameter to volumes, to turn them in raw block devices.

We should have this feature opt-in because it is alpha-stage in
kubernetes 1.9, enabled using the feature gate. Furthermore,
changes may likely happen during the k8s 1.10 cycle.

Further references:
- https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/raw-block-pv.md
- https://schd.ws/hosted_files/kccncna17/8e/Mitsuhiro_Tanino_Block_Volume_KC_CNC_NA17.pdf

Fixes #816

Signed-off-by: Francesco Romani <fromani@redhat.com>

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3547,6 +3547,10 @@
      "name"
     ],
     "properties": {
+     "mode": {
+      "description": "Mode indicates how the volume should be accessed by the VM. Supported values: block, filesystem (default).\n+optional",
+      "type": "string"
+     },
      "cloudInitNoCloud": {
       "description": "CloudInitNoCloud represents a cloud-init NoCloud user-data source.\nThe NoCloud data will be added as a disk to the vm. A proper cloud-init installation is required inside the guest.\nMore info: http://cloudinit.readthedocs.io/en/latest/topics/datasources/nocloud.html\n+optional",
       "$ref": "#/definitions/v1.CloudInitNoCloudSource"
--- a/pkg/api/v1/schema.go
+++ b/pkg/api/v1/schema.go
@@ -211,6 +211,9 @@ type VolumeSource struct {
 	// Ephemeral is a special volume source that "wraps" specified source and provides copy-on-write image on top of it.
 	// +optional
 	Ephemeral *EphemeralVolumeSource `json:"ephemeral,omitempty"`
+	// Mode indicates how the volume should be accessed by the VM. Supported values: block, filesystem (default).
+	// +optional
+	Mode string `json:"mode,omitempty"`
 }
 
 type EphemeralVolumeSource struct {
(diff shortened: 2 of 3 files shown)


cynepco3hahue:
@fromanirh FYI PR to move our code to client-go version 6 - https://github.com/kubevirt/kubevirt/pull/678

fabiand:
My general take is: We do not need to configure the type (raw/fs) of PV on the KubeVirt side. Instead, we can look at the PV(C) to understand if it's a fs or raw PV, and the connect it accordingly.

Thus the decision if raw or fs storage should be used, is made when the PVC is created (by setting PVC `volumeMode`).

See https://kubernetes.io/docs/concepts/storage/persistent-volumes/#raw-block-volume-support

This means, that in my understanding - and in order to keep it simple for the user - on the KubeVirt side, we do not need a new configuration item, instead the ystem (KubeVirt) just needs to decide on the type of PV how to connect it to the VM.

- If fs, then create `disk.img` and connect (we have this today)
- If block, then connect directly (this PR)

fromanirh:
@fabiand looks like I addressed this from the opposite direction! your approach is way nicer and simpler for everyone (AFAIU), so I'm all for it. I'll keep this PR open until I post the new one changing direction, after that I'll close this one.

cynepco3hahue:
@fromanirh You will need client-go version 6 to fetch type of PV(https://github.com/kubevirt/kubevirt/pull/678)

cynepco3hahue:
I missed that I already leaved comment, srry for noise

fabiand:
#678 was merged.

@cynepco3hahue @mpolednik any feedback on this pr or is a rebase enough?

fabiand:
@aglitke we are still one the same page that this is good to have but not critical?

mpolednik:
> My general take is: We do not need to configure the type (raw/fs) of PV on the KubeVirt side. Instead, we can look at the PV(C) to understand if it's a fs or raw PV, and the connect it accordingly.

 /signed

We shouldn't ask user to fill in something that the system we control knows.

PR #811, opened 2018-03-13
Build virtctl also for darwin and windows

### LINKED ISSUE #810: build and publish virtctl for other operating systems
Currently virtctl is only build and published for linux/amd64. It would be nice to be able to download macos and windows binaries, too. PR follows.

slintes:
This change ensures that `virtctl` binaries are build for `linux`, `darwin` and `windows` (all `amd64`), regardless on which platform the build was triggered. When I understand the travis configuration correct, all binaries will automatically be released because of the `virtctl-*` pattern.

Fixes #810

### DIFF
--- a/hack/build-go.sh
+++ b/hack/build-go.sh
@@ -71,13 +71,27 @@ for arg in $args; do
     elif [ "${target}" = "install" ]; then
         eval "$(go env)"
         BIN_NAME=$(basename $arg)
-        ARCHBIN=${BIN_NAME}-$(git describe --always --tags)-${GOHOSTOS}-${GOHOSTARCH}
+        ARCH_BASENAME=${BIN_NAME}-$(git describe --always --tags)
+        ARCHBIN=${ARCH_BASENAME}-${GOHOSTOS}-${GOHOSTARCH}
         mkdir -p ${CMD_OUT_DIR}/${BIN_NAME}
         (
             cd $arg
             go vet ./...
             go build -o ${CMD_OUT_DIR}/${BIN_NAME}/${ARCHBIN}
             (cd ${CMD_OUT_DIR}/${BIN_NAME} && ln -sf ${ARCHBIN} ${BIN_NAME})
+
+            # build virtctl binaries for other operating systems
+            if [ "${BIN_NAME}" = "virtctl" ]; then
+                if [[ "${GOHOSTOS}" != "linux" ]] || [[ "${GOHOSTARCH}" != "amd64" ]]; then
+                    GOOS=linux GOARCH=amd64 go build -o ${CMD_OUT_DIR}/${BIN_NAME}/${ARCH_BASENAME}-linux-amd64
+                fi
+                if [[ "${GOHOSTOS}" != "darwin" ]] || [[ "${GOHOSTARCH}" != "amd64" ]]; then
+                    GOOS=darwin GOARCH=amd64 go build -o ${CMD_OUT_DIR}/${BIN_NAME}/${ARCH_BASENAME}-darwin-amd64
+                fi
+                if [[ "${GOHOSTOS}" != "windows" ]] || [[ "${GOHOSTARCH}" != "amd64" ]]; then
+                    GOOS=windows GOARCH=amd64 go build -o ${CMD_OUT_DIR}/${BIN_NAME}/${ARCH_BASENAME}-windows-amd64.exe
+                fi
+            fi
         )
     else
         (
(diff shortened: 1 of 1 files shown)


ovirt-infra:

hack/build-go.sh
```diff
@@ -71,13 +71,27 @@ for arg in $args; do
...
+        ARCHBIN=${ARCH_BASENAME}-${GOHOSTOS}-${GOHOSTARCH}
         mkdir -p ${CMD_OUT_DIR}/${BIN_NAME}
         (
             cd $arg
             go vet ./...
             go build -o ${CMD_OUT_DIR}/${BIN_NAME}/${ARCHBIN}
             (cd ${CMD_OUT_DIR}/${BIN_NAME} && ln -sf ${ARCHBIN} ${BIN_NAME})
+
+            # build virtctl binaries for other operating systems
+            if [ "${BIN_NAME}" = "virtctl" ]; then
```
mpolednik: One of these will override the binary built at line 80 IIUC (in case of virtctl). The building could therefore be done as a part of an `else` block, and linked afterwards.

slintes: If you run this script on another architecture than amd64, or another OS than linux/darwin/windows, the build at line 80 wouldn't be overridden. I didn't want to break the "feature" that you always have a binary at the end, which runs on the same platform where you built it. Not sure if that is more a theoretical feature only though... :)

stu-gott: Thanks for your contribution @slintes! That said, I think @mpolednik has a point here--that `go build` command will be run twice under the normal use case, which isn't ideal. Like @mpolednik said, couldn't we use if/else or case statements to accomplish a default?

slintes: mh, no... for virtctl it even runs up to 4 times: once for the os/arch on which the build runs ($GOHOSTOS / $GOHOSTARCH), which CAN at least theoretically be different than linux/amd64, if you run the script directly and not within docker. And then I added builds for linux/darwin/windows on amd64, because IMHO those should always be released, but it will skip one of them if it matches the current os/arch.

But let's start the other way around: do you even want to have a build with $GOHOSTOS / $GOHOSTARCH, or do you just always need the linux/amd64 version? That would make it more simple:
```
        eval "$(go env)"
        BIN_NAME=$(basename $arg)
        ARCH_BASENAME=${BIN_NAME}-$(git describe --always --tags)
        LINUX_NAME=${ARCH_BASENAME}-linux-amd64
        mkdir -p ${CMD_OUT_DIR}/${BIN_NAME}
        (
            cd $arg
            go vet ./...
            
            # always build linux/amd64
            GOOS=linux GOARCH=amd64 go build -o ${CMD_OUT_DIR}/${BIN_NAME}/${LINUX_NAME}
            (cd ${CMD_OUT_DIR}/${BIN_NAME} && ln -sf ${LINUX_NAME} ${BIN_NAME})
            
            # build virtctl also for darwin and windows
            if [ "${BIN_NAME}" = "virtctl" ]; then
                GOOS=darwin GOARCH=amd64 go build -o ${CMD_OUT_DIR}/${BIN_NAME}/${ARCH_BASENAME}-darwin-amd64
                GOOS=windows GOARCH=amd64 go build -o ${CMD_OUT_DIR}/${BIN_NAME}/${ARCH_BASENAME}-windows-amd64.exe
            fi
        )
```

mpolednik: > But let's start the other way around: do you even want to have a build with $GOHOSTOS / $GOHOSTARCH, or do you just always need the linux/amd64 version? That would make it more simple:

I believe this simplification is what we need. cc @rmohr

fabiand: For the client side it does make sense to also support other architectures.

hack/build-go.sh
```diff
@@ -71,13 +71,27 @@ for arg in $args; do
...
+                if [[ "${GOHOSTOS}" != "linux" ]] || [[ "${GOHOSTARCH}" != "amd64" ]]; then
+                    GOOS=linux GOARCH=amd64 go build -o ${CMD_OUT_DIR}/${BIN_NAME}/${ARCH_BASENAME}-linux-amd64
```
rmohr: Look like we  don't need these checks at all. We would always build the same artifact, no matter on which hostos or arch we are. Looks like we could just build all all the time and then link the correct one based on the hostos and arch where we are currently running on. How time consuming is building all of them?

slintes: > link the correct one based on the hostos and arch
I think we always want to link the linux/amd64 binary, because that is used for building docker images?
Other than that that matches my comment from above about always building linux, and for virtctl also darwin and windows.
I just checked build times, and I'm a bit surprised tbh: the linux builds need a few seconds only, but the darwin and windows builds around 50 seconds each (on my macbook). That's a bit slow... I will update this PR nevertheless

fabiand:
Awesome! 🌮

PR #815, opened 2018-03-14
cluster: Add a fedora template

fabiand:

### DIFF
--- a/cluster/vm-template-fedora.yaml
+++ b/cluster/vm-template-fedora.yaml
@@ -1,0 +1,55 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: fedora-vm-template
+  annotations:
+    description: "OCP KubeVirt Red Hat Enterprise Linux 7.4 VM template"
+    tags: "kubevirt,ocp,template,linux"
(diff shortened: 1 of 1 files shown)


cynepco3hahue:
We have already a lot of stuff under the cluster, maybe it time to create two directories `cluster/vms` and `cluster/templates`
Template lgtm :smile:

fabiand:
Let me see if I can find some structure ...

rmohr:
Let's stop adding all those templates at all. We have enough examples already. If they are meant to  be used by e.g. manageIQ, then they should be included there.

fabiand:
Both existing templates are not working out of the box.

This newly introduced template does, because it's using a registry disk.

stu-gott:
I agree that items not critical to running a KubeVirt cluster probably don't belong there.

Now that we have a user-guide repository, would it be worth considering moving any example templates/manifests there? A lot of the documents have examples inline, but it might be helpful to also have a dedicated directory.

fabiand:
So, flame me. But I'll merge this template. I'll propose some changes to make the locations better.

PR #817, opened 2018-03-14
Implement convenience functions to start/stop OfflineVMs in virtctl

stu-gott:
Add start and stop commands to virtcli, which will modify the spec of an OfflineVirtualMachine to update `running` to `true` or `false`.

- [x] Unit tests
- [x] Update User Guide

### DIFF
--- a/pkg/virtctl/virtctl.go
+++ b/pkg/virtctl/virtctl.go
@@ -24,8 +24,6 @@ import (
 	"os"
 
 	flag "github.com/spf13/pflag"
-
-	"kubevirt.io/kubevirt/pkg/kubecli"
 )
 
 const (
@@ -60,14 +58,4 @@ func (o *Options) Usage() string {
 	usage := "The following options can be passed to any command:\n\n"
 	usage += o.FlagSet().FlagUsages()
 	return usage
-}
-
-func GetKubevirtClient(flags *flag.FlagSet) (kubecli.KubevirtClient, error) {
-	server, _ := flags.GetString("server")
-	kubeconfig, _ := flags.GetString("kubeconfig")
-	if (server == "") && (kubeconfig == "") {
-		return kubecli.GetKubevirtClient()
-	} else {
-		return kubecli.GetKubevirtClientFromFlags(server, kubeconfig)
-	}
 }
--- a/pkg/virtctl/virtctl.go
+++ b/pkg/virtctl/virtctl.go
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
- * Copyright 2017 Red Hat, Inc.
+ * Copyright 2017, 2018 Red Hat, Inc.
  *
  */
 
@@ -24,6 +24,13 @@ import (
 	"os"
 
 	flag "github.com/spf13/pflag"
+
+	"kubevirt.io/kubevirt/pkg/kubecli"
+)
+
+const (
+	STATUS_OK    = 0
+	STATUS_ERROR = 1
 )
 
 type App interface {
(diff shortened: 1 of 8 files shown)


fabiand:
Looks pretty straight forward to me.

@slintes do you have an opinion? ;)

slintes:
Mh, on a quick look:
- There is some code duplication between the console, vnc and the new commands (checking if the vm name is present, and getting the kubevirtclient). Would make sense to share that code imho.
- Would be nice to print a short message if the vm is already started/stopped :)

fabiand:
Good points @slintes - I'll wait for an update, otherwise this looks good to me

stu-gott:
@slintes Thanks for the feedback! I didn't really pay much attention to the other modules but I agree there probably is some room for consolidating common logic. I'll see what I can do.

I've implemented the suggested feedback/status message (still on my own branch), but am having second thoughts. Emitting a message sort of implies an error condition. Maybe log output at a higher verbosity?

slintes:
When I try to start a vm, and I get a "VM is already started" or similar message, at least I wouldn't interpret that as an error, but as a kind hint that I'm trying to do something what is not possible :)

rmohr:
> When I try to start a vm, and I get a "VM is already started" or similar message, at least I wouldn't interpret that as an error, but as a kind hint that I'm trying to do something what is not possible :)

There is pause and resume support for deployments in kubectl, could we check what they do if you want to pause an already paused deployment? I would align with that.

stu-gott:
Looks like kubectl considers that to be an error condition.

```bash
[kubevirt]$ ./cluster/kubectl.sh --namespace kube-system rollout pause deployment kube-dns
deployment "kube-dns" paused
[kubevirt]$ ./cluster/kubectl.sh --namespace kube-system rollout pause deployment kube-dns
error: deployments "kube-dns" is already paused
```

pkg/virtctl/vnc/vnc.go
```diff
@@ -31,31 +31,29 @@ import (
...
 
 	if len(flags.Args()) != 2 {
 		log.Println("VM name is missing")
-		return 1
+		return virtctl.STATUS_ERROR
 	}
 	vm := flags.Arg(1)
 
-	virtCli, err := kubecli.GetKubevirtClientFromFlags(server, kubeconfig)
+	virtCli, err := virtctl.GetKubevirtClient(flags)
```
rmohr: Why are you adding this `virtcl.GetKubevirtClientFromFlags`? `kubecli.GetKubeivirtClient` already takes `server` and `kubeconfig` flags into account.

stu-gott: This `Run` function takes `flags` as a parameter--which is technically different than the default command line flags (especially in the test suite). master and kubeconfig from the flags passed into `Run` as a parameter should be honored.

This is simply a refactor to remove duplicate code between vnc.go, console.go, and offlinevm.go. Perhaps the real issue here is that `virtctl.GetKubevirtClient` should really be located in `kubecli` and we should think about why a function named `kubecli.GetKubevirtClientFromFlags` takes two strings as parameters. Perhaps it should be named `kubecli.GetKubevirtClientFromConfig`?

stu-gott: I was refactoring the code to match my previous suggestion and realized: virtctl uses a flag called `server` while kubecli uses one called `master`...

Because this has direct client API implications we shouldn't change this capriciously, but shouldn't those be unified?

fabiand: Please check what kubectl uses and adjust our code appropriately.

stu-gott:
Note this PR will need to be reconciled with https://github.com/kubevirt/kubevirt/pull/827 as a function name in kubecli changed that hasn't been accounted for here.

rmohr:
retest this please

rmohr:
let's wait for ci.

pkg/virtctl/virtctl.go
```diff
@@ -54,3 +61,13 @@ func (o *Options) Usage() string {
 	usage += o.FlagSet().FlagUsages()
 	return usage
 }
+
+func GetKubevirtClient(flags *flag.FlagSet) (kubecli.KubevirtClient, error) {
```
rmohr: The function is not used right now ...

stu-gott: Good catch. Since we've committed to an entirely different direction, it will never be used.

fabiand:
retest this please

rmohr:
@stu-gott you need to rebase and/or run "make generate".

rmohr:
retest this please

fabiand:
retest this please

PR #830, opened 2018-03-21
Provide all relevant POD routes to the guest

### LINKED ISSUE #824: Pod on one node can not ping VM on other node
On OpenShift.

Launching a pod on on node, then trying to ping running a VM on another node fails.

- pod tp pod on another node works
- pod to vm on same node works

vladikr:
This series will provide all relevant POD routes to the guest via
DHCP option 121 (RFC4332 - https://tools.ietf.org/html/rfc3442)

Should resolve #824 

### DIFF
--- a/pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
+++ b/pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
@@ -26,6 +26,7 @@ import (
 
 	dhcp "github.com/krolaw/dhcp4"
 	dhcpConn "github.com/krolaw/dhcp4/conn"
+	"github.com/vishvananda/netlink"
 
 	"kubevirt.io/kubevirt/pkg/log"
 )
@@ -39,19 +40,22 @@ func SingleClientDHCPServer(
 	serverIface string,
 	serverIP net.IP,
 	routerIP net.IP,
-	dnsIP net.IP) error {
+	dnsIP net.IP,
+	routes *[]netlink.Route) error {
 
 	log.Log.Info("Starting SingleClientDHCPServer")
 
+	netRoutes := FormClasslessRoutes(routes, routerIP)
 	handler := &DHCPHandler{
 		clientIP:      clientIP,
 		clientMAC:     clientMAC,
 		serverIP:      serverIP.To4(),
 		leaseDuration: infiniteLease,
 		options: dhcp.Options{
-			dhcp.OptionSubnetMask:       []byte(clientMask),
-			dhcp.OptionRouter:           []byte(routerIP),
-			dhcp.OptionDomainNameServer: []byte(dnsIP.To4()),
+			dhcp.OptionSubnetMask:           []byte(clientMask),
+			dhcp.OptionRouter:               []byte(routerIP),
+			dhcp.OptionDomainNameServer:     []byte(dnsIP.To4()),
+			dhcp.OptionClasslessRouteFormat: netRoutes,
 		},
 	}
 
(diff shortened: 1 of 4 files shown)


rmohr:
retest this please

rmohr:
retest this please

cynepco3hahue:
I miss functional test to start VM's on different nodes and ping one from another.
Or create an issue and I will add it :smile:

pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
```diff
@@ -100,3 +104,31 @@ func (h *DHCPHandler) ServeDHCP(p dhcp.Packet, msgType dhcp.MessageType, options
...
+	// For example:
+	// 		routes:
+	//				10.0.0.0/8 ,  gateway: 10.1.2.3
+	//              192.168.1/24, gateway: 192.168.2.3
+	//		would result in the following structure:
+	//      []byte{8, 10, 10, 1, 2, 3, 24, 192, 168, 1, 192, 168, 2, 3}
+
+	for _, route := range *routes {
+		if route.Dst == nil {
+			continue
```
stu-gott: A route without a destination... So this code basically discards the entire route... Having only skimmed RFC4332, is this the correct thing to do? (the other option being to raise an error). To be clear, I'm not suggesting a change, I'm simply hoping to clarify.

This condition wasn't covered by unit tests. Would it be difficult to add one?

vladikr: > A route without a destination...

The only route without a destination I know is a default gateway, which is set in OptionRouter , not via OptionClasslessRouteFormat. Are there any other cases?

vladikr:
> I miss functional test to start VM's on different nodes and ping one from another.

@cynepco3hahue - It should be covered by [1], but you are welcome to improve it :)

[1] https://github.com/kubevirt/kubevirt/blob/master/tests/vm_networking_test.go#L172

pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
stu-gott: Not that I'm aware of. Thanks for clarifying!

rmohr:
> @cynepco3hahue - It should be covered by [1], but you are welcome to improve it :)
> 
> [1] https://github.com/kubevirt/kubevirt/blob/master/tests/vm_networking_test.go#L172

VM to VM would indeed be great. it would cover a very relevant additional case where something can go wrong.

cynepco3hahue:
@vladikr @rmohr will add it to my TODO list

vladikr:
[BUG: pod can not ping VM on other node](https://trello.com/c/AXEnTYnE/43-bug-pod-can-not-ping-vm-on-other-node)

vladikr:
[BUG: pod can not ping VM on other node](https://trello.com/c/AXEnTYnE/43-bug-pod-can-not-ping-vm-on-other-node)

davidvossel:
> @cynepco3hahue - It should be covered by [1], but you are welcome to improve it :)
[1] https://github.com/kubevirt/kubevirt/blob/master/tests/vm_networking_test.go#L172

these unit tests verify
* pod -> VM
* node -> VM
* VM -> google

To finish the test matrix, we should add cases for  
* VM -> VM
* VM -> pod

fabiand:
We also need to make sure that we ping from _different_ nodes

vladikr:
> We also need to make sure that we ping from different nodes

We do ;)

fabiand:
Darn. very good.

davidvossel:
retest this please

rmohr:
retest this please

mlsorensen:
I may have run into an issue related to this. It seems I'm getting a static route to my gateway, and no default gateway.

My lease looks like this:

> lease {
>   interface "eth0";
>   fixed-address 172.16.52.146;
>   option subnet-mask 255.255.255.255;
>   option routers 169.254.1.1;
>   option dhcp-lease-time 86313600;
>   option dhcp-message-type 5;
>   option domain-name-servers 192.168.0.10;
>   option dhcp-server-identifier 10.11.12.13;
>   option classless-static-routes 32.169.254.1.1 169.254.1.1;
>   renew 3 2019/05/01 23:04:52;
>   rebind 0 2020/08/16 23:48:11;
>   expire 6 2020/12/19 20:48:11;
> }

And results in routing table like this:

> [root@fedora ~]# ip route show
> 169.254.1.1 via 169.254.1.1 dev eth0 proto static

And I can't get anywhere. Maybe in the "// Filter out irrelevant routes" it should skip routes with destination IP of the gateway? Or does that defeat your purpose?

I ran some debugging and I'm getting a VIF route entry of Src=<nil> Dst.IP=169.254.1.1/32

If I do filter out the Dst == router route I'm back in business:

> [root@fedora ~]# ip route show
> default via 169.254.1.1 dev eth0

PR #838, opened 2018-03-22
Fix controller

### LINKED ISSUE #837: Sometimes two VM pods are scheduled that should be impossible
If I post a new VM, very seldom I see that two pods are getting created. That should be impossible and it causes the vm controller to just sit there and wait until someone resolved that situation.

rmohr:
@davidvossel  FYI

* Trigger a pod deletion once the vm is in a final state, or the vm has a deletion timestamp set
* Use finalizers to keep the vm around until the pod is removed from the cluster.
* Split actions and status updates in the controller
* Controller references are not used to not trigger the garbage collector, so that we have full control over the pods lifecycle (may change in the future).
* virt-handler now contains a basic heart-beat mechanism
* node controller is introduced to make nodes unschedulable for vms if virt-handler is unresponsive for some time
* node controller is introduced to move vms to final state, if virt-handler is unresponsive, the vm pods are already gone and the vm is not in a final state.

Will fix #837.


rmohr:
@davidvossel I no longer see this as WIP. I think it could go in soon. Looking forward to a first review round.

pkg/virt-controller/watch/vm.go
```diff
@@ -83,209 +126,427 @@ func (c *VMController) runWorker() {
...
+		if len(pods) == 0 {
+			controller.RemoveFinalizer(vmCopy, virtv1.VirtualMachineFinalizer)
 		}
+	}
 
-		// Defaulting and setting constants
-		// TODO move defaulting to virt-api
-		kubev1.SetObjectDefaults_VirtualMachine(vmCopy)
+	// Select the right failure reason in case we have an error
+	// TODO this reason detetion is not 100 % accurate
```
davidvossel: rather than guessing here, reason should be populated from sync().

pkg/virt-controller/watch/vm.go
```diff
@@ -83,209 +126,427 @@ func (c *VMController) runWorker() {
...
+	reason := ""
+	if len(pods) == 0 {
+		reason = "FailedCreate"
+	} else if len(pods) == 1 && vm.IsFinal() || vm.DeletionTimestamp != nil {
+		reason = "FailedDelete"
+	} else {
+		reason = "FailedHandOver"
+	}
+	controller.NewVirtualMachineConditionManager().CheckFailure(vmCopy, syncErr, reason)
```
davidvossel: is it possible for a vm to recover from any of these errors?

pkg/virt-controller/watch/vm.go
```diff
@@ -83,209 +126,427 @@ func (c *VMController) runWorker() {
...
+	}
+	return false
+}
+
+func (c *VMController) getControllerOf(pod *k8sv1.Pod) *v1.OwnerReference {
+	t := true
+	return &v1.OwnerReference{
+		Kind:               virtv1.VirtualMachineGroupVersionKind.Kind,
+		Name:               pod.Labels[virtv1.DomainLabel],
+		UID:                types.UID(pod.Annotations[virtv1.CreatedByAnnotation]),
```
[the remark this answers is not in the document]
rmohr: Hm, the informer only watches for pods which are created by us, I could still add a safety check here.

pkg/virt-controller/watch/vm.go
```diff
@@ -83,209 +126,427 @@ func (c *VMController) runWorker() {
...
+				vmCopy.Status.NodeName = pods[0].Spec.NodeName
+			} else if !podIsTerminating && !containersAreReady {
+				vmCopy.Status.Phase = virtv1.Scheduling
+			} else if podIsTerminating {
+				vmCopy.Status.Phase = virtv1.Failed
 			}
-			// the pod informer will reenqueue the key as a result of it being deleted.
-			return nil
+		} else if len(pods) == 0 && vm.IsScheduling() {
+			// someone other then the controller deleted the pod unexpectedly
```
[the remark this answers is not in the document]
rmohr: We only get to the updateStatus() function fi all expectations of pod creation/deletion/handover are fulfilled. If any expectation is still not satisfied we simply do nothing, so that should work.

pkg/virt-controller/watch/vm.go
rmohr: Yes. Normally the reason for these errors are transient connection errors. We simply re-enqueue on such errors with a backoff.

rmohr:
> One kind of broad thing I'm uncertain about is how the vm controller behaves when VMs are mid life-cycle and leader election occurs.
> 
> Since expectations are local to a specific process, it seems like it's possible during leader election for some code paths (that appear to be impossible to be hit) to get hit.

The only situation where we could have issues, is when we create/delete/hand-over the pod (so the ones protected with the expectation manager). If new controller gets elected, it will perform a full cache sync when it comes up. Since we have a guarantee that our actions are already persisted on the cluster, just not in our cache, the cache sync on the new elected controller will pick up that change too. 

Bottomline: It will work. The cruicial thing is that a new controller only starts synchronizing its cache *after* it got elected. At this time the old controller already crashed (panic because it lost the race).

pkg/virt-controller/watch/vm.go
rmohr: Played it through again. I was wrong. It is accurate.

pkg/virt-controller/watch/vm.go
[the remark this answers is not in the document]
rmohr: Oh and getting a pod which is not interesting for us will simply fail to select a vm from the cache. As a consequence nothing will happen.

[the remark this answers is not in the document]
davidvossel: :+1: ah yes, I missed that the informer is filtered

pkg/virt-controller/watch/vm.go
```diff
@@ -83,222 +126,433 @@ func (c *VMController) runWorker() {
...
+	if err != nil {
+		return err
+	}
+
+	if len(pods) > 1 {
+		logger.Reason(fmt.Errorf("More than one pod detected")).Error("That should not be possible, will not requeue")
 		return nil
+	} else if len(pods) == 1 && pods[0].Annotations[virtv1.OwnedByAnnotation] == "virt-handler" {
+		// Whenever we see that a matching pod is owned by virt-handler, we can mak handover as fulfilled
+		c.handoverExpectations.CreationObserved(key)
```
davidvossel: would it make more sense to place this 'CreationObserved' call in the updatePod() function?

pkg/virt-controller/watch/vm.go
```diff
@@ -83,222 +126,433 @@ func (c *VMController) runWorker() {
...
+	default:
+		return fmt.Errorf("unknown vm phase %v", vm.Status.Phase)
+	}
 
-		// Deep copy the object, so that we can safely manipulate it
-		vmCopy := vm.DeepCopy()
-		logger = log.Log.Object(vmCopy)
+	// Select the right failure reason in case we have an error
+	reason := ""
+	if !podExists {
```
davidvossel: We shouldn't have to guess the reason for the sync failure.

pkg/virt-controller/watch/vm.go
```diff
@@ -83,222 +126,433 @@ func (c *VMController) runWorker() {
...
+		reason = "FailedCreate"
+	} else if vm.DeletionTimestamp != nil {
+		reason = "FailedDelete"
```
davidvossel: use the const values you created at the top of this file.

pkg/virt-controller/services/template_test.go
```diff
@@ -35,21 +35,18 @@ import (
...
 		Context("launch template with correct parameters", func() {
 			It("should work", func() {
 
-				Expect(err).To(BeNil())
-				pod, err := svc.RenderLaunchManifest(&v1.VirtualMachine{ObjectMeta: metav1.ObjectMeta{Name: "testvm", Namespace: "testns", UID: "1234"}, Spec: v1.VirtualMachineSpec{Domain: v1.DomainSpec{}}})
+				pod := svc.RenderLaunchManifest(&v1.VirtualMachine{ObjectMeta: metav1.ObjectMeta{Name: "testvm", Namespace: "testns", UID: "1234"}, Spec: v1.VirtualMachineSpec{Domain: v1.DomainSpec{}}})
 
-				Expect(err).To(BeNil())
 				Expect(pod.Spec.Containers[0].Image).To(Equal("kubevirt/virt-launcher"))
 				Expect(pod.ObjectMeta.Labels).To(Equal(map[string]string{
```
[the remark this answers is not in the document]
rmohr: done

pkg/virt-controller/watch/vm.go
rmohr: Done. Placed it in updatePod and deletePod, since a graceful delete will be "redirected" from updatePod to deletePod.

pkg/virt-controller/watch/vm.go
rmohr: Done

pkg/virt-controller/watch/vm.go
rmohr: Done

rmohr:
retest this please

pkg/virt-controller/watch/node.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func NewNodeController(clientset kubecli.KubevirtClient, nodeInformer cache.SharedIndexInformer, recorder record.EventRecorder) *NodeController {
+	c := &NodeController{
+		clientset:        clientset,
+		Queue:            workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
+		nodeInformer:     nodeInformer,
+		recorder:         recorder,
+		heartBeatTimeout: 5 * time.Minute,
+		recheckInterval:  1 * time.Minute,
```
davidvossel: at some point we may want to consider making this a cli option. Reason being, if there is ever an issue with these thresholds, we could solve it with a configuration change rather than requiring a new release.

cynepco3hahue:
retest this please

rmohr:
@davidvossel the final test is there. Patch should be ready if CI agrees.

rmohr:
retest this please

rmohr:
retest this please

pkg/virt-controller/watch/node_test.go
```diff
@@ -0,0 +1,359 @@
...
+
+			addNode(node)
+
+			kubeClient.Fake.PrependReactor("patch", "nodes", func(action testing.Action) (handled bool, obj runtime.Object, err error) {
+				return true, nil, nil
+			})
+
+			vmInterface.EXPECT().List(gomock.Any()).Return(&virtv1.VirtualMachineList{}, nil)
+
+			controller.Execute()
```
cynepco3hahue: Can we check that the `node` label equals to `kubevirt.io/schedulable: false`?

rmohr: Good catch. Done

rmohr:
retest this please

Issue #898, opened 2018-04-11
prevent rsync error in make


yuvalif:
add ```sleep``` in line 35 of https://github.com/kubevirt/kubevirt/blob/master/hack/dockerized
to prevent from following error when running ```make```:
```
rsync: safe_read failed to read 1 bytes [sender]: Connection reset by peer (104)
rsync error: error in rsync protocol data stream (code 12) at io.c(276) [sender=3.1.3]
```

rmohr:
Hm, I think it is better to give fsync a timeout. E.g. 15 seconds and until then swallow the whole error output. Then do a final run and output the error. It greatly varies from run to run and host speed how  fast rsyncd is ready to accept connections (can actually take a few seconds and I don't want to delay the run for everyone).

rmohr:
We can also give it 30 seconds, then it is for sure enough. I think that within that timeout we should redirect  to /dev/null. If it is then still not ready, we should try once again and print the result to sterr, so that people can see that something is wrong. Otherwise it would just hang forever without any output.

PR #903, opened 2018-04-12
prevent rsync error in make #898

yuvalif:
could not add connection timeout, just polling for 30 seconds and redirecting output to dev/null. first time wait for 0.1 second so that 99% of the cases there is no change


hack/dockerized
```diff
@@ -33,11 +33,26 @@ trap finish EXIT
 
 RSYNCD_PORT=$(docker port $RSYNC_CID 873 | cut -d':' -f2)
 
-while ! rsync ${KUBEVIRT_DIR}/${RSYNCTEMP} "rsync://root@127.0.0.1:${RSYNCD_PORT}/build/${RSYNCTEMP}"; do
-    echo "Waiting for rsyncd to be ready"
-    sleep .1
+rsynch_fail_count=0
+
+while ! rsync ${KUBEVIRT_DIR}/${RSYNCTEMP} "rsync://root@127.0.0.1:${RSYNCD_PORT}/build/${RSYNCTEMP}" &>/dev/null; do
```
rmohr: bash has a `SECONDS` environment variable. You could set SECONDS to zero before you start the loop and just exit once SECONDS contains a value greater than 30. The auto-increment of SECONDS is a bash feature.

rmohr: also you can always use a sleep of 0.1. How long the rsync call will block is not easy to say. So in your implementation it may as well try for 60 seconds to connect.

yuvalif: i want to give indication that the system is not stuck, so i print the dots. dont think we should do that every 100ms

yuvalif: regarding SECONDS, it makes things more complex in the first check (the one that should print "Waiting for rsyncd to be ready"). do you think it is critical?

PR #934, opened 2018-04-20
virtctl: vnc: report errors from remote-viewers

fromanirh:
Related to issue https://github.com/kubevirt/kubevirt/issues/905 ,
already partially addressed in commit ce3a3188

Make it easier to understand what's wrong running

  virtctl vnc $VM

Since virtctl uses remote-viewer under the hood, this patch improves
the usability in three ways:
1. add --debug flag and pass it to remote-viewer, to have more
   informations about what's going on
2. if the invokation of remote-viewer fails, relay the output.
   It was swallowed and dropped before this patch
3. avoid the command to hang if remote-viewer fails too early.

### DIFF
--- a/pkg/kubecli/vm.go
+++ b/pkg/kubecli/vm.go
@@ -141,12 +141,12 @@ func (d *WebsocketRoundTripper) RoundTrip(r *http.Request) (*http.Response, erro
 	return resp, d.Do(conn, resp, err)
 }
 
-type wsCallbackObj struct {
-	in  io.Reader
-	out io.Writer
+type asyncWSRoundTripper struct {
+	Done       chan struct{}
+	Connection chan *websocket.Conn
 }
 
-func (obj *wsCallbackObj) WebsocketCallback(ws *websocket.Conn, resp *http.Response, err error) error {
+func (aws *asyncWSRoundTripper) WebsocketCallback(ws *websocket.Conn, resp *http.Response, err error) error {
 
 	if err != nil {
 		if resp != nil && resp.StatusCode != http.StatusOK {
@@ -156,26 +156,14 @@ func (obj *wsCallbackObj) WebsocketCallback(ws *websocket.Conn, resp *http.Respo
 		}
 		return fmt.Errorf("Can't connect to websocket: %s\n", err.Error())
 	}
+	aws.Connection <- ws
 
-	wsReadWriter := &BinaryReadWriter{Conn: ws}
-
-	copyErr := make(chan error)
-
-	go func() {
-		_, err := io.Copy(wsReadWriter, obj.in)
-		copyErr <- err
-	}()
-
-	go func() {
-		_, err := io.Copy(obj.out, wsReadWriter)
-		copyErr <- err
-	}()
-
-	err = <-copyErr
-	return err
+	// Keep the roundtripper open until we are done with the stream
+	<-aws.Done
+	return nil
 }
 
-func roundTripperFromConfig(config *rest.Config, in io.Reader, out io.Writer) (http.RoundTripper, error) {
+func roundTripperFromConfig(config *rest.Config, callback RoundTripCallback) (http.RoundTripper, error) {
 
 	// Configure TLS
 	tlsConfig, err := rest.TLSConfigFor(config)
@@ -191,13 +179,9 @@ func roundTripperFromConfig(config *rest.Config, in io.Reader, out io.Writer) (h
 		ReadBufferSize:  WebsocketMessageBufferSize,
 	}
 
-	obj := &wsCallbackObj{
-		in:  in,
-		out: out,
-	}
 	// Create a roundtripper which will pass in the final underlying websocket connection to a callback
 	rt := &WebsocketRoundTripper{
-		Do:     obj.WebsocketCallback,
+		Do:     callback,
 		Dialer: dialer,
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/virtctl/vnc/vnc.go
```diff
@@ -49,6 +51,7 @@ func NewCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
 		},
 	}
 	cmd.SetUsageTemplate(templates.UsageTemplate())
+	cmd.Flags().BoolVarP(&debug, "debug", "D", false, "enable debug output")
```
rmohr: Instead of introducing that flag, could you reuse the `-v`/`--verbose` flag? Let's say append the `--debug` flag to remote-viewer if we have a verbosity level higher or equal to 4.

pkg/virtctl/vnc/vnc.go
```diff
@@ -97,14 +97,28 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
 		viewResChan <- err
 	}()
 
+	select {
+	case err = <-viewResChan:
+		if err != nil {
+			return err
+		}
+	}
+
```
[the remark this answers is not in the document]
rmohr: What you can definitely do, and which may also fit the patter better, is adding this channel also to the selector at the bottom and make the whole ln.Accept() part execute in a go routine. Basically in one routine you have remote-viewer running, in the other go-routinge you have the proxying. In the main routing we simply wait for errors or completion on channels.

pkg/virtctl/vnc/vnc.go
```diff
@@ -112,12 +121,9 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
 
-	select {
-	case err = <-viewResChan:
-		if err != nil {
-			return err
-		}
-	}
+	glog.Infof("connection timeout: %v", LISTEN_TIMEOUT)
+	// exit early if spawning remote-viewer fails
+	ln.SetDeadline(time.Now().Add(LISTEN_TIMEOUT))
```
fromanirh: @rmohr I did this in parallel to your review, before to see your comments :) I can either keep this or do like you suggested, no strong feelings. Your call.

pkg/virtctl/vnc/vnc.go
```diff
@@ -97,22 +106,35 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
+			args = append(args, "--debug")
+			glog.Infof("remote-viewer commandline: %v", args)
+		}
+
+		cmnd := exec.Command("remote-viewer", args...)
+		output, err := cmnd.CombinedOutput()
 		if err != nil {
 			glog.Errorf("remote-viewer execution encountered an error: %v", err)
 		}
+		if err != nil || glog.V(2) {
```
rmohr: I would prefer Infof and the easier to read form:

```
if err != nil {
  glog.V(2).Infof(...)
}
```

pkg/virtctl/vnc/vnc.go
```diff
@@ -97,22 +106,35 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
+			glog.Errorf("remote-viewer: %v", string(output))
+		}
 		viewResChan <- err
 	}()
 
+	glog.Infof("connection timeout: %v", LISTEN_TIMEOUT)
+	// exit early if spawning remote-viewer fails
+	ln.SetDeadline(time.Now().Add(LISTEN_TIMEOUT))
```
rmohr: You can keep the timeout, but I would prefer to check the error channel of the command itself on the select statement at the bottom, instead of having the "racy" form.

pkg/virtctl/vnc/vnc.go
rmohr: Oh and it should be `if err ==nil`.

rmohr: Hm, what I had in mind is something like this:

```golang
if err != nil {
  glog.Errorf("remote-viewer execution failed: %v, output: %v", err, output)
} else {
  glog.V(2).("remote-viewer output: %s", output)
}

Comment (fromanirh) [reply pkg/virtctl/vnc/vnc.go]: done

File: pkg/virtctl/vnc/vnc.go [hunk 6]
```diff
@@ -76,70 +79,103 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
 	select {
 	case <-stopChan:
 	case err = <-readStop:
 	case err = <-writeStop:
 	case err = <-k8ResChan:
 	case err = <-viewResChan:
+	case err = <-listenResChan:
 	}
 
+	cmnd.Process.Kill() // avoid leak remote-viewer
```
rmohr: How is that possible? Wouldn't remove-viewer terminate as soon as our proxy server terminates?

fromanirh: Not sure how, but during my testing it happened a few time that `virtctl vnc` leaked remote-viewer processes. That *will* eventually terminate, but not before hitting timeout trying to connect to proxy. This could be confusing.

In my case, ti was caused by lack of syncronization between goroutines - this is why I introduced doneChan, which solves the issue

rmohr: ok.

However, changing the cmnd value from nil to an instance in a go-routine is not save. In theory we could get a nil exeception. There is no guarantee that the pointer update will be propagated back to the main routine.

rmohr: Further we probably also want to kill it on signals then ...

fromanirh: I decided to remove this code for the time being. Problem is - the issue is too vague. I'm not 100% clear why we leak those processes, so I believe better not commit half-baked code. We can tackle this in another PR once we clearly understand the issue.

pkg/virtctl/vnc/vnc.go
```diff
@@ -69,75 +73,116 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	var opts k8smetav1.GetOptions
+	vmObj, err := virtCli.VM(namespace).Get(vm, opts)
+	if err != nil {
+		return fmt.Errorf("Can't access VM %s: %s", vm, err.Error())
+	}
+	if !vmObj.IsRunning() {
```
rmohr: OK, I see why you have to do it this way, with the current virtcli implementation. However, the server does that validation already. We should change the virtcli interface to allow you modelling  a proper flow.

As far as I can see, the issue is this `virtCli.VM(namespace).VNC(vm, pipeInReader, pipeOutWriter)` basically blocks until we are done. So you basically have no way to see if the underlying websocket connection can be established. Therefore you need to start remote-viewer in parallel already. Let's go one step back and have a look on how that is handle in k8s:

```golang
	exec, err := remotecommand.NewSPDYExecutor(config, "POST", req.URL())
	if err != nil {
		return "", err
	}

	err = exec.Stream(remotecommand.StreamOptions{
		Stdout: &stdout,
		Stderr: &stderr,
		Tty:    false,
	})
```

As you can see, there is a split between establishing the connection, and attaching streams to the connection. If we do that, you can simply first establish the connection, check for errors, and if there are none, you can start start a go-routine which attaches the streams which whe can also attach to remote-viewer.

A more useful interaction with streams for kubevirt may look like this:

```golang
errChan :- make(chan error)

vnc, err := virtCli.VM(namespace).VNC(vm)
if err != nil {
    return fmt.Errorf("could not establish connection: %v", err)
}

go func() {
  errChan  <- vnc.Stream(VNCStreamOptions{
		Stdout: &stdout,
		Stderr: &stderr,
	})
}()
```

fromanirh: @rmohr great insight, that would improve the flow a lot. How would you like me to go forward? Should I file a different PR with the reworking of the streams or should I continue here?

rmohr: Please add it here. There is nothing better than having the improvement + the reason why the improvment is needed together :+1:

fromanirh: @rmohr sure, working on it.

pkg/kubecli/vm.go
```diff
@@ -126,56 +126,78 @@ func (s *BinaryReadWriter) err(err error) error {
...
 		}
-		return fmt.Errorf("Can't connect to websocket: %s\n", err.Error())
+		return resp, fmt.Errorf("Can't connect to websocket: %s\n", err.Error())
+	}
+
+	// Are we abusing the RoundTrip() abstraction?
+	// (see https://golang.org/pkg/net/http/#RoundTripper)
+	// We *are* "executing a single HTTP transaction, returning
+	// a Response for the provided Request."
+	// Also, it should be ok to consume the body later, in Stream().
```
rmohr: That would only work for non-interactive commands, where you would get the whole body back. A round-tripper encloses the complete request. You will have to keep the round-tripper open.

pkg/kubecli/vm.go
```diff
@@ -126,56 +126,78 @@ func (s *BinaryReadWriter) err(err error) error {
...
 			buf := new(bytes.Buffer)
 			buf.ReadFrom(resp.Body)
-			return fmt.Errorf("Can't connect to websocket (%d): %s\n", resp.StatusCode, buf.String())
+			return resp, fmt.Errorf("Can't connect to websocket (%d): %s\n", resp.StatusCode, buf.String())
 		}
-		return fmt.Errorf("Can't connect to websocket: %s\n", err.Error())
+		return resp, fmt.Errorf("Can't connect to websocket: %s\n", err.Error())
+	}
+
+	// Are we abusing the RoundTrip() abstraction?
```
rmohr: No. The RoundTripper is part of a transaction which can consist of multiple http requests. What you see here is only the request we care of. The most interesting scenario in virtcl here is to integrate with different authentication mechanisms where it very often looks like this:

1) go to authentication url X, obtain a token
2) go to acutal target url Y and send the token with it

The round-tripper allows us to only think about the request with the actual payload.

pkg/virtctl/vnc/vnc.go
rmohr: hm, misread the code before a little bit. So the connection in the SPY case is created in stream.Stream(), however the principle stays the same. In there,  they first create a connection with the roundtripper, and then start the stream outside of the round-tripper.

pkg/virtctl/vnc/vnc.go
```diff
@@ -69,75 +72,115 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
-	// The local tcp server is used to proxy the podExec websock connection to remote-viewer
-	ln, err := net.Listen("tcp", "127.0.0.1:0")
-	if err != nil {
-		return fmt.Errorf("Can't listen on unix socket: %s", err.Error())
-	}
-
-	port := ln.Addr().(*net.TCPAddr).Port
+	go func() {
+		// transfer data from/to the VM
+		k8ResChan <- vnc.Stream(kubecli.StreamOptions{
```
rmohr: The general flow and the interfaces look fine.

pkg/kubecli/vm.go
fromanirh: Got it, fixing.

rmohr:
ok to test

pkg/virtctl/console/console.go
```diff
@@ -94,8 +94,17 @@ func (c *Console) Run(cmd *cobra.Command, args []string) error {
...
 
 	go func() {
-		err := virtCli.VM(namespace).SerialConsole(vm, stdinReader, stdoutWriter)
-		resChan <- err
+		con, err := virtCli.VM(namespace).SerialConsole(vm)
+		if err != nil {
+			resChan <- err
+			return
+		}
+		defer con.Done()
```
rmohr: Looks to me like we can remove the `Done()` method from the interface and simply defer it in the `Stream()` call. I can't think of a situation where you would establish  the connection, check if there is no error, and in case that there is no error, not call `Stream()`. As far as I have seen in the k8s code that would also align with their implementation. What do you think?

rmohr:
Oh, and can you please also add a small unit tests which spans a small http test server where we write a few websocket messages back and forth? Something like https://stackoverflow.com/questions/47637308/create-unit-test-for-ws-in-golang.

Right now I am especially interested in the tests for pkg/kubecli/vm.go.

pkg/virtctl/console/console.go
fromanirh: I think there is no need in the code for Done() - Which I added trying to be forward-looking, so removing it.

fromanirh:
@rmohr sure, will add the unit test(s) in the next upload - hopefully later today. I will start from vm.go and expand from there as time/resource allow.

fromanirh:
Status update: Done() API removed, test in progress, but because my little experience with ginkgo, I'm not happy with it yet (and not sure it is good enough).

fromanirh:
status update: test about sending/receiving data still pending, others comments have been addressed.

fromanirh:
ping

rmohr:
retest this please

rmohr:
retest this please

rmohr:
retest this please

rmohr:
retest this please

PR #939, opened 2018-04-23
Pass labels from vm to pod to make service selectors work like expected

rmohr:
* Pass vm labels to pods, so that services, network policies and others can rely on vm labels, like they can on pod labels
* Add tests to verify that a service can really select a vm by its label and route the traffic properly
* As a side-effect fix netcat  based network tests which were broken (they always succeeded, no matter what happend)

### DIFF
--- a/tests/vm_networking_test.go
+++ b/tests/vm_networking_test.go
@@ -55,6 +55,8 @@ var _ = Describe("Networking", func() {
 	var inboundVM *v1.VirtualMachine
 	var outboundVM *v1.VirtualMachine
 
+	// newHelloWorldJob takes a dns entry or an IP which it will use create a pod
+	// which tries to contact the host on port 1500. It expects to receive "Hello World!" to succeed.
 	newHelloWorldJob := func(host string) *v12.Pod {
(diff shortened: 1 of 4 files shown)


rmohr:
@duyanyan FYI

rmohr:
@yuvalif FYI

rmohr:
retest this please

tests/vm_networking_test.go
```diff
@@ -83,23 +114,27 @@ var _ = Describe("Networking", func() {
 		go func() {
 			defer wg.Done()
 			defer GinkgoRecover()
-			inboundVM = createAndLogin()
+			inboundVM = createAndLogin(map[string]string{"expose": "me"})
 			expecter, _, err := tests.NewConsoleExpecter(virtClient, inboundVM, 10*time.Second)
 			defer expecter.Close()
 			Expect(err).ToNot(HaveOccurred())
-			_, err = expecter.ExpectBatch([]expect.Batcher{
+			resp, err := expecter.ExpectBatch([]expect.Batcher{
```
vladikr: +1

rmohr:
> I miss some By steps,
> like By("Creating the job pod ...)
> Except this lgtm

@cynepco3hahue tried to address that. Let me know if you want more.

rmohr:
> I did see the jobs fail many times before. However, it was mainly when there was completely no connectivity between pods/nodes.

@vladikr yes, I am not even sure if it was related to that, since it turned out that `nc` was not event installed on image.

PR #946, opened 2018-04-24
Speedup build

### LINKED ISSUE #924: [enhancement] add dependency modification checks to makefile
otherwise, it misses the point of a makefile :-)
some examples:
- no need to execute `make build` if no source file was changed
- if cluster is up, don't try to bring it up again

rmohr:
## build: golang 1.9.3 with docker

```bash
$ time make build
real	1m20.810s
user	0m0.652s
sys	0m0.335s
```
## build: golang 1.10 with docker

```bash
$ time make build
real	0m36.273s
user	0m0.610s
sys	0m0.353s
```

## build: golang 1.10 with docker + don't wait for rsyncd to terminate

```bash
$ time make build
real	0m26.492s
user	0m1.141s
sys	0m0.397s
```

## build: golang 1.10 without docker

```bash
$ time (./hack/check.sh && ./hack/build-go.sh install)
real	0m22.465s
user	0m33.317s
sys	0m4.898s
```

## tests: golang 1.10 check + build (go install build binaries which are not needed for unit tests) + test with docker

```bash
$ time make test
real	0m49.102s
user	0m1.194s
sys	0m0.437s
```

## tests: golang 1.10 check + test in docker (no go install)

```bash
$ time make test
real	0m19.583s
user	0m0.108s
sys	0m0.086s
```

## tests: golang 1.10 check + test without docker (no go install)

```bash
$ time (./hack/check.sh && ./hack/build-go.sh test)
real	0m6.139s
user	0m17.668s
sys	0m2.096s
```

## tests: golang  1.10 check + test with docker (no go install) + don't wait for rsync daemon to be stopped

```bash
$ time make test
real	0m9.888s
user	0m0.084s
sys	0m0.096s
```

## build: golang 1.10 with docker, only build virtcl for my platform

```bash
$ time make build
real	0m31.704s
user	0m0.909s
sys	0m0.330s
```

Not much difference, but it made a huge difference for golang 1.9.3

## Summary

I prepared a PR, which moves to golang 1.10 and does not run "go install" whe you do "make test" to run the unit tests and does not wait for rsyncd to be stopped. We still build virtctl with all client versions, the difference is not significant anymore with golang 1.10

The results for me are: "make build" is about 70% faster on my system, "make test" is about. When already switched to golang 1.10 and not running "go install" during "make test", the test run was about 60% faster compared to just switching to golang 1.10. Finally, our rsync daemon is bound to random ports, so we don't necessarily have to wait for it to be stopped. That improves the speed again. It cuts again 50% when taking all the other optimizations as a starting point.

Fixes #924 

@yuvalif @nertpinx FYI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ build:
 	hack/dockerized "./hack/check.sh && ./hack/build-go.sh install ${WHAT}"
 
 goveralls:
-	hack/dockerized "./hack/check.sh && ./hack/build-go.sh install && TRAVIS_JOB_ID=${TRAVIS_JOB_ID} TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST} TRAVIS_BRANCH=${TRAVIS_BRANCH} ./hack/goveralls.sh"
+	SYNC_OUT=false hack/dockerized "./hack/check.sh && TRAVIS_JOB_ID=${TRAVIS_JOB_ID} TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST} TRAVIS_BRANCH=${TRAVIS_BRANCH} ./hack/goveralls.sh"
 
 test:
 	SYNC_OUT=false hack/dockerized "./hack/check.sh && ./hack/build-go.sh test ${WHAT}"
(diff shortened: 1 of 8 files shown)


rmohr:
retest this please

PR #962, opened 2018-04-27
add expose command to virtctl

yuvalif:
manual tests are described here: https://github.com/yuvalif/kubevirt-test

some issues I'm aware of:

1. did not test the load-balancer option (currently not supported by the cluster in the dev environment)
~~2. test of node-port when target-port is defined is not working. this is not specific to the "expose" command, and fails when I just create a service~~ tested and working, my conf was wrong
~~3. unit tests are not doing too much (didn't do proper mocking)~~ added mocking
~~4. unit tests are not using the ginkgo framework~~ using ginkgo
5. ~~no integration tests added (except the manual ones in the repo above)~~
6. ~~failed to expose service for vm replica set, even manually~~ once i fixed the label issue, replica sets are working

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/962)


cluster/kubectl.sh
```diff
@@ -26,6 +26,10 @@ source ${KUBEVIRT_DIR}/hack/config.sh
 
 if [ "$1" == "console" ] || [ "$1" == "vnc" ]; then
     ${KUBEVIRT_DIR}/_out/cmd/virtctl/virtctl --kubeconfig=${kubeconfig} "$@"
+elif [ "$1" == "start" ] || [ "$1" == "stop" ]; then
```
rmohr: That is not needed anymore. You can simply us `cluster/virtctl.sh`. The other ifs are only there for backward-compatibility. Will add it to the docs too on #910.

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,186 @@
...
+	COMMAND_EXPOSE = "expose"
+)
+
+type Command struct {
+	clientConfig clientcmd.ClientConfig
+	command      string
+}
+
+// holding flag information
+var ServiceName string
```
rmohr: could you make them starting with lowerCase, since they are only supposed to be used in this module?

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	// get the client
+	virtClient, err := kubecli.GetKubevirtClientFromClientConfig(o.clientConfig)
+	if err != nil {
+		return fmt.Errorf("Cannot obtain KubeVirt client: %v", err)
+	}
+
+	// TODO: not sure what are the options used for?
+	options := k8smetav1.GetOptions{}
```
rmohr: you can use them to e.g. do a fast non-quorum read and other things. Leaving them empty does a plain quorum read from the apiserver, that is perfect.

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,186 @@
...
+	// get the VM
+	_, vm_err := virtClient.VM(namespace).Get(vmName, options)
+	if vm_err != nil {
+		// try to get offline VM this is only for better error message
+		_, ovm_err := virtClient.OfflineVirtualMachine(namespace).Get(vmName, &options)
+		if ovm_err != nil {
+			return fmt.Errorf("Error fetching VirtualMachine: %v", vm_err)
+		}
+		return fmt.Errorf("OfflineVirtualMachine: %s must be started before being exposed", vmName)
```
rmohr: To be declarative, we should not assume that it has to be started. We basically define: "if vm is started start routing traffic through this service", when looking at it from the offline virtual machine perspective.

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// get the client
+	virtClient, err := kubecli.GetKubevirtClientFromClientConfig(o.clientConfig)
+	if err != nil {
+		return fmt.Errorf("Cannot obtain KubeVirt client: %v", err)
+	}
+
+	// TODO: not sure what are the options used for?
+	options := k8smetav1.GetOptions{}
+	// get the VM
```
rmohr: Instead of this fallback logic, what we will need, is another argument:

```
virtctl expose vm (vm) # should be created if the vm exists (no fallback to ovm in any way necessary)
virtctl expose ovm (ovm) # should be creaste if the ovm exists (labels can be obtained from spec.template.metadata.labels). No check if the vm exists or not is required.
```

So we need an extra argument which can have the values `vm,virtualmachine,vms,virtualmachines` or `ovm,ovms,offlinevirtualmachine,offlinevirtualmachines`. A third one would be `vmrs,vmrss,virtualmachinereplicasets,virtualmachinereplicaset`.

```
virtctl expose vmrs (vmrs) # should be creaste if the vm replica set exists (labels can be obtained from spec.template.metadata.labels). No check if vms exist are required.
```

cluster/kubectl.sh
yuvalif: will revert back

pkg/virtctl/expose/expose.go
yuvalif: sure

pkg/virtctl/expose/expose.go
yuvalif: ok

pkg/virtctl/expose/expose.go
yuvalif: ok. will remove (also test this scenario)

pkg/virtctl/expose/expose.go
yuvalif: regarding replica sets:
1. your proposal is to use: ```vmrs.Spec.Template.ObjectMeta.Labels``` (when vmrs is the VM replica set object). Why not use: ```vmrs.Spec.Selector.MatchLabels``` ?
2. using replica set in: cluster/example/vm-replicaset-cirros.yaml there are 3 nodes being created.  To which ones of them does the cluster-ip belongs?

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,213 @@
...
+	// actually create the service
+	service := &v1.Service{
+		ObjectMeta: k8smetav1.ObjectMeta{
+			Name: serviceName,
+		},
+		Spec: v1.ServiceSpec{
+			Ports: []v1.ServicePort{
+				{Name: portName, Protocol: protocol, Port: port, TargetPort: targetPort, NodePort: nodePort},
+			},
+			Selector:       map[string]string{"kubevirt.io/domain": vmName},
```
rmohr: Checked the `kubectl expose` code. I think we should align and  not allowing exposing the object if not labels are present at all. I was not sure about how to behave in this case before, but I would align with the behaviour they have, to avoid confusions.

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+	case "vmrs", "vmrss", "virtualmachinereplicaset", "virtualmachinereplicasets":
+		// get the VM replica set
+		_, err := virtClient.ReplicaSet(namespace).Get(vmName, options)
+		if err != nil {
+			return fmt.Errorf("Error fetching VirtualMachine ReplicaSet: %v", err)
+		}
+		// in case of replica set we take the label from the replica set template
+        // same as the original vmName
+		//vmName = vmrs.Spec.Template.ObjectMeta.Labels["kubevirt.io/vmReplicaSet"]
```
rmohr: @yuvalif you had a good point regarding to use `spec.selector.matchLabels` instead of `spec.template.objectmeta.labels`. Let's do that. Kubectl does the same thing on replicasets. Also like kubectl, we should not allow exposing anything if `spec.selector.matchExpressions` is used.

pkg/virtctl/expose/expose_test.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func Test_Run(t *testing.T) {
+	var flags pflag.FlagSet
+	clientConfig := kubecli.DefaultClientConfig(&flags)
+	cmd := NewExposeCommand(clientConfig)
+	if cmd == nil {
+		t.Error("'expose' command creation failure")
+	}
+	// TODO: mock the client to not communicate with the server
```
rmohr: We have a mock-client which you can use for that purpose at https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-controller/watch/vm_test.go#L145. It shows on how to set up our mock client and the fake client for k8s resources.

rmohr: Regarding functional tests, at https://github.com/kubevirt/kubevirt/blob/master/tests/ovm_test.go#L431 is an example on how one may test virtctl subcommands on a real cluster.

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,213 @@
...
+		Example: usage(),
+		Args:    cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := Command{command: COMMAND_EXPOSE, clientConfig: clientConfig}
+			return c.RunE(cmd, args)
+		},
+	}
+
+	// flags for the "expose" command
+	cmd.Flags().StringVar(&serviceName, "name", "", "Name of the service created for the exposure of the VM")
```
phoracek: Could you distinguish this from positional arg NAME? Maybe service-name. Or change NAME to MACHINE_NAME

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,213 @@
...
+func usage() string {
+	usage := "  # Expose SSH to a virtual machine called 'myvm' as a node port (5555) of the cluster:\n"
+	usage += fmt.Sprintf("  virtctl expose vm myvm --port=5555 --target-port=22 --name=myvm-ssh --type=NodePort")
+	return usage
+}
+
+// executing the "expose" command
+func (o *Command) RunE(cmd *cobra.Command, args []string) error {
+	// first argument is type of VM: VM, offline VM or replica set VM
+	vmType := strings.ToLower(args[0])
```
phoracek: Can you use cobra's positional arguments instead?

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,213 @@
...
+	var protocol v1.Protocol
+	var targetPort intstr.IntOrString
+	var serviceType v1.ServiceType
+
+	// convert from integer to the IntOrString type
+	targetPort = intstr.FromInt(intTargetPort)
+
+	// convert from string to the protocol enum
+	switch strProtocol {
+	case "TCP":
```
phoracek: What about lower case?

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,213 @@
...
+		protocol = v1.ProtocolTCP
+	case "UDP":
+		protocol = v1.ProtocolUDP
+	default:
+		return fmt.Errorf("Unknown protocol: %s", strProtocol)
```
phoracek: "Error strings should not start with a capital letter because they'll often be prefixed before printing" https://github.com/golang/go/wiki/Errors

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,213 @@
...
+	cmd.Flags().Int32Var(&nodePort, "node-port", 0, "Port used to expose the service on each node in a cluster.")
+	cmd.Flags().StringVar(&strServiceType, "type", "ClusterIP", "Type for this service: ClusterIP, NodePort, or LoadBalancer.")
+	cmd.Flags().StringVar(&portName, "port-name", "", "Name of the port. Optional.")
+	cmd.SetUsageTemplate(templates.UsageTemplate())
+
+	return cmd
+}
+
+func usage() string {
+	usage := "  # Expose SSH to a virtual machine called 'myvm' as a node port (5555) of the cluster:\n"
```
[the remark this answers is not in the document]
rmohr: You can also use a multiline string:

```golang
usage := 
`
multiple lines
work this way
`
```

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,213 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := Command{command: COMMAND_EXPOSE, clientConfig: clientConfig}
+			return c.RunE(cmd, args)
+		},
+	}
+
+	// flags for the "expose" command
+	cmd.Flags().StringVar(&serviceName, "name", "", "Name of the service created for the exposure of the VM")
+	cmd.MarkFlagRequired("name")
+	cmd.Flags().StringVar(&clusterIP, "cluster-ip", "", "ClusterIP to be assigned to the service. Leave empty to auto-allocate, or set to 'None' to create a headless service.")
```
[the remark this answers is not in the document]
yuvalif: anything that has a default value, is, by-definition "Optional", this is why I left it out. also left it out in cases that the case the text suggest it is optional.
but, I can add in all cases if you feel that this is more clear.

pkg/virtctl/expose/expose.go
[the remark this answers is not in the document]
yuvalif: just copy&paste error. will fix

pkg/virtctl/expose/expose.go
yuvalif: kubectl is case sensitive:
```spec.ports[0].protocol: Unsupported value: "tcp": supported values: "TCP", "UDP"```

File: pkg/virtctl/expose/expose.go (hunk 10 above)
Comment (yuvalif) [reply pkg/virtctl/expose/expose.go]: maybe I'm missing something, but according to this: https://github.com/spf13/cobra#positional-and-custom-arguments I think I'm using positional arguments (calling: ```cobra.ExactArgs(2)```)?

File: pkg/virtctl/expose/expose.go (hunk 9 above)
Comment (yuvalif) [reply pkg/virtctl/expose/expose.go]: i tried to follow "kubectl expose":
```
kubectl expose (-f FILENAME | TYPE NAME) [--port=port] [--protocol=TCP|UDP] [--target-port=number-or-name] [--name=name] [--external-ip=external-ip-of-service] [--type=type] [options]
```
they use "NAME" and "--name"

File: pkg/kubecli/fake_kubevirt.go [hunk 15]
```diff
@@ -0,0 +1,114 @@
...
+// parameter to control behavior of fake objects
+var InvalidFakeClient bool
+var InvalidFakeResource bool
+var InvalidFakeService bool
+
+// holding a copy of the created service, so that the test code could verify
+var CurrentFakeService *k8sapiv1.Service
+
+// entry point for testing, could be used to override GetKubevirtClientFromClientConfig
+func GetFakeKubevirtClientFromClientConfig(cmdConfig clientcmd.ClientConfig) (KubevirtClient, error) {
```
rmohr: Why can't you use our mock client?

pkg/virtctl/expose/expose.go
rmohr: Yes. Arguably we should in all our commands use custom validators, to not just return the generic error message.

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,211 @@
...
+		protocol = v1.ProtocolTCP
+	case "UDP":
+		protocol = v1.ProtocolUDP
+	default:
+		return fmt.Errorf("unknown protocol: %s", strProtocol)
+	}
+
+	// convert from string to the service type enum
+	switch strServiceType {
+	case "ClusterIP":
```
rmohr: These are also case-sensitive in kubectl?

yuvalif:
in the latest push:
* add ginkgo UTs and mocking above client level
* piggybacked to this PR the unification of the passing of "options" between OVMs and the resr

pkg/virtctl/expose/expose.go
rmohr: still needs to be addressed.

pkg/virtctl/expose/expose.go
rmohr: then let's do the same like kubectl.

pkg/virtctl/expose/expose.go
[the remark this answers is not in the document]
rmohr: then let's do the same like kubectl.

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,211 @@
...
+		return fmt.Errorf("cannot obtain KubeVirt client: %v", err)
+	}
+
+	// does a plain quorum read from the apiserver
+	options := k8smetav1.GetOptions{}
+
+	switch vmType {
+	case "vm", "vms", "virtualmachine", "virtualmachines":
+		// get the VM
+		_, err := virtClient.VM(namespace).Get(vmName, options)
```
rmohr: In this case you have to use `spec.Labels` from the vm.

Also we should probably ignore all `kubevirt.io` labels. They are used for internal orchestration only now  and the expose result would differ from an expose on a pod.

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,211 @@
...
+		if err != nil {
+			return fmt.Errorf("error fetching VirtualMachine: %v", err)
+		}
+	case "ovm", "ovms", "offlinevirtualmachine", "offlinevirtualmachines":
+		// get the offline VM
+		_, err := virtClient.OfflineVirtualMachine(namespace).Get(vmName, options)
```
rmohr: In this case you have to use `spec.template.Labels` from the ovm.

pkg/virtctl/expose/expose.go
```diff
@@ -0,0 +1,211 @@
...
+	case "vmrs", "vmrss", "virtualmachinereplicaset", "virtualmachinereplicasets":
+		// get the VM replica set
+		vmrs, err := virtClient.ReplicaSet(namespace).Get(vmName, options)
+		if err != nil {
+			return fmt.Errorf("error fetching VirtualMachine ReplicaSet: %v", err)
+		}
+		if len(vmrs.Spec.Selector.MatchExpressions) > 0 {
+			return fmt.Errorf("cannot expose VirtualMachine ReplicaSet with match expressions")
+		}
+		vmName = vmrs.Spec.Selector.MatchLabels["kubevirt.io/vmReplicaSet"]
```
rmohr: In this case you have to use `spec.template.Labels` from the replicaset.

pkg/virtctl/expose/expose.go
rmohr: Hm, thinking again about it, we have already a special label `kubevirt.i/os` which we definitely do not want to include.

pkg/virtctl/expose/expose.go
yuvalif: so, we should probably keep it like this, unless we decide to re-do all commands?

pkg/kubecli/fake_kubevirt.go
yuvalif: because I want to more easily test all branches of my code, without running that via a mock http server.
doing that at the mock http server level would require much more complexity in the test code. note that other commands either dont have any unit-tests, or have unit tests that does not cover anything. 
this will allow adding good coverage to all commands

rmohr: The mock client does not involve a http server. Here an example on how it works:


```
		virtClient = kubecli.NewMockKubevirtClient(ctrl)
		vmInterface = kubecli.NewMockVMInterface(ctrl)
		virtClient.EXPECT().VM(k8sv1.NamespaceDefault).Return(vmInterface).AnyTimes()
                vmInterface.EXPECT().Update(myupdatedvm)
```

It uses gomock.

pkg/virtctl/expose/expose.go
yuvalif: yes. tested them

pkg/virtctl/expose/expose.go
yuvalif: if I try to follow kubectl. in: ```kubernetes/pkg/kubectl/cmd/util/factory_client_access.go``` in: ```MapBasedSelectorForObject()``` they use: ```Spec.Selector.MatchLabels``` for the replica set

rmohr: yes you are correct. `spec.selector.matchLabels` is the right thing. Thx.

pkg/kubecli/fake_kubevirt.go
yuvalif: this is based on a generated file pkg/kubecli/generated_mock_kubevirt.go
but if you look at my mock code at pkg/kubecli/fake_kubevirt.go I also need a mock for the ServiceInterface. how do i add that to the generated file?

pkg/virtctl/expose/expose.go
```diff
@@ -162,6 +162,8 @@ func (o *Command) RunE(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("error fetching VirtualMachine: %v", err)
 		}
 		serviceSelector = vm.ObjectMeta.Labels
+		// remove unwanted labels
+		delete(serviceSelector, "kubevirt.io/nodeName")
```
yuvalif: if the nodeName is added to the service selector, it fails in finding the pod

cluster/examples/vm-ephemeral.yaml
```diff
@@ -20,5 +22,5 @@ spec:
   volumes:
   - name: registryvolume
     registryDisk:
-      image: kubevirt/cirros-registry-disk-demo:devel
+        image: registry:5000/kubevirt/cirros-registry-disk-demo:devel
```
yuvalif: this was failing with:
```
Failed to pull image "kubevirt/cirros-registry-disk-demo:devel": rpc error: code = Unknown desc = Tag devel not found in repository docker.io/kubevirt/cirros-registry-disk-demo
```

cluster/examples/vm-ephemeral.yaml
```diff
@@ -3,6 +3,8 @@ kind: VirtualMachine
 metadata:
   creationTimestamp: null
   name: vm-ephemeral
+  labels:
+    kubevirt.io/vm: vm-ephemeral
```
yuvalif: with the new logic, VMs that don't have a label will not be exposed. therefore a label was added to the samples

pkg/virtctl/expose/expose.go
```diff
@@ -189,7 +196,7 @@ func (o *Command) RunE(cmd *cobra.Command, args []string) error {
 			Ports: []v1.ServicePort{
 				{Name: portName, Protocol: protocol, Port: port, TargetPort: targetPort, NodePort: nodePort},
 			},
-			Selector:       map[string]string{"kubevirt.io/domain": vmName},
+			Selector:       serviceSelector,
```
yuvalif: service selector is taken from VM, OVM or VMRS, not the hardcoded pod label

pkg/kubecli/fake_kubevirt.go
rmohr: You can use the client-go fake client for k8s objects. See for instance here: https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-controller/watch/node_test.go#L76.

pkg/virtctl/expose/expose.go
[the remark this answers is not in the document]
yuvalif: in kubectl, the only one that is explicitly marked as "Optional" is `target-port`, all the other optional flags are implicit. Will keep as is

yuvalif:
regarding the conflict in `tests/ovm_test.go`. the only thing i changed there was to use `options` instead of `&options` in the API calls. I think that you can just use the master, as in the latest version these APIs are not used

tools/vms-generator/vms-generator.go
```diff
@@ -96,6 +96,7 @@ func getBaseVm(name string) *v1.VirtualMachine {
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
+			Labels: map[string]string{"special": "key"},
```
yuvalif: this is needed for service exposure to work. following: https://github.com/kubevirt/user-guide/blob/master/workloads/virtual-machines/expose-service.md

yuvalif:
still open:
* integration tests
* use standard mocking in unit-tests
* add label per vm

yuvalif:
*[cluster/examples/vm-ephemeral.yaml, line 6 at r2](https://reviewable.io/reviews/kubevirt/kubevirt/962#-LCicExygU98rczYhyWD-r2-6:-LDBjid04vUiLuG97q1H:bb9rzuz) ([raw file](https://github.com/kubevirt/kubevirt/blob/3e5c2f29c2515dd5b91f6428caa287af574f32ef/cluster/examples/vm-ephemeral.yaml#L6)):*
<details><summary><i>Previously, rmohr (Roman Mohr) wrote…</i></summary><blockquote>

If you use `special: key` an all example vms, it will probably be a surprising result if people have more than one example deployed on the cluster. The service would start to load-balance between the vms. Could you use unique labels per example vm?
</blockquote></details>

any issue with: "special:" and then the VM name as the value?

---


*Comments from [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/962)*

rmohr:
Review status: 0 of 23 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[cluster/examples/vm-ephemeral.yaml, line 6 at r2](https://reviewable.io/reviews/kubevirt/kubevirt/962#-LCicExygU98rczYhyWD-r2-6:-LDBnptQDtJ3_9J1Dh7U:b-66xhdl) ([raw file](https://github.com/kubevirt/kubevirt/blob/3e5c2f29c2515dd5b91f6428caa287af574f32ef/cluster/examples/vm-ephemeral.yaml#L6)):*
<details><summary><i>Previously, yuvalif (Yuval Lifshitz) wrote…</i></summary><blockquote>

any issue with: "special:" and then the VM name as the value?
</blockquote></details>

No, works for me.

---


*Comments from [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/962)*

tests/expose_test.go
```diff
@@ -0,0 +1,384 @@
...
+	const testPort = 1500
+
+	Context("Expose service on a VM", func() {
+		By("Creating a VM object")
+		vm := tests.NewRandomVMWithEphemeralDiskAndUserdata(tests.RegistryDiskFor(tests.RegistryDiskCirros), "#!/bin/bash\necho 'hello'\n")
+		vm.Labels = map[string]string{"expose": "vm"}
+
+		It("Should start a VM and run a 'hello world' server on it", func() {
+			By("Starting the VM")
+			vm, err = virtClient.VM(tests.NamespaceTestDefault).Create(vm)
```
rmohr: Hm, that only works by accident. You can't assume a specific execution order for `It`. You will have to use the same pattern like in the networking tests, with the `BeforeAll` and `AfterAll`. Basically make a `BeforeAll` out of this.

tests/expose_test.go
```diff
@@ -0,0 +1,384 @@
...
+			})
+		})
+	})
+
+	Context("Expose UDP service on a VM", func() {
+		By("Creating a VM object")
+		vm := tests.NewRandomVMWithEphemeralDiskAndUserdata(tests.RegistryDiskFor(tests.RegistryDiskCirros), "#!/bin/bash\necho 'hello'\n")
+		vm.Labels = map[string]string{"expose": "udp-vm"}
+
+		It("Should start a VM and run a 'hello world' UDP server on it", func() {
```
rmohr: Same like above make a `BeforeAll` out of this.

pkg/kubecli/fake_kubevirt.go
rmohr: Let's not add a third technology for a fake client. I would prefer if you use the k8s fake client for k8s resources.

tests/expose_test.go
yuvalif: ok. will fix

tests/expose_test.go
yuvalif: will fix

tests/expose_test.go
```diff
@@ -0,0 +1,384 @@
...
+			By("Checking the number of ready replicas")
+			Eventually(func() int {
+				rs, err := virtClient.ReplicaSet(tests.NamespaceTestDefault).Get(vmrs.ObjectMeta.Name, k8smetav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				return int(rs.Status.ReadyReplicas)
+			}, 60*time.Second, 1*time.Second).Should(Equal(numberOfVMs))
+		})
+
+		It("Should add an 'hello world' server on each VM in the replica set", func() {
+			// TODO: add label to list options
```
yuvalif: @rmohr I need help here. what can i pass into the ListOption object so it would find only the VMs that belong to the replica set?
would like to get rid of line 243

tests/utils.go
```diff
@@ -91,6 +92,8 @@ const (
 
 const SubresourceTestLabel = "subresource-access-test-pod"
 
+const NetworkAccessTestLabel = "network-access-test-pod"
```
yuvalif: opps. should remove

tests/expose_test.go
```diff
@@ -0,0 +1,384 @@
...
+				&expect.BSnd{S: "echo $?\n"},
+				&expect.BExp{R: "0"},
+			}, 60*time.Second)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		Context("Expose ClusterIP service", func() {
+			const servicePort = "27017"
+			const serviceName = "cluster-ip-vm"
+			It("Should expose a Cluster IP service on a VM", func() {
```
yuvalif: seems like i have the same issue here. if the test in line 75 happens before this, then it would fail. will merge to one "It" section

pkg/kubecli/fake_kubevirt.go
yuvalif: will remove this, and use existing one

pkg/kubecli/ovm_test.go
```diff
@@ -129,11 +127,3 @@ var _ = Describe("Kubevirt OfflineVirtualMachine Client", func() {
 		server.Close()
 	})
 })
-
-func NewMinimalOVM(name string) *v1.OfflineVirtualMachine {
```
yuvalif: moved to new kubevirt_test_utils.go file

pkg/kubecli/replicaset_test.go
```diff
@@ -129,11 +127,3 @@ var _ = Describe("Kubevirt VirtualMachineReplicaSet Client", func() {
 		server.Close()
 	})
 })
-
-func NewVMReplicaSetList(rss ...v1.VirtualMachineReplicaSet) *v1.VirtualMachineReplicaSetList {
```
yuvalif: moved to new kubevirt_test_utils.go file

pkg/kubecli/vm_test.go
```diff
@@ -129,7 +129,3 @@ var _ = Describe("Kubevirt VM Client", func() {
 		server.Close()
 	})
 })
-
-func NewVMList(vms ...v1.VirtualMachine) *v1.VirtualMachineList {
```
yuvalif: moved to new kubevirt_test_utils.go file just for completeness :-)

pkg/kubecli/kubevirt_test_utils.go
```diff
@@ -0,0 +1,54 @@
...
+// (4) rest of the kubevirt mocking is automatically generated in generated_mock_kubevirt.go
+
+// InvalidMockClient is a parameter to control behavior of mock objects
+var InvalidMockClient bool
+
+// MockKubevirtClientInstance is a reference to the kubevirt client that could be manipulated by the test code
+var MockKubevirtClientInstance *MockKubevirtClient
+
+// GetMockKubevirtClientFromClientConfig is an entry point for testing, could be used to override GetKubevirtClientFromClientConfig
+func GetMockKubevirtClientFromClientConfig(cmdConfig clientcmd.ClientConfig) (KubevirtClient, error) {
```
rmohr: Could you just create another function for the error case?

pkg/virtctl/expose/expose_test.go
```diff
@@ -0,0 +1,207 @@
...
+				Expect(cmd).ToNot(BeNil())
+			})
+		})
+	})
+
+	Describe("Run an 'expose' command", func() {
+		Context("When client has an error", func() {
+			BeforeEach(func() {
+				// only in this test the client call should fail
+				kubecli.InvalidMockClient = true
```
rmohr: We should get rid of this global side-effect.

rmohr:
ok to test

pkg/kubecli/kubevirt_test_utils.go
yuvalif: ok. yes, better than the global var

yuvalif: fixed

tests/expose_test.go
```diff
@@ -39,9 +39,10 @@ func generateHelloWorldServer(vm *v1.VirtualMachine, virtClient kubecli.Kubevirt
 	defer expecter.Close()
 	Expect(err).ToNot(HaveOccurred())
 
-	serverCommand := fmt.Sprintf("screen -d -m nc -klp %d -e echo -e \"Hello World!\"\n", testPort)
+	serverCommand := fmt.Sprintf("screen -d -m nc -klp %d -e echo -e 'Hello World!'\n", testPort)
 	if protocol == "udp" {
-		serverCommand = fmt.Sprintf("screen -d -m nc -uklp %d -e echo -e \"Hello UDP World!\"\n", testPort)
+		// nc has to be in a while loop in case of UDP, since it exists after one message
+		serverCommand = fmt.Sprintf("screen -d -m sh -c \"while true\n do nc -uklp %d -e echo -e 'Hello UDP World!'\ndone\n\"\n", testPort)
```
yuvalif: the udp server exited after one message (in-spite of the -lk flags), so, to support multiple node env, added it inside a "while" loop

yuvalif:
retest this please

rmohr:
ok to test

rmohr:
retest this please

rmohr:
retest this please

tests/expose_test.go
```diff
@@ -0,0 +1,364 @@
...
+			By("Start the replica set")
+			vmrs, err = virtClient.ReplicaSet(tests.NamespaceTestDefault).Create(vmrs)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Checking the number of ready replicas")
+			Eventually(func() int {
+				rs, err := virtClient.ReplicaSet(tests.NamespaceTestDefault).Get(vmrs.ObjectMeta.Name, k8smetav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				return int(rs.Status.ReadyReplicas)
+			}, 60*time.Second, 1*time.Second).Should(Equal(numberOfVMs))
```
rmohr: Might be worth 120 seconds:

```
• Failure in Spec Setup (BeforeEach) [60.094 seconds]
Expose
/root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:58
  Expose service on a VM replica set
  /root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:232
    Expose ClusterIP service [BeforeEach]
    /root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:265
      Should create a ClusterIP service on VMRS and connect to it
      /root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:269

      Timed out after 60.000s.
      Expected
          <int>: 1
      to equal
          <int>: 2

      /root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:250
```

tests/expose_test.go
rmohr: The approach you have looks good. Why do you want the vms here?

Apart from that, you can pass in label selectors:

```golang
v12.ListOptions{LabelSelector: "key=val"})
```
This example would match all vms with the label `key` and value `val`.

yuvalif: I need the VMs because I want to run the "hello world" server on them.
I tried providing the label selector as you suggested (with the label I set into the VMRS):
```v12.ListOptions{LabelSelector: "expose=vmrs"})```
However, the list return empty.
When I print the labels of the VMs in the VMRP it seems like the label that I set on the VMRS is not inherited by them, both ```vm.Labels``` and ```vm.ObjectMeta.Labels``` has the following value:
```map[kubevirt.io/nodeName:node01 name:replicaset99mjs]```
Is this by-design, or an issue?
anyway, will keep the tes tcode as-is, since I'm testing something else here.

rmohr: That is by design. The same like with every k8s controller. Controllers have their own labels which are distinct from the objects in their template section. You will have to add your `expose=vmrs` label on the label section on the vm, to also have it there.

PR #967, opened 2018-04-27
[WIP] Add selinux check

petrkotas:
This PR add selinux check after the functest is run.
This is preliminary implementation that requires opinions on what is the best way to implement this.

Currently it is hard coded to vagrant. Is this fine?

@fabiand @stu-gott opinions?

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/967)

### DIFF
--- a/hack/functests.sh
+++ b/hack/functests.sh
@@ -19,9 +19,36 @@
 
 set -e
 
-source hack/common.sh
-source hack/config.sh
+ source hack/common.sh
+ source hack/config.sh
 
 functest_docker_prefix=${manifest_docker_prefix-${docker_prefix}}
 
+START_TIME=$(vagrant ssh master -c 'date +"%x %T"' -- -T)
 ${TESTS_OUT_DIR}/tests.test -kubeconfig=${kubeconfig} -tag=${docker_tag} -prefix=${functest_docker_prefix} -kubectl-path=${kubectl} -test.timeout 60m ${FUNC_TEST_ARGS}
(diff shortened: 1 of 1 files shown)


rmohr:
> Currently it is hard coded to vagrant. Is this fine?

We are going to remove vagrant: #910 ;)

I think that this task is a great opportunity to improve our log- and event-collection during test runs.
What I would love to see, is a `rsyslog` container which runs outside of our cluster, spawned by our client in https://github.com/kubevirt/kubevirtci, like it also does for e.g. the nfs server (https://github.com/kubevirt/kubevirtci/blob/master/gocli/cmd/run.go#L215). We can simply deploy this: https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/fluentd-daemonset-syslog.yaml and let it point to our rsyslog server.


Flow:
 1. If gocli is started with --logs-dir, we start a rsyslog daemon which writes everything to --logs-dir
 2. We deploy fluentd as part of our testing infra
 3. We run the tests
 4. We parse the logs for any artifacts which are interesting for us, e.g. selinux denials, which are also collected, since the go to /var/log/audit.log (or something like that).

fabiand:
@petrkotas Hey is there any update on this one - or a PR somewhere else?

petrkotas:
@fabiand actually the pr went kubevirtci/logging after this one is done, we can continue.

rmohr:
> @fabiand actually the pr went kubevirtci/logging after this one is done, we can continue.

Here: https://github.com/kubevirt/kubevirtci/pull/9

fabiand:
:+1:

PR #978, opened 2018-05-01
DO NOT MERGE: test dhcp activity

vladikr:

### DIFF
--- a/tests/vm_networking_test.go
+++ b/tests/vm_networking_test.go
@@ -105,6 +105,18 @@ var _ = Describe("Networking", func() {
 			vm, err = virtClient.VM(tests.NamespaceTestDefault).Get(vm.ObjectMeta.Name, v13.GetOptions{})
 			Expect(err).ToNot(HaveOccurred())
 
+			dhcpAddr := vm.Status.Interfaces[0].IP
+
+			expecterDhcp, _, err := tests.NewConsoleExpecter(virtClient, vm, 10*time.Second)
+			defer expecterDhcp.Close()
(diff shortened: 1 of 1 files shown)


vladikr:
retest this please

vladikr:
retest this please

tests/vm_networking_test.go
```diff
@@ -105,6 +105,18 @@ var _ = Describe("Networking", func() {
 			vm, err = virtClient.VM(tests.NamespaceTestDefault).Get(vm.ObjectMeta.Name, v13.GetOptions{})
 			Expect(err).ToNot(HaveOccurred())
 
+			dhcpAddr := vm.Status.Interfaces[0].IP
+
+			expecterDhcp, _, err := tests.NewConsoleExpecter(virtClient, vm, 10*time.Second)
+			defer expecterDhcp.Close()
+			Expect(err).ToNot(HaveOccurred())
+			resp1, err := expecterDhcp.ExpectBatch([]expect.Batcher{
+				&expect.BExp{R: "Sending discover..."},
```
rmohr: At the end  the vms always booted, so I don't think that his helps you. What I would love to see, is that we log and timestamp every line comming from the console. We have access to everything coming from the vm, before it passed to the Expecter here: https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L951. I would simply log every line coming in and then move it forward. If the test fails, the logs will get printed.

rmohr:
Closing this since we found the issue for #936

PR #980, opened 2018-05-02
Add test for creating VM from template via oc CLI

gszasz:
# Test for creating and running VM from a template via external oc command

## Goal

Provide test automation covering creation and execution of a VM created from an *OpenShift Template*. The test has to use the external `oc` command.

## Test cases covered

1. Generate VM JSON file from a Template via `oc-process` command.
2. Create VM from a template via `oc-create` command.
3. Spawn VMI of the VM via `oc-patch` command.

### DIFF
--- a/tests/oc_template_test.go
+++ b/tests/oc_template_test.go
@@ -1,0 +1,300 @@
+/*
+ * This file is part of the KubeVirt project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2018 Red Hat, Inc.
+ *
+ */
+
+package tests_test
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"kubevirt.io/kubevirt/pkg/api/v1"
+	"kubevirt.io/kubevirt/tests"
+)
+
+// template parameters
+const (
+	vmName     = "testvm"
+	vmCpuCores = "2"
+)
+
+const (
+	vmStartPatch = "{\"spec\":{\"running\":true}}"
+	vmStopPatch  = "{\"spec\":{\"running\":false}}"
+)
+
+var _ = Describe("VM Template", func() {
+	flag.Parse()
+
+	vmTemplateFedora := getVmTemplateFedora()
+
+	generateVmJsonFromTemplate := func(vmName string, vmTemplateJsonFile string) string {
+		By("Converting VirtualMachine Template into JSON file via oc-process command")
+		vmTemplateParams := []string{"-p", "NAME=" + vmName, "-p", "CPU_CORES=" + vmCpuCores}
+		args := append([]string{"process", "-f", vmTemplateJsonFile}, vmTemplateParams...)
+		out, err := tests.RunOcCommand(args...)
+		Expect(err).ToNot(HaveOccurred())
+		vmJsonFile, err := tests.WriteJson(vmName, out)
+		Expect(err).ToNot(HaveOccurred())
+		return vmJsonFile
+	}
+
+	createVmFromJson := func(vmName string, vmJsonFile string) {
+		var message = ""
+		By("Creating VirtualMachine from JSON file via oc-create command")
+		out, err := tests.RunOcCommand("create", "-f", vmJsonFile)
+		Expect(err).ToNot(HaveOccurred())
+		message = fmt.Sprintf("virtualmachine.kubevirt.io \"%s\" created\n", vmName)
+		Expect(out).To(Equal(message))
+
+		Eventually(func() bool {
+			out, err := tests.RunOcCommand("get", "vms")
+			Expect(err).ToNot(HaveOccurred())
+			return strings.Contains(out, vmName)
+		}, time.Duration(60)*time.Second).Should(BeTrue(), "Timed out waiting for vm to appear")
+	}
+
+	patchVm := func(vmName string, patch string) {
+		var message = ""
+		By("Starting VirtualMachine via oc-patch command")
+		out, err := tests.RunOcCommand("patch", "virtualmachine", vmName, "--type", "merge", "-p", patch)
+		Expect(err).ToNot(HaveOccurred())
+		message = fmt.Sprintf("virtualmachine.kubevirt.io \"%s\" patched\n", vmName)
+		Expect(out).To(Equal(message))
+	}
+
+	deleteVm := func(vmName string) {
+		var message = ""
+		By("Deleting the VirtualMachine via oc-delete command")
+		out, err := tests.RunOcCommand("delete", "vm", vmName)
+		Expect(err).ToNot(HaveOccurred())
+		message = fmt.Sprintf("virtualmachine.kubevirt.io \"%s\" deleted\n", vmName)
+		Expect(out).To(Equal(message))
+
+		Eventually(func() bool {
+			out, err := tests.RunOcCommand("get", "vms")
+			Expect(err).ToNot(HaveOccurred())
+			return out == "No resources found.\n"
+		}, time.Duration(60)*time.Second).Should(BeTrue(), "Timed out waiting for vm to disappear")
+	}
+
+	Context("with oc command", func() {
+		vmTemplateJsonFile := ""
+		vmJsonFile := ""
+
+		BeforeEach(func() {
+			tests.SkipIfNoOc()
+			// write testing VirtualMachine Template into JSON file
+			var err error
+			vmTemplateJsonFile, err = tests.GenerateVmTemplateJson(vmTemplateFedora)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		AfterEach(func() {
+			if vmTemplateJsonFile != "" {
+				// remove testing VirtualMachine Template JSON file
+				err := os.Remove(vmTemplateJsonFile)
+				Expect(err).ToNot(HaveOccurred())
+				vmTemplateJsonFile = ""
+			}
+			if vmJsonFile != "" {
+				// remove testing VirtualMachine JSON file
+				err := os.Remove(vmJsonFile)
+				Expect(err).ToNot(HaveOccurred())
+				vmJsonFile = ""
+			}
+		})
+
+		It("should generate a vm JSON from a template", func() {
+			generateVmJsonFromTemplate(vmName, vmTemplateJsonFile)
+		})
+
+		It("should create a vm from template", func() {
+			vmJsonFile = generateVmJsonFromTemplate(vmName, vmTemplateJsonFile)
+			createVmFromJson(vmName, vmJsonFile)
+			deleteVm(vmName)
+		})
+
+		It("should succeed to start a vmi from template", func() {
+			vmJsonFile = generateVmJsonFromTemplate(vmName, vmTemplateJsonFile)
+			createVmFromJson(vmName, vmJsonFile)
+
+			patchVm(vmName, vmStartPatch)
+
+			Eventually(func() bool {
+				out, err := tests.RunOcCommand("get", "vmis")
+				Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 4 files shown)


tests/oc_template_test.go
```diff
@@ -0,0 +1,331 @@
...
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Creating offline vm from JSON file via oc-create command")
+			out, err = tests.RunOcCommand("process", "-f", vmTemplateJsonFile, vmTemplateParams)
+			Expect(err).ToNot(HaveOccurred())
+			message = fmt.Sprintf("offlinevirtualmachine \"%s\" created", vmName)
+			Expect(out).To(Equal(message))
+
+			// this should be probably handled differently
+			time.Sleep(time.Second * 120)
```
stu-gott: the `Eventually()` command can help here. Assuming the goal here is to wait for the VM to be running, something along these lines:

```
Eventually(func() bool {
   out, err = tests.RunOcCommand("look up that vm...")
   return out == "???"
}, time.Duration(60)*time.Second).Should(Equal(true), "Timed out waiting for preset to appear")
```

stu-gott: Note the func you define here does not need to return a bool. if an error object, integer, anything is more appropriate, that's fair game. Just make the Should() comparison match what you're looking for....

tests/utils.go
```diff
@@ -1133,6 +1133,30 @@ func RunKubectlCommand(args ...string) error {
...
+	if err != nil {
+		Skip("Skip test that requires oc binary")
+	}
+}
+
+func RunOcCommand(args ...string) (string, error) {
+	// Do not try to run 'oc' command if it is not available
+	path, err := exec.LookPath("oc")
+	if err != nil {
+		return "", fmt.Errorf("can not find oc binary ")
```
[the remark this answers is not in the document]
gszasz: I have already deleted the trailing space. Thanks for noticing it.

tests/oc_template_test.go
```diff
@@ -0,0 +1,331 @@
...
+				err = os.Remove(vmTemplateJsonFile)
+				Expect(err).ToNot(HaveOccurred())
+				vmTemplateJsonFile = ""
+			}
+			if vmJsonFile != "" {
+				// remove testing vm JSON file
+				err = os.Remove(vmJsonFile)
+				Expect(err).ToNot(HaveOccurred())
+				vmJsonFile = ""
+			}
```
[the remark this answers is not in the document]
gszasz: That is a fair point. It would be indeed much cleaner to add:

```go
AfterEach(func() {
    // logout from oc CLI
    out, err = RunOcCommand("logout")
    ...
}).
```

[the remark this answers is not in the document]
rmohr: Why do we even have to do this "login"? Can't you just pass in `-server` and `kubeconfig` flags from our code?

tests/oc_template_test.go
```diff
@@ -0,0 +1,328 @@
...
+const (
+	ocUser = "system:admin"
+)
+
+var _ = Describe("VM Template", func() {
+	flag.Parse()
+
+	tests.PanicOnError(err)
+
+	// I need to figure out how to generate this JSON string from a struct using tests/types.go
```
rmohr: What exactly is blocking you there? If it is about the serialization process itself, have a look at https://blog.golang.org/json-and-go. That should contain everything needed.

tests/oc_template_test.go
rmohr: Yes Eventually is the way to go ...

Is oc-create asynchronous? If not, you should immediatley have the chance to fetch the vm.

tests/oc_template_test.go
```diff
@@ -0,0 +1,328 @@
...
+			out, err = tests.RunOcCommand("delete", "-f", vmJsonFile)
+			Expect(err).ToNot(HaveOccurred())
+			message = fmt.Sprintf("offlinevirtualmachine \"%s\" deleted", vmName)
+			Expect(out).To(Equal(message))
+
+			// this should be probably handled differently
+			time.Sleep(time.Second * 120)
+
+			By("Checking out that the vm does not exist anymore")
+			result := virtClient.RestClient().Get().Resource(tests.VmResource).Namespace(k8sv1.NamespaceDefault).Name(vmName).Do()
```
rmohr: Was this supposed to check if the vm, or the ovm existed? Either way, you can use the client like this:

```
virtCliet.VM("namespace").Get("vmname", GetOptions{})
```

rmohr: Could you use one of our test namespaces for testing the entitis? Normally you create everything in `tests.NamespaceTestDefault`.

tests/oc_template_test.go
```diff
@@ -0,0 +1,328 @@
...
+			time.Sleep(time.Second * 60)
+
+			By("Stopping an offline vm via oc-patch command")
+			out, err = tests.RunOcCommand("patch offlinevirtualmachine", vmName, "--type merge", vmStopPatch)
+			Expect(err).ToNot(HaveOccurred())
+			message = fmt.Sprintf("offlinevirtualmachine \"%s\" patched", vmName)
+			Expect(out).To(Equal(message))
+
+			// this should be probably handled differently
+			time.Sleep(time.Second * 60)
```
rmohr: Please replace all sleeps with `Eventually`.

tests/utils.go
```diff
@@ -1133,6 +1133,42 @@ func RunKubectlCommand(args ...string) error {
...
 }
 
+func SkipIfNoOc() {
+	path, err := exec.LookPath("oc")
+	if err != nil {
+		Skip("Skip test that requires oc binary")
+	}
+}
+
+func RunOcCommand(args ...string) (string, error) {
```
rmohr: That function looks now great. You will have to also add a commandline flag, so that you can do similar checks for oc, like done at https://github.com/gszasz/kubevirt/blob/0dc18ddc43a28ba29d2dee24949dba431a683602/tests/utils.go#L1108 for kubectl.

Finally, use that parameter to point our tests to the oc tool which we extract from the containerized cluster.

tests/oc_template_test.go
gszasz: @rmohr thanks for the suggestion that really makes it shorter.

tests/oc_template_test.go
```diff
@@ -0,0 +1,331 @@
...
+			By("Converting offline vm template into JSON file via oc-process command")
+			out, err = tests.RunOcCommand("process", "-f", vmTemplateJsonFile, vmTemplateParams)
+			Expect(err).ToNot(HaveOccurred())
+			vmJsonFile, err = WriteJson(vmName, out)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should create an offline vm from a template", func() {
+			var message = ""
+			// probably there is a better way how to do this without duplicating the code
```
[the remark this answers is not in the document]
gszasz: @stu-gott Do you have a link to some kind of handy how-to that would give me an example how to implement this? I love the idea.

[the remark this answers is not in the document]
cynepco3hahue: @gszasz You can find here example of table usage https://github.com/kubevirt/kubevirt/blob/a3dc90abf9741191d621f04e974fdde70368569c/tests/replicaset_test.go#L96

tests/oc_template_test.go
```diff
@@ -0,0 +1,328 @@
...
+			By("Converting offline vm template into JSON file via oc-process command")
```
cynepco3hahue: I think you can move this section to separate function under `Describe`
```
generateVmFromTemplate := func (vmName string) {
    By("Converting offline vm template into JSON file via oc-process command")
    out, err = tests.RunOcCommand("process", "-f", vmTemplateJsonFile, vmTemplateParams)
    vmJsonFile, err = WriteJson(vmName, out)
}
```
the same true for start and stop VM via `oc` command

gszasz: @cynepco3hahue I like this idea more than the `table.DescribeTable`. Thanks for the suggestion.

shiywang:
Overal looks good to me !!

tests/oc_template_test.go
```diff
@@ -0,0 +1,257 @@
...
+)
+
+var _ = Describe("VM Template", func() {
+	flag.Parse()
+
+	vmTemplate := `{
+		"kind": "Template",
+		"objects": [
+			{
+				"kind": "OfflineVirtualMachine",
```
rmohr: At least the whole `OfflineVirtualMachine` acn be created via our structs. Why can't we use the Template struct like agreed on before? If we don't test mappings between structs and manifests themselves, then raw jsons or yamls are not an option. That leads to harder maintainable code.

gszasz: @rmohr I agree. Nevertheless, the `types.go` file smuggled from the `openshift/origin` contained dependencies that were not met within `kubevirt/kubevirt`. @stu-gott suggested not bringing any KubeVirt-unrelated dependencies into the project, so I have decided to ephemerally drop this file and define Json code by the hardcoded string until I do not figure out how to resolve this properly. Note that this test will need refactoring soon anyway once offlinevirtualmachines will be renamed. Creating Json object by serialization from the template `struct` is vital in the long term. At this very moment, we need test results.

tests/oc_template_test.go
```diff
@@ -0,0 +1,257 @@
...
+		})
+
+		It("should succeed to start an offline vm", func() {
+			vmJsonFile = generateVmJsonFromTemplate(vmName, vmTemplateJsonFile)
+			createVmFromJson(vmName, vmJsonFile)
+
+			patchVm(vmName, vmStartPatch)
+
+			Eventually(func() bool {
+				out, err := tests.RunOcCommand("get vms")
```
shiywang: I think this should be `tests.RunOcCommand("get", "vms")`

tests/oc_template_test.go
```diff
@@ -0,0 +1,257 @@
...
+				Expect(err).ToNot(HaveOccurred())
+				return strings.Contains(out, vmName)
+			}, time.Duration(60)*time.Second).Should(BeTrue(), "Timed out waiting for vm to appear")
+
+			patchVm(vmName, vmStopPatch)
+
+			Eventually(func() bool {
+				out, err := tests.RunOcCommand("get vms")
```
shiywang: ditto

tests/oc_template_test.go
```diff
@@ -0,0 +1,257 @@
...
+	createVmFromJson := func(vmName string, vmJsonFile string) {
+		var message = ""
+		By("Creating offline vm from JSON file via oc-create command")
+		out, err := tests.RunOcCommand("create", "-f", vmJsonFile)
+		Expect(err).ToNot(HaveOccurred())
+		message = fmt.Sprintf("offlinevirtualmachine \"%s\" created", vmName)
+		Expect(out).To(Equal(message))
+
+		Eventually(func() bool {
+			out, err := tests.RunOcCommand("get ovms")
```
shiywang: I tried RunOcCommand this function in my own tests like `RunOcCommand("get pvc")` it throw an error, but when I change to `RunOcCommand("get", "pvc")` it works, are you sure this is correct ?

cynepco3hahue:
Can you please fix travis errors
```
go version go1.10 linux/amd64
/root/go/src/kubevirt.io/kubevirt/tests/oc_template_test.go:50:14: expected type, found '&'
/root/go/src/kubevirt.io/kubevirt/tests/oc_template_test.go:167:3: missing ',' in composite literal
make: *** [build] Error 2
```

cynepco3hahue:
@gszasz Hi Gabriel, I fixed and added some stuff to your PR, I think you can take it from here
- check that test passed locally
- maybe you can re-organize methods that I added because I just copy pasted them from `tools/vms-generator/vms-generator.go`

cynepco3hahue:
@rmohr Maybe it worth to create some common `utils` pkg, that tools and tests can use?

cynepco3hahue:
OK to test

cynepco3hahue:
@gszasz I see test skipped on openshift environment, can you please check it.

stu-gott:
@cynepco3hahue It appears that "skip if no OC" is triggered if the `oc` binary does not appear on the system path. This is a pretty reasonable approach as the test case runs the `oc` command so will fail if it's not present. I assume that binary is available in CI?

cynepco3hahue:
@stu-gott I think @gszasz  must explicitly set `oc` path via argument, otherwise, it always will fail, because we do not run our tests on master

fabiand:
@gszasz what needs to be done next to get this merged?

tests/utils.go
```diff
@@ -1361,6 +1363,48 @@ func RunKubectlCommand(args ...string) (string, error) {
 	return string(stdOutBytes), nil
 }
 
+func SkipIfNoOc() {
```
stu-gott: Wasn't there something about using the KubeCtl path here? at least in the dev environment?

stu-gott:
Looks good overall

tests/utils.go
gszasz: @stu-gott This was resolved by adding the `-oc-path` flag to the

```bash
if [[ ${TARGET} == openshift* ]]; then
    oc=${kubectl}
fi

${TESTS_OUT_DIR}/tests.test -kubeconfig=${kubeconfig} -tag=${docker_tag} -prefix=${functest_docker_prefix} -oc-path=${oc} -kubectl-path=${kubectl} -test.timeout 90m ${FUNC_TEST_ARGS}
```

 `hack/functest.sh`. The value of this flag is set to empty string when the value of `TARGET` environment variable does not begin with `openshift`.

stu-gott:
retest this please

cynepco3hahue:
retest this please

Issue #718, opened 2018-02-07
Make Kubevirt network model selectable


shiywang:
according to comments here: https://github.com/kubevirt/demo/issues/34#issuecomment-363328856

cc @fabiand 

# API
```yaml
kind: VM
spec:
  devices:
    interfaces:
      # Today, only one source: default
      # This issue is about intorducing the API to specify the NIC model
      - name: nic-ip
        model: e1000
```

# Future API considerations
Furtehr down the road the bigger picture would be alogn the lines:

```yaml
kind: VM
spec:
  devices:
    interfaces:
      # In the next iteration it's about the ability to specify the source (once we have multiple sources)
      - name: nic-ip
        model: e1000
        source:
          default:
            bind: ip  # we have this today

      # #787 is about this mode
      - name: nic-ports
        model: e1000
        source:
          default:
            bind: ports  # #787 is about this mode
        ports:  # only needed in 'ports' mode
        - port: 42
          protocol: tcp (optional)
          name: foo (optional)

        # @phoracek is working on additional nics
        # IDEA: Example for resource based nic (not part of this PR)
        - name: nic-resource
          source:
            resource: bridge.k6t.io/mybr0  ## mode for selecting which resource use
```

slintes:
Is somebody working on this already? Seems to be a nice issue to get familiar with the codebase.

This would mean to extend the Devices of the DomainSpec with Interface(s?). Should the model be the only configurable field, which will be used then in SetDefaults_Devices?

slintes:
and one more question: should the model be validated? From https://libvirt.org/formatdomain.html#elementsNICSModel: 

> Typical values for QEMU and KVM include: ne2k_isa i82551 i82557b i82559er ne2k_pci pcnet rtl8139 e1000 virtio

And what should happen then if the value doesn't match any of these?

fabiand:
In general yes, But this is a little tricky.

Today - as you saw - we do not have an `interfaces` list. The reason is that we can only support a single NIC.
In general networking is not configurable, because we don't have API for the only NIC.

Now, we also know that we want to support multiple NICs on the long run (and @phoracek is working on this).

Thus, if we solve this bug, then this implies to introduce an API which will support more than one network interface.

Thus yes, it would be about introducing an `devices.interfaces` section, which would currently allow (and by default contain) just a single entry.
But could be used in future to enumerate multiple devices.

I'll add an API exmaple to the description.

fabiand:
[...]

PR #990, opened 2018-05-04
Use virtio for cirros vms in our functests

### LINKED ISSUE #936: Cirros VM hangs very long on dhcp discover

rmohr:
Fixes #936 for our tests. The solution here is intended to buy us some time until we have decided on how to properly represent network devices on the API. If "kubevirt.io/os=cirros" is present in the annotations, kubevirt will choose virtio for the network type.

Using "kubevirt.io/os" for that purpose is not that nice, but it kind of fits into what this label is meant for: Allowing special configurations based on the operating system used. See also https://www.kubevirt.io/user-guide/#/workloads/virtual-machines/guest-operating-system-information?id=guest-operating-system-information for more context.

### DIFF
--- a/pkg/virt-launcher/virtwrap/api/converter.go
+++ b/pkg/virt-launcher/virtwrap/api/converter.go
@@ -471,6 +471,23 @@ func Convert_v1_VirtualMachine_To_api_Domain(vm *v1.VirtualMachine, domain *Doma
 		},
 	}
 
+	// Add mandatory interface
+	interfaceType := "e1000"
+	if vm.Annotations["kubevirt.io/os"] == "cirros" {
+		interfaceType = "virtio"
+	}
+
+	// For now connect every virtual machine to the pod network
+	domain.Spec.Devices.Interfaces = []Interface{{
+		Model: &Model{
+			Type: interfaceType,
+		},
+		Type: "bridge",
+		Source: InterfaceSource{
+			Bridge: DefaultBridgeName,
+		}},
+	}
+
 	return nil
 }
 
--- a/pkg/virt-launcher/virtwrap/api/converter_test.go
+++ b/pkg/virt-launcher/virtwrap/api/converter_test.go
@@ -403,6 +403,13 @@ var _ = Describe("Converter", func() {
 			Expect(vmToDomainXMLToDomainSpec(vm, c).VCPU.CPUs).To(Equal(uint32(3)))
 
 		})
+
+		It("should select virto network model if cirros", func() {
+			v1.SetObjectDefaults_VirtualMachine(vm)
+			vm.Annotations = map[string]string{"kubevirt.io/os": "cirros"}
+			domain := vmToDomain(vm, c)
+			Expect(domain.Spec.Devices.Interfaces[0].Model.Type).To(Equal("virtio"))
+		})
 	})
 })
 
--- a/pkg/virt-launcher/virtwrap/api/defaults.go
+++ b/pkg/virt-launcher/virtwrap/api/defaults.go
@@ -17,16 +17,6 @@ func SetDefaults_Devices(devices *Devices) {
 		},
 	}
 
-	// For now connect every virtual machine to the default network
-	devices.Interfaces = []Interface{{
-		Model: &Model{
-			Type: "e1000",
-		},
-		Type: "bridge",
-		Source: InterfaceSource{
-			Bridge: DefaultBridgeName,
-		}},
-	}
 }
 
 func SetDefaults_OSType(ostype *OSType) {
--- a/pkg/virt-launcher/virtwrap/api/schema.go
+++ b/pkg/virt-launcher/virtwrap/api/schema.go
@@ -551,9 +551,6 @@ func NewMinimalDomainSpec(vmName string) *DomainSpec {
 	domain.Name = vmName
 	domain.Memory = Memory{Unit: "MB", Value: 9}
 	domain.Devices = Devices{}
-	domain.Devices.Interfaces = []Interface{
-		{Type: "network", Source: InterfaceSource{Network: "default"}},
-	}
 	return domain
 }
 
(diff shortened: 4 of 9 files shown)


rmohr:
retest this please

rmohr:
retest this please

rmohr:
> should virtio be the default and os=windows be the exception that toggles e1000?

we could say, if it starts with "win", then use e1000. libos-info shows these windows tags:

```
 win1.0               | Microsoft Windows 1.0                              | 1.0      | http://microsoft.com/win/1.0            
 win10                | Microsoft Windows 10                               | 10.0     | http://microsoft.com/win/10             
 win2.0               | Microsoft Windows 2.0                              | 2.0      | http://microsoft.com/win/2.0            
 win2.1               | Microsoft Windows 2.1                              | 2.1      | http://microsoft.com/win/2.1            
 win2k                | Microsoft Windows 2000                             | 5.0      | http://microsoft.com/win/2k             
 win2k12              | Microsoft Windows Server 2012                      | 6.3      | http://microsoft.com/win/2k12           
 win2k12r2            | Microsoft Windows Server 2012 R2                   | 6.3      | http://microsoft.com/win/2k12r2         
 win2k16              | Microsoft Windows Server 2016                      | 1709     | http://microsoft.com/win/2k16           
 win2k3               | Microsoft Windows Server 2003                      | 5.2      | http://microsoft.com/win/2k3            
 win2k3r2             | Microsoft Windows Server 2003 R2                   | 5.2      | http://microsoft.com/win/2k3r2          
 win2k8               | Microsoft Windows Server 2008                      | 6.0      | http://microsoft.com/win/2k8            
 win2k8r2             | Microsoft Windows Server 2008 R2                   | 6.1      | http://microsoft.com/win/2k8r2          
 win3.1               | Microsoft Windows 3.1                              | 3.1      | http://microsoft.com/win/3.1            
 win7                 | Microsoft Windows 7                                | 6.1      | http://microsoft.com/win/7              
 win8                 | Microsoft Windows 8                                | 6.2      | http://microsoft.com/win/8              
 win8.1               | Microsoft Windows 8.1                              | 6.3      | http://microsoft.com/win/8.1            
 win95                | Microsoft Windows 95                               | 4.0      | http://microsoft.com/win/95             
 win98                | Microsoft Windows 98                               | 4.1      | http://microsoft.com/win/98             
 winme                | Microsoft Windows Millennium Edition               | 4.9      | http://microsoft.com/win/me             
 winnt3.1             | Microsoft Windows NT Server 3.1                    | 3.1      | http://microsoft.com/winnt/3.1          
 winnt3.5             | Microsoft Windows NT Server 3.5                    | 3.5      | http://microsoft.com/winnt/3.5          
 winnt3.51            | Microsoft Windows NT Server 3.51                   | 3.51     | http://microsoft.com/winnt/3.51         
 winnt4.0             | Microsoft Windows NT Server 4.0                    | 4.0      | http://microsoft.com/winnt/4.0          
 winvista             | Microsoft Windows Vista                            | 6.0      | http://microsoft.com/win/vista          
 winxp                | Microsoft Windows XP                               | 5.1      | http://microsoft.com/win/xp  
```

michalskrivanek:
why would you use e1000 by default at all? drivers should always be installed
if you want to have a different configuration for first start/installation then that should be an exception rather than default.

booxter:
I think there is a better way that we could use to completely avoid conditional: using something that is neither e1000 nor virtio and that is supported by all types of images we target. Drivers have different maintenance level, so I would recommend to use "e1000e" that is the Express version of e1000. The driver is *not* affected by the slow link ready issue, and I believe Windows images should boot just fine with it. Also, as per changelog, the e1000e type will be the default NIC type for x86 q35 machine starting qemu 2.12.

The alternative PR that switches to e1000 is: https://github.com/kubevirt/kubevirt/pull/1016

rmohr:
> Also, as per changelog, the e1000e type will be the default NIC type for x86 q35 machine starting qemu 2.12.

So we would use e1000 if it is not a q35 architecture?

rmohr:
> why would you use e1000 by default at all? drivers should always be installed
> if you want to have a different configuration for first start/installation then that should be an exception rather than default.

In general I agree but since everything for proper defaulting and exceptions from the default is not in place yet, it probably makes sense to play safe right now?

rmohr:
Closing this in favour of #1016

michalskrivanek:
@rmohr,
virtio devices are the only ones truly well maintained, none of the emulated devices are that great.
e1000 is useful for booting Windows without drivers first, same as IDE on i440fsx or SATA on q35. And it's useful for more exotic guests without virtio drivers, but that's it. I wouldn't recommend to use it for Windows beyond first boot or on any Linux OS.
Do we primarily target Windows? Or should a generic Linux OS be the one to start with?

rmohr:
@michalskrivanek well I completely agree with you. It is just that the model of the interface is not yet represented on the API (nor is a mechanism which does that in a configurable way by the user), so it can't be changed right now. That is obviously something which needs to be changed (and needs to be done before we say that KubeVirt is production ready), but until we have that I would want that operating systems simply boot. Don't you think?

michalskrivanek:
well, yes. 
though if we already have at least some defaults per osinfo then any Linux would be better off with virtio, Windows with q35 e1000e/sata should work fine. i440fx is not really needed,right?

mykaul:
@rmohr - is there an open issue on the lack of model of the NIC in the in API?

fabiand:
Yes, https://github.com/kubevirt/kubevirt/issues/718

PR #1006, opened 2018-05-08
Run artifact copier outside of docker builder

rmohr:
The artifact copier can run outside of the docker container. This way we get all the environement variable overrides which we need when we do a release.

@cynepco3hahue after that everything should really work again.


rmohr:
retest this please

PR #1007, opened 2018-05-08
Use gocli with a lower DOCKER_API_VERSION to support docker 1.12

rmohr:
We accidently only supported docker 1.13 and up with our CI tooling.


rmohr:
See https://github.com/kubevirt/kubevirtci/pull/6

stu-gott:
Works here as well

PR #1021, opened 2018-05-15
Fix secretRef resolution for cloud-init

rmohr:
We had two issues in that area:
 * secrets were double-encoded (our test double-decoded too, therfore it
went unnoticed)
 * we updated  the vm with the resolved cloud-init data on the cluster

### DIFF
--- a/pkg/cloud-init/cloud-init.go
+++ b/pkg/cloud-init/cloud-init.go
@@ -168,14 +168,11 @@ func ResolveSecrets(source *v1.CloudInitNoCloudSource, namespace string, clients
 		return err
 	}
 
-	userDataBase64, ok := secret.Data["userdata"]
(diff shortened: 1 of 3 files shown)


rmohr:
retest this please

pkg/cloud-init/cloud-init.go
```diff
@@ -168,14 +168,11 @@ func ResolveSecrets(source *v1.CloudInitNoCloudSource, namespace string, clients
 		return err
 	}
 
-	userDataBase64, ok := secret.Data["userdata"]
+	userData, ok := secret.Data["userdata"]
```
cynepco3hahue: Does the client also decode it back?

rmohr: yes. Our tests passed because we double-en and  double-decoded it ....

PR #1022, opened 2018-05-15
Make virt-api debugable again

rmohr:
By accient the commandline flags were parsed to late. As a result it was
not possible to provide kubeconfig via commandline flags.

To debug it e.g. via a UI, simply point virt-api to your cluster-config and specify your desired port:

```bash
--kubeconfig cluster/k8s-1.9.3/.kubeconfig --port 1234
```


rmohr:
retest this please

PR #1026, opened 2018-05-15
virtctl should use REST Patch instead PUT on update ovm call #863

zvikorn:
https://github.com/kubevirt/kubevirt/issues/863


ovirt-infra:

cynepco3hahue:
ok to test

pkg/virtctl/offlinevm/offlinevm.go
```diff
@@ -107,7 +109,9 @@ func (o *Command) Run(cmd *cobra.Command, args []string) error {
 
 	if ovm.Spec.Running != running {
 		ovm.Spec.Running = running
```
rmohr: That line is not needed anymore now.

pkg/virtctl/offlinevm/offlinevm.go
```diff
@@ -107,7 +109,9 @@ func (o *Command) Run(cmd *cobra.Command, args []string) error {
...
-		_, err := virtClient.OfflineVirtualMachine(namespace).Update(ovm)
+		_, err := virtClient.OfflineVirtualMachine(namespace).Patch(ovm.Name, types.MergePatchType,
+			[]byte("{\"spec\":{\"running\":true}}"))
```
rmohr: You will have to set it to false if we want to stop it.

rmohr: In general it may be beneficial now to untangle the start and stop flows in two different branches.

rmohr:
retest this please

pkg/kubecli/ovm.go
```diff
@@ -88,6 +90,19 @@ func (o *ovm) Update(offlinevm *v1.OfflineVirtualMachine) (*v1.OfflineVirtualMac
 	return updatedOvm, err
 }
 
+func (v *ovm) Patch(name string, pt types.PatchType, data []byte, subresources ...string) (result *v1.OfflineVirtualMachine, err error) {
```
rmohr: Could you add a basic unit test with e.g. json-merge type to `pkg/kubecli/ovm_test.go`?

fabiand: +1

IIRC there were issues withc patching CRDs.

rmohr:
> IIRC there were issues withc patching CRDs.

I think you are referring to the the fact that CRs don't support strategic-merge-patch?

zvikorn:
MergePatchType worked for me. I could do virtctl start and stop and VM was instantiated and deleted. Hence the code itself works. The problem I'm having is with the unit test.
At the end of the test, we compare between  a 'NewMinimalOVM' instance we create, to a 'patchedVM' which is returned from the 'patch' function we call. The 'Patch()' function in 'generated_moke_kubevirt.go' is not implemented well, so the patchedVM that is returned from ''ret := _m.ctrl.Call(_m, "Patch", _s...)' contains wrong values (it's basically not updated with spec.running='true'. Also the APIVersion and Kind metadata return empty). The arguments we send in '_m.ctrl.Call' are different than the ones we send when we call 'Update'. we send an array (_s...) instead an ovm instance.
Do we want (can we) to change the implementation of the mockup to fix it?
Also, the 'NewMinimalOvm' creates the instance with 'spec.running:false'. so when we compare it with the patchedVM, we need to patch it 'spec.running:false' (I know, it's redicoulus), otherwise it will fail (assuming we fix the first issue).
What do you think?

rmohr:
@zvikorn functionality wise you are covered. We have functional tests which do update the ovm via virtctl and that obviously works. 

What I would expect from the unit test, is that the ghttp server tests if we send a patch with the expected values. Then return a predefined ovm. Something like:


```go
ghttp.VerifyRequest("PATCH", vmPath),
ghttp.VerifyBody([]byte("myjsonmergepatchstring")),
ghttp.RespondWithJSONEncoded(http.StatusOK, myPredefinedOVM)
```

pkg/virtctl/offlinevm/offlinevm_test.go
```diff
@@ -5,10 +5,18 @@ import (
...
+	"kubevirt.io/kubevirt/pkg/api/v1"
+	"kubevirt.io/kubevirt/pkg/kubecli"
+	"github.com/golang/mock/gomock"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 var _ = Describe("OfflineVirtualMachine", func() {
 
+	var ovmInterface *kubecli.MockOfflineVirtualMachineInterface
+	var ctrl = gomock.NewController(GinkgoT())
```
rmohr: Can you always create a new instance in the `BeforeEach` block? In the `AfterEach` block you will then have to call `ctrl.Finish()`. If `Finish` is not called, the mocks would return the expected objects, but final checks, if expected calls were triggered at all are not performed.

cynepco3hahue:
@zvikorn Can you please rebase and fix Roman comments?

pkg/kubecli/vm_test.go
```diff
@@ -112,6 +113,43 @@ var _ = Describe("Kubevirt VirtualMachine Client", func() {
 		Expect(updatedVMI).To(Equal(vm))
 	})
 
+	It("should patch a OfflineVirtualMachine", func() {
```
cynepco3hahue: change `OfflineVirtualMachine -> VirtualMachine` under each test

pkg/virtctl/vm/vm_test.go
```diff
@@ -1,30 +1,60 @@
...
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	k8smetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"kubevirt.io/kubevirt/pkg/kubecli"
+	"kubevirt.io/kubevirt/tests"
 )
 
-var _ = Describe("VirtualMachine", func() {
+var _ = Describe("OfflineVirtualMachine", func() {
```
cynepco3hahue: change OfflineVirtualMachine -> VirtualMachine under each test

zvikorn: oops... missed that somehow.. Thanks

PR #1030, opened 2018-05-15
Add disk boot order to API

mlsorensen:
Per discussion on mailing list, add ability to provide boot order to disks

https://groups.google.com/d/topic/kubevirt-dev/tBhwK8Y4UE0/discussion

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3292,7 +3292,7 @@
     ],
     "properties": {
      "bootOrder": {
-      "description": "BootOrder is an integer value used to determine ordering of boot devices\n+optional",
+      "description": "BootOrder is an integer value \u003e 0, used to determine ordering of boot devices.\nDisks without a boot order are not tried if a disk with a boot order exists.\n+optional",
       "type": "integer",
       "format": "integer"
      },
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3292,7 +3292,7 @@
     ],
     "properties": {
      "bootOrder": {
-      "description": "BootOrder is an integer value \u003e 0, used to determine ordering of boot devices.\nDisks without a boot order are not tried if a disk with a boot order exists.\n+optional",
+      "description": "BootOrder is an integer value \u003e 0, used to determine ordering of boot devices.\nLower values take precedence.\nDisks without a boot order are not tried if a disk with a boot order exists.\n+optional",
       "type": "integer",
       "format": "integer"
      },
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3291,6 +3291,11 @@
      "volumeName"
     ],
     "properties": {
+     "bootOrder": {
+      "description": "BootOrder is an integer value used to determine ordering of boot devices\n+optional",
+      "type": "integer",
+      "format": "integer"
+     },
      "cdrom": {
       "description": "Attach a volume as a cdrom to the vm",
       "$ref": "#/definitions/v1.CDRomTarget"
--- a/pkg/virt-launcher/virtwrap/api/converter.go
+++ b/pkg/virt-launcher/virtwrap/api/converter.go
@@ -73,6 +73,10 @@ func Convert_v1_Disk_To_api_Disk(diskDevice *v1.Disk, disk *Disk, devicePerBus m
 		Name: "qemu",
 	}
 	disk.Alias = &Alias{Name: diskDevice.Name}
+	if diskDevice.BootOrder != 0 {
+		disk.BootOrder = &BootOrder{Order: diskDevice.BootOrder}
+	}
+
 	return nil
 }
 
--- a/pkg/virt-launcher/virtwrap/api/converter.go
+++ b/pkg/virt-launcher/virtwrap/api/converter.go
@@ -73,8 +73,8 @@ func Convert_v1_Disk_To_api_Disk(diskDevice *v1.Disk, disk *Disk, devicePerBus m
 		Name: "qemu",
 	}
 	disk.Alias = &Alias{Name: diskDevice.Name}
-	if diskDevice.BootOrder != 0 {
-		disk.BootOrder = &BootOrder{Order: diskDevice.BootOrder}
+	if diskDevice.BootOrder != nil {
+		disk.BootOrder = &BootOrder{Order: *diskDevice.BootOrder}
 	}
 
 	return nil
--- a/pkg/virt-launcher/virtwrap/api/converter_test.go
+++ b/pkg/virt-launcher/virtwrap/api/converter_test.go
@@ -30,7 +30,6 @@ import (
 
 	"fmt"
 	"os"
-	"strings"
 
 	"kubevirt.io/kubevirt/pkg/api/v1"
 )
@@ -49,13 +48,13 @@ var _ = Describe("Converter", func() {
 					},
 				},
 			}
-			var convertedDisk = `<disk device="disk" type="">
+			var convertedDisk = `<Disk device="disk" type="">
   <source></source>
   <target bus="virtio" dev="vda"></target>
   <driver name="qemu" type=""></driver>
   <alias name="mydisk"></alias>
   <boot order="1"></boot>
-</disk>`
+</Disk>`
 			xml := diskToDiskXML(kubevirtDisk)
 			fmt.Println(xml)
 			Expect(xml).To(Equal(convertedDisk))
@@ -71,12 +70,12 @@ var _ = Describe("Converter", func() {
 					},
 				},
 			}
-			var convertedDisk = `<disk device="disk" type="">
+			var convertedDisk = `<Disk device="disk" type="">
   <source></source>
   <target bus="virtio" dev="vda"></target>
   <driver name="qemu" type=""></driver>
   <alias name="mydisk"></alias>
-</disk>`
+</Disk>`
 			xml := diskToDiskXML(kubevirtDisk)
 			fmt.Println(xml)
 			Expect(xml).To(Equal(convertedDisk))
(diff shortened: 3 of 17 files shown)


ovirt-infra:

mlsorensen:
The error is:

> `The command "if [[ -n "$(git status --porcelain)" ]] ; then echo "It seems like you need to run `make`. Please run it and commit the changes"; git status --porcelain; false; fi" exited with 1.
> `

However, I'm not detecting any changes locally when I do this and I'm in sync with this branch, nothing to push.

rmohr:
It seems like the latest version of shfmt has issues with one of our shell script. #1032 should fix that. Will open an issue on the shfmt project.

rmohr:
shfmt is fixed: https://github.com/mvdan/sh/issues/241

mlsorensen:
@rmohr Sounds good. I'll look into how the functional tests work and try to get back within a few weeks at the latest.

pkg/virt-launcher/virtwrap/api/converter_test.go
```diff
@@ -406,6 +453,15 @@ var _ = Describe("Converter", func() {
 	})
 })
 
+func diskToDiskXML(disk *v1.Disk) string {
+	devicePerBus := make(map[string]int)
+	libvirtDisk := &Disk{}
+	Expect(Convert_v1_Disk_To_api_Disk(disk, libvirtDisk, devicePerBus)).To(Succeed())
+	data, err := xml.MarshalIndent(libvirtDisk, "", "  ")
+	Expect(err).ToNot(HaveOccurred())
+	return strings.ToLower(string(data))
```
rmohr: Why is there a `ToLower`?

tests/vmlifecycle_test.go
```diff
@@ -159,6 +159,32 @@ var _ = Describe("Vmlifecycle", func() {
...
+			It("should be able to boot from 'middle' disk", func() {
+				By("defining a VM with an Alpine disk")
+				vm = tests.NewRandomVMWithEphemeralDisk(tests.RegistryDiskFor(tests.RegistryDiskAlpine))
+				By("adding a Cirros Disk with boot order 1")
+				tests.AddEphemeralDiskWithBootOrder(vm, "disk1", "virtio", tests.RegistryDiskFor(tests.RegistryDiskCirros), 1)
+				By("adding a Fedora Disk")
+				tests.AddEphemeralDisk(vm, "disk2", "virtio", tests.RegistryDiskFor(tests.RegistryDiskFedora))
+
+				By("starting VM")
+				vm.Spec.Hostname = "cirros"
```
rmohr: Is it necessary to set the hostname?

tests/vmlifecycle_test.go
```diff
@@ -159,6 +159,32 @@ var _ = Describe("Vmlifecycle", func() {
...
+				obj, err := virtClient.RestClient().Post().Resource("virtualmachines").Namespace(tests.NamespaceTestDefault).Body(vm).Do().Get()
```
rmohr: You can use `vm, err = virtclient.VM(tests.NamespaceTestDefault).Create(vm)`.

tests/vmlifecycle_test.go
```diff
@@ -159,6 +159,32 @@ var _ = Describe("Vmlifecycle", func() {
...
+				Expect(err).To(BeNil())
+
+				By("waiting for VM to start")
+				_, ok := obj.(*v1.VirtualMachine)
+				Expect(ok).To(BeTrue(), "Object is not of type *v1.VM")
+				Expect(tests.WaitForSuccessfulVMStart(obj)).ToNot(BeEmpty())
```
rmohr: WaitForSuccessfulVMStart will time out and make the test fail. No need for the check. We should probably rename it to `ExpectSuccessfullVMStart` to make that clear.

tests/vmlifecycle_test.go
```diff
@@ -159,6 +159,32 @@ var _ = Describe("Vmlifecycle", func() {
 			})
 		})
 
+		Context("with boot order", func() {
+			It("should be able to boot from 'middle' disk", func() {
```
rmohr: If you let that test run twice, once booting cirros and once booting alpine by changing the boot priority this should be sufficient. You can use `table.DescribeTable` to create a test table.

tests/vmlifecycle_test.go
mlsorensen: It looks to be required to me. The LoggedInCirrosExpecter looks for 'vmName + " login:"', which seems to be 'cirros login:', and vmName is either from vm.Name (which is dynamically generated in these tests) or vm.Spec.Hostname.  I'll admit that I looked at other uses of this and it wasn't immediately clear to me how those were getting past this.

pkg/virt-launcher/virtwrap/api/converter_test.go
mlsorensen: The XML marshaler, per its documentation, only works on and returns capitalized root objects. Libvirt would handle it as-is, I could instead make the expected XML capitalized.

tests/vmlifecycle_test.go
mlsorensen: Ok, I'll make this change. I was following the existing patterns.

tests/vmlifecycle_test.go
mlsorensen: I can't seem to get the test to pass the Expect portions unless I include this to wait for the VM to start before proceeding.

tests/vmlifecycle_test.go
mlsorensen: Incoming...

tests/vmlifecycle_test.go
mlsorensen: I ditched this anyway and wrote an expecter function that can be used for the multiple OS tests.

tests/vmlifecycle_test.go
rmohr: Yes yes, plese keep the `WaitForSuccessfulVMStart(obj)`. That is perfect. I simply meant that the wrapping `Expect()` is not needed.

rmohr:
ok to test

rmohr:
retest this please

mlsorensen:
Looks like the OpenShift/Cirros combo timed out. I'll try to reproduce it and see if there's a problem.

mlsorensen:
These tests are passing over here when I run a cluster with KUBEVIRT_PROVIDER=os-3.9.0, multiple times. The Cirros image does block awhile on cloud-init data. I'm not sure if it's just a lot slower in the test environment.

cynepco3hahue:
@mlsorensen also be aware of the problem with e1000 interface type - https://github.com/kubevirt/kubevirt/issues/936
3 minutest already to big timeout for cirros image, but can you increase it to 4 minutest and I will re-run tests?

mlsorensen:
Looks like one of these timed out on an unrelated storage test. 

> • [SLOW TEST:120.533 seconds]
Storage
/root/go/src/kubevirt.io/kubevirt/tests/storage_test.go:42
  Starting a VM
  /root/go/src/kubevirt.io/kubevirt/tests/storage_test.go:122
    with Alpine PVC
    /root/go/src/kubevirt.io/kubevirt/tests/storage_test.go:123
      should be successfully started and stopped multiple times
      /root/go/src/kubevirt.io/kubevirt/vendor/github.com/onsi/ginkgo/extensions/table/table.go:92
        with CDRom PVC
        /root/go/src/kubevirt.io/kubevirt/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46
panic: test timed out after 1h30m0s

The other failed multiple tests, but still doesn't like Cirros. I'm tempted to switch to the Fedora image, but it seems there are other issues with the OpenShift functional tests that still may block the checks.

rmohr:
retest this please

mlsorensen:
I've switched Cirros out for Fedora. It adds a third disk (userdata), which I guess makes the test even better. It's also a lot faster.

rmohr:
> I've switched Cirros out for Fedora. It adds a third disk (userdata), which I guess makes the test even better. It's also a lot faster.

Oh, there was no user-data before on cirros? Well that explains why it was slow. I missed that. Now with #1047 merged (just merged a few hours ago), cirros should be the fastest one again (but it needs user-data too).

rmohr:
@mlsorensen could I ask you once more to try with cirros (+ user-data, so that it does not wait for the cloud-init via network).

mlsorensen:
Sure. There seemed to be a problem aside from userdata with Cirros, as even as it was it passed in everything but openshift. With the other changes, however, maybe things will be different.

mlsorensen:
Yay because it finally passed the new tests, boo because it failed others.

rmohr:
Especially the pvc test issue is well known. Looks fine for me now. Thx. Will have another look on Monday and merge then.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -73,6 +73,10 @@ func Convert_v1_Disk_To_api_Disk(diskDevice *v1.Disk, disk *Disk, devicePerBus m
 		Name: "qemu",
 	}
 	disk.Alias = &Alias{Name: diskDevice.Name}
+	if diskDevice.BootOrder != 0 {
```
rmohr: Hm, so the priority starts with `1`? If so, then allowing passing in 0 is a little bit confusing. I see a few options here:

 * Using a pointer, so that we can always distinguish between `0` and no value supplied. If `0` is supplied fail the validation.
 * Allow 0 also as a priority value (and add + 1 if libvirt requires it to start wit 1). With or without a pointer.

mlsorensen:
Travis is telling me I need to run 'make', but doing so results in no changes. I think it's actually this:

> package golang.org/x/tools/cover: unrecognized import path "golang.org/x/tools/cover" (https fetch: Get https://golang.org/x/tools/cover?go-get=1: dial tcp 173.194.198.141:443: i/o timeout)

rmohr:
> Travis is telling me I need to run 'make', but doing so results in no changes. I think it's actually this:
> 
>> package golang.org/x/tools/cover: unrecognized import path "golang.org/x/tools/cover" (https fetch: Get https://golang.org/x/tools/cover?go-get=1: dial tcp 173.194.198.141:443: i/o timeout)

Yes seems to be a transient error there.

api/openapi-spec/swagger.json
```diff
@@ -3291,6 +3291,11 @@
      "volumeName"
     ],
     "properties": {
+     "bootOrder": {
+      "description": "BootOrder is an integer value used to determine ordering of boot devices\n+optional",
```
rmohr: Can you also add something like `\nMust be a value greater or equal 1.` Also if lower values are tried first or later.

rmohr: And also how disks without a bootOrder are treated if some have one.

rmohr:
Ok, the failing tests are soley PVC related and is worked on separately to stabilize them.

PR #1031, opened 2018-05-15
Cleared some golint warnings

awels:
- Started cleaning up some golint warnings about missing comments
  and other stylistic items.


fabiand:
ok to test

rmohr:
@awels thanks. Could you add golint to our dockerized builder and run it in hack/check.sh? you find the dockerized builder in hack/docker-builder/Dockerfile. If you add golint there, it will be available in hack/check.sh for you. If only some of us run it they will have to permanently clean up behind others.

rmohr:
You can limit the linter to already clean up files initially, so that not everything needs to be cleaned up at once.

awels:
Something in our CI is already running golint, see: https://goreportcard.com/report/github.com/kubevirt/kubevirt

Its also running other stuff as well, my IDE just happens to do the same thing so it was fairly easy for me to find them.

awels:
I updated the dockerized builder to include golint, I decided to not limit the files, as it will not fail the test in any way. If I limit it to fixed files only no one will ever add new files.

rmohr:
> I updated the dockerized builder to include golint, I decided to not limit the files, as it will not fail the test in any way. If I limit it to fixed files only no one will ever add new files.

While I don't see why one should fix his files ever if we don't let something fail, I agree that we should not enforce linting, especially since there is nothing automatic happening (also there are false positives involved too). So could you remove it again from the builder? If we don't enforce it we should also not bug people with huge warning lists. What do you think?

awels:
So my take on this is we should fix as many as possible, 90% of the warnings are about missing comments on exported functions and interfaces. In practice this won't make any difference functionality wise, but if we have actual comments that document the function it will help new people. The rest appear to be either formatting (camel case not quite what it expects) or if XX else YY, where XX is a return so the else makes no sense. Those are also stylistic things, but I think it is important to enforce a consistent style across the project to keep it easier to read in the future.

As I mentioned above, the lint is run automatically by the go report card, so if someone is bothered by the linting (like I am) they can use that to fix it.

rmohr:
@awels there seems to be no code left in the pr.

awels:
Ugh, must have done the sync with master wrong and it cleared out the patch, and since there is no history in github, no way to recover. Man I wish we were using gerrit instead. I guess I can close it.

rmohr:
Git reflog will help you.

Issue #1036, opened 2018-05-16
ISO readonly defaulting to 'nil' ?


mlsorensen:
I've been experimenting with ISOs and I may have found an issue with the current master. Starting a VM I get this in the pod logs:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xe7c3b7]

goroutine 83 [running]:
kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api.Convert_v1_Disk_To_api_Disk(0xc421bbb1b8, 0xc421bbb3e8, 0xc421bbb048, 0xb, 0xc421bbb3a8)
	/root/go/src/kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api/converter.go:70 +0x637
```
Here is the YAML. The docs indicate that CDROMs are readonly by default, but looking at the code it seems that readonly may be 'nil'.

```
apiVersion: kubevirt.io/v1alpha1
kind: VirtualMachine
metadata:
  name: centos-iso
  labels:
    app: centos-iso
spec:
  terminationGracePeriodSeconds: 15
  domain:
    cpu:
      cores: 4
    resources:
      limits:
        cpu: 2
        memory: 5Gi
      requests:
        cpu: 2
        memory: 4Gi
    devices:
      disks:
      - name: install-iso
        volumeName: install-iso
        cdrom:
          bus: sata
  volumes:
    - name: install-iso
      persistentVolumeClaim:
        claimName: centos-live-iso
```

If I change the disk definition to this it works:

```
    devices:
      disks:
      - name: install-iso
        volumeName: install-iso
        cdrom:
          bus: sata
          readonly: true
```

stu-gott:
brilliant! nice find.

PR #1037, opened 2018-05-16
Fix readonly

stu-gott:
Fixes https://github.com/kubevirt/kubevirt/issues/1036


rmohr:
The question is, if we want it to be filled out. The intention was to not serialize the default values for readOnly if the user does not set it explicitly. So at least from that perspective the correct solution is to check if the value is nil and let the converter in virt-launcher set the right value if it is nil, instead of setting the default value on the cluster.

stu-gott:
@rmohr actually that's a really good point. Moved all logic to the DOM XML conversion side.

PR #1047, opened 2018-05-17
Make virtio the default interface model, and allow to change it with a label

booxter:
This is another attempt to make all machines happy with their network model type. This PR is a spin on the other PR that is closed now: https://github.com/kubevirt/kubevirt/pull/990 The difference here is that instead of leaving all images but cirros at e1000, we make virtio the default model type and fall back to e1000 for win* annotations.

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -943,6 +943,13 @@ func NewRandomVMWithWatchdog() *v1.VirtualMachine {
 	return vm
 }
 
+func NewRandomVMWithe1000NetworkInterface() *v1.VirtualMachine {
+	// Use alpine because cirros dhcp client starts prematurily before link is ready
+	vm := NewRandomVMWithEphemeralDisk(RegistryDiskFor(RegistryDiskAlpine))
+	vm.ObjectMeta.Labels = map[string]string{v1.InterfaceModel: "e1000"}
+	return vm
+}
+
 // Block until the specified VM started and return the target node name.
 func waitForVmStart(vm runtime.Object, seconds int, ignoreWarnings bool) (nodeName string) {
 	_, ok := vm.(*v1.VirtualMachine)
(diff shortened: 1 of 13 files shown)


itamarh:
@booxter if we know the OS, why not have the default come from a preset for the OS, rather than "if" in the code?
my understanding is presets are like config's, which is usually better than keeping logic inside the code?
and of course I assume the user can always override for a specific interface when defining the actual VM, since I prefer virtio for windows (if I can / installed the drivers)?

fabiand:
The reason for this code-level condition is that the vNIC model field is not yet exposed inthe API, thus it needs to be on the code-level.

But this bug actually caused me to preempt. the networking doc here:
https://groups.google.com/forum/#!topic/kubevirt-dev/Tk1bxwRsoXs

With the API in that document we do have the ability to set the model via presets, which would obsolete this approach.

rmohr:
How about adding `alpha.kubevirt.io/interface/model=virtio` and clearly state already right now that it is just temporary until we have it on the UI?

booxter:
The latest upload moves to alpha annotation explicitly for interface model and adds a todo to remove it later.

rmohr:
ok to test

tests/utils.go
```diff
@@ -760,6 +763,10 @@ func NewRandomVMWithEphemeralDiskAndUserdataHighMemory(containerImage string, us
 func NewRandomVMWithEphemeralDisk(containerImage string) *v1.VirtualMachine {
 	vm := NewRandomVM()
 
+	if strings.Contains(containerImage, "winrmcli") {
```
cynepco3hahue: no need, `winrmcli` container runs straightforward on a pod and not under VM
can you add this annotation under `windows_test.go` to `windowsVmSpec`

rmohr:
retest this please

booxter:
The latest PR push sets the annotation in windows_test. It looks like this test requires a disk image that is not freely available to execute, so I can't guarantee it passes. I assume CI will catch issues though.

booxter:
Hold on, there is a bug, and I should probably add more tests for the feature that are not just unit tests.

booxter:
The latest upload fixed a bug where the annotation set was not read by virt-handler making virtio used regardless. In the new upload, we are using metadata labels to set interface model; we also use interface-model instead of interface/model because otherwise API validation fail; and finally, we also add functional tests to check which NIC the guests see (both for default virtio case and for overridden e1000).

booxter:
check-patch and standard-ci failing with:

```
invalid value "KUBEVIRT_GIT_VERSION should be a valid Semantic Version\nPlease see more details here: https://semver.org" for flag -ldflags: missing =<value> in <pattern>=<value>
```

I couldn't find any related issues already reported for the failure, but I also believe my PR is not related to it. I will need more time to dig why CI fails.

rmohr:
@booxter FYI I have removed the standard-ci checks for now. They are basically a duplicate of our "vagrant-dev" lane.

PR #1053, opened 2018-05-21
Wait for virt-handler pods after the destructive test

cynepco3hahue:
Probably fixes `virt-handler` failure under https://github.com/kubevirt/kubevirt/issues/1014#issuecomment-389199091

### DIFF
--- a/tests/vmlifecycle_test.go
+++ b/tests/vmlifecycle_test.go
@@ -322,9 +322,10 @@ var _ = Describe("Vmlifecycle", func() {
 			var vm *v1.VirtualMachine
 			var nodeName string
 			var virtHandler *k8sv1.Pod
+			var virtHandlerAvailablePods int32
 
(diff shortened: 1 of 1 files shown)


rmohr:
Hm, in general I like it. However it can bite us if there is a non-schedulable master involved. Maybe we should check for non-schedulable masters?

cynepco3hahue:
@rmohr I still encounter the error with my PR, so I will need to find some other way to recognize that all current virt-handler pods are good and I will add the check for node schedulable state.

cynepco3hahue:
@rmohr Do you want me to check on taints existence or only check for label `kubevirt.io/schedulable=true` is enough?
Because with taints it can be pretty problematic to consider if it `NoSchedule`(you can have different keys with different values)

cynepco3hahue:
@rmohr I think now it more correct, and it avoids toleration checks at all.

PR #1056, opened 2018-05-22
Don't override reference ovm in case of errors

rmohr:
In some Eventually loops we were overriding the reference ovm with a
fetched entity even in cases where an error was returned. These returned
ovm entities are not properly fetched an mess up the test.

Fixes issues like

```
• Failure [452.107 seconds]
OfflineVirtualMachine
/root/go/src/kubevirt.io/kubevirt/tests/ovm_test.go:47
  A valid OfflineVirtualMachine given
  /root/go/src/kubevirt.io/kubevirt/tests/ovm_test.go:115
    should start and stop VM multiple times [It]
    /root/go/src/kubevirt.io/kubevirt/tests/ovm_test.go:333

    Timed out after 300.000s.
    Expected error:
        <*errors.errorString | 0xc4203a0570>: {
            s: "resource name may not be empty",
        }
        resource name may not be empty
    not to have occurred

    /root/go/src/kubevirt.io/kubevirt/tests/ovm_test.go:144
```

### DIFF
--- a/tests/ovm_test.go
+++ b/tests/ovm_test.go
@@ -139,21 +139,21 @@ var _ = Describe("OfflineVirtualMachine", func() {
 			updatedOVM = updatedOVM.DeepCopy()
 			updatedOVM.Spec.Running = true
 			Eventually(func() error {
-				updatedOVM, err = virtClient.OfflineVirtualMachine(updatedOVM.Namespace).Update(updatedOVM)
+				_, err := virtClient.OfflineVirtualMachine(updatedOVM.Namespace).Update(updatedOVM)
 				return err
 			}, 300*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
 
 			// Observe the VM created
 			Eventually(func() error {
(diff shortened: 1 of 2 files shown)


tests/ovm_test.go
```diff
@@ -169,7 +169,7 @@ var _ = Describe("OfflineVirtualMachine", func() {
 			updatedOVM = updatedOVM.DeepCopy()
 			updatedOVM.Spec.Running = false
 			Eventually(func() error {
-				updatedOVM, err = virtClient.OfflineVirtualMachine(updatedOVM.Namespace).Update(updatedOVM)
```
rmohr: Here was an issue.

tests/ovm_test.go
```diff
@@ -139,21 +139,21 @@ var _ = Describe("OfflineVirtualMachine", func() {
 			updatedOVM = updatedOVM.DeepCopy()
 			updatedOVM.Spec.Running = true
 			Eventually(func() error {
-				updatedOVM, err = virtClient.OfflineVirtualMachine(updatedOVM.Namespace).Update(updatedOVM)
```
rmohr: Here was an issue.

PR #1065, opened 2018-05-23
Get rid of go routines under network test

cynepco3hahue:
go routines under the test make debugging more complicated
and it does not save a lot of time, so let just get rid of them

### DIFF
--- a/tests/vm_networking_test.go
+++ b/tests/vm_networking_test.go
@@ -89,61 +89,51 @@ var _ = Describe("Networking", func() {
 	tests.BeforeAll(func() {
 		tests.BeforeTestCleanup()
 
-		inboundChan := make(chan *v1.VirtualMachine)
-		outboundChan := make(chan *v1.VirtualMachine)
+		// Create and start inbound VM
+		inboundVM = tests.NewRandomVMWithEphemeralDiskAndUserdata(tests.RegistryDiskFor(tests.RegistryDiskCirros), "#!/bin/bash\necho 'hello'\n")
+		inboundVM.Labels = map[string]string{"expose": "me"}
+		inboundVM.Spec.Subdomain = "myvm"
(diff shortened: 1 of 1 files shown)


tests/vm_networking_test.go
```diff
@@ -89,61 +89,51 @@ var _ = Describe("Networking", func() {
...
 
 			// Fetch the new VM with updated status
-			vm, err = virtClient.VM(tests.NamespaceTestDefault).Get(vm.ObjectMeta.Name, v13.GetOptions{})
+			vm, err := virtClient.VM(tests.NamespaceTestDefault).Get(networkVm.Name, v13.GetOptions{})
 			Expect(err).ToNot(HaveOccurred())
 
 			// Lets make sure that the OS is up by waiting until we can login
 			expecter, err := tests.LoggedInCirrosExpecter(vm)
 			defer expecter.Close()
 			Expect(err).ToNot(HaveOccurred())
```
rmohr: because I see it: a `defer` must be done after the err checks. Return values can be nil in case of errors.

tests/vm_networking_test.go
```diff
@@ -89,61 +89,51 @@ var _ = Describe("Networking", func() {
...
+			tests.WaitForSuccessfulVMStart(networkVm)
 
 			// Fetch the new VM with updated status
-			vm, err = virtClient.VM(tests.NamespaceTestDefault).Get(vm.ObjectMeta.Name, v13.GetOptions{})
+			vm, err := virtClient.VM(tests.NamespaceTestDefault).Get(networkVm.Name, v13.GetOptions{})
 			Expect(err).ToNot(HaveOccurred())
 
 			// Lets make sure that the OS is up by waiting until we can login
 			expecter, err := tests.LoggedInCirrosExpecter(vm)
 			defer expecter.Close()
```
rmohr: You can simply call expecter.Close(). No need to wait.

PR #1066, opened 2018-05-23
Improve ovm tests again

rmohr:
The last "fix" introduced a new issue which led to never updated ovms,
in case the ovm was modified in the meantime.

### DIFF
--- a/tests/ovm_test.go
+++ b/tests/ovm_test.go
@@ -131,17 +131,17 @@ var _ = Describe("OfflineVirtualMachine", func() {
 
 		startOVM := func(ovm *v1.OfflineVirtualMachine) *v1.OfflineVirtualMachine {
 			By("Starting the VM")
-			var err error
+
+			Eventually(func() error {
+				updatedOVM, err := virtClient.OfflineVirtualMachine(ovm.Namespace).Get(ovm.Name, &v12.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


tests/ovm_test.go
```diff
@@ -131,18 +131,18 @@ var _ = Describe("OfflineVirtualMachine", func() {
...
-			var err error
-
-			updatedOVM, err := virtClient.OfflineVirtualMachine(ovm.Namespace).Get(ovm.Name, &v12.GetOptions{})
-			Expect(err).ToNot(HaveOccurred())
 
-			updatedOVM = updatedOVM.DeepCopy()
-			updatedOVM.Spec.Running = true
 			Eventually(func() error {
-				_, err := virtClient.OfflineVirtualMachine(updatedOVM.Namespace).Update(updatedOVM)
+				updatedOVM, err := virtClient.OfflineVirtualMachine(ovm.Namespace).Get(ovm.Name, &v12.GetOptions{})
```
cynepco3hahue: hm, why do we need to run the update under `Eventually`?

rmohr: If the cluster updates the ovm document, the resource version is increased. If do not send this new resource version with the PUT, then we are notified that we should retry. We should do a patch here. I am waiting for #1026. This will expose `Patch` on our client. Then we can get rid of the `Eventually` code.

cynepco3hahue: ok thx for explanation

Issue #1067, opened 2018-05-23
Network API


fabiand:
KubeVirt needs to gain an API to specify the guest facing aspects of networking.

In addition this API should allow to connect to different backends.

As there are a few ways to connect backends and frontends, some options around how to connect them should also be exposed.

Design: https://docs.google.com/document/d/10rXr91aqn8MvVcLgHw33WX8BaQwHPZERp25PHxoZGgw/edit#heading=h.nbihdrjh9f07

fabiand:
Scope of this card is to settle on an API scheme. Different issues the cover implementing different backends and frontends and connections.

fabiand:
Closing this issue, as it's merged.

PR #1075, opened 2018-05-25
Introducing the Network API basics

vladikr:
This patch begins the implementation of the evolved network API, suggested by #1067
It introduces Networks to the VM spec and Interfaces to VM spec devices.
- Networks
A list of networks that represent a physical device that can be attached to the Virtual Machine network interface.
- Interfaces
A list of Virtual interfaces that can be created for the virtual machine. The interface also provides a mechanism that will be used to connect to the physical device specified by the network.

However, this PR doesn't attempt to implement multi-interface/network support, but rather it is trying to keep the existing functionality in which we still support a single pod network interface.

It also re-works the networking code base to prepare it for an addition of various interface types, each with its own attachment mechanism.
Introduces an interface that implements Plug and Unplug to simplify the connection of interfaces. 

Note: By not specifying any interfaces and networks, pod network interface will be bridged and it's MAC and IP will be delegated to the VM.

```yaml
kind: VM
spec:
  devices:
    interfaces:
      - name: red
        bridge: ##
          delegateIp: true # offers the ip in case that the source has an ip
  networks:
  - name: red
    pod: {} # Stock pod network
```


ovirt-infra:
Some tests failed

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -0,0 +1,239 @@
...
+func (l *PodInterface) Plug(iface *v1.Interface, network *v1.Network, domain *api.Domain) error {
+	precond.MustNotBeNil(domain)
+	initHandler()
+
+	driver, err := getBinding(iface)
+	if err != nil {
+		return err
+	}
+
+	// There should alway be a pre-configured interface for the default pod interface.
```
stu-gott: It still might be worth double checking that there is one and returning an error. Otherwise this will panic.

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -0,0 +1,315 @@
...
+func (h *NetworkUtilsHandler) GetMacDetails(iface string) (net.HardwareAddr, error) {
+	currentMac, err := lmf.GetCurrentMac(iface)
+	if err != nil {
+		log.Log.Reason(err).Errorf("failed to get mac information for interface: %s", iface)
+		return nil, err
+	}
+	return currentMac, nil
+}
+
+// RandomizeMac returns a random MAC address
```
stu-gott: Is it worth allowing the first three octets to be set arbitrarily? Even for virtual devices, generally only the last 24 bits are truly random.

e.g. KVM guest mac addresses should start with "52:54:00"

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	scanner := bufio.NewScanner(strings.NewReader(content))
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, nameserverPrefix) {
+			nameserver := re.FindString(line)
+			if nameserver != "" {
+				nameservers = append(nameservers, net.ParseIP(nameserver).To4())
```
stu-gott: In a pod config in 2018, it might be reasonable to assume in general that the addresses listed will be IPv4, but I think there might be a better forward-looking approach. This line in particular is more likely than others to suffer from issues even if we're not truly ready to support IPv6 yet.

pkg/virt-launcher/virtwrap/network/common.go
fabiand: We should use some prefix, using qemus sounds good.

vladikr: Yep, I'll get rid of this method - I've missed it when I was removing the proxy. 
However, good point! I'll change SetRandomMacAddr to SetRandomMacAddrSameVendor and will get the prefix from the pod interface.

vladikr:
The latest update adds some API schema tests, removes networkName from the interface struct and uses same vendor prefix when randomizing pod nic mac address.

booxter:
Review status: 0 of 23 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[api/openapi-spec/swagger.json, line 3296 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEAzjwb77VC2YdWFNhi:-LEAzjwb77VC2YdWFNhj:buylg25) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/api/openapi-spec/swagger.json#L3296)):*
> ```JSON
>      },
>      "interfaces": {
>       "description": "Interfaces describe network interface which can be added to the vm",
> ```

-> describe network interaces

Also, the language should change from 'can' to 'are': -> which are added to the vm (comments in this file generally apply to other places where the info is duplicated).

---

*[api/openapi-spec/swagger.json, line 3620 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEAwqeB1zXIbseJLiqR:-LEAwqeB1zXIbseJLiqS:b-m7v83m) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/api/openapi-spec/swagger.json#L3620)):*
> ```JSON
>      },
>      "name": {
>       "description": "Logical name of the interface as well as a reference to an associated networks\nMust match the Name of a Network",
> ```

-> to the associated network

---

*[api/openapi-spec/swagger.json, line 3627 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEAxpe65E1hDWSa_kYO:-LEAxpe65E1hDWSa_kYP:b-u8do36) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/api/openapi-spec/swagger.json#L3627)):*
> ```JSON
>    "v1.InterfaceBridge": {
>     "properties": {
>      "delegateIp": {
> ```

do we need to have a default value provided for the attribute for documentation purposes?

---

*[pkg/api/v1/schema.go, line 620 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEB-m9I0dyflQOsjbZ0:-LEB-m9I0dyflQOsjbZ1:bi3bg4e) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/pkg/api/v1/schema.go#L620)):*
> ```Go
> 	// Must match the Name of a Network
> 	Name string `json:"name"`
> 	// BinindMethod specifies the method which will be used to connect the interface to the guest
> ```

BinindMethod -> BindingMethod here and everywhere

---

*[pkg/api/v1/schema.go, line 625 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEB-uLw1jRFNmvOdFr7:-LEB-uLw1jRFNmvOdFr8:b-jkjruo) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/pkg/api/v1/schema.go#L625)):*
> ```Go
> }
> 
> // Represents the method which will be used to connet the interface to the guest.
> ```

-> connect

---

*[pkg/api/v1/schema_test.go, line 332 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEB0K4nACeA1lj-1hmu:-LEB0K4nACeA1lj-1hmv:b-mc49zs) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/pkg/api/v1/schema_test.go#L332)):*
> ```Go
> 			buf, err := json.MarshalIndent(*exampleVM, "", "  ")
> 			Expect(err).To(BeNil())
> 			fmt.Println(string(buf))
> ```

are these remnants to remove?

---

*[pkg/virt-launcher/virtwrap/network/common.go, line 133 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEB5ZTjElJDXS1A9XVa:-LEB5ZTjElJDXS1A9XVb:br2sp0o) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/pkg/virt-launcher/virtwrap/network/common.go#L133)):*
> ```Go
> 	changed, err := lmf.SpoofMacRandom(iface, false)
> 	if err != nil {
> 		log.Log.Reason(err).Errorf("failed to spoof MAC for iface: %s", iface)
> ```

probably a nit but it's better to avoid using jargon like 'iface' in messages. Replace with 'interface' maybe. (Same suggestion below.)

---

*[pkg/virt-launcher/virtwrap/network/common.go, line 142 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEB7JJeEl3vvIAF62a9:-LEB7JJeEl3vvIAF62aA:b-p4vg2z) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/pkg/virt-launcher/virtwrap/network/common.go#L142)):*
> ```Go
> 			return nil, err
> 		}
> 		log.Log.Reason(err).Errorf("updated Mac for iface: %s - %s", iface, mac)
> ```

Please use MAC not Mac. The latter is an Apple computer. :)

---

*[pkg/virt-launcher/virtwrap/network/network_test.go, line 54 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEBFFkk4YYbWUGjKS7Z:-LEBFFkk4YYbWUGjKS7_:bxy20en) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/pkg/virt-launcher/virtwrap/network/network_test.go#L54)):*
> ```Go
> 			api.SetObjectDefaults_Domain(domain)
> 			iface := getDefaultNetworkInterface()
> 			defaultNet := &v1.Network{
> ```

I think you could have another helper for getDefaultNetwork and reuse it here and in network.go to avoid duplication

---

*[pkg/virt-launcher/virtwrap/network/network_test.go, line 65 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEBFdz3CQcvg7N2hsIE:-LEBFdz42Rwa7xdT7mZA:bshwthn) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/pkg/virt-launcher/virtwrap/network/network_test.go#L65)):*
> ```Go
> 			Expect(err).To(BeNil())
> 		})
> 		It("should fail when no network is sepecified", func() {
> ```

specified

---

*[pkg/virt-launcher/virtwrap/network/network_test.go, line 72 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEBGLRW8WbqbXLwOOfE:-LEBGLRX78ru_LSekqs3:br7wkp8) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/pkg/virt-launcher/virtwrap/network/network_test.go#L72)):*
> ```Go
> 			vm.Spec.Domain.Devices.Interfaces = []v1.Interface{*iface}
> 			err := SetupNetworkInterfaces(vm, domain)
> 			Expect(err).To(Equal(fmt.Errorf("no networks were specified for interface %s", iface.Name)))
> ```

I wonder if there is a better way to distinguish between errors than just comparing strings of log messages. (preferably type constrained)

---

*[pkg/virt-launcher/virtwrap/network/podinterface.go, line 59 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEBHhLA4e6cw9a9xROF:-LEBHhLA4e6cw9a9xROG:bixs5x6) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/pkg/virt-launcher/virtwrap/network/podinterface.go#L59)):*
> ```Go
> 
> 	// There should alway be a pre-configured interface for the default pod interface.
> 	defaultIconf := domain.Spec.Devices.Interfaces[0]
> ```

I don't think in general there is (will be) a guarantee that the pod network interface is the first in the list though.  I understand that this code is already present in the tree but now that the PR changes assumptions to enable support for multiple networks / interfaces, I don't think we can any longer leave it as-is. It seems like we may need to have a helper (getDefaultPodInterface?) that would find the interface that matches a pod network.

There are some complications to solve along the way; for example, is it possible to have two interfaces matching pod network? It doesn't seem like current implementation is capable of doing it, or even should support it. If so, do we have API level validators to forbid this situation? Maybe I've missed them.

---

*[pkg/virt-launcher/virtwrap/network/podinterface.go, line 71 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEBJWeG7bqVO86RvKBf:-LEBJLz36yyyH2d5DVX3:b-hmwffj) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/pkg/virt-launcher/virtwrap/network/podinterface.go#L71)):*
> ```Go
> 			return err
> 		}
> 		if err := driver.preparePodNetworkInterfaces(); err != nil {
> ```

Again, I understand that it was the same before your PR but considering that we add support for multiple networks, shouldn't we revisit the name of the prepare function? It doesn't seem like it acts, or should act, on multiple interfaces, then why does it refer to the interface in plural? That's confusing, especially in scope of the ongoing networking API work.

---

*[pkg/virt-launcher/virtwrap/network/podinterface_test.go, line 106 at r1](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEBMIofClHHJmvaiRd3:-LEBMIofClHHJmvaiRd4:b-68rauk) ([raw file](https://github.com/kubevirt/kubevirt/blob/bb8dc4ec402984de6b6487392c799e8a30d00d1d/pkg/virt-launcher/virtwrap/network/podinterface_test.go#L106)):*
> ```Go
> 		mockNetwork.EXPECT().LinkSetUp(dummy).Return(nil)
> 		mockNetwork.EXPECT().LinkAdd(bridgeTest).Return(nil)
> 		mockNetwork.EXPECT().LinkByName(api.DefaultBridgeName).Return(bridgeTest, nil)
> ```

Some of these checks don't validate behavior but internal implementation. For example, why should we care that LinkByName was called? What we probably care is that LinkSetUp was later called with the right device name, and only that. I think it should be possible to skip some of the checks that are of no relevance to function behavior.

---


*Comments from [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/1075)*

booxter:
I agree with Yuval there should be additional integration tests to cover the changes. I understand that existing tests may already catch a number of potential regressions, but having more complex definitions validated end to end would be nice (for what I understand, the current code will only work with implicit and explicit pod definition; the former is already covered, the latter would benefit from additional test case(s)).

pkg/virt-launcher/virtwrap/network/common.go
stu-gott: Upon re-reading my comment, I'm not sure I was completely clear about what my concern is: When I look at the `/etc/resolv.conf` on my system there's three "nameserver" lines. One is IPv4 and two are IPv6. The same thing can and will happen in clusters/pods where KubeVirt is deployed. Since we're consuming the contents of `/etc/resolv.conf` (and the addresses contained there come from sources beyond the scope of KubeVirt) we can't assume IPv4-only addressing schemes. Ideally we'd use IPv6 addresses found here, but at a minimum this block needs to at least check for and discard them.

booxter:
:lgtm:

Things to follow up with are:

- additional integration test case using explicit pod network definition;
- (maybe) using "default" keyword in swagger definition for delegateIp;
- rename InterfaceBiningMethod into InterfaceBindingMethod.

The way I understood it, the PR was not meant to support multiple network definitions, and this piece belongs to a separate PR to merge after this one. That other piece will unblock support for additional network types (proxy, DP-based). We just decided I am taking that missing piece.

---

Review status: all files reviewed at latest revision, all discussions resolved.

---

*[api/openapi-spec/swagger.json, line 3628 at r3](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEFhh7nEICy7fiYrxQ0:-LEFhh7nEICy7fiYrxQ1:b-eebx1z) ([raw file](https://github.com/kubevirt/kubevirt/blob/670fb2ede4c71845434539332ce807a00dc51750/api/openapi-spec/swagger.json#L3628)):*
> ```JSON
>     "properties": {
>      "delegateIp": {
>       "description": "offers the source interface ip to the guest interface\nDefaults to true",
> ```

Thanks for adding the default to the description though I was more thinking about a special "default" key that swagger supports to document default values for attributes.

---

*[pkg/api/v1/deepcopy_generated.go, line 777 at r3](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-LEFiAXP6mTkjgxYEQIi:-LEFiAXP6mTkjgxYEQIj:b-5stt07) ([raw file](https://github.com/kubevirt/kubevirt/blob/670fb2ede4c71845434539332ce807a00dc51750/pkg/api/v1/deepcopy_generated.go#L777)):*
> ```Go
> func (in *Interface) DeepCopyInto(out *Interface) {
> 	*out = *in
> 	in.InterfaceBiningMethod.DeepCopyInto(&out.InterfaceBiningMethod)
> ```

InterfaceBin*d*ingMethod :)

---


*Comments from [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/1075#-:-LEFjHGEAZoZilADjnoq:b5ebqmx)*

booxter:
There is one test failure related to network connectivity when pinging google.com with max MTU allowed, worth checking before merging.

vladikr:
retest this please

vladikr:
https://github.com/kubevirt/user-guide/pull/72

fabiand:
@booxter so - how does reviewable look? :)

vladikr:
retest this please

api/openapi-spec/swagger.json
```diff
@@ -3601,6 +3608,28 @@
...
+     },
+     "name": {
+      "description": "Logical name of the interface as well as a reference to the associated networks\nMust match the Name of a Network",
+      "type": "string"
+     }
+    }
+   },
+   "v1.InterfaceBridge": {
+    "properties": {
+     "delegateIp": {
```
rmohr: Can we introduce `delegateIp` once it is used?

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -0,0 +1,303 @@
...
+	currentMac, err := lmf.GetCurrentMac(iface)
+	if err != nil {
+		log.Log.Reason(err).Errorf("failed to get mac information for interface: %s", iface)
+		return nil, err
+	}
+	return currentMac, nil
+}
+
+// SetRandomMacAddrSameVendor changes the MAC address for a agiven interface
+func (h *NetworkUtilsHandler) SetRandomMacAddrSameVendor(iface string) (net.HardwareAddr, error) {
```
rmohr: Can we call that something like MoveMacToVirtualMachineInterface? Still not perfect and not describing everything but less cryptic. I would add the random mac to the description.

api/openapi-spec/swagger.json
vladikr: Yes, I'll remove it for now.

pkg/virt-launcher/virtwrap/network/common.go
rmohr: Great!

vladikr:
> Add the default interface and network to the VirtualMachine defaulter

I'll follow up on that in another PR, if it's okay?

vladikr:
ok to test

booxter:
For the record, I'm adding a functional test for the explicit pod interface spec case here: https://github.com/booxter/kubevirt/commit/c91edb7abda5cb59501baf8c391cb835aa874b36 I will propose it for review once this PR is merged, or as part of the PR that will add support for multiple interface definitions (depending on how quick we'll merge the base PR).

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -0,0 +1,244 @@
...
+
+	"github.com/vishvananda/netlink"
+
+	"kubevirt.io/kubevirt/pkg/api/v1"
+	"kubevirt.io/kubevirt/pkg/log"
+	"kubevirt.io/kubevirt/pkg/precond"
+	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
+)
+
+var bridgeFakeIP = "169.254.75.86/32"
```
SchSeba: I think we need this to be configurable from the user side.
We also need to create a range not a single ip address one for the bridge ip (vm default gateway) and one for the vm ip address

pkg/api/v1/defaults.go
```diff
@@ -151,6 +152,34 @@ func setDefaults_DiskFromMachineType(obj *VirtualMachine) {
 	}
 }
 
+func setDefaults_NetworkInterface(obj *VirtualMachine) {
+	interfaces := obj.Spec.Domain.Devices.Interfaces
+	if len(interfaces) == 0 {
```
rmohr: I would do that differently:

 * First check if `networks` is empty or does not contain a pod network.
 * If one of the above is true:
      -  Add the network with a specific name.
      - Add an interface with the same name.
 * Else, do nothing.

That seems to be the more correct approach.

vladikr:
retest this please

vladikr:
I'll resubmit without the latest functional test. We'll add it as a follow-up.

vladikr:
Failures are not related to networking

fabiand:
retest this please

booxter:
retest this please

booxter:
I will propose the functional test case dropped in a follow-up PR. It wasn't hard to fix it to accommodate for new VM defaults.

fabiand:
sgtm @booxter 

The test failures seem to be unrelated but rather flakyness.

Anything else needed?

fabiand:
The tests pass on two platforms. The other test failures seem to be unrelated.

jerry7z:
tests/registry_disk_test.go also failed by this. vm spac changed.

PR #1088, opened 2018-05-29
Move interface-model annotation to annotations

rmohr:
The kubevirt.io/interface-model annotation was accidentally implemented as
a label. This would influence commands like `virtctl expose` in an
unexpected way.


rmohr:
retest this please

PR #1090, opened 2018-05-29
Really fix virt-handler test

cynepco3hahue:
Need to check that all DS values equal to desired one


cynepco3hahue:
@davidvossel The problem was that next test failed on https://github.com/kubevirt/kubevirt/issues/1014#issuecomment-389199091
- `virt-handler` reduces the number of desired pods
- test finished but DS wasn't fast enough to up new `virt-handler` pods
- next test started and choose `virt-handler` pod that DS will replace by updated pod

cynepco3hahue:
I think it does not the single test when we do not provide normal tear down

PR #1094, opened 2018-05-30
Apply default values

### LINKED ISSUE #1068: server error. command Launcher.Sync failed: virError
Deployed a VM from [working template](https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/master/manifests/working-template.yml), and in status I see the following
```
[root@cnv-executor-vparekhmiq-master1 ~]# oc get vms vparekh.test-432001 -n default -o yaml
apiVersion: kubevirt.io/v1alpha1
kind: VirtualMachine
metadata:
  annotations:
    presets.virtualmachines.kubevirt.io/presets-applied: kubevirt.io/v1alpha1
  clusterName: ""
  creationTimestamp: 2018-05-23T12:29:38Z
  finalizers:
  - foregroundDeleteVirtualMachine
  generateName: vparekh.test-432001
  generation: 1
  labels:
    kubevirt.io/nodeName: cnv-executor-vparekhmiq-master1.example.com
  name: vparekh.test-432001
  namespace: default
  ownerReferences:
  - apiVersion: kubevirt.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: OfflineVirtualMachine
    name: vparekh.test-432001
    uid: 55d8b7ac-5e81-11e8-b92b-fa163e5e0d6b
  resourceVersion: "34131"
  selfLink: /apis/kubevirt.io/v1alpha1/namespaces/default/virtualmachines/vparekh.test-432001
  uid: f58db24d-5e84-11e8-b92b-fa163e5e0d6b
spec:
  domain:
    cpu:
      cores: 1
    devices:
      disks:
      - disk: {}
        name: registrydisk
        volumeName: registryvolume
      - disk: {}
        name: cloudinitdisk
        volumeName: cloudinitvolume
    firmware:
      uuid: 10e10d24-8fc9-51c0-bb00-2d1787c1c656
    machine:
      type: q35
    resources:
      requests:
        memory: 1Gi
  volumes:
  - name: registryvolume
    registryDisk:
      image: kubevirt/fedora-cloud-registry-disk-demo:latest
  - cloudInitNoCloud:
      userDataBase64: I2Nsb3VkLWNvbmZpZwpwYXNzd29yZDogYXRvbWljCnNzaF9wd2F1dGg6IFRydWUKY2hwYXNzd2Q6IHsgZXhwaXJlOiBGYWxzZSB9Cg==
    name: cloudinitvolume
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: 2018-05-23T12:29:53Z
    message: 'server error. command Launcher.Sync failed: virError(Code=20, Domain=20,
      Message=''missing target information for device /var/run/libvirt/kubevirt-ephemeral-disk/registry-disk-data/default/vparekh.test-432001/disk_registryvolume/disk-image.qcow2'')'
    reason: Synchronizing with the Domain failed.
    status: "False"
    type: Synchronized
  interfaces:
  - ipAddress: x.x.x.x
  nodeName: cnv-executor-vparekhmiq-master1.example.com
  phase: Scheduled
```

slintes:
Fixes #1068

### DIFF
--- a/pkg/virt-controller/watch/preset.go
+++ b/pkg/virt-controller/watch/preset.go
@@ -152,6 +152,9 @@ func (c *VirtualMachinePresetController) initializeVirtualMachine(vm *kubev1.Vir
 		if !success {
 			logger.Object(vm).Warning("Marking VM as failed")
 			vm.Status.Phase = kubev1.Failed
+		} else {
+			logger.Object(vm).Info("Setting default values on VirtualMachine")
+			kubev1.SetObjectDefaults_VirtualMachine(vm)
 		}
 	} else {
 		logger.Object(vm).Infof("VM is excluded from VirtualMachinePresets")
--- a/pkg/virt-controller/watch/preset.go
+++ b/pkg/virt-controller/watch/preset.go
@@ -153,7 +153,7 @@ func (c *VirtualMachinePresetController) initializeVirtualMachine(vm *kubev1.Vir
 			logger.Object(vm).Warning("Marking VM as failed")
 			vm.Status.Phase = kubev1.Failed
 		} else {
-			logger.Object(vm).Info("Setting default values on VirtualMachine")
+			logger.Object(vm).V(4).Info("Setting default values on VirtualMachine")
 			kubev1.SetObjectDefaults_VirtualMachine(vm)
 		}
 	} else {
--- a/pkg/virt-controller/watch/preset_test.go
+++ b/pkg/virt-controller/watch/preset_test.go
@@ -1021,6 +1021,35 @@ var _ = Describe("VM Initializer", func() {
 			// Prove that the VM was initialized
 			Expect(isVirtualMachineInitialized(vm)).To(BeTrue())
 		})
+
+		It("should set default values to VM", func() {
+			vm := v1.NewMinimalVM("testvm")
+			vm.Spec = v1.VirtualMachineSpec{
+				Domain: v1.DomainSpec{
+					Devices: v1.Devices{
+						Disks: []v1.Disk{
+							{Name: "testdisk"},
+						},
+					},
+				},
+			}
+
+			// Register the expected REST call
+			server.AppendHandlers(
+				ghttp.CombineHandlers(
+					ghttp.VerifyRequest("PUT", "/apis/kubevirt.io/v1alpha1/namespaces/default/virtualmachines/testvm"),
+					ghttp.RespondWithJSONEncoded(http.StatusOK, vm),
+				),
+			)
+
+			err := app.vmPresetController.initializeVirtualMachine(vm)
+
+			Expect(err).ToNot(HaveOccurred())
+
+			disk := vm.Spec.Domain.Devices.Disks[0]
+			Expect(disk.Disk).ToNot(BeNil(), "DiskTarget should not be nil")
+			Expect(disk.Disk.Bus).ToNot(BeEmpty(), "DiskTarget's bus should not be empty")
+		})
 	})
 })
 
--- a/tests/vmdefaults_test.go
+++ b/tests/vmdefaults_test.go
@@ -1,0 +1,83 @@
+/*
+ * This file is part of the KubeVirt project
+ *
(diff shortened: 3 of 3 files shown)


pkg/virt-controller/watch/preset.go
```diff
@@ -152,6 +152,9 @@ func (c *VirtualMachinePresetController) initializeVirtualMachine(vm *kubev1.Vir
 		if !success {
 			logger.Object(vm).Warning("Marking VM as failed")
 			vm.Status.Phase = kubev1.Failed
+		} else {
+			logger.Object(vm).Info("Setting default values on VirtualMachine")
```
rmohr: That may deserve a higher verbosity level. Maybe `V(4)`?

rmohr:
ok to test

pkg/virt-controller/watch/preset.go
slintes: ok, trying to understand how the logger works...
- both `verbosityLevel` and `currentVerbosityLevel` are `0` or whatever is given with verbosityFlag by default... that means info logs are always logged by default because both values are equal, right? Wouldn't it make more sense to set `verbosityLevel` to 2 or 3 by default?
- when I change the `currentVerbosityLevel` with `V(x)`, it will stay at that level until i call `V(x)` again. Is that on purpose?

rmohr: You have to use `V(X)` to specify the minimum vervosity level which needs to be configured to make the message appear. The minimum level can be set via the `-v` flag. So if you configure the logger with `-v 3` on the commandline, all  messages which have used `V(3)` or below will be logged. In other words, the more logs you want to see the higher you set the verbosity level on the commandline.

rmohr: > when I change the currentVerbosityLevel with V(x), it will stay at that level until i call V(x) again. Is that on purpose?

No. It should only be set for invocations after `V()`, e.g. `V(3).Info(""). That looks like a bug.

So this is the desired behaviour:

```
mylogger := logger.Object(vm).V(3) # sets 3 for mylogger
mylogger.Info("lal") # uses 3 for logging
logger.Object(vm).Info("lala") # should use 0
mylogger.V(5).Info("blub") # should use 5
mylogger.Info("lal") # uses still 3
```

rmohr: When looking into the sourcecode, we basically want a copy of the struct but right now it returns a pointer to it.

slintes: Ok, will add the V(4) in a second.
I will have a look into the copy/pointer topic tomorrow and create an issue if needed.
I still think that the initialization of `verbosityLevel` and `currentVerbosityLevel` could be improved... will also create an issue about it and explain.
Thx for reviewing!

pkg/virt-controller/watch/preset.go
```diff
@@ -152,6 +152,9 @@ func (c *VirtualMachinePresetController) initializeVirtualMachine(vm *kubev1.Vir
 		if !success {
 			logger.Object(vm).Warning("Marking VM as failed")
 			vm.Status.Phase = kubev1.Failed
+		} else {
+			logger.Object(vm).V(4).Info("Setting default values on VirtualMachine")
+			kubev1.SetObjectDefaults_VirtualMachine(vm)
```
cynepco3hahue: We need to be careful to not override preset changes with our default values, maybe does it better to apply default values before preset values?

slintes: defaults are only set when sth is missing, it won't override anything. So I think first setting presets, and then defaults, is the correct order. Agree?

cynepco3hahue: If we assume that `SetObjectDefaults_VirtualMachine` does not have bugs, I agree :smiley: :

rmohr: Yes, presets first. We also want to ensure that structs set or added by presets get filled with missing defaults.

slintes:
added functional test

rmohr:
retest this please

Issue #395, opened 2017-08-29
Support for additional architectures


fabiand:
We might want to build KubeVirt for [arm64 and ppc64le (as requested)](https://github.com/kubevirt/kubevirt/pull/384#issuecomment-325285624) as well.

- [ ] Multi arch support for binaries
- [ ] Reflect arch in containers
- [ ] minikube on desired arch
- [ ] vagrant and tests on desired arch

mpolednik:
How does arm64 make sense when KVM barely works there? How is kubernetes doing on those architectures? We should not do the mistake of building for architectures that don't really work (where working means running docker, KVM, kubernetes). I'm not sure about vagrant *right now* either, because it's gonna be slow (emulated) and ppc64le doesn't have nested virtualization. Not sure about arm in this regard.

fabiand:
This ticket is intentionally named to support additional architectures, without naming them.
It's to keep in mind that we should consider this in our tools, build systems, and ci.

I'd also put it up to discussion what arches we need to support - and who implements this support - but so far arm64 and ppc64le were explicitly requested (see the added link).

fabiand:
To stretched atm, closing this.

PR #1097, opened 2018-05-31
Make possible to consume hugepages

cynepco3hahue:
- [x] add unittests
- [x] add functional tests
- [x] add validation on memory and hugepages resources

Open Questions:
Do we want to make hugepages and memory requests
mutually exclusive?
- yes, so we will need to pass some defaul amount
of memory to virt-launcher pod(QEMU overhead)
- no, we will need to validate, that memory requests
and VM requests have the same value


mpolednik:
Something to consider: 2 MiB and 1 GiB hugepages are x86_64 sizes, and even there 1 GiB pages require pdpe1gb CPU flag. It might make sense to build the code robust enough to handle sizes for different platforms too - Kubernetes needs to be clever enough to select a node that supports the requested size, and from that point on KubeVirt doesn't need additional logic other than convert the pod requirement to a VM specification.

cynepco3hahue:
@mpolednik I remember we had the issue https://github.com/kubevirt/kubevirt/issues/395 about building KubeVirt packages for additional architectures, so I do not see the reason to support additional architectures until we consider that we want them. And currently, we do not have a proper CI to verify that it will work with other huge pages sizes except 2Mi and 1Gi.

michalskrivanek:
@cynepco3hahue  mpolednik's comment still stands.  It's fine if it now works in CI only for certain size, but we do want to consider other sizes too, please do not hardcode sizes in any way.

rmohr:
> @cynepco3hahue mpolednik's comment still stands. It's fine if it now works in CI only for certain size, but we do want to consider other sizes too, please do not hardcode sizes in any way.

Fine for me. This PR is finally the one which should start propagating scheduling errors and delays from the pods to the vm. @cynepco3hahue could you add this here? Should be pretty simple to add. Basically you have to look for the `PodScheduled` condition in virt-controller and update our vm condition with that:

```
  conditions:
  - lastProbeTime: null
    lastTransitionTime: 2018-06-08T07:16:01Z
    message: '0/1 nodes are available: 1 Insufficient memory.'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
```

cynepco3hahue:
ok to test

cynepco3hahue:
retest this please

cynepco3hahue:
OK to test

fabiand:
@cynepco3hahue could you please provide a user-guide update ot get the PAI level insights?

cynepco3hahue:
retest this please

cynepco3hahue:
@fabiand Done - https://github.com/kubevirt/user-guide/pull/82

fabiand:
I have an API change request in here: https://github.com/kubevirt/user-guide/pull/82#pullrequestreview-127548719

pkg/api/v1/schema.go
```diff
@@ -56,6 +56,9 @@ type DomainSpec struct {
 	// CPU allow specified the detailed CPU topology inside the vm.
 	// +optional
 	CPU *CPU `json:"cpu,omitempty"`
+	// Hugepages allow to specify desired hugepages size for the vm usage.
+	// +optional
+	Hugepages *Hugepages `json:"hugepages,omitempty"`
```
rmohr: Can't we simply express that the same way like in k8s unter `resources`?

```
apiVersion: v1
kind: Pod
metadata:
  generateName: hugepages-volume-
spec:
  containers:
  - image: fedora:latest
    command:
    - sleep
    - inf
    name: example
    volumeMounts:
    - mountPath: /hugepages
      name: hugepage
    resources:
      limits:
        hugepages-2Mi: 100Mi
  volumes:
  - name: hugepage
    emptyDir:
      medium: HugePages
```

We can still validatate that only huge-pages or memory is used.

fabiand: In the K8s context a user requests that resource, and he will explicitly get it (as a device).

In the VM context however, the user requests a resource, that is memory, and hugepages are merely a _different backend_ for the requested memory.

IOW: The broad ppicture is that the reqource requests outline what the user will get in the context of his workload. In pods he can have memory and hugepages. But for VMs he will always just get a virtual memory - which can then either be backed by hugepages or not -at least inside the VM he will only have memory.

This is why it is alright to just have a memory requests field, and not have the user to request something else than memory.

Hwoever, if we see that this can not stand over th elong run, then let's revisit the API.

rmohr: > IOW: The broad ppicture is that the reqource requests outline what the user will get in the context of his workload. In pods he can have memory and hugepages. But for VMs he will always just get a virtual memory - which can then either be backed by hugepages or not -at least inside the VM he will only have memory.

Only allowing specifying one of those should be good enough. Simple validation check. Especially from the point of view you bring up, it is strange that you specify hugepages in the domain section, since as you say inside the VM it is only memory, but you need to specify it as a guest detail.

> In the K8s context a user requests that resource, and he will explicitly get it (as a device).

Also in the vm case he explicitly asks for it. If it will not be available the vm will not be scheduled. Very explicit.

> This is why it is alright to just have a memory requests field, and not have the user to request something else than memory.

This proposed implementation in fact moves parts of the scheduling decision regarding to resources into the guest details and hides what will be used by the scheduler.

> Hwoever, if we see that this can not stand over th elong run, then let's revisit the API.
> 
>  Write

Yes.

cynepco3hahue: To be honest resource usage for hugepages will include some different validations that I wanted to avoid.
In the case when you specified hugepages:
- no memory request
- no memory limit
- hugepage request without limit
- different hugepage request and limit
I think k8s resource API does not suit well for resources without overcommit, because it leaves some corner cases, that user will reveal only after the start of the pod

rmohr:
retest this please

cynepco3hahue:
retest this please

cynepco3hahue:
retest this please

pkg/api/v1/schema.go
```diff
@@ -94,6 +97,23 @@ type CPU struct {
...
+}
+
+// Hugepages allow to use hugepages for the VirtualMachineInstance instead of regular memory.
+// ---
+// +k8s:openapi-gen=true
+type Hugepages struct {
+	// PageSize specifies the hugepage size.
+	PageSize string `json:"pageSize,omitempty"`
```
rmohr: Can you add something like `for instance for x86_64 valid values are 1Gi and 2Mi`. To give a hint of the formatting and possible values.

pkg/virt-controller/watch/vmi_test.go
```diff
@@ -324,6 +324,38 @@ var _ = Describe("VirtualMachineInstance watcher", func() {
 			table.Entry(", ready and in failed state", k8sv1.PodFailed, true),
 			table.Entry(", ready and in pending state", k8sv1.PodPending, true),
 		)
+
+		Context("when pod failed to schedule", func() {
+			It("should set scheduling pod condition on the VirtualMachineInstance", func() {
```
rmohr: Can you add  a test which will remove the condition?

tests/vmi_configuration_test.go
```diff
@@ -105,6 +108,93 @@ var _ = Describe("Configurations", func() {
 				Expect(err).ToNot(HaveOccurred())
 			}, 300)
 		})
+
+		Context("with hugepages", func() {
```
rmohr: Can you add a functiona test which creates an unschedulable pod (or adjust one of the existing ones) and check for the schedulable condition?

pkg/api/v1/schema.go
cynepco3hahue: done

pkg/virt-controller/watch/vmi_test.go
cynepco3hahue: Done

tests/vmi_configuration_test.go
cynepco3hahue: Done

rmohr:
@cynepco3hahue I see the unschedulable size test fail on a timeout in two lanes.

cynepco3hahue:
@rmohr Yes I see, I need to skip this test if no hugepages resources appear under the node at all, under 1.9 hugepages still alpha.

cynepco3hahue:
retest this please

Issue #1100, opened 2018-06-03
Incorrect calculation of a VM memory


cynepco3hahue:
I create VM with the specification:
```
...
    resources:
      requests:
        memory: 64Mi
...
```
I assume that VM domain will have 64Mi of memory, but it has 65Mi
```
</metadata>
  <memory unit='KiB'>66560</memory>
  <currentMemory unit='KiB'>66408</currentMemory>
```
When we talk about regular memory, it does not really important, but in case of hugepages, it will fail to start a VM at all.

I think the problem occurs under quantity conversion
```
func QuantityToMegaByte(quantity resource.Quantity) Memory {
	return Memory{
		Value: uint(quantity.ToDec().ScaledValue(6)),
		Unit:  "MB",
	}
}
```

fabiand:
Good catch!

fabiand:
Assigning this to myself to hand it off.

vladikr:
@cynepco3hahue isn't because we add a memory overhead?
Did you request any cpus as well?

cynepco3hahue:
@vladikr We calculate memory overhead for a virt-launcher pod, a VM still must receive the value that we specified under VM spec.
No I did not requested CPU's.

mfranczy:
As @cynepco3hahue mentioned before, there is a problem with
```
Value: uint(quantity.ToDec().ScaledValue(6))
Unit:  "MB",
```
if you will request let's say 67108864 bytes of memory (which is equal to 64MB) then you will receive different value because of scaling and rounding.

To resolve that problem i suggest to change it to:
```
Value: uint64(quantity.AsInt64())
Unit:  "B",
```
and pass memory size in bytes to have exact size which you requested.
It will require change as well in ```Memory.Unit``` type from ```uint``` to ```uint64```

I checked that solution quickly by changing scaling to 0
```
Value: uint(quantity.ToDec().ScaledValue(0))
Unit: "B"
``` 
and it works, below result from my virtual machine:
```
...
<memory unit='KiB'>65536</memory>
<currentMemory unit='KiB'>65536</currentMemory>
<vcpu placement='static'>1</vcpu>
...
```

@cynepco3hahue @fabiand any suggestions or thoughts to this?
If not I will create PR with updated tests

cynepco3hahue:
@mfranczy No objections from my side

fabiand:
Sounds promising, I'd expect some test to the edge cases :)

cynepco3hahue:
[...]

PR #1106, opened 2018-06-06
Fixed memory calculation for virtual machine 

mfranczy:
Bugfix for issue: https://github.com/kubevirt/kubevirt/issues/1100

Changes:
- changed memory unit from MB to B
- extended tests for memory calculation


pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -505,10 +505,14 @@ func SecretToLibvirtSecret(vm *v1.VirtualMachine, secretName string) string {
 	return fmt.Sprintf("%s-%s-%s---", secretName, vm.Namespace, vm.Name)
 }
 
-func QuantityToMegaByte(quantity resource.Quantity) Memory {
+func QuantityToByte(quantity resource.Quantity) Memory {
+	memorySize, _ := quantity.AsInt64()
+	if memorySize < 0 {
```
cynepco3hahue: When can you receive here negative value?

mfranczy: I have added this condition because ParseQuantity accepts negative values, for example:
```ParseQuantity("-64Mi")``` aftter conversion to ```Int64``` i will have value ```-67108864``` and then there is cast to ```uint64``` which gives us result ```18446744073642442752```.

Being honest I don't expect users to specify negative memory values for vm :), but there was this possibility and that's why i added this edge condition, just to be safe.

Comment (cynepco3hahue) [reply pkg/virt-launcher/virtwrap/api/converter.go]: Uh, interesting if k8s accept negative values.

Comment (mfranczy) [reply pkg/virt-launcher/virtwrap/api/converter.go]: hmm i checked only this ```ParseQuantity``` function, maybe they have some validation at some level, I will check real scenario with it.

Comment (rmohr) [reply pkg/virt-launcher/virtwrap/api/converter.go]: Can you add a check for negative quantities in `pkg/virt-api/validating-webhook/validating-webhook.go`? If we would see negative values here, I would expect that virt-handler would reject starting the vm.

Comment (mfranczy) [reply pkg/virt-launcher/virtwrap/api/converter.go]: Sure I will do this.

File: pkg/virt-api/validating-webhook/validating-webhook.go [hunk 2]
```diff
@@ -321,6 +321,17 @@ func validateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
 		return causes
 	}
 
+	// Validate requested memory size if value is not negative
+	if spec.Domain.Resources.Requests.Memory().Value() < 0 {
```
rmohr: Can you do the same for spec.Domain.Resources.Limits.Memory().Value()? Also Please add a unit test.

mfranczy: sure :)

pkg/virt-launcher/virtwrap/api/converter.go
rmohr: Also remove the negative number flip. It should be enought to return an error from the converter with a reason able error message.

fabiand:
Is there actually already a PR on kubevirt/kubevirt to create the template?

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -321,6 +321,25 @@ func validateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
 		return causes
 	}
 
+	// Validate memory size if values are not negative
+	if spec.Domain.Resources.Requests.Memory().Value() < 0 {
```
mfranczy: I was thinking to prepare one function for checking quantity for memory insted of ```if ... else if...``` although I was following code style from volume/disks checks which is simple. Should I change it? I think it might be worth to refactor to one func for negative values when cpu quantity will be here as well.

rmohr: It's very easy to follow like you did it. Good enough for now.

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -321,6 +321,25 @@ func validateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
...
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: fmt.Sprintf("%s is specified as negative value", field.Child("domain", "resources", "requests", "memory").String()),
+			Field:   field.Child("domain", "resources", "requests", "memory").String(),
+		})
+		// We won't process anything with negative requests.memory size
+		return causes
+	} else if spec.Domain.Resources.Limits.Memory().Value() < 0 {
```
rmohr: Here you will need an `if` instead of `else if`. The `requests` value can be set independent of `limits`.

pkg/virt-api/validating-webhook/validating-webhook.go
mfranczy: Done

pkg/virt-launcher/virtwrap/api/converter.go
mfranczy: Done

mfranczy: Oh I said done, but it's not, I will return error from the converter... i just removed this flip... ok I am fixing this

pkg/virt-api/validating-webhook/validating-webhook.go
rmohr: Since you mentioned CPU, we will need validation for it too. Maybe in a follow-up PR? ;)

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -321,6 +321,25 @@ func validateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
...
 
+	// Validate memory size if values are not negative
+	if spec.Domain.Resources.Requests.Memory().Value() < 0 {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: fmt.Sprintf("%s is specified as negative value", field.Child("domain", "resources", "requests", "memory").String()),
+			Field:   field.Child("domain", "resources", "requests", "memory").String(),
+		})
+		// We won't process anything with negative requests.memory size
+		return causes
```
rmohr: Don't return early. We want to go through the whole struct and collect all issues we can find. We can then report all issues at once to the user.

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -321,6 +321,25 @@ func validateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
...
+	} else if spec.Domain.Resources.Limits.Memory().Value() < 0 {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: fmt.Sprintf("%s is specified as negative value", field.Child("domain", "resources", "limits", "memory").String()),
+			Field:   field.Child("domain", "resources", "limits", "memory").String(),
+		})
+		// We won't process anything with negative limits.memory size
+		return causes
```
rmohr: Don't return early. We want to go through the whole struct and collect all issues we can find. We can then report all issues at once to the user.

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -321,6 +321,25 @@ func validateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
 		return causes
 	}
 
+	// Validate memory size if values are not negative
+	if spec.Domain.Resources.Requests.Memory().Value() < 0 {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: fmt.Sprintf("%s is specified as negative value", field.Child("domain", "resources", "requests", "memory").String()),
```
rmohr: Let's align with the error messages from k8s:

```
The Pod "nginx" is invalid: spec.containers[0].resources.requests[memory]: Invalid value: "-4Mi": must be greater than or equal to 0
```

The relevant part is: `"-4Mi": must be greater than or equal to 0`.

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -321,6 +321,25 @@ func validateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
...
+			Field:   field.Child("domain", "resources", "requests", "memory").String(),
+		})
+		// We won't process anything with negative requests.memory size
+		return causes
+	} else if spec.Domain.Resources.Limits.Memory().Value() < 0 {
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: fmt.Sprintf("%s is specified as negative value", field.Child("domain", "resources", "limits", "memory").String()),
```
rmohr: Also this one can be align with k8s: `"-4Mi": must be greater than or equal to 0`.

pkg/virt-api/validating-webhook/validating-webhook.go
mfranczy: Done

pkg/virt-api/validating-webhook/validating-webhook.go
mfranczy: Done

pkg/virt-api/validating-webhook/validating-webhook.go
mfranczy: Done

pkg/virt-api/validating-webhook/validating-webhook.go
mfranczy: Done

pkg/virt-api/validating-webhook/validating-webhook.go
mfranczy: Done

pkg/virt-api/validating-webhook/validating-webhook.go
mfranczy: Yeah I think I can do that, if this PR is good enough to merge then I will have to do something new so I can check CPU as well and improve validation where is needed.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -505,11 +507,15 @@ func SecretToLibvirtSecret(vm *v1.VirtualMachine, secretName string) string {
 	return fmt.Sprintf("%s-%s-%s---", secretName, vm.Namespace, vm.Name)
 }
 
-func QuantityToMegaByte(quantity resource.Quantity) Memory {
-	return Memory{
-		Value: uint(quantity.ToDec().ScaledValue(6)),
-		Unit:  "MB",
+func QuantityToByte(quantity resource.Quantity) (Memory, error) {
+	memorySize, _ := quantity.AsInt64()
+	if memorySize < 0 {
```
cynepco3hahue: you already added this check under validation, so you can remove

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -321,6 +321,25 @@ func validateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
...
+
+	if spec.Domain.Resources.Limits.Memory().Value() < 0 {
+		causes = append(causes, metav1.StatusCause{
+			Type: metav1.CauseTypeFieldValueInvalid,
+			Message: fmt.Sprintf("%s '%s': must be greater than or equal to 0.", field.Child("domain", "resources", "limits", "memory").String(),
+				spec.Domain.Resources.Limits.Memory()),
+			Field: field.Child("domain", "resources", "limits", "memory").String(),
+		})
+	}
+
```
cynepco3hahue: Do we want also to validate that limit is greater than the request?

pkg/virt-launcher/virtwrap/api/converter.go
rmohr: I asked for that check. My reasoning was that we currently have nothing which checks if the validation webhook really works. I am unsure.

pkg/virt-api/validating-webhook/validating-webhook.go
rmohr: Good point. we need that check too.

pkg/virt-launcher/virtwrap/api/converter.go
cynepco3hahue: Srry, missed your comment

pkg/virt-api/validating-webhook/validating-webhook.go
cynepco3hahue: By the way, do we have a possibility to validate virt-launcher pod spec with k8s validator on VM validation stage?
It can save use some job.

mfranczy: I have added validation if limit is greater than request

mfranczy:
I think I added everything what was requested to this PR and ofc I tested this as well.

cynepco3hahue:
OK to test

rmohr:
ok to test

PR #1108, opened 2018-06-06
Fix some bumps in the getting started guide

rmohr:
* Make that mentioned vm examples work again
 * Mention nested virtualization as a requirement


rmohr:
@mfranczy FYI

PR #1109, opened 2018-06-06
Better default log verbosity initialization

### LINKED ISSUE #1098: Proposal: better initialization of log verbosity level
Neither that important nor urgent, but this catched my eyes when looking into our logging implementation:

Currently we initialize both `verbosityLevel` and `currentVerbosityLevel` level to the same value, either what's given with the verbosity command line flag, or `0` [1].
Also, info level logs are actually logged when `currentVerbosityLevel <= verbosityLevel` [2]. That means as long as the `V(currentVerbosityLevel)` func isn't called, all `Info(...)` logs are always logged, even when the verbosity flag is specified with `--v=0`... not sure if that is what we want, and what the user expects.

Proposal:

- According to [3] verbosity level `2` is the "recommended default log level for most systems". So if no verbosity flag is specified, set `verbosityLevel` to `2`.
- I think `currentVerbosityLevel` should be initialized indepent of `verbosityLevel`, always to the same value. A good value might be `2` or `3` (see [3]):
    - `2` means: "Useful steady state information about the service and important log messages that may correlate to significant changes in the system."
    - `3` means: "Extended information about changes. More info about system state changes."
    I'd propose to use `2`, because
        - that would mean no change in behaviour compared to now when no verbosity flag is provided: `Info(...)` logs without `V(...)` will be logged because `currentVerbosityLevel` == `verbosityLevel` == `2`.
        - if verbosity flag `0` or `1` is provided, you see less logs, which is very likely expected behaviour imho 

Wdyt?

[1] https://github.com/kubevirt/kubevirt/blob/master/pkg/log/log.go#L97
[2] https://github.com/kubevirt/kubevirt/blob/master/pkg/log/log.go#L172
[3] https://github.com/kubernetes/community/blob/master/contributors/devel/logging.md

slintes:
Fixes #1098

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -87,15 +87,21 @@ func MakeLogger(logger log.Logger) *FilteredLogger {
 	if verbosityFlag := flag.Lookup("v"); verbosityFlag != nil {
 		defaultVerbosity, _ = strconv.Atoi(verbosityFlag.Value.String())
 	} else {
-		defaultVerbosity = 0
+		// "the practical default level is V(2)"
+		// see https://github.com/kubernetes/community/blob/master/contributors/devel/logging.md
+		defaultVerbosity = 2
 	}
+
+	// This verbosity will be used for info logs without setting a custom verbosity level
+	defaultCurrentVerbosity := 2
+
 	return &FilteredLogger{
 		logContext:            log.NewContext(logger),
 		component:             defaultComponent,
 		filterLevel:           defaultLogLevel,
 		currentLogLevel:       defaultLogLevel,
 		verbosityLevel:        defaultVerbosity,
-		currentVerbosityLevel: defaultVerbosity,
+		currentVerbosityLevel: defaultCurrentVerbosity,
 	}
 }
 
--- a/cmd/virt-launcher/virt-launcher.go
+++ b/cmd/virt-launcher/virt-launcher.go
@@ -230,6 +230,9 @@ func main() {
 	gracePeriodSeconds := flag.Int("grace-period-seconds", 30, "Grace period to observe before sending SIGTERM to vm process")
 	allowEmulation := flag.Bool("allow-emulation", false, "Allow fallback to emulation if /dev/kvm is not present")
 
+	// set new default verbosity, was set to 0 by glog
+	flag.Set("v", "2")
+
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
 
--- a/pkg/log/log_test.go
+++ b/pkg/log/log_test.go
@@ -152,29 +152,35 @@ func TestInfoCutoff(t *testing.T) {
 func TestVerbosity(t *testing.T) {
 	setUp()
 	log := MakeLogger(MockLogger{})
-	if err := log.SetVerbosityLevel(2); err != nil {
+
+	assert(t, log.verbosityLevel == 2, "Default verbosity should be 2")
+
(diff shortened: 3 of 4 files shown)


pkg/log/log.go
```diff
@@ -87,15 +87,21 @@ func MakeLogger(logger log.Logger) *FilteredLogger {
 	if verbosityFlag := flag.Lookup("v"); verbosityFlag != nil {
 		defaultVerbosity, _ = strconv.Atoi(verbosityFlag.Value.String())
 	} else {
-		defaultVerbosity = 0
+		// "the practical default level is V(2)"
+		// see https://github.com/kubernetes/community/blob/master/contributors/devel/logging.md
```
rmohr: Ha, great that you found that. I was looking for something like that several times!

rmohr:
ok to test

rmohr:
@slintes are you also going to tackle the issue that if you call `V(X)` that the verbosity for the whole logger changes, and not juts for the message after `V(X)`?

slintes:
@rmohr that wasn't an issue, works as designed when I look at the tests. I only misunderstood it from looking at the code

pkg/log/log.go
```diff
@@ -87,15 +87,21 @@ func MakeLogger(logger log.Logger) *FilteredLogger {
...
+	defaultCurrentVerbosity := 2
+
 	return &FilteredLogger{
 		logContext:            log.NewContext(logger),
 		component:             defaultComponent,
 		filterLevel:           defaultLogLevel,
 		currentLogLevel:       defaultLogLevel,
 		verbosityLevel:        defaultVerbosity,
-		currentVerbosityLevel: defaultVerbosity,
+		currentVerbosityLevel: defaultCurrentVerbosity,
```
rmohr: Why can't we use `defaultVerbosity` here too?

slintes: Because then all info logs (without any call to `V()`) would always be logged, because `verbosityLevel == currentVerbosityLevel`. But when I set `verbosityLevel` to `0` or `1` with the verbosity flag, I probably don't want to see all of them, only those who explicitly use `V(0)` / `V(1)`.

rmohr:
I see

```
• Failure [31.623 seconds]
Vmlifecycle
/root/go/src/kubevirt.io/kubevirt/tests/vmlifecycle_test.go:45
  Creating a VM
  /root/go/src/kubevirt.io/kubevirt/tests/vmlifecycle_test.go:59
    should attach virt-launcher to it [It]
    /root/go/src/kubevirt.io/kubevirt/tests/vmlifecycle_test.go:71

    Timed out after 11.747s.
    Expected
        <string>: level=error timestamp=2018-06-06T14:24:07.754783Z pos=network.go:152 component=virt-launcher msg="Updated Mac for iface: eth0 - 6e:d9:42:ce:61:eb"
        
    to contain substring
        <string>: Found PID for qemu
```

failing across all lanes. It may be related to the logger changes, since here we are really checking for the log message to appear. Could you run the test locally and check?

slintes:
mh, I see what's going wrong, but not why yet: for some reason the verbosity command line flag defaults to `0` during runtime. And since I changed the `defaultCurrentVerbosity` to 2, info logs are not logged anymore in this case, and the tests looking for those logs fail.
When I add `"--v", "2",` to `virt-launcher` in https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-controller/services/template.go#L172, the tests succeed. But I'd prefer to somehow set the default value of that flag to `2`, will try to find out how to do that tomorrow.

PR #1111, opened 2018-06-06
Remove only KubeVirt secrets on test cleanup

cynepco3hahue:
Before it, we deleted all secrets from test namespace, that force k8s to re-create all necessary tokens for service accounts, in some cases re-creation wasn't fast enough and tests failed with the error:
```
Error creating pod: No API token found for service account \"default\", retry after the token is automatically created and added to the service account
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -127,13 +127,16 @@ const (
 )
 
 const (
-	iscsiIqn        = "iqn.2017-01.io.kubevirt:sn.42"
-	iscsiSecretName = "iscsi-demo-secret"
+	iscsiIqn = "iqn.2017-01.io.kubevirt:sn.42"
 )
 
 const (
 	defaultDiskSize        = "1Gi"
 	defaultWindowsDiskSize = "30Gi"
+)
+
+const (
+	SecretLabel = "kubevirt.io/secret"
 )
 
 const VmResource = "virtualmachines"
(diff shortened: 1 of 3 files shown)


cynepco3hahue:
OK to test

tests/utils.go
```diff
@@ -654,9 +651,6 @@ func cleanNamespaces() {
 		// Remove all Pods
 		PanicOnError(virtCli.CoreV1().RESTClient().Delete().Namespace(namespace).Resource("pods").Do().Error())
 
-		// Remove all VM Secrets
-		PanicOnError(virtCli.CoreV1().RESTClient().Delete().Namespace(namespace).Resource("secrets").Do().Error())
```
rmohr: I think that we  should keep that, but limit it to secrets which are marked with `kubevirt.io`.

tests/utils.go
```diff
@@ -127,15 +127,18 @@ const (
...
+	iscsiIqn = "iqn.2017-01.io.kubevirt:sn.42"
 )
 
 const (
 	defaultDiskSize        = "1Gi"
 	defaultWindowsDiskSize = "30Gi"
 )
 
+const (
+	SecretLabel = "kubevirt.io/secret"
```
rmohr: We should align on one label for all resources, but good enough for now.

Issue #1103, opened 2018-06-05
Can not create presets from demo-content.yaml


lukas-bednar:
Kubevirt version
```
kubevirt v0.5.1-alpha.2
```
OCP version
```
[root@cnv-executor-lbednar-master1 ~]# oc version 
oc v3.10.0-0.47.0
kubernetes v1.10.0+b81c8f8
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://172.16.0.16:8443
openshift v3.10.0-0.56.0
kubernetes v1.10.0+b81c8f8
```
I am trying to deploy `demo-content.yaml` file from `v0.5.1-alpha.2` release on my cluster in addition to kubevirt, but I am failing on following error.
```
[root@cnv-executor-lbednar-master1 ~]# oc apply -f demo-content.yaml 
[...]

PR #1115, opened 2018-06-07
fix VirtualMachinePreset fields in demo-content manifest

tchughesiv:
resolves https://github.com/kubevirt/kubevirt/issues/1103


stu-gott:
@tchughesiv First off, thanks for the contribution! What @rmohr was referring to is this: There's a module under `pkg/virt-laincher/virtwrap/manager.go`. In this module, there's a function named `Convert_v1_VirtualMachine_To_api_Domain` where a libvirt specific Domain object is synthesized based on the contents of the virtual machine. Rather than forcing the user to add boilerplate lines such as `devices: {}` to every last mainfest, it would be better to simply map a default somewhere in the Domain creation process.

glancing at that module, two lines in particular jump out at me as problematic:

on or around line 383:
`for _, disk := range vm.Spec.Domain.Devices.Disks {`

and on about line 401:
`if vm.Spec.Domain.Devices.Watchdog != nil {`

Neither of those lines make any effort to check if `vm.Spec.Domain.Devices != nil`.

Remember, the vm object passed into this function came from a global cache (so don't modify it). A little bit of "if/then" logic to protect against illegal pointer dereferencing should be all that's needed here.

tchughesiv:
@stu-gott tks, will do

rmohr:
> @tchughesiv First off, thanks for the contribution! What @rmohr was referring to is this: There's a module under pkg/virt-laincher/virtwrap/manager.go.

Hm, not really. What I meant is that this validation function https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-api/validating-webhook/validating-webhook.go#L301 in the validation webhook is not only applied to a VM, it is also applied on templates for OfflineVirtualMachine and Presets. Especially since our presets also allow setting cpu and memory, our templates do not need to be complete-valid vm specifications. Here we differ from k8s presets, where presets only add something  to already valid and complete pods.

We need to do the following:

 * This is what we should fix in this case: Have a less strict validation function for  the preset spec (only validate parts which we allow to override on the vm, maybe even rejecting the preset if fields are present which are not considered by the preset). 
 * For the VirtualMachine we need to apply presets and the defaulter in the validation function, to see if the result will be valid specification (that does not need to happen to fix this issue but is a no-brainer).
 * For OfflineVirtualMachine it is very tricky, since we allow with our current preset implementation posting incomplete VirtualMachine specs. This means that in one moment an OfflineVirtualMachine can start fine (because the preset is right), the next time it can't.

tchughesiv:
@rmohr @stu-gott 
Based on my initial efforts, the issue appears to have more to do w/ this feild being required in the DomainSpec struct -
https://github.com/kubevirt/kubevirt/blob/496634218c737d606d2841d98263a784fc8fe83e/pkg/api/v1/schema.go#L72

Efforts around improvements to the validation code go nowhere because of the above.  Changing it to the following resolves the issue but I'd imagine isn't ideal as it is shared with vm and ovm as well:
```golang
Devices Devices `json:"devices,omitempty"`
```

Another option could be a preset specific DomainSpec struct?

rmohr:
> Another option could be a preset specific DomainSpec struct? maybe "DomainPresetSpec"?

Yes you are right. As far as I can see in the code, right now CPU, Firmware, Clock, Timer and Watchdog can be overwritten. So only these things should be present there.

tchughesiv:
@rmohr the new ```DomainPresetSpec``` has to match the fields in _DomainSpec_ ... otherwise we experience issues w/ validation and vmi/vm merging. Let me know if anything else should change.

rmohr:
ok to test

rmohr:
retest this please

tchughesiv:
@rmohr just realized I left out one field in the new preset struct... Memory:
https://github.com/tchughesiv/kubevirt/blob/0055adfff5a5fc005aa746f5aac5d27f565e0669/pkg/api/v1/schema.go#L59-L61

Working on a PR now to fix it.

PR #1121, opened 2018-06-08
Added doc about our logging, removed debug log level

slintes:
Signed-off-by: Marc Sluiter <marc@slintes.net>

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -5,7 +5,7 @@ That struct offers several functions for providing logs with different log level
 
 ## Log Levels and Verbosity
 
-The Log struct offers several functions for logging with a decent log level: `Debug(msg)`, `Info(msg)`, `Warning(msg)`, `Error(msg)`, `Critical(msg)`.
+The Log struct offers several functions for logging with a decent log level: `Info(msg)`, `Warning(msg)`, `Error(msg)`, `Critical(msg)`.
 These functions also exist in a second flavor, where you can use a formatted string with additional arguments.
 
 Additionally you can set the verbosity with `V(verbosity)`, which takes effect on info level messages.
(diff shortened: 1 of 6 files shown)


docs/logging.md
```diff
@@ -0,0 +1,34 @@
+# Logging
+
+For logging use the `Log` struct provided in the `log` package (see `pkg/log`).
+That struct offers several functions for providing logs with different log levels, verbosity and content.
+
+## Log Levels and Verbosity
+
+The Log struct offers several functions for logging with a decent log level: `Debug(msg)`, `Info(msg)`, `Warning(msg)`, `Error(msg)`, `Critical(msg)`.
```
rmohr: `Debug` should be dropped. We have verbosity in combination with Info. glog does not have it either and I personally find it confusing when to use what or what the difference should be (e.g at which verbosity level does debug start?).

slintes: Completely agree that debug is confusing as it is now. Will remove it.

slintes: @rmohr should I also remove it from code right now, and replace existing calls to it with `V(4).Info(...)`?

rmohr: Would be great.

rmohr:
ok to test

slintes:
@rmohr are there known issues with the test environments? Locally all tests pass.

slintes:
retest this please

rmohr:
@slintes So the results regarding to your code change are fine. In two cases we have well known flakyness (iscsi demo target does not come up and the no-error error from libvirt (error code 0). I have seen the 504 return error before but only added it right now to https://github.com/kubevirt/kubevirt/issues/1014. The same for the failing network test (where we seem to log a byte-array).

rmohr:
@slintes the no-error error code issue may be interesting to look at. There seems to be an issue somewhere in libvirt-go. In the c-binding an error with code `0` means that there is no error. In the go-world `nil` is the equivalent. libvirt-go therefore tries to always return nil, but it seems that somewhere  a `0` error slips through.

PR #1124, opened 2018-06-09
Added integration networking test for implicit pod network definition

booxter:
All current integration tests use specs with interface / network explicitly
specified.  This patch adds connectivity test cases for specs with implicit pod
network enabled.

This is a follow-up to PR #1075.

### DIFF
--- a/pkg/api/v1/defaults.go
+++ b/pkg/api/v1/defaults.go
@@ -129,7 +129,7 @@ func SetDefaults_VirtualMachineInstance(obj *VirtualMachineInstance) {
 		obj.Spec.Domain.Machine.Type = "q35"
 	}
 	setDefaults_DiskFromMachineType(obj)
-	setDefaults_NetworkInterface(obj)
+	SetDefaults_NetworkInterface(obj)
 }
 
 func setDefaults_DiskFromMachineType(obj *VirtualMachineInstance) {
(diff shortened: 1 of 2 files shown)


tests/vm_networking_test.go
```diff
@@ -97,48 +99,66 @@ var _ = Describe("Networking", func() {
...
+		// same but without pod network definition
+		inboundVM_no_podnet, outboundVM_no_podnet = prepareVMs()
 
-		outboundVM, err = virtClient.VM(tests.NamespaceTestDefault).Get(outboundVM.Name, v13.GetOptions{})
-		Expect(err).ToNot(HaveOccurred())
+		for _, networkVm := range []*v1.VirtualMachine{inboundVM_no_podnet, outboundVM_no_podnet} {
+			// sanity check the default spec uses explicit pod network
+			Expect(networkVm.Spec.Domain.Devices.Interfaces).ToNot(BeZero())
+			Expect(networkVm.Spec.Networks).ToNot(BeZero())
+			networkVm.Spec.Domain.Devices.Interfaces = []v1.Interface{}
```
vladikr: wouldn't the vm be created at this point?
I'm probably missing something, but I can't see how this will affect the test.

booxter: Yeah you are right, I don't know what I was thinking.

booxter: I think I solved the issue in the latest version. Now I configure default Networks / Interfaces before creating the resources. I don't like the approach we take with all these New*VM* functions, so instead of creating yet another one helper, I pass a boolean flag into prepareVMs and act on it if it's true by adding pod interface to the spec.

PR #1125, opened 2018-06-09
Support multiple interface / network definitions

booxter:
This is a partial solution to support multiple interfaces. Note that the PR does *not* enable a way to specify multiple interfaces in the spec since we don't have any interface types but pod at this time.

The reason why this PR does not include new integration tests is because we can't at the moment validate multiple interfaces end to end due to reasons described above. To facilitate testing, the PR includes some unit tests though. Later PRs to support additional interface types will need to include integration tests with multiple interfaces.

The reason why we don't want to hold off the PR before we have all code to validate multiple interfaces support end to end is because multiple authors will need to work on different new interface types in parallel (proxy, dp...), and this basic bit is commonly needed for all of them, and we don't want to duplicate it in these parallel efforts.

### DIFF
--- a/pkg/virt-launcher/virtwrap/network/common.go
+++ b/pkg/virt-launcher/virtwrap/network/common.go
@@ -170,20 +170,24 @@ func initHandler() {
 	}
 }
 
-func setCachedInterface(ifconf *api.Interface) error {
+func getInterfaceCacheFile(name string) string {
+	return fmt.Sprintf(interfaceCacheFile, name)
+}
+
+func setCachedInterface(name string, ifconf *api.Interface) error {
 	buf, err := json.MarshalIndent(&ifconf, "", "  ")
 	if err != nil {
 		return fmt.Errorf("error marshaling interface cache: %v", err)
 	}
-	err = ioutil.WriteFile(interfaceCacheFile, buf, 0644)
+	err = ioutil.WriteFile(getInterfaceCacheFile(name), buf, 0644)
 	if err != nil {
 		return fmt.Errorf("error writing interface cache %v", err)
 	}
 	return nil
 }
 
-func getCachedInterface() (*api.Interface, error) {
-	buf, err := ioutil.ReadFile(interfaceCacheFile)
+func getCachedInterface(name string) (*api.Interface, error) {
+	buf, err := ioutil.ReadFile(getInterfaceCacheFile(name))
 	if err != nil {
 		if os.IsNotExist(err) {
 			return nil, nil
--- a/pkg/virt-launcher/virtwrap/network/common_test.go
+++ b/pkg/virt-launcher/virtwrap/network/common_test.go
@@ -21,10 +21,13 @@ package network
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
+	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
 )
 
 var _ = Describe("Common Network Methods", func() {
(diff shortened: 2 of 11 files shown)


pkg/virt-launcher/virtwrap/network/common_test.go
```diff
@@ -89,4 +92,20 @@ var _ = Describe("Common Network Methods", func() {
 			Expect(err).To(BeNil())
 		})
 	})
+	Context("Function setCachedInterface()", func() {
+		It("should allow partial search domains", func() {
```
booxter: Eh, this description needs a change...

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -45,6 +45,16 @@ type PodInterface struct{}
 
 func (l *PodInterface) Unplug() {}
 
+func findPodInterface(ifaces []api.Interface) (int, error) {
+	for i, iface := range ifaces {
+		// TODO:(ihar) find a more robust way to detect pod interface
```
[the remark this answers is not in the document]
booxter: You mean "Pod" set to non-nil, right? If that's what you mean, I don't think this applies here because the api.Network type is not the same as v1.Network and doesn't have anything specific to indicate whether it's a pod network. This module is in virt-launcher and not a direct representation of VirtualMachineSpec.

There is probably a better way of doing it, but it would require somehow passing additional information about in-spec network definitions related to domain networks from api layer.

booxter:
The latest upload covers all Roman's comment except the one about a better way to find a pod interface which I addressed with a comment in the thread. There is also a small TODO to cover, and maybe functional test failures, in case they show up.

booxter:
retest this please

pkg/virt-launcher/virtwrap/network/podinterface.go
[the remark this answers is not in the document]
rmohr: Yes I see. The proper way is with libvirt 3.9 with this fix: https://bugzilla.redhat.com/show_bug.cgi?id=1434451. One of the reasons why we have a mandatory name field on all devices is that we can set this name as qemu alias. You would then search for the device name like described above on the vm. Then you take the name and search for the interface in the libvirt struct where this name is set as device alias.

fabiand:
> Since the default podNetwork will be added automatically, make sure that it gets a random name and not hard-coded "default" to avoid collisions.

It's getting added automatically if no other network (no network at all) is defined.
If at least one network is defined, then it will not be added.

Is this above true?

if it is, then we could stick to the defined named `default`, not?

rmohr:
> It's getting added automatically if no other network (no network at all) is defined.
> If at least one network is defined, then it will not be added.

We agreed on adding an explicit opt-out with a field: `autoAttachPodInterface: false`. See https://docs.google.com/document/d/10rXr91aqn8MvVcLgHw33WX8BaQwHPZERp25PHxoZGgw/edit

fabiand:
"
Default configuration
If not spec.domain.interfaces list is provided, then a default vNIC with a connection to the stock Kubernets pod network will be connected.

VM without network interfaces
The user can opt-out of the default connection mechanism by setting spec.domain.autoAttachPodInterface to false:
"

Thus in the current situation it would only be created if the list of interfaces is empty.
To opt out of the _empty list_ behavior we have the switch.

IOW If there are interafces defined, then the default network would not be added.

I'm just quoting the doc - not saying it should stay like this.

I need to think a little baout it to make my mind up if we should (unconditionally) always add the default network, or just in case of an empty list.

rmohr:
> Thus in the current situation it would only be created if the list of interfaces is empty.
> To opt out of the empty list behavior we have the switch.

Then you don't need the random name.

rmohr:
> I'm just quoting the doc - not saying it should stay like this.

It is fine. I misread it.

booxter:
The latest upload has the following changes:

- we now set alias for domain interfaces to corresponding spec interface names, and use the alias as a back reference to spec interface / network in podinterface.go instead of relying on clunky "br1" / "bridge" type matching;

- fixed converter code to honor spec interfaces (and not add a pod network regardless of spec, as it did before);

- we no longer add a pod interface if networks are explicitly defined in spec;

- since we no longer add a default interface unless the spec list is empty, we no longer have a risk of name clash, so the code to randomize default interface names is ripped off (I still stored it in a private branch for if / when we decide it's a better path forward.)

booxter:
The latest upload is just a rebase after VM -> VMI rename.

booxter:
retest this please

booxter:
The latest upload is another rebase to accommodate for new git conflicts. @rmohr @vladikr may I ask you to review the PR? Thanks!

rmohr:
retest this please

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -170,20 +170,24 @@ func initHandler() {
...
+	return fmt.Sprintf(interfaceCacheFile, name)
+}
+
+func setCachedInterface(name string, ifconf *api.Interface) error {
 	buf, err := json.MarshalIndent(&ifconf, "", "  ")
 	if err != nil {
 		return fmt.Errorf("error marshaling interface cache: %v", err)
 	}
-	err = ioutil.WriteFile(interfaceCacheFile, buf, 0644)
+	err = ioutil.WriteFile(getInterfaceCacheFile(name), buf, 0644)
```
rmohr: Nothing which you introduced in this PR, but why are we writing the config to a file? If virt-launcher crashes the pod will go to finished. We can as well keep it in memory.

booxter: That I don't know. As you can see, I am still pretty bad at broad picture of all components and how they interact. :) I will look into it.

Issue #1078, opened 2018-05-25
Allow a VM to be tied into an Istio based service mesh


fabiand:
> As a user I want to be able to tie my VM (and the workload within) into my regular Istio service mesh, assuming that it will behave just like a pod

- [x] Work out a compatible networking mode
- [x] Work out how side cars can be injected
- [x] Work out how istio can be injected

fabiand:
This is done, and works with auto-injection.

PR #787, opened 2018-03-02
[WIP] Pod networking proxy

vladikr:
WIP

Re-work networking codebase to allow attaching various interface types
Introduce a proxy mode that will allow users to bind to a specific VM ports 

```yaml
kind: VM
spec:
  devices:
    interfaces:
      - name: nic-ip
        networkName: red
        proxy: ## EITHER THIS
          ports:
          - name: ssh
        bridge: ## OR THIS
          delegateIp: true # offers the ip in case that the source has an ip
        Macvtap: ## would error with pod
          mode: vepa
  networks:
  - name: red
    pod: {} # Stock pod network
```

PR #1138, opened 2018-06-11
Slirp interface implementation for pod network

SchSeba:
Create slirp interface and proxy network
Create new network type Proxy.
Create new interface type slirp using the slirp configuration of qemu.

Related to issue - [1078](https://github.com/kubevirt/kubevirt/issues/1078)

Design the connection API - Done
Write unit test - Inprogress
Write functional tests
Add the new connection to the user-guide
Get all PRs merged
Basic implementation of the connection method

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/1138)


vladikr:
Great progress!

After an initial brief look, I think we should change the place where proxy is being referenced.
I think we have agreed that `proxy` should be a binding method under an `interface` and not a `network`, as it is laid out in the networking API document[1].
I have a previous implementation of an earlier proxy attempt, it should contain the logic as how everything should be setup; I would suggest to take a look at [2]

[1] https://docs.google.com/document/d/10rXr91aqn8MvVcLgHw33WX8BaQwHPZERp25PHxoZGgw/edit#heading=h.d2sg8ncc0555

[2] https://github.com/kubevirt/kubevirt/pull/787/files#diff-4cc299b396934683b5f4388deaff9e77

SchSeba:
@vladikr Thanks for the comment I will take a look.
If I move the proxy from then network section to the interface (change the slirp name) I Should use the pod network?

I write it this way because we may have different types of proxy implementations that the user can use for every interface he create(slirp,nat,transparent)

vladikr:
> If I move the proxy from then network section to the interface (change the slirp name) I Should use the pod network?

Yea, as it's one only network we support at the moment.

vladikr:
> I write it this way because we may have different types of proxy implementations that the user can use for every interface he create(slirp,nat,transparent)

That shouldn't be a problem. We could model additional modes in the proxy struct in the future if we will need it.

fabiand:
Amazing work.

@SchSeba could you please add yaml example to see the API in action.

/me still wonders if we should call it `proxy` or something more specific. Let#s see.

fabiand:
In the recent patchset the changes was made to rename the api from `slirp` to `proxy` in order to align with the design document.
However, the design document was explicitly speaking about a proxy - which we do not resolve right now, but probably get back to at som epoint.
Thus here it is justified to explicitly use `slirp` within the API - also to take it out later on once we have a solution which is providing the same set of features.

The initial part of the paragraph was discussed with @SchSeba on irc.

cluster/examples/vm-slirp.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - disk:
+          bus: virtio
+        name: cloudinitdisk
+        volumeName: cloudinitvolume
+      interfaces:
+      - name: testSlirp
+        slirp:
+          Ports:
+          - name: http
+            podPort: 80
```
fabiand: Guided by the question: What does a user want to do? Expose a port.

In that sense we should primarily ask for the port to expose, and provide an optional mechanism to override the podPort.

Thus:

```yaml
name: http
port: 80
protocol: TCP
podPort: 8080  # If it should bind to != 80 on the pod.
```

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -489,15 +489,54 @@ func Convert_v1_VirtualMachine_To_api_Domain(vm *v1.VirtualMachine, domain *Doma
...
+		// prepare networks map
+		networks := map[string]*v1.Network{}
+		for _, network := range vm.Spec.Networks {
+			networks[network.Name] = network.DeepCopy()
+		}
+
+		domain.Spec.Devices.Interfaces = make([]Interface, 0)
+		for _, iface := range vm.Spec.Domain.Devices.Interfaces {
+			if _, ok := networks[iface.Name]; !ok {
+				return fmt.Errorf("Fail to match the Name of a Network")
```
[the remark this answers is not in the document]
SchSeba: Sure I will add validation on the validation-webhook inside the virt-api just want to make sure the api looks good

cmd/virt-launcher/Dockerfile
```diff
@@ -39,6 +39,9 @@ RUN chmod 0640 /etc/sudoers.d/kubevirt
 # Allow qemu to bind privilaged ports
 RUN setcap CAP_NET_BIND_SERVICE=+eip /usr/bin/qemu-system-x86_64
 
+# Allow proxy interface to ping
+RUN echo 'net.ipv4.ping_group_range = 107 107' >> /etc/sysctl.conf
```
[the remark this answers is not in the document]
SchSeba: I will remove this line from the dockerfile

pkg/api/v1/schema.go
```diff
@@ -628,12 +628,32 @@ type Interface struct {
 // +k8s:openapi-gen=true
 type InterfaceBindingMethod struct {
 	Bridge *InterfaceBridge `json:"bridge,omitempty"`
+	Slirp  *InterfaceSlirp  `json:"slirp,omitempty"`
```
vladikr: I think it should be name `Proxy` as it was agreed in the API doc.

SchSeba: I Change it before but @fabiand tell me to change it.
I can change it again but we need to agree before I change it again.

SchSeba: I think it's ok to leave the name slirp because is not a real proxy implementation.

pkg/virt-launcher/virtwrap/network/network.go
```diff
@@ -80,9 +80,14 @@ func SetupNetworkInterfaces(vm *v1.VirtualMachine, domain *api.Domain) error {
 }
 
 // a factory to get suitable network interface
-func getNetworkClass(network *v1.Network) (NetworkInterface, error) {
-	if network.Pod != nil {
+func getNetworkClass(iface *v1.Interface, network *v1.Network) (NetworkInterface, error) {
```
vladikr: I envisioned it differently. I think here we just need to get the correct interface, according to the network source type and plug() it. How do we plug it is determined by the interface attributes. (`bridge`, `proxy`, `direct` (macvtap) or `passthrough`) As in the API doc [1]
Since `proxy` is only being used with the pod network source, I think, we should add it a new binding mechanism. I have an example for it in an older PR [2]

[1] https://docs.google.com/document/d/10rXr91aqn8MvVcLgHw33WX8BaQwHPZERp25PHxoZGgw
[2] https://github.com/kubevirt/kubevirt/pull/787/files#diff-d3ed8eb8cca020f37cab207c079f2627R257

pkg/api/v1/schema.go
vladikr: Yea, I'm not against it. I'm just not sure whether the term `slirp` helps people understand the way this interface will be connected. Perhaps, `user` or `userspace` is a better term?

pkg/virt-launcher/virtwrap/network/network.go
SchSeba: I saw this pull request but in the plug function there are calls that the slirp interface cant implement, like start dhcp, set Interface Routes, etc (all of it is handle by the qemu process) so I cant use the same binding mechanism the bridge interface used.

vladikr: I think you can just override these methods and keep the relevant methods empty.
Would that work?

SchSeba: I can try to do it. I need to change the getCachedInterface function? 
Because the slirp interface configuration is in the qemu command line and not as a regular domain interface.

pkg/api/v1/schema.go
fabiand: Let's please stick to `slirp` - it will help people to understand what it is about.
it's also the most precise name, and we probably come up with a more non-qemu solution where the term "userspace" or "proxy" is more suited.

As we discussed slirp is a step primarily to unblock, thus having this "weird" name here is fine - and as said we avoid naming conflicts with eventually more sophisticated solutions.

Btw - in the scope of the design doc slirp was not discussed, that's why it's fine to take this name.
I'll actually add slirp to the doc.

pkg/virt-launcher/virtwrap/network/network.go
vladikr: yea, it's tricky. I didn't check, but IIRC libvirt had an interface type="user" to setup slirp, isn't it sufficient?

If not, I suppose the best way forward, in this case, is to separate everything after the getBinding to a setup() method and override this method in SlirpPodInterface. So in the Plug method we will have something like 
```
driver, err := getBinding(iface)
if err != nil {
	return err
}
err = driver.setup()
if err != nil {
	return err
}
``` 

not ideal though.

SchSeba: Yes they have an interface type user but you can only config the subnet . [1]
 
I use the qemu command line because I need to add the dnssearchname and the portforward parameters.
Cant find anything in the libvirt documentation to add this parameters as part of the interface configuration.

[1] https://libvirt.org/formatdomain.html#elementsNICSSlirp

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -34,11 +36,14 @@ import (
 
 var bridgeFakeIP = "169.254.75.86/32"
 
+// DefaultProtocol is the default port protocol
+const DefaultProtocol string = "TCP"
+
+// DefaultVMCIDR is the default CIRD for vm network
+const DefaultVMCIDR = "10.0.2.0/24"
```
booxter: Seems like it's slirp specific. If so, shouldn't the name reflect that?

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -34,11 +36,14 @@ import (
...
+
 type BindMechanism interface {
-	discoverPodNetworkInterface() error
```
booxter: I think you could retain the interface here. We may want to use it later for other types and not reimplement call order / error handling again. Of course, some of these functions will be no-ops in slirp case, but that's ok.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -242,3 +262,133 @@ func (b *BridgePodInterface) createDefaultBridge() error {
...
+	}
+
+	err = p.commitConfiguration()
+	if err != nil {
+		return err
+	}
+
+	err = setCachedCommandLine(p.domain.Spec.QEMUCmd)
+	if err != nil {
+		panic(err)
```
booxter: why do we want to abrupt process on this specific error?

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -242,3 +262,133 @@ func (b *BridgePodInterface) createDefaultBridge() error {
...
+
+func (p *SlirpPodInterface) setup() error {
+	commandLine, err := getCachedQemuCommandLine()
+	if err != nil {
+		return err
+	}
+
+	if commandLine != nil {
+		p.domain.Spec.QEMUCmd = commandLine
+
```
booxter: nit: remove the newline

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -242,3 +262,133 @@ func (b *BridgePodInterface) createDefaultBridge() error {
...
+		return nil
+	}
+
+	portForwardMap := make(map[int32]string)
+
+	for _, forwardPort := range p.iface.Slirp.Ports {
+		protocol := DefaultProtocol
+
+		if forwardPort.Port == 0 {
+			return fmt.Errorf("Port must be configure")
```
booxter: -> Port must be configure_d_

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -242,3 +262,133 @@ func (b *BridgePodInterface) createDefaultBridge() error {
...
+		}
+
+		// Check protocol, its case sensitive like kubernetes
```
booxter: This validation code is more or less the same as in webhook package. I suggest we move it into a function and reuse from both places.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -242,3 +262,133 @@ func (b *BridgePodInterface) createDefaultBridge() error {
...
+func (p *SlirpPodInterface) configVMCIDR() error {
+	if p.network.Pod == nil {
+		return fmt.Errorf("Slirp works only with pod network")
+	}
+
+	vmNetworkCIDR := ""
+	if p.network.Pod.VMNetworkCIDR != "" {
+		_, _, err := net.ParseCIDR(p.network.Pod.VMNetworkCIDR)
+		if err != nil {
+			return fmt.Errorf("Failed parsing CIDR")
```
booxter: You could helpfully include the value that failed in the log message

pkg/virt-launcher/virtwrap/network/podinterface.go
yuvalif: @booxter , what other types of networks with pod backend will need these steps?
E.g. if we will implement proxy based approach, the steps are different. 
Do you also suggest to drop the "setup()" method?
If we have an interface with 2 implementations and one of them has empty implementation in all the calls except one, then this is a strong indication that the abstraction is done at the wrong level. 
If there are several types of pod-bridge solution, we can have a BridgePodInterface with these calls - a real interface, not the struct at line 83.

pkg/api/v1/schema.go
```diff
@@ -628,12 +628,32 @@ type Interface struct {
...
+	Ports []Port `json:",omitempty"`
+}
+
+// Port repesents a port to expose from the virtual machine
+// Default protocol TCP
+// Default vmPort is PodPort
+// ---
+// +k8s:openapi-gen=true
+type Port struct {
+	Name     string `json:"name,omitempty"`
```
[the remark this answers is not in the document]
SchSeba: I think it's better to remove it. just like kubernetes does.
Only port is mandatory
```
 - name: my-nginx
        image: nginx
        ports:
        - containerPort: 80
```

pkg/api/v1/schema.go
```diff
@@ -658,4 +678,8 @@ type NetworkSource struct {
 // Represents the stock pod network interface
 // ---
 // +k8s:openapi-gen=true
-type PodNetwork struct{}
+type PodNetwork struct {
+	// CIDR for vm network
+	// Default 10.0.2.0/24 if not specified
```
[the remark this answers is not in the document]
SchSeba: It's not slirp specific, we can use it on any interface solution the need to create a network subnet inside the pod.
like proxy solution or nat.

pkg/virt-launcher/virtwrap/network/podinterface.go
SchSeba: It's not slirp specific, we can use it on any interface solution the need to create a network subnet inside the pod.
like proxy solution or nat.

pkg/virt-launcher/virtwrap/network/podinterface.go
SchSeba: I will just remove it from the virt-lancher, Because we already check the configuration in the virt-api and I dont see a reason to recheck it inside the virt lancher

booxter: Defensive programming; the approach in general would also help with upgrades when different components move separately and may incorporate different rules. But yes I have no problem with removing it for now, we'll need a more thought through approach to all that not specific to this PR.

pkg/virt-launcher/virtwrap/network/podinterface.go
SchSeba: Let me implement it with this bind mechanism and make a review about it.

booxter: Right, there is no disagreement about the current interface being not optimal. But making changing it part of this PR seems to be not appropriate (unless we can't make it work). A separate PR for this specific discussion would be very helpful.

pkg/virt-launcher/virtwrap/network/podinterface.go
booxter: Ack. BTW what's the rationale to allow users change the CIDR? Do we envision some images not working with our CIDR for some reason?

SchSeba: Because we may conflict with there network subnets.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -495,38 +499,167 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
 
-	findNetwork := func(nets []v1.Network, name string) (*v1.Network, error) {
-		for _, net := range nets {
-			if net.Name == name {
-				return &net, nil
-			}
-		}
-		return nil, fmt.Errorf("failed to find network %s", name)
+	networks := map[string]*v1.Network{}
+	for _, network := range vmi.Spec.Networks {
```
vladikr: I can't see a clear benefit for it being a map or a loop - we still have to loop the same amount of times anyway.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -34,11 +34,18 @@ import (
...
+
+// DefaultVMCIDR is the default CIDR for vm network
+const DefaultVMCIDR = "10.0.2.0/24"
+
 type BindMechanism interface {
 	discoverPodNetworkInterface() error
 	preparePodNetworkInterfaces() error
-	startDHCPServer()
-	decorateInterfaceConfig(ifconf *api.Interface)
+	libvirtConfig() error
```
vladikr: I'd prefer not to mention `libvirt` specifically, rather more like decorateConfig. To me `decorate` makes it more generic, rather than `set` or `update`

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -249,3 +287,57 @@ func (b *BridgePodInterface) createDefaultBridge() error {
 
 	return nil
 }
+
+type SlirpPodInterface struct {
+	iface           *v1.Interface
+	network         *v1.Network
```
vladikr: I can't see where is this being used. Why do we need network here?

vladikr: If both have structs have the same attributes perhaps it can simply be 
```
type SlirpPodInterface struct {
	BridgePodInterface
}
```

pkg/virt-launcher/virtwrap/network/podinterface.go
SchSeba: No problem will be change to decorateConfig

pkg/virt-launcher/virtwrap/network/podinterface.go
SchSeba: I will remove the network variable.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -59,62 +66,68 @@ func (l *PodInterface) Plug(iface *v1.Interface, network *v1.Network, domain *ap
...
+			panic(err)
+		}
+	}
 
 	return nil
 }
 
-func getBinding(iface *v1.Interface) (BindMechanism, error) {
-	vif := &VIF{Name: podInterface}
+func getBinding(iface *v1.Interface, network *v1.Network, domain *api.Domain) (BindMechanism, error) {
```
vladikr: I suppose we don't need the `network *v1.Network` here as well, right?

vladikr:
@booxter could you please take another look? Perhaps you have other concerns.

vladikr:
ok to test

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -495,65 +499,172 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+				},
+				Alias: &Alias{
+					Name: iface.Name,
+				},
+			}
+			domain.Spec.Devices.Interfaces = append(domain.Spec.Devices.Interfaces, domainIface)
+
+		} else if iface.Slirp != nil {
+			// TODO: maybe add interface model to vm spec
+			// Slirp configuration works only with e1000 or rtl8139
```
booxter: You say only e1000 and rtl work, but then you error on virtio only. There are other options supported by libvirt; should you be explicit and error unless it's e1000 / rtl?

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -495,65 +499,172 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+		return nil
+	}
+
+	for _, forwardPort := range iface.Slirp.Ports {
+
+		if forwardPort.Port == 0 {
+			return fmt.Errorf("Port must be configured")
+		}
+
+		// Check protocol, its case sensitive like kubernetes
```
booxter: not sure where this comment fits, since we don't seem to look / care about case of the protocol anywhere here.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -495,65 +499,172 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+		vmNetworkCIDR = DefaultVMCIDR
+	}
+
+	// Insert configuration to qemu commandline
+	qemuArg.Value += fmt.Sprintf(",net=%s", vmNetworkCIDR)
+
+	return nil
+}
+
+func configDNSSearchName(qemuArg *Arg) error {
```
booxter: I think you should reuse ParseSearchDomains instead of reimplementing it here.

pkg/virt-launcher/virtwrap/api/converter_test.go
```diff
@@ -534,6 +534,97 @@ var _ = Describe("Converter", func() {
 			vmi.Spec.Networks = append(vmi.Spec.Networks, *net)
 			Expect(Convert_v1_VirtualMachine_To_api_Domain(vmi, &Domain{}, c)).ToNot(Succeed())
 		})
+
+		It("should add tcp if protocol not exist", func() {
```
booxter: nit: my understanding is that we should define separate Contexts if we target different code. Test cases above are for convertion functions; you added several cases specifically for configPortForward; I think it would make sense to have them separate in a new Context.

pkg/virt-launcher/virtwrap/network/common_test.go
```diff
@@ -92,20 +93,38 @@ var _ = Describe("Common Network Methods", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+
+			var cached_iface api.Interface
+			isExist, err := readFromCachedFile(ifaceName, interfaceCacheFile, &cached_iface)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isExist).To(Equal(true))
+
+			Expect(iface).To(Equal(cached_iface))
+		})
+		It("should persis qemu arg payload", func() {
```
booxter: ->persist

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -170,36 +170,36 @@ func initHandler() {
...
-		return nil, err
+		return false, err
 	}
-	ifconf := api.Interface{}
-	err = json.Unmarshal(buf, &ifconf)
+
+	err = json.Unmarshal(buf, &inter)
 	if err != nil {
-		return nil, fmt.Errorf("error unmarshaling interface: %v", err)
+		return false, fmt.Errorf("error unmarshaling command line: %v", err)
```
booxter: Command line? I think you we use these functions in two contexts: for qemu CLI args and for interface cache; it would be incorrect to log this message for interface cache; (same issue but vice versa in write function).

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -59,62 +66,68 @@ func (l *PodInterface) Plug(iface *v1.Interface, network *v1.Network, domain *ap
...
 		}
 
-		defaultIconf := interfaces[podInterfaceNum]
-		driver.startDHCPServer()
-
 		// After the network is configured, cache the result
 		// in case this function is called again.
-		driver.decorateInterfaceConfig(&defaultIconf)
-		err = setCachedInterface(iface.Name, &defaultIconf)
+		// Use this section
```
booxter: What does this comment mean?

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -59,62 +66,68 @@ func (l *PodInterface) Plug(iface *v1.Interface, network *v1.Network, domain *ap
...
 	}
 
-	interfaces := domain.Spec.Devices.Interfaces
-	podInterfaceNum, err := findInterfaceByName(interfaces, iface.Name)
-	if err != nil {
-		return err
-	}
-
-	if ifconf == nil {
+	if !isExist {
```
booxter: hm. if we only care about file existence, why do we even bother to marshal / unmarshal?

pkg/api/v1/schema.go
```diff
@@ -659,18 +659,18 @@ type InterfaceBridge struct{}
...
 type InterfaceSlirp struct {
 	// List of ports to be forwarded to the virtual machine
-	Ports []Port `json:",omitempty"`
+	Ports []Port `json:"ports,omitempty"`
 }
 
 // Port repesents a port to expose from the virtual machine
 // Default protocol TCP
-// Default vmPort is PodPort
+// Default Port is PodPort
```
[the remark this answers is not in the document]
booxter: Also, consider adding trailing dots to each line in spirit of https://github.com/kubevirt/kubevirt/pull/1188

booxter:
Is this PR still work-in-progress, as the title suggests? It seems like we are close to getting it complete and merged, so maybe change the title to reflect the progress.

pkg/virt-launcher/virtwrap/api/converter.go
SchSeba: I cant use the function because I get a circular import.
I removed the function from the common file and pass it to the convert file.

pkg/virt-launcher/virtwrap/api/converter.go
SchSeba: removed

pkg/virt-launcher/virtwrap/api/converter.go
SchSeba: I cant find any documentation about all the interfaces that are supported be the slirp interface so I will change the if to check for only e1000 / rtl

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -495,65 +499,172 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+			domain.Spec.Devices.Interfaces = append(domain.Spec.Devices.Interfaces, domainIface)
+
+		} else if iface.Slirp != nil {
+			// TODO: maybe add interface model to vm spec
+			// Slirp configuration works only with e1000 or rtl8139
+			slirpInterfaceType := "e1000"
+			if interfaceType != "virtio" {
+				slirpInterfaceType = interfaceType
+			} else {
+				log.Log.Infof("The network interface type was changed from virtio to e1000 due to unsupported interface type by qemu slirp network")
```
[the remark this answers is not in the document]
SchSeba: sure

pkg/api/v1/schema.go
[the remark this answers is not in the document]
SchSeba: done

pkg/virt-launcher/virtwrap/api/converter_test.go
SchSeba: No problem I will move it to a new context

pkg/virt-launcher/virtwrap/network/common_test.go
SchSeba: done

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -494,6 +494,14 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 	}
 
 	getInterfaceType := func(iface *v1.Interface) string {
+		if iface.Slirp != nil {
+			// Slirp configuration works only with e1000 or rtl8139
+			if iface.Model != "e1000" && iface.Model != "rtl8139" {
+				log.Log.Infof("The network interface type of %s was changed from virtio to e1000 due to unsupported interface type by qemu slirp network", iface.Name)
```
booxter: I don't think the log message is now correct? It says it changed from virtio but there are other options.

tests/vmi_slirp_interface_test.go
```diff
@@ -0,0 +1,108 @@
...
+				virtClient,
+				vmiPod,
+				vmiPod.Spec.Containers[1].Name,
+				[]string{"netstat", "-tnlp"},
+			)
+			log.Log.Infof("%v", output)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(strings.Contains(output, "0.0.0.0:80")).To(BeTrue())
+		})
+		It("should return \"Hello Word\" when connecting to localhost on port 80", func() {
```
booxter: Hello world

booxter:
Nothing major, we could merge (there is one issue with a log message but we could tackle it separately later too).

booxter:
@vladikr I think this is ready to merge though note there are no fresh CI results. I would suggest to squash on merge since commits are not logically separated anyway.

tests/vmi_slirp_interface_test.go
fabiand: or Microsoft Word

vladikr:
Some functional tests failed, but not because of anything related to this PR.

vladikr:
retest this please

pkg/virt-launcher/virtwrap/api/converter.go
SchSeba: remove the virtio word

tests/vmi_slirp_interface_test.go
SchSeba: fix the typo sorry :)

SchSeba:
retest this please

fabiand:
Please eliminate the WIP in the description.

Looks great

PR #1146, opened 2018-06-12
fix PVC path in storage doc

thiagodasilva:
Found a small error with the PVC path and ended up fixing a few nits.


fabiand:
Thanks!

PR #1150, opened 2018-06-13
openapi-spec: add name and namespaces param for patch operation

lukas-bednar:
Fixes: kubevirt/client-python#20

Signed-off-by: Lukas Bednar <lbednar@redhat.com>


rmohr:
@lukas-bednar as always great. I wonder if there is already somewhere something which would generate the whole swagger spec vor CRDs, including the rest paths, ... Maybe we stumble across something.

rmohr:
ok to test

rmohr:
retest this please

PR #1152, opened 2018-06-13
Correct a few more VM rename artifacts

stu-gott:


rmohr:
Ok, missed that travis failed.

PR #1155, opened 2018-06-14
Indicate need for X11 when running virtctl over ssh

karmab:


docs/getting-started.md
```diff
@@ -209,3 +209,5 @@ to start a remote session with `remote-viewer`.
 
 `cluster/virtctl.sh` is a wrapper around `virtctl`. `virtctl` brings all
 virtual machine specific commands with it. It is supplement to `kubectl`.
+
+Note: if accessing your cluster through ssh, be sure to forward your X11 session in order to launch virtctl
```
rmohr: It is only related to `virtctl vnc`, could you make that clearer? Btw. was that not clear enough in the kubevirt-dev email and the user-guide?

karmab: i believe it's crystal clear for old school linux users ( i feel old now), but better to be explicite

docs/getting-started.md
```diff
@@ -209,3 +209,5 @@ to start a remote session with `remote-viewer`.
 
 `cluster/virtctl.sh` is a wrapper around `virtctl`. `virtctl` brings all
 virtual machine specific commands with it. It is supplement to `kubectl`.
+
+Note: if accessing your cluster through ssh, be sure to forward your X11 session in order to launch *virtctl vnc*
```
rmohr: **Note:** if accessing your cluster through ssh, be sure to forward your X11 session in order to launch `virtctl vnc`.

PR #1156, opened 2018-06-14
Update storage doc to use cdi.io

awels:
- Updated document to use cdi.io annotation instead of kubevirt.io
- Updated document to use v1alpha2.

### DIFF
--- a/docs/devel/storage.md
+++ b/docs/devel/storage.md
@@ -58,8 +58,8 @@ metadata:
   labels:
     app: containerized-data-importer
   annotations:
-    kubevirt.io/storage.import.endpoint: "https://download.cirros-cloud.net/0.4.0/cirros-0.4.0-x86_64-disk.img"   # Required.  Format: (http||s3)://www.myUrl.com/path/of/data
-    kubevirt.io/storage.import.secretName: "" # Optional.  The name of the secret containing credentials for the data source
+    cdi.kubevirt.io/storage.import.endpoint: "https://download.cirros-cloud.net/0.4.0/cirros-0.4.0-x86_64-disk.img"   # Required.  Format: (http||s3)://www.myUrl.com/path/of/data
+    cdi.kubevirt.io/storage.import.secretName: "" # Optional.  The name of the secret containing credentials for the data source
 spec:
   accessModes:
   - ReadWriteOnce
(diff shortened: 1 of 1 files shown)


stu-gott:
Documentation change. No CI impact.

aglitke:
Let's hold off on this until the next CDI release.  We need to adopt this change in an orderly fashion.

docs/devel/storage.md
```diff
@@ -76,7 +76,7 @@ The CDI controller will see the annotation and spawn an 'importer' pod that does
 
 Now that we have a populated PVC, we can create a VM using the disk image. For instance using this yaml:
 ```YAML
-apiVersion: kubevirt.io/v1alpha1
 kind: VirtualMachine
```
thiagodasilva: IIUC, VirtualMachine needs to be changed to VirtualMachineInstance

awels: No, VMs are what used to be OfflineVMs, and VM instances are what used to be VMs. Were are defining the specification of the VM here, so the kind should be Virtual Machine, when we start this VMs, we in essence create a VM instance.

thiagodasilva: I tried running the example just a minute ago and got the following error:
```
[thiago@localhost kubevirt]$ ./cluster/kubectl.sh create -f /tmp/vm.yaml 
error: unable to recognize "/tmp/vm.yaml": no matches for kind "VirtualMachine" in version "kubevirt.io/v1alpha1"
```
Then while checking the kubevirt examples[1] I notcied that it had changed to VirtualMachineInstance. Making this change fixed it for me.

[1] - https://github.com/kubevirt/kubevirt/blob/master/cluster/examples/vmi-ephemeral.yaml

awels: Right, that is why this PR is changing it to v1alpha2. Again VirtualMachine is the definition, and VirtualMachineInstance is the running instance of a VirtualMachine. The example you are looking at is creating a running VM, this example is like [1]. Creating the definition and then you can launch it at a later time.

[1] https://github.com/kubevirt/kubevirt/blob/master/cluster/examples/vm-cirros.yaml

awels: No you are right, I was confused, it should be a VirtualMachineInstance, I have updated the PR to match.

rmohr:
Would it make sense to have this doc in the cdi repo and link from here?

awels:
It would if it only deals with cdi configuration (which right now is a large part of it) but it also helps people configure a hostPath provisioner, and in the near future will also include the configuration of the host assisted cloning (once that is complete).

rmohr:
Is this now accurate and should be merged until we have DataVolume support merged, or should this be closed?

ovirt-infra:

Issue #1141, opened 2018-06-11
changes to build-go.sh block building rpms outside of git repo


sreichar:
When attempting to build the 0.6.0 release based RPM, a failure is seen since version.sh is using git.

RPM builds can not access external systems.

Version.sh is being called in build-go.sh which appears to have changed since previous release.

fabiand:
@sreichar can you please provide a paste of the error an dthe command you used to trigger it?

sreichar:
Here is a pointer to the scratch build I attempted:

http://download-node-02.eng.bos.redhat.com/brewroot/work/tasks/9063/16679063/build.log


Mock Version: 1.1.16
Mock Version: 1.1.16
ENTER do(['bash', '--login', '-c', 'rpmbuild -bs --target x86_64 --nodeps builddir/build/SPECS/kubevirt.spec'], False, '/mnt/build/mock/cnv-1.1-rhel-7-build-4110699-2545021/root/', None, 86400, True, False, 248, 135, None, logger=<mockbuild.trace_decorator.getLog object at 0xa54cd0>)
Executing command: ['bash', '--login', '-c', 'rpmbuild -bs --target x86_64 --nodeps builddir/build/SPECS/kubevirt.spec']
Building target platforms: x86_64
Building for target x86_64
Wrote: /builddir/build/SRPMS/kubevirt-0.6.0-1.git.0.cf13e2b.src.rpm
Child returncode was: 0
LEAVE do --> 

ENTER do(['bash', '--login', '-c', 'rpmbuild -bb --target x86_64 --nodeps builddir/build/SPECS/kubevirt.spec'], False, '/mnt/build/mock/cnv-1.1-rhel-7-build-4110699-2545021/root/', None, 86400, True, False, 248, 135, None, logger=<mockbuild.trace_decorator.getLog object at 0xa54cd0>)
Executing command: ['bash', '--login', '-c', 'rpmbuild -bb --target x86_64 --nodeps builddir/build/SPECS/kubevirt.spec']
Building target platforms: x86_64
Building for target x86_64
Executing(%prep): /bin/sh -e /var/tmp/rpm-tmp.EqCFg6
+ umask 022
+ cd /builddir/build/BUILD
+ cd /builddir/build/BUILD
+ rm -rf kubevirt-0.6.0
+ /usr/bin/gzip -dc /builddir/build/SOURCES/kubevirt-0.6.0-1.git.0.cf13e2b.tar.gz
[...]

PR #1158, opened 2018-06-14
Default SOURCE_DATE_EPOCH to avoid using 'git'

rthallisey:
In downstream builds ```SOURCE_DATE_EPOCH=$(git show -s --format=format:%ct HEAD)``` fails because git isn't installed.  Make it so if this command fails, log the error and continue.  ```SOURCE_DATE_EPOCH``` gets set to "" on a failure.

fixes: https://github.com/kubevirt/kubevirt/issues/1141

### DIFF
--- a/hack/version.sh
+++ b/hack/version.sh
@@ -27,6 +27,9 @@
 #        "dirty" indicates source code changes after the git commit id
 #        "archive" indicates the tree was produced by 'git archive'
 #    KUBEVIRT_GIT_VERSION - "vX.Y" used to indicate the last release version.
+#    KUBEVIRT_SOURCE_DATE_EPOCH - unix timestamp.  Set to ' ' to generate using
+#          'date' instead of 'git'.
+#
 
 # Grovels through git to set a set of env variables.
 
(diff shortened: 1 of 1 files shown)


hack/version.sh
```diff
@@ -100,7 +100,7 @@ function kubevirt::version::ldflag() {
 function kubevirt::version::ldflags() {
     kubevirt::version::get_version_vars
 
-    SOURCE_DATE_EPOCH=$(git show -s --format=format:%ct HEAD)
+    set +e SOURCE_DATE_EPOCH=$(git show -s --format=format:%ct HEAD)
```
rmohr: Can you instead do `SOURCE_DATE_EPOCH=${KUBEVIRT_SOURCE_DATE_EPOCH-$(git show -s --format=format:%ct HEAD)}` and add `KUBEVIRT_SOURCE_DATA_EPOCH` to the three env variables on top?

hack/version.sh
```diff
@@ -27,6 +27,9 @@
 #        "dirty" indicates source code changes after the git commit id
 #        "archive" indicates the tree was produced by 'git archive'
 #    KUBEVIRT_GIT_VERSION - "vX.Y" used to indicate the last release version.
+#    KUBEVIRT_SOURCE_DATE_EPOCH - unix timestamp.  Set to ' ' to generate using
```
rmohr: you can also set it to `date +%s` to make it more explicit that you want the current build time.

Issue #773, opened 2018-02-27
Update Preset unit-tests in order to align with coding best practice


fabiand:
The preset unit-tests need an update, see https://github.com/kubevirt/kubevirt/pull/652#pullrequestreview-99638237

fabiand:
@stu-gott what's the status of this issue?

stu-gott:
@rmohr You mentioned that the VM Preset Unit tests were written "the old way". What specifically is your concern?

rmohr:
> @rmohr You mentioned that the VM Preset Unit tests were written "the old way". What specifically is your concern?

better to keep the tests in a consistent style to make improvements faster. We are not using the ghttp based approach anymore. We have  mocked worker queues which allow "relatively" easy testing of complex scenarios.

PR #1169, opened 2018-06-15
Use Gomock instead of ghttp in VMI Preset unit tests

stu-gott:
Fixes https://github.com/kubevirt/kubevirt/issues/773

### DIFF
--- a/pkg/virt-controller/watch/preset_test.go
+++ b/pkg/virt-controller/watch/preset_test.go
@@ -21,27 +21,26 @@ package watch
 
 import (
 	"fmt"
-	"net/http"
 	"testing"
-	"time"
 
+	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/onsi/gomega/ghttp"
 
 	k8sv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	k8smetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/cache/testing"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 
 	"kubevirt.io/kubevirt/pkg/api/v1"
 	"kubevirt.io/kubevirt/pkg/controller"
 	"kubevirt.io/kubevirt/pkg/kubecli"
-	"kubevirt.io/kubevirt/pkg/log"
+	"kubevirt.io/kubevirt/pkg/testutils"
 )
 
 type Event struct {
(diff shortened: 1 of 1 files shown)


pkg/virt-controller/watch/preset_test.go
```diff
@@ -21,27 +21,27 @@ package watch
...
 	. "github.com/onsi/gomega"
-	"github.com/onsi/gomega/ghttp"
 
 	k8sv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	k8smetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/watch"
+	//"k8s.io/apimachinery/pkg/watch"
```
j-griffith: Think you missed this one?

stu-gott: Good catch. thanks

pkg/virt-controller/watch/preset_test.go
```diff
@@ -884,19 +847,22 @@ var _ = Describe("VirtualMachineInstance Initializer", func() {
...
-				),
-			)
+			virtClient.EXPECT().VirtualMachineInstance(k8smetav1.NamespaceDefault).Return(vmiInterface)
 
-			err := app.vmiPresetController.initializeVirtualMachine(vmi)
+			vmiInterface.EXPECT().Update(gomock.Any()).Do(func(arg interface{}) {
+				vmi := arg.(*v1.VirtualMachineInstance)
+				val, found := vmi.Annotations["virtualmachinepreset.kubevirt.io/test-preset"]
+				Expect(found).To(BeTrue(), "preset should have been applied")
+				Expect(val).To(Equal("kubevirt.io/v1alpha2"))
```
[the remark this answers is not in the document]
davidvossel: I see, you're calling the controller.initializeVirtualMachine() and passing the vmi object directly in instead of controller.Execute()

Is there a reason we can't follow the same pattern in vm_test.go where the VM key is added to the queue and controller.Execute() is the entry point for the test?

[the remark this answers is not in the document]
stu-gott: Well that codepath is exercised directly a couple of times actually. But for the rest of the tests I took the more direct (and readable) approach of calling `initializeVirtualMachine` directly.

That said. I think you're right. there's no value in testing the values twice as they don't differ for any of the cases calling `initializeVirtualMachine`.

pkg/virt-controller/watch/preset_test.go
```diff
@@ -1034,21 +993,23 @@ var _ = Describe("VirtualMachineInstance Initializer", func() {
...
+				_, found := vmi.Annotations[initializerMarking]
+				Expect(found).To(BeTrue(), "vmi should have been initialized")
 
-			disk := vmi.Spec.Domain.Devices.Disks[0]
-			Expect(disk.Disk).ToNot(BeNil(), "DiskTarget should not be nil")
-			Expect(disk.Disk.Bus).ToNot(BeEmpty(), "DiskTarget's bus should not be empty")
+				Expect(vmi.Status.Phase).ToNot(Equal(v1.Failed))
+			}).Return(nil, nil)
+
+			err := vmiPresetController.initializeVirtualMachine(vmi)
```
j-griffith: app.vmiPresetController ?

stu-gott: What do you mean? anything referencing `app` was removed.

j-griffith: Yeah sorry, my bad.  I missed the removal at L#1013, but figured it out when going through the full file.  Meant to cancel that comment but was exiting everything out in a hurry and added instead :)  My bad.

stu-gott:
retest this please

PR #1180, opened 2018-06-19
Always remove scheduled error conditions from vms if pod gets scheduled

### LINKED ISSUE #1175: launcher in progress, but status still unscheduled
I have the situation that a VM's status is not scheduled (missing ready node), but the launcher is actually already launching:

```
$ oc get pods --all-namespaces | grep launcher ; oc get -o yaml vmi testvm | tail
myproject               virt-launcher-testvm-nt7q9       0/2       ContainerCreating   0          1m
    name: cloudinitvolume
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: 2018-06-18T19:04:27Z
    message: '0/1 nodes are available: 1 MatchNodeSelector.'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
  phase: Scheduling
```

rmohr:
Whenever the vm is in the scheduling phase, check the pod scheduled
condition and react to it. Previously the condition only got removed
after the Pod was scheduled and ready. It can take a lot of time for a
Pod to become ready, therefore always check the condition.

Fixes #1175

### DIFF
--- a/pkg/virt-controller/watch/vmi_test.go
+++ b/pkg/virt-controller/watch/vmi_test.go
@@ -104,6 +104,13 @@ var _ = Describe("VirtualMachineInstance watcher", func() {
 		})
 	}
 
+	ignorePodUpdates := func() {
+		kubeClient.Fake.PrependReactor("update", "pods", func(action testing.Action) (handled bool, obj runtime.Object, err error) {
+			update, _ := action.(testing.UpdateAction)
+			return true, update.GetObject(), nil
+		})
+	}
+
 	shouldExpectVirtualMachineSchedulingState := func(vmi *v1.VirtualMachineInstance) {
 		vmiInterface.EXPECT().Update(gomock.Any()).Do(func(arg interface{}) {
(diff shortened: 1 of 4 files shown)


pkg/virt-controller/watch/vmi_test.go
```diff
@@ -356,13 +356,13 @@ var _ = Describe("VirtualMachineInstance watcher", func() {
 			})
 		})
 
-		Context("when pod owned by virt-handler", func() {
-			It("should remove scheduling pod condition from the VirtualMachineInstance", func() {
+		Context("when Pod recovers from scheduling issues", func() {
```
cynepco3hahue: I do not sure about this tests, I think "shouldExpectPodHandover" not really must be part of the check, I expect something like:
- if the pod has scheduling false condition and the VM not, the condition must be added to the VM after the Execute
- if the pod does not have scheduling false condition and the VM has, the condition must be removed from the VM after the Execute

rmohr: Tried to address it with `ignorePodUpdates` and `testutils.IgnoreEvents`.

PR #1188, opened 2018-06-19
Add dots between sentences in api definition

booxter:
It seems like a common mistake to miss dots between sentences which results in
descriptions like: "ReadOnly Defaults to true". Which is sad.

This patch rights all the wrong. It consistently updates all descriptions for
all fields and structures to always end all sentences with dots, even if it's
just a single sentence.

While at it, several typos were fixed for affected descriptions, but fixing all
of them is not a goal of this patch.

Note: this patch doesn't guarantee new descriptions won't have this issue.
Reviewer, stay vigilant!

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3333,21 +3333,21 @@
    "v1.CDRomTarget": {
     "properties": {
      "bus": {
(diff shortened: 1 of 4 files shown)


fabiand:
Please rebase, then it can be moved fwd as #1167 is merged.

Issue #1184, opened 2018-06-19
Networking test is flakey


stu-gott:
As noted in: https://github.com/kubevirt/kubevirt/pull/1099

One networking test has odd behavior. It appears to run twice when run by itself (possibly indicating that something in the test forks). One of the runs passes and one fails.

```
hack/dockerized "hack/build-func-tests.sh"
sha256:06de53ac7ea69ecc45d2182f0509857fd8061a8d7b9654afdb6129b36463d340
go version go1.10 linux/amd64

go version go1.10 linux/amd64
Compiling tests...
    compiled tests.test
hack/functests.sh
Running Suite: Tests Suite
==========================
Random Seed: 1529422164
Will run 2 of 20 specs

SSSSSS
------------------------------
• [SLOW TEST:58.094 seconds]
Networking
/root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:48
  VirtualMachineInstance attached to implicit pod network
  /root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:363
    with a service matching the vmi exposed
    /root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:276
       should be able to reach the vmi based on labels specified on the vmi
      /root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:296
------------------------------
SSSSSSSSSTEP: starting a pod which tries to reach the vmi via the defined service
STEP: waiting for the pod to report a successful connection attempt
[...]

PR #1190, opened 2018-06-19
Fix tests

stu-gott:
Fix a couple functional tests. Split out from https://github.com/kubevirt/kubevirt/pull/1187 as they're needed regardless of that PR.

### DIFF
--- a/tests/vmi_lifecycle_test.go
+++ b/tests/vmi_lifecycle_test.go
@@ -608,7 +608,7 @@ var _ = Describe("VMIlifecycle", func() {
 			})
 
 			It("should request a TUN device but not KVM", func() {
-				err := virtClient.RestClient().Post().Resource("virtualmachines").Namespace(tests.NamespaceTestDefault).Body(vmi).Do().Error()
+				err := virtClient.RestClient().Post().Resource("virtualmachineinstances").Namespace(tests.NamespaceTestDefault).Body(vmi).Do().Error()
 				Expect(err).To(BeNil())
 
 				listOptions := metav1.ListOptions{}
--- a/tests/vmi_lifecycle_test.go
+++ b/tests/vmi_lifecycle_test.go
@@ -740,8 +740,9 @@ var _ = Describe("VMIlifecycle", func() {
 				nodeList, err := virtClient.CoreV1().Nodes().List(listOptions)
 				Expect(err).ToNot(HaveOccurred())
 
-				Expect(len(nodeList.Items)).ToNot(Equal(0))
-
+				if len(nodeList.Items) == 0 {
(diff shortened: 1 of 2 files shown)


cynepco3hahue:
@stu-gott I see we persistently have the test that fails on `k8s-1.10.3-release`
```
[Fail] VMIlifecycle Creating a VirtualMachineInstance VM Accelerated Mode [It] Should provide KVM via plugin framework 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_lifecycle_test.go:747
```
I think it related to KVM PR, can you please check it?

cynepco3hahue:
retest this please

tests/vmi_lifecycle_test.go
```diff
@@ -740,23 +726,16 @@ var _ = Describe("VMIlifecycle", func() {
 				nodeList, err := virtClient.CoreV1().Nodes().List(listOptions)
 				Expect(err).ToNot(HaveOccurred())
 
-				Expect(len(nodeList.Items)).ToNot(Equal(0))
-
+				if len(nodeList.Items) == 0 {
```
cynepco3hahue: Hm does it possible to have a cluster without nodes?

tests/vmi_lifecycle_test.go
```diff
@@ -740,23 +726,16 @@ var _ = Describe("VMIlifecycle", func() {
...
-
+				_, ok := node.Status.Allocatable[services.KvmDevice]
 				Expect(ok).To(BeTrue(), "KVM devices not allocatable on node: %s", node.Name)
-				// The number of devices allocated could change in the future, but it's
-				// for sure not 0
-				Expect(int(allocatableKvm.Value())).ToNot(Equal(0))
-
-				capacityKvm, ok := node.Status.Capacity[services.KvmDevice]
 
+				_, ok = node.Status.Capacity[services.KvmDevice]
```
cynepco3hahue: Why do you remove value check? I expect that capacity will equal to max devices constant, 110 if I remember correct

cynepco3hahue:
@stu-gott Looks much better

stu-gott:
retest this please

tests/vmi_lifecycle_test.go
stu-gott: It didn't serve a purpose actually. what we really needed to know is that the key was there at all--which the previous line already proves. Since that second (superfluous) test usually caused the test suite to fail without good reason, I removed it.

stu-gott:
The single test failure in this case is an unrelated test that's failed randomly lately.

https://github.com/kubevirt/kubevirt/issues/1184

PR #1192, opened 2018-06-20
Validate hostname and sanitize when its taken from the VM name

yanirq:
When creating a vm, sanitize hostname field and cloud-init if no hostname was specified on the vm spec. 

Fixes https://github.com/kubevirt/kubevirt/issues/1092

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ vendor/**/*_test.go
 **/polarion.xml
 tools/manifest-templator/manifest-templator
 tools/vms-generator/vms-generator
+.coverprofile
--- a/Makefile
+++ b/Makefile
@@ -44,8 +44,14 @@ deps-update:
 docker: build
 	hack/build-docker.sh build ${WHAT}
 
-publish: docker
+push: docker
 	hack/build-docker.sh push ${WHAT}
+
+publish: docker verify-build
+	hack/build-docker.sh push ${WHAT}
+
+verify-build:
+	hack/verify-build.sh
 
 manifests:
 	hack/dockerized "DOCKER_PREFIX=${DOCKER_PREFIX} DOCKER_TAG=${DOCKER_TAG} ./hack/build-manifests.sh"
--- a/cluster/ephemeral-provider-common.sh
+++ b/cluster/ephemeral-provider-common.sh
@@ -48,7 +48,7 @@ function build() {
 
     # Build everyting and publish it
     ${KUBEVIRT_PATH}hack/dockerized "DOCKER_TAG=${DOCKER_TAG} KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER} ./hack/build-manifests.sh"
-    make build docker publish
+    make push
 
     # Make sure that all nodes use the newest images
     container=""
--- a/hack/verify-build.sh
+++ b/hack/verify-build.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
+function report_dirty_build() {
+    set +e
+    echo "Build is not clean:"
+    cluster/virtctl.sh version
+    git status
+    exit 1
+}
+
+# Check that "clean" is reported at least once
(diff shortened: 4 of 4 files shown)


yanirq:
@rmohr  other than treating the dot for sanitization , do we want to also to remove all non alphanumeric lower case characters  / ''-''  ?

rmohr:
> @rmohr other than treating the dot for sanitization , do we want also to remove all non alphanumeric lower case characters / ''-'' ?

`-` should not be harmful. The `.` seems to be the only offending character which k8s names support: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

rmohr:
@yanirq I see two network tests failing on two test lanes:

```
[Fail] Networking VirtualMachineInstance attached to explicit pod network with a service matching the vmi exposed [It]  should be able to reach the vmi based on labels specified on the vmi 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:306

[Fail] Expose Expose UDP service on a VM Expose NodePort UDP service [It] Should expose a NodePort service on a VM and connect to it 
/root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:221
```

yanirq:
@rmohr seems like these are flaky tests. Letting CI run again.
Question :  what is the desired default behaviour when we don't add a subdomain to the spec ?

rmohr:
> Question : what is the desired default behaviour when we don't add a subdomain to the spec ?

It should still pass the sanitized vm name or the hostname to cloud-init and the pod.

pkg/virt-controller/services/template.go
```diff
@@ -303,7 +303,7 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 
 	containers = append(containers, container)
 
-	hostName := vmi.Name
+	hostName := strings.Split(vmi.Name, ".")[0]
```
rmohr: I think that we really should have one common helper function between cloud-init and pod.

rmohr:
retest this please

rmohr:
@yanirq will run the tests again. The last status was deleted. Once my comment is addressed it should be good to go in.

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -86,7 +85,7 @@ func (l *LibvirtDomainManager) preStartHook(vmi *v1.VirtualMachineInstance, doma
 	// generate cloud-init data
 	cloudInitData := cloudinit.GetCloudInitNoCloudSource(vmi)
 	if cloudInitData != nil {
-		hostName := strings.Split(vmi.Name, ".")[0]
+		hostName := util.SanitizeDNSLabel(vmi.Name)
```
rmohr: Almost there. If you now also move the fallback

```golang
		if vmi.Spec.Hostname != "" {
 			hostName = vmi.Spec.Hostname
 		}
```
into this helper we have the whole hostname logic isolated.

pkg/virt-launcher/virtwrap/util/sanitize.go
```diff
@@ -21,8 +21,18 @@ package util
...
+
+	"kubevirt.io/kubevirt/pkg/api/v1"
 )
 
-func SanitizeDNSLabel(value string) string {
-	return strings.Split(value, ".")[0]
+// Get sanitized hostname according to DNS label rules
+// In case vmi.Spec.Hostname has a value then it already passed
+// the webhook validation for DNS label rules.
+func SanitizeHostname(vmi *v1.VirtualMachineInstance) string {
```
rmohr: you will have to move this function to something like `pkg/util`. If you move it to `virt-launcher`, then this will pull in libvirt-cbindings which has the consequence that virt-controller starts requiring libvirt libraries.

That is also the reason why virt-controller never started in the last ci-run.

yanirq: I started that with pkg/util and removed it,  will bring it back now.  we didn't seem to have a util package up until now

PR #1197, opened 2018-06-21
Allow setting custom pkgdirs for cross-compilation

rmohr:
If building outside of our container environment, it can happen that
builds fail because the user does not have permissions to create the
default pkgdir locations uder ${GOROOT}. To overcome this limitation
allow setting an optional KUBEVIRT_GO_BASE_PKGDIR variable.

Example of what one can see if it can't write the standard locations:

```
go install runtime/internal/sys: mkdir /usr/lib/golang/pkg/darwin_amd64: permission denied
```

```release-note
NONE
```

### DIFF
--- a/hack/build-go.sh
+++ b/hack/build-go.sh
@@ -57,6 +57,13 @@ if [ $# -eq 0 ]; then
     fi
 fi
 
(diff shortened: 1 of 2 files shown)


rmohr:
retest this please

PR #1015, opened 2018-05-11
[WIP] virtctl ssh experiments [skip ci]

rmohr:
virtctl ssh experiments. Aimed to deliver two flows:
 * pure websocket proxying to kubectl with usual authentication on the clientside.
 * websocket connection with server-side ssh authentication, based on the v4 streaming protocol which is also used for e.g. `kubectl exec`.

The PR contains the following WIP code to illustrate possiblities:

Example: Store the credentials alongside a cloud-init secret:

```bash
$ cluster/kubectl.sh create secret generic cirros-nocloud --from-file=userdata=startup-script.sh --from-literal=ssh.user=cirros --from-literal=ssh.password=gocubsgo --from-literal=ssh.port=22
```

Create this VM:

```yaml
apiVersion: kubevirt.io/v1alpha1
kind: VirtualMachine
metadata:
  name: vm-nocloud
spec:
  domain:
    devices:
      disks:
      - disk:
          bus: virtio
        name: registrydisk
        volumeName: registryvolume
      - disk:
          bus: virtio
        name: cloudinitdisk
        volumeName: cloudinitvolume
      - disk:
          bus: virtio
        name: emptydisk
        volumeName: emptydiskvolume
    resources:
      requests:
        memory: 64M
  volumes:
  - name: registryvolume
    registryDisk:
      image: kubevirt/cirros-registry-disk-demo:devel
  - cloudInitNoCloud:
      secretRef:
        name: cirros-nocloud
    name: cloudinitvolume
  - emptyDisk:
      capacity: 2Gi
    name: emptydiskvolume
```

Access the VM:

```bash
# client-side ssh via golang ssh libraries
$ virtctl ssh -it cirros@vm-nocloud -- /bin/sh -l
cirros@vm-nocloud1`s password:
# servers-side ssh via k8s-exec protocol v4
$ virtctl wssh -it vm-nocloud -- /bin/sh -l
```

PR #1209, opened 2018-06-21
launcher: Drop console hack

### LINKED ISSUE #834: Remove ocp hack once console websocket is introduced
#819 needs to be reverted then.

fabiand:
We temporarily carried a console hack in order to allow the OpenShift
Webconsole to conncet to a VM's console. But since then we added a
API subresource which can be accessed, and it's up to the webconsole
to adjust to this new feature.

Originally introduced with #819

Fixes #834 

```release-note
A friendly hack we carried around for OpenShift Web Console is getting removed by this change.
```


fabiand:
I'd liekt o get this into v0.7.0

fabiand:
Right - Just to understand it correctly:

The current kubevirt console code uses a different transport than the k8s
container console.

The work in #1015 contains partial work to adjust the kubevirt code to have
the same transport as k8s container console.

Correct?

if so, then yes, dropping it in 0.8 is also fine for me.

rmohr:
> The work in #1015 contains partial work to adjust the kubevirt code to have
> the same transport as k8s container console.

yes it contains the necessary glue-code (the ssh sessions there are transported with the streaming protocol). That part can be copied and connected to the vm-console stream.

rmohr:
@mareklibra the discussion in that PR might be interesting for you.

fabiand:
Could you remind me what /hold means?

rmohr:
> Could you remind me what /hold means?

Updated the comment, was bad practice that I did not also add the reason. Just meant as a marker that although the PR may be fine we have reasons to not merge it right now. I had the impression we agreed on removing it once we provide the k8s streaming protocol support.

fabiand:
Just noting it here as well: The stream protocol has an issue. but there is another workaround which is to just launch the socket-connector from the terminal which is created in the webconsole for the VM serial console.

PR #1210, opened 2018-06-22
Move binaries to /usr/bin

rthallisey:
- Move binaries to /usr/bin to line up with downstream builds.
- Move virt-launcher scripts to ```/usr/share/kubevirt/virt-launcher/```
- Remove chown commands in the Dockerfile since the scripts have the proper permissions
- Condense the COPY scripts into a single layer

```release-note
NONE
```


rmohr:
ok to test

PR #1215, opened 2018-06-24
move port configuration to the interface level and add them to the pod manifest

### LINKED ISSUE #1211: Istio doesn't create the ingress traffic configuration
Istio init container looks for the container ports, if no port is configure istio doesn't configure the ingress traffic. [istio-sidecar-injector-configmap](https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/templates/sidecar-injector-configmap.yaml#L42)

We need to add the ports configured by the slirp to the compute container inside our virt-lancher pod.

SchSeba:
This PR move the port configuration from the slirp level up to the interfaces level and add them into the compute container ports inside the virt-lancher pod.

```yaml
interfaces:
      - name: rednet
        slirp: {}
        ports:
        - name: http
          port: 80
          protocol: TCP      
```

Primary usage is for the istio integration, when the istio sidecar is injected it checks if there are ports exposed by the pod, if not istio doesn't create the ingress traffic rules.

Fixes #1211

**Release note**:
```release-note
Ports configure for vms are now propagated as ContainerPorts to the virt-lancher pod.
Services can reference configured ports by name.
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/1215)


pkg/virt-controller/services/template_test.go
```diff
@@ -428,6 +428,52 @@ var _ = Describe("Template", func() {
...
+					},
+					Spec: v1.VirtualMachineInstanceSpec{Domain: domain},
+				}
+
+				pod, err := svc.RenderLaunchManifest(&vmi)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(len(pod.Spec.Containers)).To(Equal(1))
+				Expect(len(pod.Spec.Containers[0].Ports)).To(Equal(3))
+				Expect(pod.Spec.Containers[0].Ports[0].Name).To(Equal("tcp-80"))
```
booxter: Ideally, we should check that the array *contains* the expected ports but regardless of their order. But I guess it's harder to implement the check then.

SchSeba:
please retest this

SchSeba:
Checked the CI logs.

[Fail] Configurations VirtualMachineInstance definition with 3 CPU cores [It] should report 3 cpu cores under guest OS 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_configuration_test.go:78

Not relevant to this PR.

pkg/virt-controller/services/template.go
```diff
@@ -401,6 +406,38 @@ func getMemoryOverhead(domain v1.DomainSpec) *resource.Quantity {
...
 
+func getPortsFromVMI(vmi *v1.VirtualMachineInstance) []k8sv1.ContainerPort {
+	ports := make([]k8sv1.ContainerPort, 0)
+
+	configuredPorts := make(map[string]struct{})
+	for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
+		if iface.Slirp != nil && iface.Slirp.Ports != nil {
+			for _, port := range iface.Slirp.Ports {
+				if port.Protocol == "" {
+					port.Protocol = "TCP"
```
rmohr: You could copy it over blindly too since we follow the same pattern like on the pod.

SchSeba: No problem

pkg/virt-controller/services/template.go
```diff
@@ -401,6 +406,38 @@ func getMemoryOverhead(domain v1.DomainSpec) *resource.Quantity {
...
+				}
+				if port.PodPort == 0 {
```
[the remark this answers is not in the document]
rmohr: Yes, but why do we have it?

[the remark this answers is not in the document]
rmohr: Was there an example of an issue or a specific usecase we need to support here?

pkg/virt-controller/services/template.go
SchSeba: The copy will work but I need the protocol to be configured, I use it to create a unique name for the containerPort struct.

SchSeba:
About moving the port section to the interface I followed the [network API document](https://docs.google.com/document/d/10rXr91aqn8MvVcLgHw33WX8BaQwHPZERp25PHxoZGgw/edit#heading=h.kuh9o4iepyxb) when I create the slirp interface struct.

If we don't move the ports to be in the interface level we can have problems if we have two different interface with the same port name.

I wasn't sure about the name field. Its not mandatory now, as I said only the port is required, we want to make the name field  also mandatory?

The podPort is to give the user move flexibility, he can use it or not just like the hostPort in the container configuration. 
One scenario is to create multiple ports on the pod that will forward the traffic to the same port inside the virtual machine.

rmohr:
> About moving the port section to the interface I followed the network API document when I create the slirp interface struct.

Yes, still I think it needs to be challenged now, because with doing the mapping it will have an impact which goes beyond the scope of the slirp backend.

> If we don't move the ports to be in the interface level we can have problems if we have two different interface with the same port name.

Yes that would have to be validated. Also with the current implementation it would have to be validated as soon as other backends allow ports too.

> 
> I wasn't sure about the name field. Its not mandatory now, as I said only the port is required, we want to make the name field also mandatory?

It is also not mandatory on pods. Since pods don't have an issue with having the same containerPort exposed multiple times, the only thing you have to do is checking for duplicate names in the validation webhook (as far as I have seen that does not happen right now). You don't need to generate names.

> 
> The podPort is to give the user move flexibility, he can use it or not just like the hostPort in the container configuration.

If we add `hostPort` then mapping starts making sense (would be great if we add it). E.g. two vms on the same node have port 80 hard-coded and you want both to be accessible via the host IP. But then I would do the mapping between `port` and `hostPort`.

> One scenario is to create multiple ports on the pod that will forward the traffic to the same port inside the virtual machine.

What is the benefit if that comes from the same interface (with other words network)? Btw. services can do that for you on the cluster level. There you can then map 5 different ports to one port on one vm. If different hostnames are need you can also spread that across services.

vladikr:
My only concern with moving the ports section to interface level is that we will lose the fine-grained control of which ports each interface can map. On the other hand, at this point, it doesn't seem to matter.
It probably makes more sense to have the ports on a higher level now,  as we could re-introduce the ports section under slirp/proxy later if we will have a use case for it.

SchSeba:
No problem I will implement this tomorrow and update the pr

booxter:
> My only concern with moving the ports section to interface level is that we will lose the fine-grained control of which ports each interface can map

Why is it? Wouldn't you still have a separate ports section per interface, just on a top level?

pkg/virt-controller/services/template.go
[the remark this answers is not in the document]
fabiand: My 2ct - going with _no_ remapping for now (aka VM port == pod port) sounds good for now.

Containers have a similar situation: All processes in all containers need to share on network namespace. If there is a collision, the application needs to solve it (bind to a different port) for VMs we can assume the VM for now.

rmohr:
> My only concern with moving the ports section to interface level is that we will lose the fine-grained control of which ports each interface can map.

I think that @booxter is right here. I also think that we should make sure that the port names (if provided) are unique across all interfaces.

> It probably makes more sense to have the ports on a higher level now, as we could re-introduce the ports section under slirp/proxy later if we will have a use case for it.

Yep I think you are right, we can add specific port-mapping logic to the backend section of slirp if the requirement comes up.

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -553,6 +554,14 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 						})
 					}
 
+					if forwardPort.Port < 0 || forwardPort.Port > 65536 {
+						causes = append(causes, metav1.StatusCause{
+							Type:    metav1.CauseTypeFieldValueRequired,
```
booxter: I think you should use a different cause type (CauseTypeFieldValueInvalid).

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -565,15 +574,17 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
...
-						causes = append(causes, metav1.StatusCause{
-							Type:    metav1.CauseTypeFieldValueInvalid,
-							Message: fmt.Sprintf("Port and protocol combination must be unique"),
-							Field:   field.Child("domain", "devices", "interfaces").Index(idx).Child("ports").Index(portIdx).String(),
-						})
-					}
+					if forwardPort.Name != "" {
+						if _, ok := portForwardMap[forwardPort.Name]; ok {
+							causes = append(causes, metav1.StatusCause{
+								Type:    metav1.CauseTypeFieldValueInvalid,
```
booxter: CauseTypeFieldValueDuplicate?

pkg/virt-api/validating-webhook/validating-webhook_test.go
```diff
@@ -771,15 +776,42 @@ var _ = Describe("Validating Webhook", func() {
 
 			causes := ValidateVirtualMachineInstanceSpec(k8sfield.NewPath("fake"), &vm.Spec)
 			Expect(len(causes)).To(Equal(1))
-			Expect(causes[0].Field).To(Equal("fake.domain.devices.interfaces[0].ports[1]"))
+			Expect(causes[0].Field).To(Equal("fake.domain.devices.interfaces[0].ports[0]"))
 		})
-		It("should reject networks with a pod network source and slirp interface with two same Ports and Protocol", func() {
+		It("should reject interface with two port with the same name", func() {
```
booxter: -> two ports

pkg/virt-api/validating-webhook/validating-webhook_test.go
```diff
@@ -790,7 +822,26 @@ var _ = Describe("Validating Webhook", func() {
 
 			causes := ValidateVirtualMachineInstanceSpec(k8sfield.NewPath("fake"), &vm.Spec)
 			Expect(len(causes)).To(Equal(1))
-			Expect(causes[0].Field).To(Equal("fake.domain.devices.interfaces[0].ports[2]"))
+			Expect(causes[0].Field).To(Equal("fake.domain.devices.interfaces[1].ports[0].name"))
+		})
+		It("should allow interface with two same Ports and Protocol", func() {
```
booxter: with two same ports and different protocols

pkg/virt-api/validating-webhook/validating-webhook_test.go
```diff
@@ -771,15 +776,42 @@ var _ = Describe("Validating Webhook", func() {
...
-			Expect(causes[0].Field).To(Equal("fake.domain.devices.interfaces[0].ports[1]"))
+			Expect(causes[0].Field).To(Equal("fake.domain.devices.interfaces[0].ports[0]"))
 		})
-		It("should reject networks with a pod network source and slirp interface with two same Ports and Protocol", func() {
+		It("should reject interface with two port with the same name", func() {
 			vm := v1.NewMinimalVMI("testvm")
 			vm.Spec.Domain.Devices.Interfaces = []v1.Interface{v1.Interface{
 				Name: "default",
 				InterfaceBindingMethod: v1.InterfaceBindingMethod{
-					Slirp: &v1.InterfaceSlirp{Ports: []v1.Port{{Port: 80}, {Protocol: "UDP", Port: 80}, {Protocol: "TCP", Port: 80}}},
```
booxter: You seem to have lost a test case for when there are three ports, and first and last are duplicates.

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -565,15 +574,17 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 						forwardPort.Protocol = "TCP"
 					}
 
-					if _, ok := portForwardMap[fmt.Sprintf("%s-%d", forwardPort.Protocol, forwardPort.Port)]; ok {
-						causes = append(causes, metav1.StatusCause{
-							Type:    metav1.CauseTypeFieldValueInvalid,
-							Message: fmt.Sprintf("Port and protocol combination must be unique"),
-							Field:   field.Child("domain", "devices", "interfaces").Index(idx).Child("ports").Index(portIdx).String(),
```
booxter: You no longer validate that there are no duplicate port definitions for the same interface. Was the change intended?

pkg/virt-controller/services/template.go
```diff
@@ -401,6 +405,28 @@ func getMemoryOverhead(domain v1.DomainSpec) *resource.Quantity {
...
+	ports := make([]k8sv1.ContainerPort, 0)
+
+	for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
+		if iface.Ports != nil {
+			for _, port := range iface.Ports {
+				if port.Protocol == "" {
+					port.Protocol = "TCP"
+				}
+
+				ports = append(ports, k8sv1.ContainerPort{Protocol: k8sv1.Protocol(port.Protocol), Name: port.Name, ContainerPort: port.Port})
```
booxter: If that was indeed the intent to allow duplicate ports in API, would it make sense to dedup them here to avoid passing duplicates to container spec? (I am still puzzled why we may want to accept duplicates in API so maybe this comment is moot.)

pkg/virt-controller/services/template_test.go
```diff
@@ -428,6 +428,56 @@ var _ = Describe("Template", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(len(pod.Spec.Containers)).To(Equal(1))
+				Expect(len(pod.Spec.Containers[0].Ports)).To(Equal(0))
+			})
+			It("Should create a port list in the pod manifest", func() {
+				slirpInterface := v1.InterfaceSlirp{}
+				ports := []v1.Port{{Name: "http", Port: 80}, {Protocol: "UDP", Port: 80}, {Port: 90}, {Name: "other-http", Port: 80}}
+				domain := v1.DomainSpec{}
+				domain.Devices.Interfaces = []v1.Interface{{Name: "testnet", Ports: ports, InterfaceBindingMethod: v1.InterfaceBindingMethod{Slirp: &slirpInterface}}}
```
booxter: Please add a case, or modify this one, to execute with multiple interfaces, to check all ports for all interfaces are combined and passed forward. Also, since you allow duplicate ports for the same interface in API, maybe also add a case to pass duplicate ports.

api/openapi-spec/swagger.json
```diff
@@ -4259,23 +4256,22 @@
...
    },
    "v1.Port": {
-    "description": "Port repesents a port to expose from the virtual machine.\nDefault protocol TCP.\nDefault port is PodPort.",
+    "description": "Port repesents a port to expose from the virtual machine.\nDefault protocol TCP.\nDefault podPort is port.",
     "required": [
      "port"
     ],
     "properties": {
      "name": {
+      "description": "If specified, this must be an IANA_SVC_NAME and unique within the pod. Each\nnamed port in a pod must have a unique name. Name for the port that can be\nreferred to by services.\n+optional",
```
yuvalif: is it optional in case of slirp? could we define slirp without any port?

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -595,11 +597,13 @@ func createSlirpNetwork(iface v1.Interface, network v1.Network, domain *Domain)
 }
 
 func configPortForward(qemuArg *Arg, iface v1.Interface) error {
-	if iface.Slirp.Ports == nil {
+	if iface.Ports == nil {
 		return nil
 	}
 
-	for _, forwardPort := range iface.Slirp.Ports {
+	// Cant be duplicated ports forward or the qemu process will crash
```
booxter: -> Can't

api/openapi-spec/swagger.json
```diff
@@ -4259,23 +4256,22 @@
...
       "type": "string"
      },
-     "podPort": {
-      "type": "integer",
-      "format": "int32"
-     },
      "port": {
+      "description": "Number of port to expose for the virtual machine.\nThis must be a valid port number, 0 \u003c x \u003c 65536.",
```
yuvalif: dont think that 0 and 65536 are valid port numbers. should be 1-65,535

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -609,13 +613,10 @@ func configPortForward(qemuArg *Arg, iface v1.Interface) error {
 			forwardPort.Protocol = DefaultProtocol
 		}
 
-		// Check if PodPort is configure If not Get the same Port as the vm port
-		if forwardPort.PodPort == 0 {
-			forwardPort.PodPort = forwardPort.Port
+		if _, ok := configuredPorts[fmt.Sprintf("%s-%d", forwardPort.Protocol, forwardPort.Port)]; !ok {
+			qemuArg.Value += fmt.Sprintf(",hostfwd=%s::%d-:%d", strings.ToLower(forwardPort.Protocol), forwardPort.Port, forwardPort.Port)
+			configuredPorts[fmt.Sprintf("%s-%d", forwardPort.Protocol, forwardPort.Port)] = struct{}{}
```
booxter: nit: you could construct it once in line before the if and then reuse in both if and here.

tests/vmi_slirp_interface_test.go
```diff
@@ -43,17 +44,31 @@ var _ = Describe("Networking", func() {
...
-			ports := []v1.Port{{Port: 80}}
+			ports := []v1.Port{{Name: "http", Port: 80}}
 			vmi = tests.NewRandomVMIWithSlirpInterfaceEphemeralDiskAndUserdata(tests.RegistryDiskFor(tests.RegistryDiskCirros), "#!/bin/bash\necho 'hello'\n", ports)
 			vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
 			Expect(err).ToNot(HaveOccurred())
 			tests.WaitForSuccessfulVMIStartIgnoreWarnings(vmi)
 			generateHelloWorldServer(vmi, virtClient, 80, "tcp")
 		})
-
+		It("should contains containerPort in the pod manifest", func() {
```
booxter: nit: should contain

api/openapi-spec/swagger.json
yuvalif: ignore. i din't see the "<" sign

pkg/virt-controller/services/template_test.go
```diff
@@ -428,6 +428,56 @@ var _ = Describe("Template", func() {
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "testvmi", Namespace: "default", UID: "1234",
+					},
+					Spec: v1.VirtualMachineInstanceSpec{Domain: domain},
+				}
+
+				pod, err := svc.RenderLaunchManifest(&vmi)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(len(pod.Spec.Containers)).To(Equal(1))
```
yuvalif: as @booxter mentioned before. this may break if something changes in internal implementation even if all functionality remain the same. you should probably search the map

api/openapi-spec/swagger.json
SchSeba: We only need to port number just like kubernetes does.
the name is not mandatory.

pkg/virt-api/validating-webhook/validating-webhook_test.go
SchSeba: We stick with the kubernetes configuration.
In kubernetes you can configure the same port multiple time, the only thing that need to be unique is the name (if you gave one)

pkg/api/v1/openapi_generated.go
```diff
@@ -1047,30 +1045,27 @@ func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenA
...
 			Schema: spec.Schema{
 				SchemaProps: spec.SchemaProps{
-					Description: "Port repesents a port to expose from the virtual machine. Default protocol TCP. Default port is PodPort.",
+					Description: "Port repesents a port to expose from the virtual machine. Default protocol TCP. Default podPort is port.",
 					Properties: map[string]spec.Schema{
 						"name": {
 							SchemaProps: spec.SchemaProps{
-								Type:   []string{"string"},
-								Format: "",
+								Description: "If specified, this must be an IANA_SVC_NAME and unique within the pod. Each named port in a pod must have a unique name. Name for the port that can be referred to by services.",
```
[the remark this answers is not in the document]
SchSeba: That mean you can write the port name when you expose a service and kubernetes will find the port and protocol by this name

SchSeba:
I will change the name of the PR.
Now we allow to connect multiple interfaces to the same network this is intended ?
If yes we need to rethink about the port because it put more complex like:
1. two ports with the same name in different interfaces ? (allow/reject)
2. two ports with the same name in different interfaces but for the same network ?  (allow/reject)

Maybe even move the ports as part of the pod network configuration?
I don't see why exposed ports from network like DP need to be configure on the pod spec (kubernetes doesnt need to know about it), But maybe for the Network CR (Multus) we need to add this ports to the pod specs.

Please comment what you think about it.

booxter:
I believe it was not an intent to allow multiple interfaces backed by the same network (that being said, it's allowed to declare two networks with identical properties but unique names).

As for whether we should put ports for non-pod interfaces into the pod spec, 1) I don't think so; and 2) if that's somehow the case, we can revisit it later when we add support for these advanced network types. It's better to be explicit right now and NOT process ports from non-pod interfaces.

I don't think we should move ports to global level above interfaces, because users may want to have different services exposed on different interfaces.

Also, tactically, I would suggest to split this PR into two separate pieces. One would move ports definition up to interface level. Another would pass the ports into pod spec for pod interfaces. These two pieces seem to be loosely coupled and would merit independent review and merge.

fabiand:
> I think that @booxter is right here. I also think that we should make sure that the port names (if provided) are unique across all interfaces.

Why would we want to enforce uniqueness across interfaces?

> This PR move the port configuration from the slirp level up to the interfaces level and add them into the compute container ports inside the virt-lancher pod.

There are many things here which are unclear to me:

- If on the interface level, will ports be mandatory?
- If on the interface level, will SLRIP require the interface ports to be defined (even just an empty list)?
- When will ports be added to the pod spec? Just for SLIRP connections? Just for pod network (if it's used)?
- Is now the uniqueness for one port across all interfaces required?

- What does it semantically mean if we define ports for an L2 interface where we don't know TPC/UDP is available at all?

This PR changed a bit which now makes it hard to review it.

My gut feeling btw: Ports on interface _could_ be ok, but definetly optional - just for connections which require it like slirp, it needs to be required. Ports for L2 interfaces are problematic, as thy provide little value, and can only represent a subset of what L2 could provide - but maybe that's still valuable. Ports must not be required to be unique across interfaces.

SchSeba:
> Why would we want to enforce uniqueness across interfaces?

The uniqueness is across interfaces with pod network but we only allow one pod network.
For multiple kubernetes networks we need to use something like multus and this is an other backend not pod.

>There are many things here which are unclear to me:
If on the interface level, will ports be mandatory?

There are not mandatory just like in kubernetes

> If on the interface level, will SLRIP require the interface ports to be defined (even just an empty list)?

There no ports anymore in the slirp level. In the future we can add a forward configuration inside the slirp to forward a different port to the vm (but we cant find any use case for this now)

> When will ports be added to the pod spec? Just for SLIRP connections? Just for pod network (if it's used)?

We add them for the pod network now, in the future we may have to add them for the multus network backend to.
Container ports are not mandatory but we need them for istio and if the user configure a name to the port when he expose the pod he can use the name and not the port number in the expose command this is why only the names must be unique.

> Is now the uniqueness for one port across all interfaces required?

Now we can only have one pod network connected to one interface as @booxter said.
After this PR I will work on this because now the interface names are not unique (you can configure multiple interfaces connected to the same pod network but this make the vm to crush)

> What does it semantically mean if we define ports for an L2 interface where we don't know TPC/UDP is available at all?

That's the reason why we don't what to add the ports from the DP network and only from networks inside the kubernetes cluster like pod and multus backends 

> This PR changed a bit which now makes it hard to review it.

> My gut feeling btw: Ports on interface _could_ be ok, but definetly optional - just for connections which require it like slirp, it needs to be required. Ports for L2 interfaces are problematic, as thy provide little value, and can only represent a subset of what L2 could provide - but maybe that's still valuable. Ports must not be required to be unique across interfaces.

Even with the slirp interface the user can decide if he want to expose ports or not (maybe the vm only send traffic) then if the user not define exposed ports istio will configure the service mesh to only allow egress traffic from the vm.

rmohr:
@fabiand 

> What does it semantically mean if we define ports for an L2 interface where we don't know TPC/UDP is available at all?

For L2 networks it would be not used by us in any way right now but since it is structured, external services on that L2 service may use it if they like (e.g. setting up firewalls, ...). Being able to document it is in my oppinon valuable.

> Ports must not be required to be unique across interfaces.

Are you talking about the port number or the port name?

fabiand:
> The uniqueness is across interfaces with pod network but we only allow one pod network.

If it's only uniqueness across pod networks, then it sounds good.

> There no ports anymore in the slirp level.

That's fine. If SLIRP is used, then this should require the user to either specify ports or an empty ports list. If no port configuration (on the interface level) is provided then there should be an error - to highlight that slirp relies on ports.

> We add them for the pod network now, in the future we may have to add them for the multus network backend to.

My question was more: How do we decide when ports need to be added to the pod spec?

> After this PR I will work on this because now the interface names are not unique (you can configure multiple interfaces connected to the same pod network but this make the vm to crush)

On what will you start to work? What is the goal?

> That's the reason why we don't what to add the ports from the DP network and only from networks inside the kubernetes cluster like pod and multus backends

Right. But by moving the ports definition to the interface level a user can specify ports for a DP backed interface.

> Even with the slirp interface the user can decide if he want to expose ports or not (maybe the vm only send traffic) then if the user not define exposed ports istio will configure the service mesh to only allow egress traffic from the vm.

Right. IIUIC you are saying: The VM can expose ports, but then it can be that only few or none of them are relayed through istio. Whic his fine.
But how would this be configured? Are you saiyng this should be done now, or rather saying this can be an option on the long run?

fabiand:
> Ports must not be required to be unique across interfaces.
> 
> Are you talking about the port number or the port name?

Numbers in this case.
But unique port names make sense.

fabiand:
In short - I don't block on putting ports on interfaces, as it's not critical, and because I don't have a better solution atm.

But it leaves a .. not so nice taste, as interfaces so far were only about frontend configuration (aka what the guest is seeing). With ports there is now something which is not relevant for the guest.

rmohr:
> But it leaves a .. not so nice taste, as interfaces so far were only about frontend configuration (aka what the guest is seeing). With ports there is now something which is not relevant for the guest.

I see what you mean. I think that there are reason for and against each section to put that information. If you want to express that "through this interface these ports and services are available", then it more belongs to the vms interface. If you want to express that "this technology-backend for slirp thakes vm-port a and connects it to network x" then it is better of in the backend.

rmohr:
I personally prefer the first one because it is more service orientated in my eyes but the other one does not hurt either.

SchSeba:
> That's fine. If SLIRP is used, then this should require the user to either specify ports or an empty ports list. If no port configuration (on the interface level) is provided then there should be an error - to highlight that slirp relies on ports.

I don't think it need to be an error if the user not configure any port, as I said the vm may only generate traffic and the user need to use the slirp interface because the services the vm communicate with are inside the istio service mesh

> My question was more: How do we decide when ports need to be added to the pod spec?

To be 100% kubernetes compatibility any time we connect the interface to the kubernetes network we need to add the ports to the container configuration because the user may use them to create services or sidecars may need them (like istio) 

> On what will you start to work? What is the goal?

Make 1to1 relation between network and interface.

Now this manifest is allowed but make the vm to crush

```yaml
interfaces:
- name: red
  bridge: {}
- name red
  bridge {}
....
network:
- name: red
  pod: {}
```

> Right. But by moving the ports definition to the interface level a user can specify ports for a DP backed interface.

Yes but this is ok we are not going to configure them to the pod but the user may use them to configure is external network (firewall, ips and other network devices) automatically. (he can create a controller to watch on virtual machines creation with DP and configure is network)

> Right. IIUIC you are saying: The VM can expose ports, but then it can be that only few or none of them are relayed through istio. Whic his fine.
But how would this be configured? Are you saiyng this should be done now, or rather saying this can be an option on the long run?

Just like a pod this is up to the user. He can configure only the istio ports or all the ports the pod expose. For us it's the same.

fabiand:
> Now this manifest is allowed but make the vm to crush

Oh yes. names must be unique in the interface list.

SchSeba:
Can we run the tests again?

SchSeba:
@booxter @rmohr Can you make another review please ?

pkg/api/v1/schema.go
```diff
@@ -693,21 +695,26 @@ type InterfaceBridge struct{}
...
-type InterfaceSlirp struct {
-	// List of ports to be forwarded to the virtual machine.
-	Ports []Port `json:"ports,omitempty"`
-}
+type InterfaceSlirp struct{}
 
 // Port repesents a port to expose from the virtual machine.
 // Default protocol TCP.
-// Default port is PodPort.
+// Port is mandatory.
```
booxter: It would be slightly less confusing if it would say: "The `port` field is mandatory." because the API field is named "port" (lower case).

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -579,15 +589,17 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
...
-						})
-					}
+					if forwardPort.Name != "" {
+						if _, ok := portForwardMap[forwardPort.Name]; ok {
+							causes = append(causes, metav1.StatusCause{
+								Type:    metav1.CauseTypeFieldValueDuplicate,
+								Message: fmt.Sprintf("Duplicate value: %s", forwardPort.Name),
```
booxter: it could also indicate it's a duplicate "name of the port" and not just a vague "value"

rmohr:
@SchSeba could you update the PR description?

rmohr:
> @SchSeba could you update the PR description?

@SchSeba everyting fine. Misread it.

SchSeba:
@rmohr What I need to add for the description?

cluster/examples/vmi-slirp.yaml
```diff
@@ -19,12 +19,11 @@ spec:
...
-          ports:
-          - name: http
-            podPort: 80
-            port: 80
-            protocol: TCP
+        ports:
+        - name: http
+          port: 80
+          protocol: TCP
+        slirp: {}
```
rmohr: I more and more have the feeling that `slirp` and `bridge` should go to the `network` section under `pod`. Maybe we can discuss that too outside of this PR. @booxter @SchSeba.

SchSeba: I think it will be best to speak about it outside of this PR. We may also create a post about it in the kubevirt group to get more comments.

SchSeba:
The failing tests don't see relevant to this PR.

PR #1226, opened 2018-06-26
Skip virt-handler crush test on k8s version < 1.10.3

cynepco3hahue:
`kubelet` does not update allocated device if device plugin was restarted on k8s version < `1.10.3`, so we need to skip `virt-handler` crash tests under specific k8s version.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -32,6 +32,7 @@ import (
 	"reflect"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/google/goexpect"
@@ -48,6 +49,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/rand"
+	k8sversion "k8s.io/apimachinery/pkg/version"
 
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/remotecommand"
@@ -1434,4 +1436,22 @@ func GetNodeWithHugepages(virtClient kubecli.KubevirtClient, hugepages k8sv1.Res
 		}
 	}
 	return nil
+}
+
+// SkipIfNoVersion will skip tests if it runs on k8s version less than 1.10.3
+func SkipIfNoVersion(virtClient kubecli.KubevirtClient) {
+	version, err := virtClient.RestClient().Get().AbsPath("/version").DoRaw()
+	Expect(err).ToNot(HaveOccurred())
+
+	var info k8sversion.Info
+
+	err = json.Unmarshal(version, &info)
+	Expect(err).ToNot(HaveOccurred())
+
+	gitVersion := strings.Split(info.GitVersion, ".")
+	gitMinorVersion, err := strconv.Atoi(strings.Split(gitVersion[2], "+")[0])
+	Expect(err).ToNot(HaveOccurred())
+	if gitMinorVersion < 3 {
+		Skip(fmt.Sprintf("Test does not supported under the version %s", info.String()))
+	}
 }
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -1438,20 +1438,23 @@ func GetNodeWithHugepages(virtClient kubecli.KubevirtClient, hugepages k8sv1.Res
 	return nil
 }
 
-// SkipIfNoVersion will skip tests if it runs on k8s version less than 1.10.3
-func SkipIfNoVersion(virtClient kubecli.KubevirtClient) {
-	version, err := virtClient.RestClient().Get().AbsPath("/version").DoRaw()
+// SkipIfVersionBelow will skip tests if it runs on an environment with k8s version below specified
+func SkipIfVersionBelow(message string, expectedVersion string) {
+	virtClient, err := kubecli.GetKubevirtClient()
+	PanicOnError(err)
+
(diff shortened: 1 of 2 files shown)


tests/vmi_lifecycle_test.go
```diff
@@ -297,6 +297,8 @@ var _ = Describe("VMIlifecycle", func() {
 
 		Context("when virt-handler crashes", func() {
 			It("should recover and continue management", func() {
+				tests.SkipIfNoVersion(virtClient)
```
rmohr: Can we make something like `tests.SkipIfVersionBelow("1.10.4", "Device plugins have issue X in these versions.")`? Otherwise we will for sure forget why we skipped which tests and for which version.

tests/utils.go
```diff
@@ -1435,3 +1437,21 @@ func GetNodeWithHugepages(virtClient kubecli.KubevirtClient, hugepages k8sv1.Res
...
+
+	var info k8sversion.Info
+
+	err = json.Unmarshal(version, &info)
+	Expect(err).ToNot(HaveOccurred())
+
+	gitVersion := strings.Split(info.GitVersion, ".")
+	gitMinorVersion, err := strconv.Atoi(strings.Split(gitVersion[2], "+")[0])
+	Expect(err).ToNot(HaveOccurred())
+	if gitMinorVersion < 3 {
```
rmohr: The check looks like it would also fail if we have e.g. `1.11.2`.

cynepco3hahue:
@rmohr You are right in both cases, will do the method more genereic

tests/utils.go
```diff
@@ -1435,3 +1437,25 @@ func GetNodeWithHugepages(virtClient kubecli.KubevirtClient, hugepages k8sv1.Res
 	}
 	return nil
 }
+
+// SkipIfVersionBelow will skip tests if it runs on an environment with k8s version below specified
+func SkipIfVersionBelow(virtClient kubecli.KubevirtClient, message string, version string) {
```
rmohr: You can fetch a new client. Should be cheap. We do that with other helper functions too, so that you don't have to permanently pass clients around..

tests/utils.go
```diff
@@ -1435,3 +1437,25 @@ func GetNodeWithHugepages(virtClient kubecli.KubevirtClient, hugepages k8sv1.Res
...
+
+	var info k8sversion.Info
+
+	err = json.Unmarshal(response, &info)
+	Expect(err).ToNot(HaveOccurred())
+
+	gitVersion := strings.Split(info.GitVersion, "+")[0]
+	curVersion := strings.Split(gitVersion, ".")
+	expectedVersion := strings.Split(version, ".")
+	Expect(len(expectedVersion)).To(Equal(len(curVersion)))
```
rmohr: I think that you can simply use `<` on `gitVersion`. Just make sure that you strip te leading `v`.

E.g. `"v1.9" < "v1.20.3"`

cynepco3hahue: done

stu-gott:
I'm not sure this PR is the correct fix for the symptom in question. On my system, I tried the following steps:

 * make cluster-down
 * export KUBEVIRT_PROVIDER=k8s-1.10.3
 * make cluster-sync
 * kubectl get nodes -o yaml | grep 110
 * kubectl -n kube-system get pods
 * kubectl -n kube-system delete pod <virt-handler>
 * kubectl get nodes -o yaml | grep 110

At least on my system, the system recovers just fine. I still see 110 allocatable devices after manually deleting the virt-handler pod.

Are these steps correct to expose the bug this PR is trying to address? If so, the problem is something else.

cynepco3hahue:
retest this please

cynepco3hahue:
> I'm not sure this PR is the correct fix for the symptom in question. On my system, I tried the following steps:
> 
> make cluster-down
> export KUBEVIRT_PROVIDER=k8s-1.10.3
> make cluster-sync
> kubectl get nodes -o yaml | grep 110
> kubectl -n kube-system get pods
> kubectl -n kube-system delete pod
> kubectl get nodes -o yaml | grep 110
> At least on my system, the system recovers just fine. I still see 110 allocatable devices after manually deleting the virt-handler pod.
> 
> Are these steps correct to expose the bug this PR is trying to address? If so, the problem is something else.

`os-3.10.0` runs with `k8s-1.10.0` that has troubles with re-register device plugin in the case of a crash, so I skip all `virt-handler` crash tests in the case when k8s version less than `1.10.3`

tests/utils.go
```diff
@@ -1435,3 +1437,24 @@ func GetNodeWithHugepages(virtClient kubecli.KubevirtClient, hugepages k8sv1.Res
...
+
+	var info k8sversion.Info
+
+	err = json.Unmarshal(response, &info)
+	Expect(err).ToNot(HaveOccurred())
+
+	curVersion := strings.Split(info.GitVersion, "+")[0]
+	curVersion = strings.Trim(curVersion, "v")
+
+	if curVersion < expectedVersion {
```
stu-gott: On second thought, I'm not entirely comfortable with this. This line is doing a lexical comparison for "less than" on semantic version strings. If almost feels like it gets the right answer by accident.

cynepco3hahue: @stu-gott I discussed it with Roman, string comparison convert each character to ASCII and compare it, so until we do not changes under ASCII table all must work as expected :smile:

rmohr: The comment here is also helpful in this context: https://golang.org/src/strings/compare.go

PR #1232, opened 2018-06-27
Provide additional registry disks

karmab:
```release-note
the following registry disk images are added:

- arch-registry-disk-demo
- debian-registry-disk-demo
- gentoo-registry-disk-demo
- ubuntu-registry-disk-demo
```

### DIFF
--- a/images/arch-container-disk-demo/Dockerfile
+++ b/images/arch-container-disk-demo/Dockerfile
@@ -1,0 +1,24 @@
+#
+# This file is part of the KubeVirt project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


simon3z:
@karmab I know it's more maintenance work, but do we feel we should keep a valid checksum/SHA to make sure the images are correctly downloaded and they're not broken/tampered with?

simon3z:
> @simon3z You mean performing a checksum in the dockerfile to ensure it was downloaded correctly?

@fabiand yes, something like:

```
curl ... && sha512sum ...
```

So that if the downloaded image is not what we expect we have an early failure/feedback (already mentioned it's more maintenance work).
Or any other more advanced functional test to ensure images are booting.

rmohr:
> Do we want to include an automated way to build and sync these optional registry disks so people can test with them?
> 
> Also, it might be nice for us to host a repo of common registryDisks for people to try out.

@davidvossel I want to pick up your comment. Let's create an extra repo on quay for all of these images.

rmohr:
> @karmab I know it's more maintenance work, but do we feel we should keep a valid checksum/SHA to make sure the images are correctly downloaded and they're not broken/tampered with?

That would be very valuable and we should definitely do that.

fabiand:
Any update, otherwise it will get closed.

karmab:
@fabiand  i ve updated it this week actually, i think so let me know whether the md5sum is mandatory in the Docker file and what s missing for this to get merged

PR #1234, opened 2018-06-27
Remove default kubeconfig value

stu-gott:
Default points to vagrant, which is now invalid. Default is not used
by any provider.

/release-note-none


stu-gott:
fixes a concern noted in https://github.com/kubevirt/kubevirt/issues/844

PR #1237, opened 2018-06-27
Make sure that the VMI replica set will remove VM's in the final state

### LINKED ISSUE #1107: VirtualMachineReplicaSet does not delete finished VMs
VirtualMachineReplicaSet quietly assumes that a restartPolicy on the vm restarts failed vms and does not delete failed vms. We don't have policies on VMs. As a consequence the replicaset creates new vm instances, but it does not delete vms in a final state.

How to reproduce:
 * Create a replicaset with e.g. 2 replicas
 * Delete the pods of the vm (will make the vms go to a final state)
 * You will now see 4 vms (two fresh ones and two in a final state)

Expected:
 * The controller should remove the two vms in final state and not just create new instances

cynepco3hahue:
Now VMI replica set will delete VM's in the final state

Fixes #1107 

**Release note**:

```release-note
VMI replica set will remove VM's in the final state
```

### DIFF
--- a/pkg/virt-controller/watch/replicaset.go
+++ b/pkg/virt-controller/watch/replicaset.go
@@ -185,8 +185,6 @@ func (c *VMIReplicaSet) execute(key string) error {
 		return err
 	}
 
-	vmis = c.filterActiveVMIs(vmis)
-
 	// If any adoptions are attempted, we should first recheck for deletion with
 	// an uncached quorum read sometime after listing VirtualMachines (see kubernetes/kubernetes#42639).
 	canAdoptFunc := controller.RecheckDeletionTimestamp(func() (v1.Object, error) {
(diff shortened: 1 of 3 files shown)


cynepco3hahue:
@rmohr Yes, will do it, it really will make `scale` logic less complicated.

pkg/virt-controller/watch/replicaset_test.go
```diff
@@ -306,12 +307,13 @@ var _ = Describe("Replicaset", func() {
 			vmiFeeder.Modify(modifiedVMI)
 
 			// Expect the re-crate of the VirtualMachineInstance
+			rsInterface.EXPECT().Update(rsCopy).Times(1)
```
rmohr: I see no new unit test which explicitly addresses the cleanup. Can you add one or two?

pkg/virt-controller/watch/replicaset.go
```diff
@@ -341,7 +352,7 @@ func (c *VMIReplicaSet) scale(rs *virtv1.VirtualMachineInstanceReplicaSet, vmis
 // filterActiveVMIs takes a list of VMIs and returns all VMIs which are not in a final state
 func (c *VMIReplicaSet) filterActiveVMIs(vmis []*virtv1.VirtualMachineInstance) []*virtv1.VirtualMachineInstance {
 	return filter(vmis, func(vmi *virtv1.VirtualMachineInstance) bool {
-		return !vmi.IsFinal()
+		return !vmi.IsFinal() && vmi.DeletionTimestamp == nil
```
rmohr: Since that change is introduce here, it might make sense to copy the two tests from #1238 too.

pkg/virt-controller/watch/replicaset.go
```diff
@@ -469,7 +487,7 @@ func (c *VMIReplicaSet) updateVirtualMachine(old, cur interface{}) {
 	}
 
 	labelChanged := !reflect.DeepEqual(curVMI.Labels, oldVMI.Labels)
-	if curVMI.DeletionTimestamp != nil {
+	if curVMI.DeletionTimestamp != nil || curVMI.IsFinal() {
```
rmohr: That looks wrong. A vm which is final is not a delete.

cynepco3hahue: @rmohr Without it, the code does not call to `c.deleteVirtualMachine` that update expectations, in this case, `needSync` variable will have `false` value and the code will skip clean and scale stages.

pkg/virt-controller/watch/replicaset.go
cynepco3hahue: I think we can just rebase your PR after this one will be merged

pkg/virt-controller/watch/replicaset_test.go
cynepco3hahue: will add it

pkg/virt-controller/watch/replicaset.go
rmohr: If that's the case, then there is something else wrong. A deletion expectation for a UID is only fulfilled if you either see a real delete (object immediately disappears from the cluster, in this case `deleteVirtualMachine` is directly invoked by the informer) or if you get a chache update that the deletion timestamp was set (that's the situation in this line).

rmohr:
retest this please

rmohr:
retest this please

PR #1238, opened 2018-06-27
Add edge case tests for VMIRS when vms go down

rmohr:
Tests two cases:
 * Replicasets are meant for availability. Test that as soon as a vmi is
marked for deletion we will create a replacement.
 * If a VMI is already marked for deletion, don't delete it again

Adds test for changes in #1237.

```release-note
NONE
```

### DIFF
--- a/pkg/virt-controller/watch/replicaset_test.go
+++ b/pkg/virt-controller/watch/replicaset_test.go
@@ -174,6 +174,42 @@ var _ = Describe("Replicaset", func() {
 			// TODO test for missing 5
 		})
 
+		It("should already create replacement VMIs once it discovers that a VMI is marked for deletion", func() {
+			rs, vmi := DefaultReplicaSet(10)
+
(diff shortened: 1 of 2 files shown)


rmohr:
@cynepco3hahue rebased. The PR now only adds tests for some VMIRS edge cases.

pkg/virt-controller/watch/replicaset_test.go
```diff
@@ -174,6 +174,42 @@ var _ = Describe("Replicaset", func() {
 			// TODO test for missing 5
 		})
 
+		It("should already create replacement VMIs once it discovers that a VMI is marked for deletion", func() {
+			rs, vmi := DefaultReplicaSet(10)
```
stu-gott: A minor point, could you use named variables for the total count of 10 here and the expected difference of 7 below? It took a moment to figure out where the magic number 7 came from.

PR #1249, opened 2018-07-01
Add possibility to specify VM CPU model

### LINKED ISSUE #1229: Allow setting CPU models
As a user caring about performance I want to set my (v)CPU model to a specific CPU family or host pass-through in order to get best performance.

/kind enhancement

cynepco3hahue:
Add possibility to specify a VM CPU model and vendor, if one of them not specified fallback to `host-model` mode.
- add two fields to CPU API, model, and vendor
- fallback to "host-model" mode, if vendor or model not specified

Fixes #1229

- [x] add unittests
- [x] add functional tests

I do not have any validation for CPU model under this PR, will need to add it in following PR's.

**Release note**:

```release-note
Give the possibility to specify VM CPU model and vendor
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3352,6 +3352,14 @@
      "cores": {
       "description": "Cores specifies the number of cores inside the vmi.\nMust be a value greater or equal 1.",
       "type": "integer"
+     },
+     "model": {
+      "description": "Model specifies the CPU model inside the VMI.\n+optional",
+      "type": "string"
+     },
+     "vendor": {
+      "description": "Vendor specifies the CPU vendor inside the VMI.\n+optional",
+      "type": "string"
      }
     }
    },
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3354,7 +3354,7 @@
       "type": "integer"
      },
      "model": {
-      "description": "Model specifies the CPU model inside the VMI.\n+optional",
+      "description": "Model specifies the CPU model inside the VMI.\nList of available models https://github.com/libvirt/libvirt/blob/master/src/cpu/cpu_map.xml.\n+optional",
       "type": "string"
      }
     }
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3356,10 +3356,6 @@
      "model": {
       "description": "Model specifies the CPU model inside the VMI.\n+optional",
       "type": "string"
-     },
-     "vendor": {
-      "description": "Vendor specifies the CPU vendor inside the VMI.\n+optional",
-      "type": "string"
      }
     }
    },
--- a/pkg/api/v1/schema.go
+++ b/pkg/api/v1/schema.go
@@ -123,6 +123,12 @@ type CPU struct {
 	// Cores specifies the number of cores inside the vmi.
 	// Must be a value greater or equal 1.
 	Cores uint32 `json:"cores,omitempty"`
+	// Model specifies the CPU model inside the VMI.
+	// +optional
+	Model string `json:"model,omitempty"`
+	// Vendor specifies the CPU vendor inside the VMI.
+	// +optional
+	Vendor string `json:"vendor,omitempty"`
 }
 
 // Memory allows specifying the VirtualMachineInstance memory features.
--- a/pkg/api/v1/schema.go
+++ b/pkg/api/v1/schema.go
@@ -126,9 +126,6 @@ type CPU struct {
 	// Model specifies the CPU model inside the VMI.
 	// +optional
 	Model string `json:"model,omitempty"`
-	// Vendor specifies the CPU vendor inside the VMI.
-	// +optional
-	Vendor string `json:"vendor,omitempty"`
 }
 
 // Memory allows specifying the VirtualMachineInstance memory features.
--- a/pkg/api/v1/schema.go
+++ b/pkg/api/v1/schema.go
@@ -124,6 +124,7 @@ type CPU struct {
 	// Must be a value greater or equal 1.
 	Cores uint32 `json:"cores,omitempty"`
 	// Model specifies the CPU model inside the VMI.
+	// List of available models https://github.com/libvirt/libvirt/blob/master/src/cpu/cpu_map.xml.
 	// +optional
 	Model string `json:"model,omitempty"`
 }
--- a/pkg/virt-launcher/virtwrap/api/converter.go
+++ b/pkg/virt-launcher/virtwrap/api/converter.go
@@ -445,14 +445,26 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 	}
 
 	if vmi.Spec.Domain.CPU != nil {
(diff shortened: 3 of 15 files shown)


itamarh:
Can you please describe logic behind vendor being needed (I could see arch being relevant, but its also implied from the cpu model, or from the host-model)?

cynepco3hahue:
@itamarh I think currently we do not need it, so I will remove it. It can be useful in the case when you specified the number of CPU features and want to restrict CPU model to a specific vendor.

api/openapi-spec/swagger.json
```diff
@@ -3352,6 +3352,10 @@
      "cores": {
       "description": "Cores specifies the number of cores inside the vmi.\nMust be a value greater or equal 1.",
       "type": "integer"
+     },
+     "model": {
+      "description": "Model specifies the CPU model inside the VMI.\n+optional",
+      "type": "string"
```
fabiand: Just like the interface models we should validate this against a list of CPU models (+ host-model).

Do we have a way to validate this value against a set of values?

cynepco3hahue: @fabiand We can copy available models from https://github.com/wiedi/libvirt/blob/master/src/cpu/cpu_map.xml, but do we want to do it? It likes to validate CPU features, I believe if someone specifies model he at least must know what he is doing.
But if you prefer strict validation, I can add it.

fabiand: First: I'll not block.
But it's something we need to solve, because: If in this case the user puts a weird value in, then it needs som digging (aka look at a few entities) to find the root cause for the VM to not be launched.

I actually wonder,  @booxter , did you settle on an approach to validate nic models? (the problem is similar)

fabiand:
Setting it to critical, as it is very important for performance numbers.

api/openapi-spec/swagger.json
michalskrivanek: agreed with fabiand it's not really required right now, but once we add it let's please use an API, not a copy of internal .xml

booxter: @fabiand for network models, we have a closed set of possible values that is more narrow than what qemu would actually allow.

In addition to concerns about debugging complexity, another thing to consider is that allowing *any* CPU model / vendor may expose us to more CVEs specific to a particular hardware model / vendor if they are ever found in underlying components (libvirt / qemu). Controlling the list potentially reduces the attack area.

cynepco3hahue: @michalskrivanek I do not copy libvirt XML with supported CPU types, so we do not have any validation for now.
I also do not see the possibility to use API to fetch available CPU models from libvirt, because we do not run libvirt inside a node. We can fetch capabilities from `virt-launcher` pod, but it will happen after validation stage, so it does not really help us.

tests/vmi_configuration_test.go
```diff
@@ -234,6 +235,94 @@ var _ = Describe("Configurations", func() {
...
+									{Key: "kubernetes.io/hostname", Operator: kubev1.NodeSelectorOpIn, Values: []string{nodes.Items[0].Name}},
+								},
+							},
+						},
+					},
+				},
+			}
+		})
+
+		Context("when CPU model definied", func() {
```
booxter: defined, here and below

tests/vmi_configuration_test.go
```diff
@@ -234,6 +235,94 @@ var _ = Describe("Configurations", func() {
...
+				By("Expecting the VirtualMachineInstance console")
+				expecter, err := tests.LoggedInCirrosExpecter(cpuVmi)
+				Expect(err).ToNot(HaveOccurred())
+				defer expecter.Close()
+
+				By("Checking the CPU model under the guest OS")
+				_, err = expecter.ExpectBatch([]expect.Batcher{
+					&expect.BSnd{S: fmt.Sprintf("grep %s /proc/cpuinfo\n", vmiModel)},
+					&expect.BExp{R: "model name"},
+				}, 120*time.Second)
```
booxter: why do we need 120 seconds to check that? (same question below)

api/openapi-spec/swagger.json
fabiand: @cynepco3hahue having "smart" logic is a next step. if we do initial validation, then it should just be a whitelist. Nothing smarter.

Let's do smaller steps.

fabiand: Let's make it short: No semantic validation here. 

Can be up next,let's close this gap for now.

tests/vmi_configuration_test.go
cynepco3hahue: @booxter below I have `Context("when CPU model not definied"`

tests/vmi_configuration_test.go
cynepco3hahue: no reason, will reduce to 10 seconds

api/openapi-spec/swagger.json
michalskrivanek: > Can be up next

yep, that work needs to be started asap (with virtuned, ideally). it's not easy, but we'll be running into similar validation issues again and again.

cynepco3hahue:
retest this please

pkg/api/v1/schema.go
```diff
@@ -123,6 +123,9 @@ type CPU struct {
 	// Cores specifies the number of cores inside the vmi.
 	// Must be a value greater or equal 1.
 	Cores uint32 `json:"cores,omitempty"`
+	// Model specifies the CPU model inside the VMI.
+	// +optional
```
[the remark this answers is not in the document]
cynepco3hahue: Sure will add the link to https://github.com/wiedi/libvirt/blob/master/src/cpu/cpu_map.xml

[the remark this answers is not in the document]
rmohr: Make sure that you use the libvirt/libvirt repo.

vladikr:
looks good to me.

tests/vmi_configuration_test.go
booxter: I meant that "definied" -> "defined".

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -445,14 +445,26 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+				Threads: 1,
+			}
+			domain.Spec.VCPU = &VCPU{
+				Placement: "static",
+				CPUs:      vmi.Spec.Domain.CPU.Cores,
+			}
+		}
+
+		// Set VM CPU model and vendor
+		if vmi.Spec.Domain.CPU.Model != "" && vmi.Spec.Domain.CPU.Vendor != "" {
```
berrange: FYI, setting "Vendor" for a libvirt CPU model is pretty unusual/uncommon. I can't think of a compelling reason to mandate that really. Allowing just model name should be sufficient.

tests/vmi_configuration_test.go
cynepco3hahue: oh will create small PR for it

pkg/virt-launcher/virtwrap/api/converter.go
fabiand: I actually thought that the PR changed to only set the model ..

PR #1250, opened 2018-07-02
Refactor the Template test for multiple VM Templates

gszasz:
Templates test package should be run against multiple templates, thus it had to undergo the complete redesign and refactoring in order to meet the future requirements.

## Test cases covered:

With given Fedora VM Template:

  1. Generate VM JSON file from the Template via `oc-process` command.
  2. Create VM from the VM JSON file via `oc-create` command.
  3. Launch VMI from the VM via `oc-patch` command.
  4. Terminate the VMI via `oc-patch` command.

The test can be now easily extended for any number of different VM Templates.

### Changed

- top-level `Describe` changed to "Templates" to reflect the section in KubeVirt documentation and to simplify regex for Ginkgo focus
- `tests/oc_template_test.go` renamed to `tests/template_test.go`
- dependencies between individual test cases are properly handled by the `BeforeEach`, `JustBeforeEach` and nested `Context` blocks
- reduced number of the closure variables
- Template parameters are initialized distributed using the `TemplateParams` type instead of the set of constants

### Added

- `TemplateMeta` type for parametrized template metadata to reduce redundancy in the handling of multiple templates.
- wrapper functions for the individual `oc` command calls
- `Context` for a given Template

### Deprecated

- `tests.WriteJson()` function in `utils.go`

---

The PR entirely refactors functional tests in `tests/oc_template_test.go` and related functions in `tests/utils.go` to simplify their further extensions that are already defined by testing requirements.  Besides the refactoring efforts, there was also added an explicit test case covering termination of a VMI using external `oc-patch` command. We need the PR to make tests compliant with the future requirements.

The PR is not related to any open issue.

```release-note
NONE
```


cynepco3hahue:
ok to test

gszasz:
@cynepco3hahue Originally, I did not include the renaming into my first commit, but it turned out that it would be hard to see specific changes anyway, as I rewrote the vast majority of the original code from scratch and `git diff` output was very confusing. I concluded that in these specific circumstances, it is fair to introduce renaming into the commit.

PR #1272, opened 2018-07-04
Don't validate origin header on websocket upgrades

### LINKED ISSUE #1242: Bearer token authentication for websockets does not work for kubevirt subresources

/kind bug

**What happened**:

When opening a websocket connection to our subresources from the browser, one has to provide the bearer token in protocol header of websockets:

```
Sec-WebSocket-Protocol: base64url.bearer.authorization.k8s.io.bXl0b2tlbg, base64.binary.k8s.io
```

More details can be found here: https://github.com/kubernetes/kubernetes/pull/47740.

If the token is provided this way, it seems to be ignored by the authentication-proxy in the k8s apiserver and we are forwarded to the aggregated-apiserver as user `system:anonymous`.

Authentication should succeed.

A command like this should lead to a successful authentication:

```
curl --insecure -H "Connection: Upgrade" -H "Upgrade: websocket" -H "Sec-WebSocket-Version: 13" -H "Sec-WebSocket-Key: p1PITMtgfujjqx+O3obU0Q==" \
  -H "Sec-WebSocket-Protocol: base64url.bearer.authorization.k8s.io.BASE64token, base64.binary.k8s.io" \
  https://localhost:8443/apis/subresources.kubevirt.io/v1alpha2/namespaces/kube-system/virtualmachineinstances/vmi-nocloud/console
```

**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

rmohr:

The apiserver does CORS checks for all aggregated servers if it is
configured for CORS checks. The aggregated apiserver does not have to do it and can't do it.

Without this change websockets can't be used from browsers, since they send an origin header and then the request gets rejected by virt-api.

Fixes #1242 

```release-note
Don't validate Origin headers in the aggregated apiserver. It does not have a well known server name to validate against. The k8s apiserver can do that in a central location. This allows browsers to connect to our console and vnc subresources.
```

### DIFF
--- a/tests/vnc_test.go
+++ b/tests/vnc_test.go
@@ -29,6 +29,15 @@ import (
 
 	"time"
 
+	"net/http"
+
+	"github.com/gorilla/websocket"
+	"github.com/onsi/ginkgo/extensions/table"
+	"k8s.io/client-go/rest"
+
+	"fmt"
+
+	"kubevirt.io/kubevirt/pkg/api/v1"
 	"kubevirt.io/kubevirt/pkg/kubecli"
 	"kubevirt.io/kubevirt/pkg/log"
 	"kubevirt.io/kubevirt/tests"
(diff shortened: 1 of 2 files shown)


rmohr:
need to add a test.

rmohr:
@cynepco3hahue could you have another look? It contains now tests too.

tests/vnc_test.go
```diff
@@ -110,5 +117,67 @@ var _ = Describe("VNC", func() {
...
+
+func (t checkUpgradeRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	GinkgoRecover()
+	conn, resp, err := t.Dialer.Dial(r.URL.String(), r.Header)
+	if err != nil {
+		return nil, err
+	}
+	Expect(resp.StatusCode).To(Equal(http.StatusSwitchingProtocols))
+
+	defer conn.Close()
```
cynepco3hahue: I think it must be before `Expect(resp.StatusCode).To(Equal(http.StatusSwitchingProtocols))`

PR #1277, opened 2018-07-05
Take into account namespace memory limits when starting a Virtual Machine

### LINKED ISSUE #970: Limits issue?
Currently, it seems memory for the VM is created via the 'spec.domain.resources.requests' field. The limit for the virt-launcher pod seems to come from the namespace default (kubectl get limits default). If the default limits are lower than the requests, the virt-launcher pod does not create.

Reproduce: "kubectl edit limits default", set memory to 128Mi, then try to create a VM with 1Gi of memory.

I propose that the virt-launcher pod's 'spec.domain.resources.limits.memory' be set to the request field. It seems the VM is already slightly less than the request, so hopefully there's enough buffer in the pod for running bash or whatever else might be needed, but there may need to be additional buffer accounted for in this number.

Alternatively, change the memory field to come from the limit rather than the request, in which case kubernetes will automatically populate the request field to match.

Likewise, it seems like setting CPU via spec.domain.cpu.cores does not change the virt-launcher's CPU request/limit beyond the namespace default. This may be by design, for example one could create an 8 core VM that is throttled down to 100m CPU time, but I'm not sure.

vladikr:
This patch simply copies the relevant default namespace memory limits value to the Virtual Machine spec, if it hasn't been provided by the user.
This value is also being taken into account during the VM memory overhead calculation.
This is particularly useful for the users to see this value on the VM spec, when namespace defaults prevent the VM from starting. 

Fixes #970 

Apparently, k8s allows setting multiple namespace defaults for the same value. Eventually, it applies the minimal value to the container.
```
Resource Limits
 Type       Resource  Min  Max  Default Request  Default Limit  Max Limit/Request Ratio
 ----       --------  ---  ---  ---------------  -------------  -----------------------
 Container  memory    -    -    32Mi             64Mi           -
 Container  memory    -    -    32Mi             32Mi           -
 Container  cpu       -    -    1                1              -
```

```release-note
Make resource memory limits, set for namespaces, visible on VMIs
```

### DIFF
--- a/pkg/controller/virtinformers.go
+++ b/pkg/controller/virtinformers.go
@@ -189,7 +189,7 @@ func (f *kubeInformerFactory) LimitRanges() cache.SharedIndexInformer {
 	return f.getInformer("limitrangeInformer", func() cache.SharedIndexInformer {
 		restClient := f.clientSet.CoreV1().RESTClient()
 		lw := cache.NewListWatchFromClient(restClient, "limitranges", k8sv1.NamespaceAll, fields.Everything())
-		return cache.NewSharedIndexInformer(lw, &k8sv1.LimitRangeList{}, f.defaultResync, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+		return cache.NewSharedIndexInformer(lw, &k8sv1.LimitRange{}, f.defaultResync, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
 	})
 }
 
--- a/pkg/controller/virtinformers.go
+++ b/pkg/controller/virtinformers.go
@@ -65,6 +65,9 @@ type KubeInformerFactory interface {
 
 	// Watches for ConfigMap objects
 	ConfigMap() cache.SharedIndexInformer
+
+	// Watches for LimitRange objects
+	LimitRanges() cache.SharedIndexInformer
 }
 
 type kubeInformerFactory struct {
@@ -179,6 +182,14 @@ func (f *kubeInformerFactory) ConfigMap() cache.SharedIndexInformer {
 		fieldSelector := fields.OneTermEqualSelector("metadata.name", "kubevirt-config")
 		lw := cache.NewListWatchFromClient(restClient, "configmaps", systemNamespace, fieldSelector)
 		return cache.NewSharedIndexInformer(lw, &k8sv1.ConfigMap{}, f.defaultResync, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	})
+}
+
+func (f *kubeInformerFactory) LimitRanges() cache.SharedIndexInformer {
+	return f.getInformer("limitrangeInformer", func() cache.SharedIndexInformer {
+		restClient := f.clientSet.CoreV1().RESTClient()
+		lw := cache.NewListWatchFromClient(restClient, "limitranges", k8sv1.NamespaceAll, fields.Everything())
+		return cache.NewSharedIndexInformer(lw, &k8sv1.LimitRangeList{}, f.defaultResync, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
 	})
 }
 
--- a/pkg/virt-controller/watch/preset.go
+++ b/pkg/virt-controller/watch/preset.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	k8sv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	k8smetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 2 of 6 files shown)


pkg/virt-controller/services/template.go
```diff
@@ -150,6 +193,22 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 	resources := k8sv1.ResourceRequirements{}
 	vmiResources := vmi.Spec.Domain.Resources
 
+	isMemoryFiledExist := func(resource k8sv1.ResourceList) bool {
+		_, ok := resource[k8sv1.ResourceMemory]
+		return ok
+	}
+
+	// Copy namespace memory limits (if exists) to the VM spec
+	if vmiResources.Limits == nil || !isMemoryFiledExist(vmiResources.Limits) {
```
rmohr: Can you move the whole `vmi` manipulation logic to the preset controller? This is the location where all this magic should happen. The preset controller will then in the future be moved to the modifying webhook once introduce it.

pkg/virt-controller/services/template.go
```diff
@@ -70,6 +71,48 @@ func IsEmulationAllowed(store cache.Store) (bool, error) {
 	return useEmulation, nil
 }
 
+var GetNamespaceLimitRangesList = func(namespace string) (*k8sv1.LimitRangeList, error) {
```
rmohr: We should avoid rest-calls inside controller loops. Especially, if they would be done on every vmi creation. Can you use an informer instead and fetch the values from a cache?

pkg/virt-controller/services/template.go
```diff
@@ -150,6 +193,22 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 	resources := k8sv1.ResourceRequirements{}
 	vmiResources := vmi.Spec.Domain.Resources
 
+	isMemoryFiledExist := func(resource k8sv1.ResourceList) bool {
```
stu-gott: Field? was this supposed to be named `isMemoryFieldExist`

vladikr: Yup, thanks :)

vladikr:
The latest uses a limitRanges informer and moves the handling to the presets controller.

pkg/virt-controller/watch/preset_test.go
```diff
@@ -765,6 +765,8 @@ var _ = Describe("VirtualMachineInstance Initializer", func() {
 		var recorder *record.FakeRecorder
 
 		var vmiPresetQueue *testutils.MockWorkQueue
+		var resourcesInformer cache.SharedIndexInformer
+		//var resourcesCache *framework.FakeControllerSource
```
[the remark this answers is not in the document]
vladikr: done

pkg/virt-controller/watch/preset.go
```diff
@@ -156,6 +160,9 @@ func (c *VirtualMachinePresetController) initializeVirtualMachine(vmi *kubev1.Vi
 		} else {
 			logger.Object(vmi).V(4).Info("Setting default values on VirtualMachine")
 			kubev1.SetObjectDefaults_VirtualMachineInstance(vmi)
+
+			//handle namespace limitrange default values
```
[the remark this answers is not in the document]
vladikr: done

vladikr:
> set a VM requested memory bigger than cluster limit range and do not set a VM memory limit
> I think in this case we will fail on the pod creation because a pod limit will be smaller than a pod requested memory.

Thanks! 

Yea, exactly. It will fail. :)

The original issue was filed because the VM creation was failing, for above case, but the user couldn't see the reason (namespace limits applied) on the VM spec.
The user would only get a messages `Invalid value: "64M": must be less than or equal to memory limit`

We've  then agreed to express these limits in the VM spec too, so the user would see the applied limits.
```
Resources:
      Limits:
        Memory:  32Mi
      Requests:
        Memory:  64M
```

That's the reason why I didn't explicitly check that scenario. From my POV this PR is only about copying the limits to the VM spec and take it into account for overhead calculation ( which we previously missed).

I can probably add a functional test. Do you think it's necessary?

rmohr:
> That's the reason why I didn't explicitly check that scenario. From my POV this PR is only about copying the limits to the VM spec and take it into account for overhead calculation ( which we previously missed).

If we want to limit this PR to just that, it is fine by me but then it fixes #970 only half. I would definitely expect that I can't even post such a VM, like I also can't post a Pod which violates that. We can solve it immediately be checking it in the validation webhook or defer it until we have #1278.

vladikr:
I've created a follow-up issue #1348 to completely reject the VM spec if the requested resources are higher than the namespace limits.

pkg/controller/virtinformers.go
```diff
@@ -182,6 +185,14 @@ func (f *kubeInformerFactory) ConfigMap() cache.SharedIndexInformer {
 	})
 }
 
+func (f *kubeInformerFactory) LimitRanges() cache.SharedIndexInformer {
+	return f.getInformer("limitrangeInformer", func() cache.SharedIndexInformer {
+		restClient := f.clientSet.CoreV1().RESTClient()
+		lw := cache.NewListWatchFromClient(restClient, "limitranges", k8sv1.NamespaceAll, fields.Everything())
+		return cache.NewSharedIndexInformer(lw, &k8sv1.LimitRangeList{}, f.defaultResync, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
```
rmohr: You added a namespace indexer to the informer, which is good. You can make us of the extra index further down.

pkg/virt-controller/watch/preset.go
```diff
@@ -172,6 +179,55 @@ func (c *VirtualMachinePresetController) initializeVirtualMachine(vmi *kubev1.Vi
...
+
+}
+
+func getNamespaceLimits(namespace string, limitrangeInformer cache.SharedIndexInformer) (*resource.Quantity, error) {
+	finalLimit := &resource.Quantity{Format: resource.BinarySI}
+
+	objListKeys := limitrangeInformer.GetStore().ListKeys()
+	// there can be multiple LimitRange values set for the same resource in
+	// a namespace, we need to find the minimal
+	for _, key := range objListKeys {
```
rmohr: Here you should be able to do something like:

```
limits, err := limitrangeInformer.GetIndexer().ByIndex(cache.NamespaceIndex, namespace)
for limit := range limits {
...
}
```

if you make use of the namespace indexer.

pkg/virt-controller/watch/preset.go
```diff
@@ -172,6 +179,55 @@ func (c *VirtualMachinePresetController) initializeVirtualMachine(vmi *kubev1.Vi
...
+		if strings.HasPrefix(key, namespace) {
+			item, _, _ := limitrangeInformer.GetStore().GetByKey(key)
+			for _, val := range item.(*k8sv1.LimitRange).Spec.Limits {
+				mem := val.Default.Memory()
+				if val.Type == k8sv1.LimitTypeContainer {
```
rmohr: Do we also have the case into account where the limits apply to a `Pod` as a whole?

pkg/virt-controller/watch/preset.go
vladikr: Nice, thanks! I wasn't aware of this.

pkg/virt-controller/watch/preset.go
vladikr: This is another limitation but works differently. Afaiu, it's not possible to express `defaults` on a `Pod` level, only `min`, and `max`. However, I didn't account for this case at all as it's not a resource that we can copy to the VM spec.

Overall, my thinking in this PR was that the namespace limits are applied on the pod but not visible for the user. We then copy it to the VM spec for visibility. But if it fails for another reason, it will be visible for the user. I'm not sure that we want to replicate all the pod validations to the vm, or am I getting it wrong?

rmohr: Probably then also something for #1278 then. :+1:

vladikr:
retest this please

PR #1281, opened 2018-07-05
Make libvirt calls thread safe

cynepco3hahue:
In our tests we saw that sometime, calls to libvirt
methods return "Missing error" error, it happens when
libvirt call from another go routine override libvirt
last error.

Fixes #

**Release note**:

```release-note
Make calls to libvirt methods thread safe
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/cli/libvirt.go
+++ b/pkg/virt-launcher/virtwrap/cli/libvirt.go
@@ -24,6 +24,7 @@ package cli
 import (
 	"fmt"
 	"io"
+	"runtime"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


rmohr:
I am also not sure what to do with the event loop. It can block for a very long time. If we use the thread lock, then  the lock is off limits for other go routines, although it does nothing.

cynepco3hahue:
@rmohr With `VirDomain` it will be more complicated because we just use some interface and `Domain` from libvirt library implement it, so I will need to create an additional struct that will wrap libvirt domain and will implement `VirDomain` interface.

By event loop do you mean `startDomainEventMonitoring`?

cynepco3hahue:
retest this please

gbenhaim:
Since version `0.6.0` libvirt is thread safe https://wiki.libvirt.org/page/FAQ#Is_libvirt_thread_safe.3F
Are you sure that this change is needed?

cynepco3hahue:
@gbenhaim Sorry for misleading, I talk here about `libvirt-go` client and this PR https://github.com/berrange/libvirt-go/tree/thread-safe-errors must fix the problem, so I think I will just wait for @berrange PR.

gbenhaim:
@cynepco3hahue +1

PR #1287, opened 2018-07-08
Fix interface name validation

SchSeba:
reject non unique interface names by the virt-api webhook validation. 

**Release note**:

```release-note
Fix duplicate interface name.
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/1287)


pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -558,6 +560,17 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 				})
 			}
 
+			// Check if the interface name is unique
+			if _, networkAlreadyUsed := networkInterfaceMap[iface.Name]; networkAlreadyUsed {
+				causes = append(causes, metav1.StatusCause{
+					Type:    metav1.CauseTypeFieldValueDuplicate,
+					Message: fmt.Sprintf("Only one interface can be connected to one spesefic network"),
```
rmohr: s/spesefic/specific/

rmohr:
ok to test

pkg/virt-api/validating-webhook/validating-webhook.go
SchSeba: Done

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -472,7 +472,7 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 	} else if getNumberOfPodInterfaces(spec) > 1 {
 		causes = append(causes, metav1.StatusCause{
 			Type:    metav1.CauseTypeFieldValueDuplicate,
-			Message: fmt.Sprintf("multiple pod interfaces in %s", field.Child("interfaces").String()),
+			Message: fmt.Sprintf("multiple pod networks in %s", field.Child("interfaces").String()),
```
booxter: why the change? getNumberOfPodInterfaces doesn't care if you have multiple pod networks as long as you have just one interface that is linked to a pod network. (meaning, unused network definitions are ok).

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -558,6 +560,17 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 				})
 			}
 
+			// Check if the interface name is unique
+			if _, networkAlreadyUsed := networkInterfaceMap[iface.Name]; networkAlreadyUsed {
+				causes = append(causes, metav1.StatusCause{
+					Type:    metav1.CauseTypeFieldValueDuplicate,
+					Message: fmt.Sprintf("Only one interface can be connected to one specific network"),
```
booxter: It looks like the message doesn't correspond to what we actually validate for here: we validate that no two interfaces have the same name, not that they are connected to the same network.

SchSeba: Yes but if they have the same name its mean they are connected to the same network. This is why I wrote this message

pkg/virt-api/validating-webhook/validating-webhook.go
SchSeba: Yes I just have a problem with "pod interface" ( pod is a network type and not a interface type) I change the message tell me if now its sounds good please.

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -472,7 +472,7 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 	} else if getNumberOfPodInterfaces(spec) > 1 {
 		causes = append(causes, metav1.StatusCause{
 			Type:    metav1.CauseTypeFieldValueDuplicate,
-			Message: fmt.Sprintf("multiple pod networks in %s", field.Child("interfaces").String()),
+			Message: fmt.Sprintf("more then one interface is connected to a pod network in %s", field.Child("interfaces").String()),
```
booxter: then => than

booxter:
I would also suggest to squash patches into one before we merge if possible because "Fixed a typo" commit messages are not very helpful, and the number of tiny commits makes git bisect and git log output longer.

SchSeba:
I squash all the patches into one tell me if its look good now. Thanks

booxter:
I am not sure about a bunch of commits we ended up. Specifically, why is the commit message for 5d8ac59dd25a082c819ff845bca13a10dd4d9282 so messed? (It looks like it contains a number of messages from other commits.) There are a bunch of commits in the series that would be better off being part of existing commits, f.e. "Change test error message" and "Merge prev commits" should be squashed into the original commits they are trying to fix.

I dunno, maybe it's just me and I should give up on trying to get our git tree cleaner. Maybe github is not a platform to have it. It's just very hard to understand at this point what is actually included with the PR since we don't have a clear base to compare to; github UI is completely broken for reviewing updates; and git log -p origin/master.. shows commits of unclear intent (with unclear / intermixed commit messages).

If the only change from the previous iteration is fixing the typo I pointed out, and if we don't care about commit messages / git tree structure then I guess it's good to go.

SchSeba:
The squash make a mess to this PR I will open a new one with everything in order.

rmohr:
> The squash make a mess to this PR I will open a new one with everything in order.

Just force-push on the PR. You don't have to create new ones if you mess something up by accident.

booxter:
Right, git push --force would do the job for you.

PR #1294, opened 2018-07-12
Changed virt-controller to DeamonSet

### LINKED ISSUE #975: Let our HA configuration scale automatically when masters are added or deleted
Picking up https://github.com/kubevirt/kubevirt/issues/727#issuecomment-366983555 from @alezzandro.

We can make our HA setup scale automatically, if we do the following:

 * Use a DaemonSet
 * Give it the right toleration: `tolerations: [key: node-role.kubernetes.io/master]`
 * Add the right label selector by matching [node-role.kubernetes.io/master](https://github.com/kubernetes/kubernetes/blob/f542fe12837560d991218289e8297dac7362c7ba/cmd/kubeadm/app/constants/constants.go#L88)

vatsalparekh:

This PR changes `virt-controller` from `Deployment` to `DeamonSet`, making it to schedule only on the master, and on all masters if there are HA setup with more then one master
**Which issue(s) this PR fixes** *
Fixes #975 

**Release note**:

```release-note
Changed virt-controller from Deployment to DeamonSet
```

### DIFF
--- a/manifests/dev/virt-controller.yaml.in
+++ b/manifests/dev/virt-controller.yaml.in
@@ -20,9 +20,6 @@ metadata:
   labels:
     kubevirt.io: "virt-controller"
 spec:
-  selector:
-    matchLabels:
-      kubevirt.io: virt-controller
   template:
     metadata:
       annotations:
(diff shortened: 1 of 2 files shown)


rmohr:
ok to test

cynepco3hahue:
We have different node labels on OS and k8s:
- `node-role.kubernetes.io/master=` on k8s
- `node-role.kubernetes.io/master=true` on openshift
So I think you will need to use affinity instead of `nodeSelector`,
```
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
      - matchExpressions:
        - key: node-role.kubernetes.io/master
          operator: Exists
```

cynepco3hahue:
retest this please

cynepco3hahue:
@vatsalparekh Looks like you need to adapt `[Fail] LeaderElection Start a VirtualMachineInstance when the controller pod is not running [It] should success 
/root/go/src/kubevirt.io/kubevirt/tests/controller_leader_election_test.go:57` test to the changes I can see it fails under all runs

PR #1299, opened 2018-07-12
Add KubeVirt custom fail handler

cynepco3hahue:
Custom Fail handler will print cluster status and 15 last line from kubevirt pods logs.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -1567,3 +1567,38 @@ func GetNodeLibvirtCapabilities(nodeName string) string {
 
 	return output
 }
+
+// KubevirtFailHandler call ginkgo.Fail with printing the additional information
+func KubevirtFailHandler(message string, callerSkip ...int) {
+	virtClient, err := kubecli.GetKubevirtClient()
+	PanicOnError(err)
+
+	for _, ns := range []string{metav1.NamespaceSystem, NamespaceTestDefault} {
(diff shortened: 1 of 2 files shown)


cynepco3hahue:
@rmohr if you have ideas for another useful information that I can print on error, give me know

tests/utils.go
```diff
@@ -1567,3 +1567,38 @@ func GetNodeLibvirtCapabilities(nodeName string) string {
...
+			if strings.HasPrefix(pod.Name, "virt-launcher") {
+				containerName = "compute"
+			}
+			logsRaw, err := virtClient.CoreV1().Pods(metav1.NamespaceSystem).GetLogs(
+				pod.Name, &k8sv1.PodLogOptions{
+					TailLines: &tailLines,
+					Container: containerName,
+				},
+			).DoRaw()
+			if err == nil {
```
stu-gott: If err != nil, that means the logs couldn't be fetched, right? Is that a normal situation? If not should we mention that we couldn't get them?

cynepco3hahue: @stu-gott I think `Fail` handler runs in separate go routine, so when you coming to fetch logs from `virt-launcher` pod, it possible that it already does not exists, but we just want to continue with the standard `Fail` handler flow so I do not fail on fetch logs problem.

PR #1302, opened 2018-07-12
Fix stdci configuration for the project

danielBelenky:
Signed-off-by: Daniel Belenky <dbelenky@redhat.com>

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #1306, opened 2018-07-12
Setup stdci1

rmohr:
Run tests on kubevirt tests on stdci-vms.

### DIFF
--- a/automation/check-patch.k8s-1.10.3-dev.mounts
+++ b/automation/check-patch.k8s-1.10.3-dev.mounts
@@ -1,0 +1,1 @@
+check-patch.mounts
--- a/automation/check-patch.k8s-1.10.3-dev.packages
+++ b/automation/check-patch.k8s-1.10.3-dev.packages
@@ -1,0 +1,1 @@
+check-patch.packages.el7
--- a/automation/check-patch.k8s-1.10.3-dev.sh
+++ b/automation/check-patch.k8s-1.10.3-dev.sh
@@ -1,0 +1,1 @@
+check-patch.sh
--- a/automation/check-patch.k8s-1.10.3-release.mounts
+++ b/automation/check-patch.k8s-1.10.3-release.mounts
@@ -1,0 +1,1 @@
+check-patch.mounts
--- a/automation/check-patch.k8s-1.10.3-release.packages
+++ b/automation/check-patch.k8s-1.10.3-release.packages
@@ -1,0 +1,1 @@
+check-patch.packages.el7
--- a/automation/check-patch.k8s-1.10.3-release.sh
+++ b/automation/check-patch.k8s-1.10.3-release.sh
@@ -1,0 +1,1 @@
+check-patch.sh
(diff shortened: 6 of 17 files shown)


rmohr:
retest this please

rmohr:
ci test please

rmohr:
@ifireball I see

```
Error response from daemon: driver failed programming external connectivity on endpoint k8s-1.10.3-dnsmasq (8a9f317e4cdcb5a2c8d6b68a80d1eb5c0bbf4f53de892a24d4ef64be6ed63313):  (iptables failed: iptables --wait -t nat -A DOCKER -p tcp -d 127.0.0.1 --dport 8443 -j DNAT --to-destination 172.17.0.2:8443 ! -i docker0: iptables: No chain/target/match by that name.
```

on some vms. In the last run on all but nut only on some in previous runs. May be related to docker/firewall issues.

ifireball:
@rmohr isn't that iptables command running inside the k8s VMs your system brings up? I'll need more details in order to figure out if/how the host configuration could cause this...

rmohr:
@ifireball no. We just start our dnsmasq container on your vm. Docker tries to do that iptables command, not us.

ifireball:
@rmohr I see we`re trying to run on VMs here rather then BMs... lets stabilize things on BMs first and the research using VMs ok?

rmohr:
We just start the dnsmasq container with the option to expose all ports in general `-P` and optionally add some more ports with `-p`. And it is privielged (`--privileged`). That's it.

ifireball:
@rmohr ok I managed to reproduce this with:

     sudo docker run -it --rm --privileged -P -p 4444:3333 centos

The puzzling thing is that this works fine on the BMs... I really wonder where is this difference coming from, given that they're both setup the same way with the same scripts...

ifireball:
@rmohr ah it may be an fc28 thing... since all the threads that failed with this issue ran on FC28 VMs.  (Note that 'check-patch.k8s-1.10.3-release.el7.x86_64' which ran on EL7 failed on something else).

Lets try forcing it to use EL7 VMs by adding to following to the YAML:

     runtime-requirements:
        host-distro: same

But again maybe we should just drop the commit that allows it to use VMs and focus on BMs first....

rmohr:
Would `some` have to be `el7` or `EL7` or does it not matter?

rmohr:
Or is it literally `same`?

ifireball:
literally {{same}} to tell it that the host should have the same OS as the CI userspace.

ifireball:
I the meantime I managed to reproduce and find a solution, can see details here:

https://ovirt-jira.atlassian.net/browse/OVIRT-2340

I'll patch our system to prevent this from happening in the future.

rmohr:
@ifireball does the check-patch.yaml look correct? It still shows the same errors.

rmohr:
ci test please

ifireball:
@rmohr since you have an stdci.yaml file in the project root, it should go there

rmohr:
Nesting is too deep. My tests show that running our vms on vms degrades performance too much (we talk about minutes to start e.g. cirros).

ifireball:
please repone this and set it to run on BMs - this PR was originally meant to generally enable STDCI, not just on VMs

rmohr:
@ifireball you mean #1302.

PR #1309, opened 2018-07-12
Support bridge.delegateNetworkToGuest

booxter:
**What this PR does / why we need it**: add support for delegateIp field for bridge interface type.

Fixes #

**Release note**:

```release-note
Added support for delegateIp field for bridge interface type. Defaults to true. If false, KubeVirt will not delegate the pod IP and MAC addresses to the guest. The guest should then set up IP networking by other means (static or external DHCP server).
```


booxter:
I am currently struggling with retaining actual connectivity inside the pod when delegateIp is false. The IP address and routing tables are restored as expected; but when I ping gateway, and ARP request is sent to it, while I see in tcpdump that ARP reply is received with correct MAC address (of eth0), ping still fails; and 'arp -e' doesn't show a MAC address for the gateway IP address. It's not clear to me how it's possible to receive a reply but not update the table.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -182,16 +182,23 @@ func (b *BridgePodInterface) discoverPodNetworkInterface() error {
...
-		return err
+		if err != nil {
+			log.Log.Reason(err).Errorf("failed to delete link for interface: %s", podInterface)
+			return err
+		}
 	}
 
 	// Set interface link to down to change its MAC address
-	err = Handler.LinkSetDown(b.podNicLink)
+	err := Handler.LinkSetDown(b.podNicLink)
```
vladikr: I'm just wondering why do you think we should randomize the mac address of the pod even when delegateIp is false?

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -208,15 +215,38 @@ func (b *BridgePodInterface) preparePodNetworkInterfaces() error {
 		return err
 	}
 
-	if err := b.createDefaultBridge(); err != nil {
+	if !delegateIp {
+		err = b.setupRouting()
```
vladikr: When do the routes disappear? Is it when we add the bridge or when we bring down eth0 to randomize the mac?

booxter: When eth0 going down.

pkg/virt-launcher/virtwrap/network/podinterface.go
booxter: That's partially because we haven't solved custom MAC address for OpenShift SDN yet. If we don't move the allocated MAC address to VIF (and hence replace it on br1 with a random one) then no traffic coming from the guest will pass rigid ARP flows the SDN sets specific to the allocated MAC address.

Maybe we should leave it as-is and wait until the issue is solved for the SDN in general.

booxter:
The latest upload should be ready for review / merge. @vladikr @SchSeba @rmohr 

One thing I am not sure about is the name of the field (delegateIp) since it guards delegation for both MAC and IP addresses.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -181,39 +181,79 @@ func (b *BridgePodInterface) discoverPodNetworkInterface() error {
...
 		return err
 	}
 
-	if err := b.createDefaultBridge(); err != nil {
-		return err
+	if delegateIp {
+		// Start DHCP server to lease the allocated IP address
+		b.startDHCPServer()
+	} else {
+		// Restore default gateway inside the pod
```
vladikr: Is it because we've moved the ip from the nic to the bridge?
I'm wondering what about the other routes or these shouldn't be affected?

booxter: Yeah good comment, it's possible we need to restore other routes. I will need to check. Maybe some routes point to other nodes and I will miss them now.

booxter:
@vladikr delegateID is a bit vague. Maybe "delegateNetworkProperties" or "delegateNetworkSettings"?

booxter:
@fabiand please chime in on the question of whether we should stick to delegateIp or rename the field into something more generic and correct, like delegateNetworkProperties. Thanks.

booxter:
In the latest upload, I 1) restore all in-pod eth0 routes, using br1; and 2) rename the field into delegateNetworkToGuest.

Yeah the name is ugly so welcome with better suggestions.

booxter:
Thanks for the suggestion, indeed it would be nice to validate not only that the pod can to talk to external resources, but also that pods can inter-communicate. Ideally, we would have two scenarios: with pods on the same node, and with pods on different nodes (assuming there are multiple compute nodes). There are examples of test cases that schedule pods to different nodes: https://github.com/kubevirt/kubevirt/blob/master/tests/vmi_networking_test.go#L277 that we could use as inspiration.

We are going to hold off the PR because now I hear that we may want to ban some features, including delegateNetworkToGuest=false mode, for pod interfaces, instead of accommodating for them. (btw while I agree the default pod interface behavior should accommodate for all networking backends, I don't agree that we should necessarily ban additional features like custom MAC addresses or delegateNetworkToGuest=false for pod interfaces; especially if there is technical way to make it work without breaking external SDN assumptions about MAC address etc.)

I will try to get a definite answer on the path we are taking for these advanced features in relation to pod interfaces and will report back here. In the meantime, there is no good reason to spend time writing additional functional test cases for something that may well be deleted.

booxter:
The decision du jour is that we are not going to support pure L2 connectivity for pod interfaces. As for additional L2 networks, looking at https://github.com/kubevirt/kubevirt/pull/1428/files#diff-d3ed8eb8cca020f37cab207c079f2627, the idea is to detect IP address on the interface and delegate it if present, otherwise assume pure L2 connectivity. In which case, we don't need an additional flag to enable / disable it.

Due to that, I now don't believe that this whole PR, and the flag it attempts to implement, is helpful.

PR #1311, opened 2018-07-13
Allow build-time configuration of Image Pull Policy

benwatcisco:
This adds the ability to set container "Image Pull Policy" at build time.  This can be helpful in a developer workflow by setting the policy to "Always".  In such a case worker nodes will always pull the latest kubevirt images, removing the need for the developer to either use ephemeral Docker tags or to explicitly pull on each worker node.

None

```release-note
User can set the ImagePullPolicy that will be applied to manifests by setting the IMAGE_PULL_POLICY environment variable.
This defaults to "IfNotPresent" (previous behavior)
```

### DIFF
--- a/hack/dockerized
+++ b/hack/dockerized
@@ -14,7 +14,7 @@ SYNC_VENDOR=${SYNC_VENDOR:-false}
 TEMPFILE=".rsynctemp"
 
 # Build the build container
-(cd ${DOCKER_DIR} && docker build . -q -t ${BUILDER})
+(cd ${DOCKER_DIR} && docker build . -t ${BUILDER})
 
 # Create the persistent docker volume
 if [ -z "$(docker volume list | grep ${BUILDER})" ]; then
--- a/hack/build-manifests.sh
+++ b/hack/build-manifests.sh
@@ -44,6 +44,7 @@ for arg in $args; do
         --namespace=${namespace} \
         --docker-prefix=${manifest_docker_prefix} \
         --docker-tag=${docker_tag} \
+        --image-pull-policy=${image_pull_policy} \
         --generated-manifests-dir=${KUBEVIRT_DIR}/manifests/generated/ \
         --input-file=${KUBEVIRT_DIR}/manifests/$arg >${outfile}
 
@@ -51,6 +52,7 @@ for arg in $args; do
         --namespace="{{ namespace }}" \
         --docker-prefix="{{ docker_prefix }}" \
         --docker-tag="{{ docker_tag }}" \
+        --image-pull-policy="{{ image_pull_policy }}" \
         --generated-manifests-dir=${KUBEVIRT_DIR}/manifests/generated/ \
         --input-file=${KUBEVIRT_DIR}/manifests/$arg >${template_outfile}
 done
@@ -63,6 +65,7 @@ find ${MANIFEST_TEMPLATES_OUT_DIR}/ -type f -exec sed -i {} -e '${/^$/d;}' \;
 export namespace=${namespace}
 export docker_tag=${docker_tag}
 export docker_prefix=${manifest_docker_prefix}
+export image_pull_policy=${image_pull_policy}
 
 TMP_DIR=$(mktemp -d)
 cleanup() {
--- a/hack/config-default.sh
+++ b/hack/config-default.sh
@@ -5,3 +5,4 @@ docker_tag=${DOCKER_TAG:-latest}
 master_ip=192.168.200.2
 network_provider=flannel
 namespace=kube-system
+image_pull_policy=${IMAGE_PULL_POLICY:-IfNotPresent}
--- a/hack/config.sh
+++ b/hack/config.sh
@@ -1,5 +1,5 @@
 unset binaries docker_images docker_prefix docker_tag manifest_templates \
-    master_ip network_provider kubeconfig manifest_docker_prefix namespace
+    master_ip network_provider kubeconfig manifest_docker_prefix namespace image_pull_policy
 
 KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-${PROVIDER}}
 KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-k8s-1.10.3}
@@ -14,4 +14,4 @@ test -f "hack/config-provider-${KUBEVIRT_PROVIDER}.sh" && source hack/config-pro
 test -f "hack/config-local.sh" && source hack/config-local.sh
 
 export binaries docker_images docker_prefix docker_tag manifest_templates \
-    master_ip network_provider kubeconfig namespace
+    master_ip network_provider kubeconfig namespace image_pull_policy
--- a/manifests/dev/virt-api.yaml.in
+++ b/manifests/dev/virt-api.yaml.in
@@ -40,7 +40,7 @@ spec:
       containers:
         - name: virt-api
           image: {{.DockerPrefix}}/virt-api:{{.DockerTag}}
-          imagePullPolicy: IfNotPresent
+          imagePullPolicy: {{.ImagePullPolicy}}
           command:
               - "virt-api"
               - "--port"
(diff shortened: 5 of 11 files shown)


rmohr:
@benwatcisco I see, here is the part I was missing in #1314. I am not sure if that is the right thing to do. I was thinking in the past about this too. I then decided to have the push/pull logic inside the providers and pre-pull all images to all nodes, in order to make the test executions and the developer experience more consistent. It is very hard to write tests with reasonable timeout expectations if you don't know if images are going to be pulled or not. Also as an engineer I found it hard to differentiate between pull-delays and real issues during development.

Pre-pulling might however not be an option for big clusters. So I am not against it either. It looks like it is a necessity for the external provider to work.

hack/dockerized
```diff
@@ -24,12 +31,19 @@ fi
 # Make sure that the output directory exists
 docker run -v "${BUILDER}:/root:rw,z" --rm ${BUILDER} mkdir -p /root/go/src/kubevirt.io/kubevirt/_out
 
+# Add all of the environment variables that may be needed within the build container
+append_envvar_file KUBEVIRT_PROVIDER ${KUBEVIRT_PROVIDER}
```
rmohr: In order to inject these values for the external provider, it should be ablot to create your own config file, like we do e.g. here: https://github.com/kubevirt/kubevirt/blob/master/cluster/local/provider.sh#L35. The name of the file should be `hack/config-provider-external.sh`.

You should not have to modify `hack/dockerized`.

benwatcisco: This doesn't work for me.  When a script is run "dockerized", for example "build_manifest.sh", unless you pass the provider in as an environment variable, the script doesn't know its provider and thus the per-provider script will not get called.
```
Before calling, in dockerized KUBEVIRT_PROVIDER=external
go version go1.10 linux/amd64
+ source hack/common.sh
++++ dirname 'hack/common.sh[0]'
+++ cd hack/../
+++ pwd
++ KUBEVIRT_DIR=/root/go/src/kubevirt.io/kubevirt
++ OUT_DIR=/root/go/src/kubevirt.io/kubevirt/_out
++ VENDOR_DIR=/root/go/src/kubevirt.io/kubevirt/vendor
++ CMD_OUT_DIR=/root/go/src/kubevirt.io/kubevirt/_out/cmd
++ TESTS_OUT_DIR=/root/go/src/kubevirt.io/kubevirt/_out/tests
++ APIDOCS_OUT_DIR=/root/go/src/kubevirt.io/kubevirt/_out/apidocs
++ MANIFESTS_OUT_DIR=/root/go/src/kubevirt.io/kubevirt/_out/manifests
++ MANIFEST_TEMPLATES_OUT_DIR=/root/go/src/kubevirt.io/kubevirt/_out/templates/manifests
++ PYTHON_CLIENT_OUT_DIR=/root/go/src/kubevirt.io/kubevirt/_out/client-python
++ KUBEVIRT_PROVIDER=k8s-1.10.3
++ KUBEVIRT_NUM_NODES=1
```

Part of the problem is that the same "dockerized" environment is being used for provider-independent things like building binaries, and for provider-dependent things like building the manifests.  IMHO nothing run in this particular container should have to know about the test environment.  Maybe only build templatized (i.e. Jinja2) manifests inside the container, then apply the data outside, i.e. in the provider-aware environment.  I'll try to get something like this working.

rmohr: > This doesn't work for me. When a script is run "dockerized", for example "build_manifest.sh", unless you pass the provider in as an environment variable, the script doesn't know its provider and thus the per-provider script will not get called.

Ah yes you are right. What we do is something like this for other providers in the `build` function of the provider:

```
${KUBEVIRT_PATH}hack/dockerized KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER} ./hack/build-manifests.sh"
```

> IMHO nothing run in this particular container should have to know about the test environment. Maybe only build templatized (i.e. Jinja2) manifests inside the container, then apply the data outside, i.e. in the provider-aware environment. I'll try to get something like this working.

I see the difficulty with passing in the data. However it is very valuable that everything runs in containers. Maybe the suggestion above is good enough. Although I see that it is not perfect. Maybe we can wait with refactoring until we see the need for more providers? I hope that we don't need much more. The `external` provider seems to be the generic missing one.

hack/docker-builder/Dockerfile
```diff
@@ -8,11 +8,19 @@ RUN curl --output /etc/yum.repos.d/fedora-virt-preview.repo \
...
 
 ENV GOPATH="/go" GOBIN="/usr/bin"
+ENV KUBEVIRT_VERSION=${KUBEVIRT_VERSION}
+ENV KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER}
+ENV DOCKER_PREFIX=${DOCKER_PREFIX}
+ENV DOCKER_TAG=${DOCKER_TAG}
+ENV IMAGE_PULL_POLICY=${IMAGE_PULL_POLICY}
+ENV TRAVIS_JOB_ID=${TRAVIS_JOB_ID}
+ENV TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST}
+ENV TRAVIS_BRANCH=${TRAVIS_BRANCH}
```
rmohr: is there a specific reason why you added those? For documentation?

rmohr: Oh, now I get it. Might be a good idea to do it this way instead of using the Makefile for passing it.

rmohr:
ok to test

hack/docker-builder/Dockerfile
rmohr: Is this something which requires a specific docker version? The environment variables do not seem to be passed through. For instance the `coveralls` upload on travis fails because it can't see TRAVIS_JOB_ID, ...

benwatcisco: @rmohr, sorry, this doesn't work.  Premature push on my part.  There doesn't seem to be a way to define container environment variables at build time in Docker, only some kludgy argument passing or at run time via the '--env' parameter.  Consider this broken while I find a better way.  Thanks for the attention :)

rmohr: Don't worry. I think that if you would remove the part which tries to pass environment variables through and make sure that you pass the new environment variable as usual via the Makefile, we could bring this PR in and defer the environment variable discussion to #1352.

benwatcisco:
@rmohr : Environment stuff removed.

rmohr:
retest this please

hack/dockerized
```diff
@@ -14,7 +14,7 @@ SYNC_VENDOR=${SYNC_VENDOR:-false}
 TEMPFILE=".rsynctemp"
 
 # Build the build container
-(cd ${DOCKER_DIR} && docker build . -q -t ${BUILDER})
+(cd ${DOCKER_DIR} && docker build . -t ${BUILDER})
```
rmohr: Hm, without the `-q` starting `hack/dockerized` is very verbose.

benwatcisco: Well, with the -q it is completely silent, sometimes for several minutes.  My preference is to always have some feedback, but if you insist, I can put it back.

rmohr: Ok, so lets do the following: If "JOB_NAME" (jenkins) or "TRAVIS_BUILD_ID" (travis) is defined, then add the `-q`. Otherwise not.

Only parts of the testing infra cache the builder and it is normally only of interest when something goes wrong during creating the builder, where you then get reasonable output. For devs it is either cached (so almost no extra output), or the situation you describe is encountered, where you did not see output for minutes. There it is probably always preferred to see the progress.

What do you think?

An alternative would also be to simply print a message like `Creating the build container. The initial uncached build may take several minutes.`

rmohr:
ci test please

benwatcisco:
Hi Roman,

Sorry for the delay, I just got back from vacation.

Thanks for merging this.  I’ve created another pull request with handling for the quiet flag:
https://github.com/kubevirt/kubevirt/pull/1412

This seems like a good way of doing things.

--Ben

PR #1314, opened 2018-07-13
Add a new 'external' provider

benwatcisco:
Trivial addition - this new provider is just an external, configured Kubernetes cluster.  Along with an external private Docker registry, this allows for a simple build/push/publish/test developer workflow.

Signed-off-by: Ben Warren <bawarren@cisco.com>

Adds a "provider" for developers who have access to an existing Kubernetes cluster.

None

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster/external/README.md
+++ b/cluster/external/README.md
@@ -1,0 +1,29 @@
+# External Kubernetes Provider
+
+This provider works with an existing, provisioned Kubernetes cluster.
+An external Docker registry is recommended for serving images.
+Unlike with other providers, lifecycles of the cluster and registry are not managed.
+The build machine should be a client of the cluster.
+
+## Verifying connectivity
+
+```bash
+export KUBEVIRT_PROVIDER=external
+export DOCKER_PREFIX=myregistry:5000/kubevirt
+export KUBECONFIG=mycluster.conf
+export IMAGE_PULL_POLICY=Always
+make cluster-up
+```
+
+## Building and pushing to the registry
+
+```bash
+make cluster-build
+```
+
+## Installing Kubevirt artifacts on the cluster
+
+```bash
+make cluster-sync
+```
+
--- a/cluster/external/provider.sh
+++ b/cluster/external/provider.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+function _kubectl() {
+    kubectl "$@"
+}
+
+function prepare_config() {
+    cat >hack/config-provider-external.sh <<EOF
+docker_tag=devel
+docker_prefix=${DOCKER_PREFIX}
+manifest_docker_prefix=${DOCKER_PREFIX}/kubevirt
+image_pull_policy=${IMAGE_PULL_POLICY:-Always}
+EOF
+}
+
+# The external cluster is assumed to be up.  Do a simple check
+function up() {
+    prepare_config
+    _kubectl version >/dev/null
+    if [ $? -ne 0 ]; then
+        echo -e "\n*** Unable to reach external cluster.  Please check configuration ***"
+        echo -e "*** Type \"kubectl config view\" for current settings               ***\n"
+        exit 1
+    fi
+    echo "Cluster is up"
+}
+
+function down() {
+    echo "Not supported by this provider"
+}
+
+function build() {
+    # Build code and manifests
+    ${KUBEVIRT_PATH}hack/dockerized "DOCKER_TAG=${DOCKER_TAG} KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER} ./hack/build-manifests.sh"
+    make push
+}
--- a/docs/env-providers.md
+++ b/docs/env-providers.md
@@ -47,6 +47,20 @@ export KUBEVIRT_PROVIDER=local # choose this provider
 make cluster-up
 ```
 
+## External
+
+Uses an existing (external) Kubernetes cluster.
+
+Requires:
+ * A working Kubernetes cluster with properly configured worker nodes.
+ * A running docker daemon
+
+Usage:
+
+```bash
+export KUBEVIRT_PROVIDER=external # choose this provider
+make cluster-up
+```
 ## New Providers
 
  * Create a `cluster/$KUBEVIRT_PROVIDER` directory


ovirt-infra:

cluster/external/provider.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+function _kubectl() {
+    kubectl "$@"
```
cynepco3hahue: It a lot of assumptions here:
- your local machine must have `kubectl` binary
- `kubectl` binary must use the correct `KUBECONFIG`

So you need to specify it under README or fetch it from an external cluster(I prefer second option)

benwatcisco: Sure, this does require some configuration on the part of the user, but it also requires the user to explicitly state the provider, so there's little danger of somebody doing this by accident.  How about I have it fail out with a useful message if there isn't enough info configured?

rmohr: I think that it is just consistent that everything needs is prepared outside of kubevirt, since it is an `external` provider.

benwatcisco: I added some error messaging.  Hopefully this is good enough

1. Normal behavior:
```
bash-3.2$ KUBEVIRT_PROVIDER=external make cluster-up
./cluster/up.sh
Cluster is up
```
2. Misconfigured cluster (simulated by modifying my ~/.kube/config file)
```
bash-3.2$ KUBEVIRT_PROVIDER=external make cluster-up
./cluster/up.sh
The connection to the server 10.193.228.153:6444 was refused - did you specify the right host or port?

*** Unable to reach external cluster.  Please check configuration ***
*** Type "kubectl config view" for current settings               ***

```
3. Missing **kubectl** binary
```
bash-3.2$ PATH=/usr/bin KUBEVIRT_PROVIDER=external make cluster-up
./cluster/up.sh
cluster/external/provider.sh: line 4: kubectl: command not found

*** Unable to reach external cluster.  Please check configuration ***
*** Type "kubectl config view" for current settings               ***

```

cluster/external/provider.sh
```diff
@@ -0,0 +1,24 @@
...
+    fi
+    echo "Cluster is up"
+}
+
+function down() {
+    echo "Not supported by this provider"
+}
+
+function build() {
+    echo "Not supported by this provider"
```
rmohr: To understand the context on how you want to use the provider: Is your intent to do the whole `make cluster-sync` stuff which builds the code, removes old versions from the cluster, pushes new images to the registry and deploys the new manifests, externally?

In the other providers we set a proper `docker_prefix` in the config which points to the used docker registry and let `make cluster-sync` do all required updates.

benwatcisco: @rmohr: here's how I currently do things:
make build --> compile stuff
make push --> push images to my Docker registry (based on DOCKER_PREFIX and DOCKER_TAG)
make cluster-sync --> resolve kubevirt installation (remove existing manifests, add new ones)

We're currently using make as a really crappy shell script language, and having make call scripts that in turn call make is less than ideal.  If we can improve the performance of the dockerized operations, we can use make properly with dependencies, and then maybe 'cluster sync' or some other wrapper target can do everything in one shot.

Until then, this works, and the workflow is fairly straightforward.

rmohr: > We're currently using make as a really crappy shell script language, and having make call scripts that in turn call make is less than ideal.

@benwatcisco oh, we use the Makefile extra only as well-known entrypoints, not in the sense of make. Further the functions for providers are extra not called in the dockerized environment so that you can for instance also use the Makefile targets there. For as long as you don't call `cluster-sync` from in there again, you should be able to do there what you want.

rmohr: So taking the flow which you described here:

> make build --> compile stuff
> make push --> push images to my Docker registry (based on DOCKER_PREFIX and DOCKER_TAG)
> make cluster-sync --> resolve kubevirt installation (remove existing manifests, add new ones)

You could make your `build` function look like this:

```bash
function build() {
 # does "build" in docker, create from the result outside of docker, pushes to the registry specified with DOCKER_PREFIX and DOCKER_TAG
make push
# creates manifests under consideration of DOCKER_PREFIX and DOCKER_TAG
make manifests
}
```

You could still save one extra docker call by somehow merging the `build` and `manifests` phase. In my case that would only save about two seconds.

But maybe I silll miss something.

rmohr:
@benwatcisco a general question: Are subsequent container starts very slow for you? Only the first call which obviously has to build the docker-builder and also has to sync the whole git-repo into a docker-volume is supposed to have a big overhead. After that the docker overhead should be fairly small.

cluster/external/provider.sh
j-griffith: @benwatcisco The workflow you describe is nice and simple, I like it.  I do find that the point @rmohr makes here about just wrapping the steps up in a ```make build``` extremely nice, and more importantly consistent with the other providers which is a bug plus.  Are you still interested in this effort?  I'm happy to pick it up if you don't have the time, or to help test/review it if you're still into it.  Thanks!!

benwatcisco: @j-griffith I would like to see this code pulled in to master, and am willing to push it through.  I personally don't like the idea of calling make this way, and will try to come up with something better, but now now :)

benwatcisco:
@rmohr : Hey Roman.  I just realized I didn't answer your last question.  I don't notice container startup to be particularly slow once the container is created.
I just added a call to make in the provider script.  Is anything else needed to get this integrated?

docs/env-providers.md
```diff
@@ -47,6 +47,20 @@ export KUBEVIRT_PROVIDER=local # choose this provider
...
+Uses an existing (external) Kubernetes cluster.
+
+Requires:
+ * A working Kubernetes cluster with properly configured worker nodes.
+ * A running docker daemon
+
+Usage:
+
+```bash
+export KUBEVIRT_PROVIDER=external # choose this provider
```
rmohr: If you agree with the two comments above, then configuring the provider would look like this:

```
export KUBEVIRT_PROVIDER=external
export KUBEVIRT_EXTERNAL_REGISTRY=external.registry.com
make cluster-up
make cluster-sync
```

cluster/external/provider.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# The external cluster is assumed to be up.  Do a simple check
+function up() {
+    _kubectl version >/dev/null
+    if [ $? -ne 0 ]; then
+        echo -e "\n*** Unable to reach external cluster.  Please check configuration ***"
+        echo -e "*** Type \"kubectl config view\" for current settings               ***\n"
+        exit 1
+    fi
+    echo "Cluster is up"
```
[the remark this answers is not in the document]
benwatcisco: Hard coding the image pull policy here isn't right IMHO.  If you apply my other pull request (https://github.com/kubevirt/kubevirt/pull/1352)
none of this stuff is necessary.

cluster/external/provider.sh
```diff
@@ -0,0 +1,25 @@
...
+}
+
+function down() {
+    echo "Not supported by this provider"
+}
+
+function build() {
+    # Build code and manifests
+    make all
```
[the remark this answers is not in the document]
benwatcisco: I don't understand the difference, since the 'all' rule essentially does the same thing (build code and manifests), and abstracting it into a make rule seems cleaner, although I'm not a big fan of how make->script->make flow but that's neither here nor there.
I'm not fully convinced the push belongs here, but since anybody using this workflow will probably have an external registry, pushing will likely always follow a build.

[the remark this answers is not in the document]
rmohr: Actually it needs to look like this:

```
${KUBEVIRT_PATH}hack/dockerized "DOCKER_TAG=${DOCKER_TAG} KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER} ./hack/build-manifests.sh"
make push
```

Only the manifest build step needs this extra step to pass in the provider.

The background is soley that. All other providers do the same thing right now.
The push needs to be part of it too, since this method is only called when you invoke `make cluster-sync`. It's purpose is to bring the cluster and the changed code at  the same state.

cluster/external/provider.sh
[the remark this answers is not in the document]
rmohr: The wole purpose of the providers is that you can run `make cluster-up` once and that then subsequent repreted calls to `make cluster-sync` synchronize the cluster with your code changes.

If you don't hardcode `Always` here, then you would only get changes synchronized to the nodes on the first invocation of `make cluster-sync`. On the second `make cluster-sync` invocation, you would only push the images to the registry but the newly created manifests would never pull the new manifests to the nodes.

If you only want to build the code and once push it to an external cluster then I would not use providers. It is really only for the developer flow.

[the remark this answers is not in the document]
benwatcisco: My point is that image pull policy is now mutable by an environment variable (IMAGE_PULL_POLICY) so it doesn't make sense to hard code it.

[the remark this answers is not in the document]
rmohr: I understand that. I tried to explain why it is important to explicitly hardcode it in this case. There is no other reasonable value than `Always` in the context of this provider. Otherwise, if you ever forget to set it to `Always`, you will have awkward results when running `make cluster-sync`.

cluster/external/provider.sh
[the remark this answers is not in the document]
benwatcisco: OK, if the PR https://github.com/kubevirt/kubevirt/pull/1352 is dead in the water, I see why this is necessary.  If we can get it in soon then simple make calls here will suffice.

benwatcisco:
@rmohr : changes made as you requested.
Please let me know if you'd like me to rebase (i.e. squash)

benwatcisco:
@rmohr : squashed and pushed.  I added a README file after noticing that all other providers have one.  Hopefully it's not controversial.

cluster/external/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Verifying connectivity
+
+```bash
+export KUBEVIRT_PROVIDER=external
+export DOCKER_PREFIX=myregistry:5000/kubevirt
+export KUBECONFIG=mycluster.conf
+export IMAGE_PULL_POLICY=Always
+make cluster-up
+```
```
j-griffith: @benwatcisco thanks for adding this, I was going to ask for it in the last set or at least as a follow up.  Works well, thanks!

rmohr:
@benwatcisco thx for squashing. The added README looks great.

PR #1323, opened 2018-07-16
virtctl console wait until the virtual machine is in running state

### LINKED ISSUE #1001: Make virtctl console wait for console to become ready
When I start a VM, I would like to immediately connect to its serial console to see all the output it may produce during boot. But I can't do it right away, because I get the error:

```
[ihar@ws kubevirt]$ ./cluster/virtctl.sh console ovm-cirros                                                                                                                                     
Escape sequence is ^]                                                                                                                                                                                       
                     Can't connect to websocket (400): Unable to connect to VM because phase is Scheduling instead of Running                                                                               
```

To mitigate the impact of it, I need to use a "while true; do virtctl console ...; done" command to get connected as soon as possible.

While I appreciate that the infrastructure to provide console output may be not ready at this point, I think virtctl could spin / wait for console to become ready and connect me to it at the earliest possibility, so that I don't miss some of boot messages.

SchSeba:
virtctl console wait until the virtual machine is in running state

Fixes #1001 

**Release note**:

```release-note
virtctl console wait until the virtual machine is in running state
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/1323)

### DIFF
--- a/pkg/kubecli/vmi.go
+++ b/pkg/kubecli/vmi.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/gorilla/websocket"
 
@@ -246,7 +247,21 @@ func (v *vmis) VNC(name string) (StreamInterface, error) {
 	return v.asyncSubresourceHelper(name, "vnc")
 }
 func (v *vmis) SerialConsole(name string) (StreamInterface, error) {
-	return v.asyncSubresourceHelper(name, "console")
+	con, err := v.asyncSubresourceHelper(name, "console")
+	for err != nil {
+		if asyncSubresourceError, ok := err.(*AsyncSubresourceError); ok {
+			if asyncSubresourceError.GetStatusCode() == http.StatusBadRequest {
+				// Sleep to prevent denial of service on the api server
+				time.Sleep(500 * time.Millisecond)
+				con, err = v.asyncSubresourceHelper(name, "console")
+			} else {
+				return nil, asyncSubresourceError
+			}
+		} else {
+			return nil, err
+		}
+	}
+	return con, nil
 }
 
 type AsyncSubresourceError struct {
--- a/pkg/kubecli/vmi.go
+++ b/pkg/kubecli/vmi.go
@@ -252,17 +252,18 @@ type connectionStruct struct {
 	err error
 }
 
-func (v *vmis) SerialConsole(name string, timeout int) (StreamInterface, error) {
-	timeoutChan := time.Tick(time.Duration(timeout) * time.Minute)
+func (v *vmis) SerialConsole(name string, timeout time.Duration) (StreamInterface, error) {
+	timeoutChan := time.Tick(timeout)
 	connectionChan := make(chan connectionStruct)
+	isWaiting := true
 
 	go func() {
 		con, err := v.asyncSubresourceHelper(name, "console")
-		for err != nil {
+		for err != nil && isWaiting {
 			if asyncSubresourceError, ok := err.(*AsyncSubresourceError); ok {
 				if asyncSubresourceError.GetStatusCode() == http.StatusBadRequest {
 					// Sleep to prevent denial of service on the api server
-					time.Sleep(500 * time.Millisecond)
+					time.Sleep(1 * time.Second)
 					con, err = v.asyncSubresourceHelper(name, "console")
 				} else {
 					connectionChan <- connectionStruct{con: nil, err: asyncSubresourceError}
@@ -273,11 +274,14 @@ func (v *vmis) SerialConsole(name string, timeout int) (StreamInterface, error) 
 				return
 			}
 		}
-		connectionChan <- connectionStruct{con: con, err: nil}
+		if isWaiting {
+			connectionChan <- connectionStruct{con: con, err: nil}
+		}
 	}()
 
 	select {
 	case <-timeoutChan:
+		isWaiting = false
 		return nil, fmt.Errorf("Timeout trying to connect to the virtual machine instance")
 	case conStruct := <-connectionChan:
 		return conStruct.con, conStruct.err
--- a/pkg/kubecli/vmi.go
+++ b/pkg/kubecli/vmi.go
@@ -246,22 +246,42 @@ func (ws *wsStreamer) Stream(options StreamOptions) error {
 func (v *vmis) VNC(name string) (StreamInterface, error) {
 	return v.asyncSubresourceHelper(name, "vnc")
 }
-func (v *vmis) SerialConsole(name string) (StreamInterface, error) {
-	con, err := v.asyncSubresourceHelper(name, "console")
-	for err != nil {
-		if asyncSubresourceError, ok := err.(*AsyncSubresourceError); ok {
-			if asyncSubresourceError.GetStatusCode() == http.StatusBadRequest {
-				// Sleep to prevent denial of service on the api server
-				time.Sleep(500 * time.Millisecond)
-				con, err = v.asyncSubresourceHelper(name, "console")
+
+type connectionStruct struct {
+	con StreamInterface
+	err error
+}
+
+func (v *vmis) SerialConsole(name string, timeout int) (StreamInterface, error) {
+	timeoutChan := time.Tick(time.Duration(timeout) * time.Minute)
+	connectionChan := make(chan connectionStruct)
+
+	go func() {
+		con, err := v.asyncSubresourceHelper(name, "console")
+		for err != nil {
+			if asyncSubresourceError, ok := err.(*AsyncSubresourceError); ok {
+				if asyncSubresourceError.GetStatusCode() == http.StatusBadRequest {
+					// Sleep to prevent denial of service on the api server
+					time.Sleep(500 * time.Millisecond)
+					con, err = v.asyncSubresourceHelper(name, "console")
+				} else {
+					connectionChan <- connectionStruct{con: nil, err: asyncSubresourceError}
+					return
+				}
 			} else {
-				return nil, asyncSubresourceError
+				connectionChan <- connectionStruct{con: nil, err: err}
+				return
 			}
-		} else {
-			return nil, err
 		}
+		connectionChan <- connectionStruct{con: con, err: nil}
+	}()
+
+	select {
+	case <-timeoutChan:
+		return nil, fmt.Errorf("Timeout trying to connect to the virtual machine instance")
+	case conStruct := <-connectionChan:
+		return conStruct.con, conStruct.err
 	}
-	return con, nil
 }
 
 type AsyncSubresourceError struct {
--- a/pkg/kubecli/vmi.go
+++ b/pkg/kubecli/vmi.go
@@ -249,6 +249,19 @@ func (v *vmis) SerialConsole(name string) (StreamInterface, error) {
 	return v.asyncSubresourceHelper(name, "console")
 }
 
+type AsyncSubresourceError struct {
+	err        string
+	StatusCode int
+}
+
+func (a *AsyncSubresourceError) Error() string {
+	return a.err
+}
+
+func (a *AsyncSubresourceError) GetStatusCode() int {
+	return a.StatusCode
+}
+
 func (v *vmis) asyncSubresourceHelper(name string, resource string) (StreamInterface, error) {
 
 	done := make(chan struct{})
@@ -276,7 +289,7 @@ func (v *vmis) asyncSubresourceHelper(name string, resource string) (StreamInter
 		response, err := wrappedRoundTripper.RoundTrip(req)
 
 		if err != nil {
-			errChan <- err
+			errChan <- &AsyncSubresourceError{err: err.Error(), StatusCode: response.StatusCode}
 			return
 		}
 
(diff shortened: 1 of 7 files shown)


pkg/virtctl/console/console.go
```diff
@@ -93,19 +93,53 @@ func (c *Console) Run(cmd *cobra.Command, args []string) error {
 	writeStop := make(chan error)
 	readStop := make(chan error)
 
+	// Wait until the virtual machine is in running phase, user interrupt or timeout
+	runningChan := make(chan error)
+	timeoutChan := time.Tick(time.Duration(timeout) * time.Minute)
```
booxter: this ticker goroutine never stops though we don't need it running while interacting with the console. I think we should switch to time.NewTicker and explicitly Stop() it when it's no longer needed.

pkg/virtctl/console/console.go
```diff
@@ -93,19 +93,53 @@ func (c *Console) Run(cmd *cobra.Command, args []string) error {
...
+		runningChan <- nil
 		resChan <- con.Stream(kubecli.StreamOptions{
 			In:  stdinReader,
 			Out: stdoutWriter,
 		})
 	}()
 
+	select {
+	case <-timeoutChan:
+		return fmt.Errorf("Timeout trying to connect to the virtual machine instance")
```
booxter: there is a chance now that a timeout happens while the goroutine above actually succeeds to connect. Is there anything to clean up in this particular case? Particularly, doesn't con.Stream hold a connection state that would be better off closed with explicit close() call?

SchSeba: Because this is a command line program if the timeout or the interrupt channel is active the program will exit so the connection will be closed anyway.

SchSeba:
> Nice work. One thing to improve beyond the code is commit message. It currently has a very long title that wraps. There is a common guideline about titles being quite short, and descriptions expanding on the titles. For example, see: https://chris.beams.io/posts/git-commit/

Sure no problem I will make the tittle shorter.

> It would also make sense to refer to the issue you are solving in the commit messages. For that, you may refer to https://help.github.com/articles/closing-issues-using-keywords/ It's just putting "Closes #..." in the commit message. Then whenever we merge the PR, the issue will be closed. (And also, references between PR and the issue will be established.)

About the closes our bot is making the same thing by adding the `Fixes #1001` in the PR

pkg/virtctl/console/console.go
SchSeba: Done

pkg/virtctl/console/console.go
```diff
@@ -93,19 +93,53 @@ func (c *Console) Run(cmd *cobra.Command, args []string) error {
...
+	timeoutChan := time.Tick(time.Duration(timeout) * time.Minute)
+	waitInterrupt := make(chan os.Signal, 1)
+	signal.Notify(waitInterrupt, os.Interrupt)
+
 	go func() {
 		con, err := virtCli.VirtualMachineInstance(namespace).SerialConsole(vmi)
-		if err != nil {
-			resChan <- err
-			return
+		if strings.Contains(fmt.Sprint(err), "not found") {
```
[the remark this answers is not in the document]
SchSeba: Done

pkg/virtctl/console/console.go
```diff
@@ -93,19 +93,53 @@ func (c *Console) Run(cmd *cobra.Command, args []string) error {
...
+			runningChan <- err
+		}
+
+		for err != nil {
+			// Sleep to prevent denial of service on the api server
+			time.Sleep(500 * time.Millisecond)
+			con, err = virtCli.VirtualMachineInstance(namespace).SerialConsole(vmi)
```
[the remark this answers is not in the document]
SchSeba: I changed the check now its inside the for loop.

pkg/virtctl/console/console.go
```diff
@@ -93,19 +93,60 @@ func (c *Console) Run(cmd *cobra.Command, args []string) error {
...
+	timeoutChan := time.NewTicker(time.Duration(timeout) * time.Minute)
+	waitInterrupt := make(chan os.Signal, 1)
+	signal.Notify(waitInterrupt, os.Interrupt)
+
 	go func() {
 		con, err := virtCli.VirtualMachineInstance(namespace).SerialConsole(vmi)
-		if err != nil {
-			resChan <- err
-			return
+		for err != nil {
```
stu-gott: Sorry for the tardy review. I thought I had reviewed yesterday but my comment seems to have disappeared.

It might be worth looking into the "k8s.io/apimachinery/pkg/util/wait" library. There's a PollImmediate function that already handles the busywork of polling on an interval with a timeout/error conditions. Using that could possibly simplify this code a bit.

pkg/virtctl/console/console.go
booxter: Do you mean some deferred callback handles it for us? If so, great.

booxter:
The latest code changes are spot on, thanks (I would split the ticker change from error type change into separate commits since they are not both related to the commit message of the commit they are lumped into), but that's probably a minor issue.

The commit message title for one of the commits is still too long so it would be nice to see this fixed. Also to clarify, the reason why we may want to mention the issue number in PR commit messages and not just PR description on github.com is that in the former case, we have a direct link to the issue; while in the latter case, there are several levels of indirection (the reader needs to first identify which merge commit introduced a commit in the master tree; then follow the PR number from the merge commit message to the PR; then finally follow the link to the issue). [It's not a major thing, but may alleviate some of the pain in the future.]

One thing that I've missed in the previous round of review, and I am sorry for not pointing it out right away, is that the PR lacks new tests for the change. I think it would benefit from a functional test case that would start a VMI and immediately call to console and check that the tool actually connects to the VMI (after some time when the guest is up and ready) and not bails out on the user with an error, as it did before the PR. I haven't checked if we already have functional test cases for the CLI, but if not, that's an invitation to add some.

Thanks.

pkg/virtctl/console/console.go
SchSeba: I take a look into the function and make the implementation but this not simplify the code.
This help with the timeout but not with the user interrupt.
Also we need to create a new goroutine after the PollImmediate success to create a new connection to the console service for the connection stream.

SchSeba:
@booxter 
> The commit message title for one of the commits is still too long so it would be nice to see this fixed.
Changed.

> One thing that I've missed in the previous round of review, and I am sorry for not pointing it out right away, is that the PR lacks new tests for the change. I think it would benefit from a functional test case that would start a VMI and immediately call to console and check that the tool actually connects to the VMI (after some time when the guest is up and ready) and not bails out on the user with an error, as it did before the PR. I haven't checked if we already have functional test cases for the CLI, but if not, that's an invitation to add some.

I Changed how the console functional tests works by removing the waitForVMIStart call before we call the console.

We don't have functional test cases for the CLI and I am not sure how to test the console because if I call the console command from the code a get an error that he cant find a `tty` related to this line `terminal.MakeRaw(int(os.Stdin.Fd()))`

tests/utils.go
```diff
@@ -1130,10 +1131,21 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
 	expecterReader, expecterWriter := io.Pipe()
 	resCh := make(chan error)
 	go func() {
-		con, err := virtCli.VirtualMachineInstance(vmi.ObjectMeta.Namespace).SerialConsole(vmi.ObjectMeta.Name)
-		if err != nil {
-			resCh <- err
-			return
+		con, err := virtCli.VirtualMachineInstance(vmi.Namespace).SerialConsole(vmi.Name)
```
booxter: I think the idea of making functional tests implicitly validate console waiting for scheduler is great and better than having one or two cases to cover this behavior explicitly.

But it doesn't seem that you really validate the change in the tool here. This code below is a duplicate of what is added in the tool; you don't reuse the code from there. I think the right thing to do is to reuse the code from the tool here. It may require spinning up some of the tool code into a separate helper function ready to use.

As-is, I suspect the tests changes would pass even without the fix in the tool.

booxter:
I understand that if you just call the tool from inside tests directly then you get `terminal.MakeRaw(int(os.Stdin.Fd()))` called, which would be wrong. I think it suggests that you may need to spin up the goroutine, or its part, that establishes connection to the console (or fails when the timeout is reached or the VMI is not found) into a separate function that you would be able to call to from both the tool `main` body as well as from `tests/utils.go`.

stu-gott:
retest this please

pkg/kubecli/vmi.go
```diff
@@ -246,7 +247,21 @@ func (v *vmis) VNC(name string) (StreamInterface, error) {
 	return v.asyncSubresourceHelper(name, "vnc")
 }
 func (v *vmis) SerialConsole(name string) (StreamInterface, error) {
-	return v.asyncSubresourceHelper(name, "console")
+	con, err := v.asyncSubresourceHelper(name, "console")
+	for err != nil {
```
booxter: Isn't it an issue that the function may never exit? The change may hang functional tests indefinitely, among other things.

cynepco3hahue:
@SchSeba you need to use affinity from k8s API to avoid
```
# kubevirt.io/kubevirt/tests_test
tests/console_test.go:113:26: undefined: "kubevirt.io/kubevirt/pkg/api/v1".Affinity
make: *** [goveralls] Error 2
```

SchSeba:
Look like there is a problem with the travis-ci.
Complain on `git status --porcelain` and `tests/console_test.go:113:26: undefined: "kubevirt.io/kubevirt/pkg/api/v1".Affinity` but the tests runs without any problem on my local environment and my git status is empty.

cynepco3hahue:
@SchSeba Rebase on master I saw Roman merged some change to use k8s `Affinity` instead of creating our own.

SchSeba:
Rebase do the magic thanks @cynepco3hahue :)

pkg/kubecli/vmi.go
```diff
@@ -245,8 +246,55 @@ func (ws *wsStreamer) Stream(options StreamOptions) error {
...
+				connectionChan <- connectionStruct{con: nil, err: err}
+				return
+			}
+		}
+		connectionChan <- connectionStruct{con: con, err: nil}
+	}()
+
+	select {
+	case <-timeoutChan:
+		return nil, fmt.Errorf("Timeout trying to connect to the virtual machine instance")
```
booxter: it's nice that we eventually exit but shouldn't we also shut down the goroutine?

tests/utils.go
```diff
@@ -1130,7 +1130,7 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
 	expecterReader, expecterWriter := io.Pipe()
 	resCh := make(chan error)
 	go func() {
-		con, err := virtCli.VirtualMachineInstance(vmi.ObjectMeta.Namespace).SerialConsole(vmi.ObjectMeta.Name)
+		con, err := virtCli.VirtualMachineInstance(vmi.Namespace).SerialConsole(vmi.Name, int(timeout))
```
booxter: I think now that you reuse the same timeout both to wait for serial console to start AND later for the expecter, you should track how much time it took to connect to the serial console, and reduce the timeout that is used for expecter by that much. Otherwise this function may take significantly longer than expected by the caller, in case when it took significant time for serial console to connect.

pkg/kubecli/vmi.go
SchSeba: I don't think we need to stop the goroutine because we have 3 scenarios:
1. the helper return the con or error i will exit the goroutine
2. timeout this return an error an the virctl exit
3. user interrupt the virtctl exit

all of them will stop the goroutine.

tests/utils.go
SchSeba: if it took a lot of time to connect the test will exit because of the test time out no?

I don't think we need to make this complex because the test have timeout for all the test scenario (create vmi, wait for console, wait for the expecter).

pkg/kubecli/vmi.go
booxter: And when it's used in a functional test, do you terminate the goroutine (before the whole suite exits)? I don't think it's generate a good idea to leak it to outside even if the caller will eventually exit.

tests/utils.go
booxter: Yes but your timeout semantics is now broken (in some cases): the function receives a timeout argument but then doesn't satisfy it (by taking up to twice the expected time). While indeed it probably works (=passes tests) in most cases, it's not ideal.

tests/utils.go
```diff
@@ -1129,13 +1129,15 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
...
 	expecterReader, expecterWriter := io.Pipe()
 	resCh := make(chan error)
-	go func() {
-		con, err := virtCli.VirtualMachineInstance(vmi.Namespace).SerialConsole(vmi.Name, int(timeout))
-		if err != nil {
-			resCh <- err
-			return
-		}
 
+	startTime := time.Now()
```
booxter: Nice!

booxter:
@vladikr I think this PR is good to go; I am not sure about CI failures. I tried to deduce what failed in the jobs but couldn't find a place to extract more detailed logs for the failures. Are these failures known?

vladikr:
ok to test

vladikr:
not sure why kubevirt functional tests didn't run.

SchSeba:
@booxter @vladikr All the tests looks good now.

booxter:
I don't have full exposure as to which jobs are supposed to pass, but I see that 1.10.3 all pass (except openshift that is expected to fail at least because of custom MAC address issue; and other failures don't seem related either). I will take it upon myself to merge this patch. I hope that's ok.

pkg/kubecli/vmi.go
```diff
@@ -245,8 +246,59 @@ func (ws *wsStreamer) Stream(options StreamOptions) error {
...
+	connectionChan := make(chan connectionStruct)
+	isWaiting := true
+
+	go func() {
+		con, err := v.asyncSubresourceHelper(name, "console")
+		for err != nil && isWaiting {
+			if asyncSubresourceError, ok := err.(*AsyncSubresourceError); ok {
+				if asyncSubresourceError.GetStatusCode() == http.StatusBadRequest {
+					// Sleep to prevent denial of service on the api server
+					time.Sleep(1 * time.Second)
```
rmohr: I think it is better to use watches instead of polling. That should be much cheaper once the connection is established.

PR #1324, opened 2018-07-16
Support disconnected instances

booxter:
**What this PR does / why we need it**: adds support for network isolated instances.

Part of #1067.

**Release note**:

```release-note
Added support for autoAttachPodInterface field in devices section to disable automatic attacment of a pod network interface to an instance.
```

### DIFF
--- a/pkg/api/v1/schema.go
+++ b/pkg/api/v1/schema.go
@@ -170,6 +170,8 @@ type Devices struct {
 	Watchdog *Watchdog `json:"watchdog,omitempty"`
 	// Interfaces describe network interfaces which are added to the vm.
 	Interfaces []Interface `json:"interfaces,omitempty"`
+	// Whether to attach a pod network interface. Defaults to true.
+	AutoattachPodInterface *bool `json:"autoattachPodInterface,omitempty"`
 }
 
 // ---
--- a/tests/vmi_networking_test.go
+++ b/tests/vmi_networking_test.go
@@ -447,4 +447,25 @@ var _ = Describe("Networking", func() {
 			checkMacAddress(deadbeafVMI, deadbeafVMI.Spec.Domain.Devices.Interfaces[0].MacAddress, "localhost:~#")
 		})
 	})
+
+	Context("VirtualMachineInstance with disabled automatic attachment of interfaces", func() {
+		It("should not configure any external interfaces", func() {
+			By("checking loopback is the only guest interface")
+			autoAttach := false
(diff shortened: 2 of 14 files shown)


tests/vmi_networking_test.go
```diff
@@ -447,4 +447,26 @@ var _ = Describe("Networking", func() {
...
+			_, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(detachedVMI)
+			Expect(err).ToNot(HaveOccurred())
+			waitUntilVMIReady(detachedVMI, tests.LoggedInCirrosExpecter)
+
+			err := tests.CheckForTextExpecter(detachedVMI, []expect.Batcher{
+				&expect.BSnd{S: "\n"},
+				&expect.BExp{R: "\\$ "},
+				&expect.BSnd{S: "ls /sys/class/net/\n"},
+				&expect.BExp{R: "lo"},
+				&expect.BExp{R: "\\$ "},
```
rmohr: I think that this test would also succeed if other network devices are present for as long as "lo" would appear in the list.

booxter: You are right, I didn't expect it to silently "skip" eth0 entry.

booxter:
The latest upload updates the functional test case to check the number of interfaces exposed into the guest instead of their particular names. This simplifies the check since we can just check that there is only a single interface (which is always lo) instead of being bothered by the order of listdir() result and building a more complex output matching regexp to accommodate for the random order.

pkg/api/v1/schema.go
```diff
@@ -170,6 +170,8 @@ type Devices struct {
 	Watchdog *Watchdog `json:"watchdog,omitempty"`
 	// Interfaces describe network interfaces which are added to the vm.
 	Interfaces []Interface `json:"interfaces,omitempty"`
+	// Whether to attach a pod network interface. Defaults to true.
+	AutoAttachPodInterface *bool `json:"autoAttachPodInterface,omitempty"`
```
rmohr: Not a huge thing  but I see that in k8s something equivalent is called `automountServiceAccountToken` and not `autoMountServiceAccountToken`. Do we want to use `autoattach` instead of `autoAttach`?

booxter:
/hold

Will rename the field as per Roman's suggestion. Consistency is important.

pkg/api/v1/schema.go
booxter: Done.

booxter:
The latest upload changes autoAttachPodInterface -> autoattachPodInterface.

booxter:
I am sorry for the last upload, it had some missing changes that 'make generate' should have generated. This is now fixed.

PR #1334, opened 2018-07-17
Fix typo in virtwrap converter

j-griffith:
In virt-launcher/virtwrap/api/converter.go there's a function for
converting a FilesystemVolumeSource, but it's name is "Covert"
instead of "Convert".

This fixes that naming and adds a proper comment to the function.


slintes:
@j-griffith answering your question from IRC here: the new ovirt ci jobs can be ignored for now, that's work in progress.

slintes:
ok to test

slintes:
OK to test

PR #1342, opened 2018-07-19
Add l2 interface option

### LINKED ISSUE #1329: Provide network device plugin for basic L2 bridging
> As a user I need a simple wya to bridge my VM into a local network.

The requirement here is to have a simple network device plugin which allows me to bridge a vNIC onto a local network using a dedicated physical NIC of the host.

Design: https://docs.google.com/document/d/1IrAKDph9qEX7kcx04uUQy6MYHn6vratOi_UfGNIjTfI/edit#

yuvalif:
This PR is intendent to allow exposing an existing L2 host network to the VM in addition to the k8s one.

Fixes #1329 

**Still missing:**
~~- integration tests~~
- unit tests
~~- documentation (will be a separate PR in the user-guide repo)~~
- The only "protocol" currently supported is "Ethernet" - which imply "bridged" binding without delegating the IP/MAC
  - We SHOULD also support "PCI", which would imply device passthrough
  - We MAY also support "IP" - more discussion is needed here, at device plugin level
  - Once we support more options, we will have to reconcile the "protocol" capabilities of the pod interface with the (optional) binding configuration of the interface

**Release Notes:**
```release-note
New network type was added to the virtual machine specification: "resource network". 
This is intended at networks exposed by using a network device plugin. 
The configuration of the network will include the name of the device plugin and the name of the network resource it should expose, using the following format: <device-plugin-name>/<network-name>

In the yaml this will have the following format:
networks:
  - name: red 
    resource:
      resourceName: bridge.network.kubevirt.io/red

To communicate the name of the pod interface that the device plugin creates, and what kind of networking this interface exposes to the pod, the device plugin will set a JSON encoded environment variable to the pod. 
The name of this environment variable must start with: "NETWORK_INTERFACE_RESOURCES_", and assumes the following JSON format for example:
{
  "name": "bridge.network.kubevirt.io/red", 
  "interfaces": [{"name": "eth-red01", "protocol": "Ethernet"}, {"name": "eth-red02", "protocol": "Ethernet"}]
}

Other than the definition of the environment variable, no other requirements exists on the device plugin, or how the host network is managed or created.

The following device plugin is compliant with the above: https://github.com/kubevirt/kubernetes-device-plugins/blob/master/docs/README.bridge.md

Notes:
- The device plugin allows multiple pod interfaces to be exposed for the same network (eth-red01 and eth-red02 in the example above), however, inside the VMI, we will expose only one interface.
- If multiple device plugins expose the same network needed by the virtual machine, only one of them will be used, and a single interface per network will be created on the virtual machine.
- In the virtual machine specification, only interface with "bridge" binding is allowed to use a network of type "resource". Any other combination is considered invalid.
- The only "protocol" currently supported is "Ethernet" - which implies "bridged" binding without delegating IP/MAC addresses.
```


gageorsburn:
The device plugin bridge allowed for passing in a count of interfaces, are we not going to allow the same functionality here?

I'm working with virtual routers and will probably need at least 10 interfaces per router. I could define 10 bridges on the host but it would be more operational overhead than being able to specify a count of interfaces I want from that bridge on the host.

yuvalif:
@gageorsburn this could be easily done from the device plugin side. It is already supported there, and it would be easy to pass that from the VMI network definition to the device plugin definition in the launcher pod. This would assume, though, that all of these interfaces are connected to the same "network" (the host bridge in our case), but will just be exposed as multiple interfaces on the guest VM.
For example:
```yaml
networks:
  - name: red 
    resource:
      resourceName: bridge.network.kubevirt.io/red
      resourceCount: 3
```
Note that at VMI level, there will be a single "interface" definition, even though multiple interfaces will be generated on the VM.

yuvalif:
@gageorsburn I've captured that requirement here: https://github.com/kubevirt/kubevirt/issues/1366
Once this PR is merged, it will be straight forward to add it.

tools/vms-generator/README.md
```diff
@@ -0,0 +1,3 @@
+# vms-generator
+This is the tool which generated the YAML files under: `cluster/example`. Which, in turn, used for manual and integration testing.
```
booxter: -> cluster/examples

booxter:
I've posted some comments, but really, the only one that is worth special attention is the one about how we construct in-pod interface names in terms of their lengths (Linux has low limit on the number of characters allowed and may either fail or truncate names that are too long).

pkg/virt-launcher/virtwrap/network/resourceinterface.go
```diff
@@ -0,0 +1,192 @@
...
+	Protocol string `json:"protocol"`
+}
+
+type ResourceConfiguration struct {
+	Name       string                      `json:"name"`
+	Interfaces []PodInterfaceConfiguration `json:"interfaces"`
+}
+
+func (l *ResourceInterface) cachePodInterfacesFromEnvironment() {
+	// TODO: no actual caching is happening here, so we can refresh. should e add rate limit here?
```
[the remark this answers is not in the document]
yuvalif: any comment on the question? do we want to cache the env vars?

pkg/virt-launcher/virtwrap/network/resourceinterface.go
```diff
@@ -0,0 +1,192 @@
...
+			resourceName := pair[1]
+
+			if len(conf.Interfaces) == 0 {
+				log.Log.Warningf("Environment variable %s, contains no pod interfaces", varName)
+				continue
+			}
+
+			// if configuration does not exists for a resource
+			// just take the first pod interface of the first device plugin which configures this network
+			if l.Configuration[resourceName] == nil {
```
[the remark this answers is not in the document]
yuvalif: if there are multiple plugins that expose the same network needed by the VM, i don't think there is any specific requirement around consistency etc.
I think the only issue, is that we may fall on some bad configuration (e.g interface does not really exists), and we would not try others (that may be good).
but this seems to me like overengineering this mechanism.
maybe future enhancement, if requested.

pkg/virt-launcher/virtwrap/network/resourceinterface.go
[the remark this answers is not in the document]
booxter: Is it called often? I actually thought it's a one time per container, so why would we cache?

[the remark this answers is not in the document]
yuvalif: only on start. and in a loop until the start is successful - so backoff is already part of the process

yuvalif:
- The main reason to a device plugin is to take advantage of the existing k8s scheduling mechanism. Currently k8s don't have a dedicated mechanism for networks as host resources, so we utilize (abuse...) device plugins for that

- Keep it simple
  - We can introduce the "host bridge" device plugin as integral part of kubevirt (similarly to how kvm is done)
  - However, unlike the kvm case, where there is no value from having an external device plugin, in the networking case, supporting an external plugin is required. Such an external networking device plugin (not the one we are using) may extend further into the host network, or even beyond the host, in the configuration and setup work it does. We do not want to preclude plugins provided externally from working with kubevirt
 - Not sure which privileges are introduced here for the launcher? A pod may request a resource from a privileged device plugin without being privileged by itself
 - The main value with allowing abstraction of the host-to-pod part of the functionality and separating from kubevirt (although it is core) is that it allows for integration with more advanced network device plugins, covering functionality that we would not like to integrate into kubevirt

Seems like it would be good to make the "host bridge" device plugin integral part of kubevirt, as this will make simple deployments simple, simplify testing etc.
However, would still keep the API, to allow for L2 networks to be exposed via any other device plugin that is compliant with the defined API.

rmohr:
> However, would still keep the API, to allow for L2 networks to be exposed via any other device plugin that is compliant with the defined API.

Let's discuss that on the community meeting. I can just think right now of three reasonable approaches on how to add network to the vm:

 * (bridge) tap
 * macvtap
 * full device passthrough

I think we can come up with something less architecture-intrusive to cover these use-cases. Preferrable only on the host-side (nothing like this json which is passed in to virt-launcher at start-time, instead virt-handler mostly).

gageorsburn:
In my opinion, it should be pluggable. This doesn't have to be the only option for L2 connectivity but if I want to use a bridge device plugin there should be a way for kubevirt to expose/allow me to do so. Right before @yuvalif opened this PR i was working on the same code to achieve this until a better solution for L2 was released.

tests/vmi_bridge_test.go
```diff
@@ -0,0 +1,480 @@
...
+			job.Spec.NodeSelector = nodeSelector
+			job, err = virtClient.CoreV1().Pods(tests.NamespaceTestDefault).Create(job)
+			Expect(err).ToNot(HaveOccurred())
+			phase = waitForPodToFinish(job)
+			// don't check address setting result
+		}
+
+		// add red and blue bridges to host
+		addBridgeToHost("red")
+		addBridgeToHost("blue")
```
duyanyan: I have a small question, do we need to remove the bridges after finishing the testing?

tests/vmi_bridge_test.go
```diff
@@ -0,0 +1,480 @@
...
+						"name":      name,
+					},
+				},
+			}
+			_, err = virtClient.CoreV1().ServiceAccounts(tests.NamespaceTestDefault).Create(&serviceAccount)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Setting SCC for the service account is only needed for OpenShift cluster
+			// Not verifying result since this is expected to fail on anything other than OpenShift
+			tests.RunKubectlCommand("adm", "policy", "add-scc-to-user", "privileged", "-n",
```
gonzolino: I am trying to build this patch, but every attempt fails with:

```
tests/vmi_bridge_test.go:77:4: undefined: tests.RunKubectlCommand
```

Is there a method missing, am I doing something wrong or is this OpenShift specific?

duyanyan: @gonzolino I think it is caused by this pr https://github.com/kubevirt/kubevirt/pull/1408

gonzolino: Ah yes thank for the hint! Exchanging `RunKubectlCommand` with `RunCommand("kubectl"` did the trick 👍

yuvalif:
Device plugins should not be used (for now) as a mean to provide L2 networking to kubevirt.
Discussion and reasons are detailed here: https://groups.google.com/forum/#!topic/kubevirt-dev/AgcZF4YtmtQ

PR #1349, opened 2018-07-19
Initialize disk.img for empty persistent volume claims attached to vm

### LINKED ISSUE #991: Initialize empty filesystem-based PVs
Currently, PVs must be manually prepped with a disk.img before they can be successfully used with a KubeVirt VM. There are often cases where users intend to attach an empty PV to their VM during new deploys, so this process can be handled automatically.

Per [discussion](https://groups.google.com/d/topic/kubevirt-dev/hybg5umVA0A/discussion), If 1) a volume source is a PersistentVolumeClaimVolumeSource, 2) is a filesystem, and 3) the volume is empty, the kube-launcher should create a new, empty disk.img automatically. Take filesystem overhead into account and ensure disk is not larger than actual usable FS space.

mfranczy:
It gives a possibility to attach empty PVC to VM without manually created disk.img

Fixes #991

After a talk with @rmohr we have decided to add a new volume source type
which is a host disk that is created based on defined PVC.

Scenario:
1. the user is attaching empty PVC to VM
2. virt-handler is checking the size of the PVC (it has to do that, I don't have this information)
3. virt-handler is changing PVC to HostDisk (it initializes new HostDisk structure)
4. virt-launcher is checking if VMI object has HostDisk volume defined
5. virt-launcher is checking if disk.img exists, if not, then disk.img will be created and saved on the host

**Release note**:

```release-note
Initializing an empty PVC by using a new volume source type which is HostDisk. 
```


mfranczy:
This PR is still in progress, I have to add additional checks like if there is enough space for disk.img, take filesystem overhead into account and add tests.

However I think we can discuss if the solution is acceptable.

pkg/api/v1/schema.go
```diff
@@ -32,6 +32,20 @@ import (
...
 
+// Represents a disk created on a cluster level based on PersistentVolumeClaim
+// --
+// +k8s:openapi-gen=true
+type HostDisk struct {
+	// PersistentVolumeClaimVolumeSource represents a reference to a PersistentVolumeClaim in the same namespace.
+	// Directly attached to the vmi via qemu.
+	// More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#persistentvolumeclaims
+	// +optional
+	PersistentVolumeClaim *v1.PersistentVolumeClaimVolumeSource `json:"persistentVolumeClaim,omitempty"`
```
rmohr: I think we can model `HostDisk` pretty close to `HostPath` from `k8s: https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

So something like:

```yaml
  - name: test-disk
    hostPath:
      # path to disk
      path: /data/disk.img
      # this says that the disk must be there
      type: Disk
```

and

```yaml
  - name: test-disk
    hostPath:
      # path to disk
      path: /data/disk.img
      # this says that the disk will be created if it is not there
      type: DiskOrCreate
      # disk size which will be used if the disk needs to be created
      capacity: 20Gi
```

pkg/virt-handler/vm.go
```diff
@@ -414,6 +415,23 @@ func (d *VirtualMachineController) execute(key string) error {
...
 
+func (d *VirtualMachineController) injectHostDisks(vmi *v1.VirtualMachineInstance) error {
+	// host disks are created based on PersistenVolumeClaims
+	for i := range vmi.Spec.Volumes {
+		if volumeSource := &vmi.Spec.Volumes[i].VolumeSource; volumeSource.PersistentVolumeClaim != nil {
+			pvcSize, err := hostdisk.GetPVCSize(volumeSource.PersistentVolumeClaim.ClaimName, vmi.Namespace, d.clientset)
+			if err != nil {
+				return err
+			}
+			volumeSource.HostDisk = v1.HostDisk{
```
rmohr: If you model the HostDisk like mentioned above, you can replace the PVC with the capacity of the PV, derive the disk-path based on the VMI definition and set the type to `DiskOrCreate`.

pkg/api/v1/schema.go
rmohr: When we go that route, we would also have to consider the `HostDisk` if it exists on the VMI itself natively by adding a HostPath mount to the pod.

pkg/virt-handler/vm.go
mfranczy: Ok I will do it.

pkg/api/v1/schema.go
mfranczy: Ok cool, and just to be on the same page, I understand from examples that you want to use hostDisk name in yaml definition (since hostPath type already exists), so it should look like this:
```
volumes:
  - name: test-disk
    hostDisk:
      path: /data/disk.img
      type: DiskOrCreate
      capacity: 20Gi
```
am I right?

rmohr: Yes, the snipped you provided would be translated to something like this on the pod template:

```
volumes:
  - name  test-disk
    hostPath:
      path: /data
      type: DirectoryOrCreate
```

pkg/host-disk/host-disk.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+const pvcBaseDir = "/var/run/kubevirt-private/vmi-disks"
+
+func getDiskImgPath(volumeName string) string {
+	return path.Join(pvcBaseDir, volumeName, "disk.img")
+}
+
+func calculateRawImgSize(quantity resource.Quantity) int64 {
+	// TODO: take fs overhead into account
```
mlsorensen: If you have the path of the mounted FS, you can do something like:

```
func dirMegabytesAvailable(path string) (uint64, error) {
       var stat syscall.Statfs_t
       err := syscall.Statfs(path, &stat)
       if err != nil {
               return 0, err
       }

       return (stat.Bavail * uint64(stat.Bsize) >> 20), nil
}
```

This will tell you what space is actually there on the FS, minus things like admin reserve (5% on ext, 0% of xfs).
I would not rely on the PVC size from the API, or at least sanity check it against the actual system FS available.

rmohr: > This will tell you what space is actually there on the FS, minus things like admin reserve (5% on ext, 0% of xfs).
> I would not rely on the PVC size from the API, or at least sanity check it against the actual system FS available.

@mlsorensen For me personally it is not really clear what to do in such a situation. I guess the correct thing would be to fail starting the vm. I thought about the root reservation which you also mentioned in the issue itself. It looks to me more like a bug in the storage provisioner than something the application should think of. What do you think?

pkg/host-disk/host-disk.go
```diff
@@ -0,0 +1,80 @@
...
+func CreateHostDisks(vmi *v1.VirtualMachineInstance) error {
+	// TODO: add checks:
+	// - if there is enough space
+	for _, volume := range vmi.Spec.Volumes {
+		if hostDisk := volume.VolumeSource.HostDisk; hostDisk.PersistentVolumeClaim != nil {
+			size := strconv.FormatInt(calculateRawImgSize(hostDisk.Capacity), 10)
+			file := getDiskImgPath(volume.Name)
+
+			if _, err := os.Stat(file); os.IsNotExist(err) {
+				if err := exec.Command("qemu-img", "create", "-f", "raw", file, size).Run(); err != nil {
```
mlsorensen: If you want to avoid exec, it's simple to create a raw sparse image. See this little proof of concept. Up to you.

```
package main

import (
  "os"
)

func createSparseRaw(fileName string, gb int) {
        offset := int64(gb << 30) - 1
        f,_ := os.Create(fileName)
        defer f.Close()
        f.WriteAt([]byte{0}, offset)
}

func main() {
        createSparseRaw("disk.img", 10)
}
```

When I run this I get:

```
$ go run sparse.go
$ qemu-img info disk.img
image: disk.img
file format: raw
virtual size: 10G (10737418240 bytes)
disk size: 4.0K
```

pkg/host-disk/host-disk.go
mlsorensen: I think maybe this implementation has gone beyond what I had originally envisioned.  I had figured that the PVC would simply have a size, and under certain circumstances a disk,img would be created that matched the PVC's full size. In this case there would be no need to have HostDisk with its own capacity.

In this case it seems perhaps that we have a PVC of a predetermined size(say 100G), and we may be asking for a host disk of a different capacity, for example a 10G HostDisk to be created on it.  If this is the design, then yes, you could sanity check that the PVC's available space (via statfs) can hold it to avoid errors as the disk image fills up. I can think of a few issues that could arise if the intention is to share the PV for multiple disk images or data and only the sparse sizes are accounted for.

If the intention is not to share, then I wonder why HostDisk has its own size. Simply use the PVC size or detect it's usable space as above.

At any rate, please take or leave my comment as you wish. I was  just giving an idea of how one may implement a  "TODO: take fs overhead into account". I think I need to go back and understand the implementation a little better.

rmohr: > If the intention is not to share, then I wonder why HostDisk has its own size. Simply use the PVC size or detect it's usable space as above.

@mlsorensen I understand the confusion. Also lies within my responsibility since I asked @mfranczy to work into this direction. The `HostDisk` is in general an independent feature. So if you specify a `HostDisk` on the VM, then we would mount a HostPath and create the disk if requested with the specified size.

For the PVC initialization, the host-disk also works as a vehicle between virt-handler and virt-launcher to pass in the size which from the PVC. The rational behind this is that for virt-launcher it does not make any difference if the mounted path is coming from a PVC or from a hostDisk.

I find the `HostDisk` a useful missing feature and just found that it is a nice base for the PVC initialization work.

mfranczy: @rmohr  @mlsorensen I have added new host disk volume source as independent feature.
If there is a ```PersistentVolumeClaim``` it's gonna be translated to ```HostDisk```,
and you are able to use ```HostDisk``` by itself definition.
Now i am writing detailed comment how it works exaclty...

@mlsorensen I have to update PR according to your comments about creating sparse file.

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -101,6 +102,12 @@ func (l *LibvirtDomainManager) preStartHook(vmi *v1.VirtualMachineInstance, doma
 		return domain, err
 	}
 
+	// create images for volumes with persistent volume claims attached
```
mfranczy: fix that comment

pkg/api/v1/schema.go
```diff
@@ -289,6 +303,9 @@ type Volume struct {
 // ---
 // +k8s:openapi-gen=true
 type VolumeSource struct {
+	// HostDisk represents a disk created on a cluster level based on a PersistentVolumeClaim
```
mfranczy: fix that comment

pkg/host-disk/host-disk.go
mfranczy: @mlsorensen  regarding your comment:

>If you have the path of the mounted FS, you can do something like
...
>I would not rely on the PVC size from the API ...

I thought about that, but I have to rely on the PVC size from the API, if I wouldn't then I have the free space from the node. I used vmi-pvc.yaml from examples and checked how PVC is mounted into container. I guess we have to improve this?

Executed from the pvc container with the path of the mounted FS:
```
$ df -Ph /run/kubevirt-private/vmi-disks/pvcvolume
Filesystem                       Size  Used Avail Use% Mounted on
/dev/mapper/VolGroup00-LogVol00   38G  4.2G   34G  12% /run/kubevirt-private/vmi-disks/pvcvolume
```

Executed from the node
```
$ df -Ph /
Filesystem                       Size  Used Avail Use% Mounted on
/dev/mapper/VolGroup00-LogVol00   38G  4.2G   34G  12% /
```

mlsorensen: I see, so this is all related to using the node's filesystem for disk images.  In my world we only have PVCs that are independent block devices, and will eventually consume those directly, but in the mean time the option was to use it entirely. The above doesn't work if your mounted FS is not dedicated to the PVC.

pkg/api/v1/schema.go
```diff
@@ -32,6 +32,20 @@ import (
  ATTENTION: Rerun code generators when comments on structs or fields are modified.
 */
 
+// Represents a disk created on the cluster level
```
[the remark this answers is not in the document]
mfranczy: Yes, you're right, I will change that, it's on the node.

pkg/virt-handler/vm.go
```diff
@@ -441,6 +442,28 @@ func (d *VirtualMachineController) execute(key string) error {
 	return nil
 }
 
+func (d *VirtualMachineController) replacePVCByHostDisk(vmi *v1.VirtualMachineInstance) error {
+	// if PVC is defined then it is replaced by HostDisk
+	// PersistenVolumeClaim is mounted into pod as directory from node filesystem
+	// it is not attached as independent block device
+	for i := range vmi.Spec.Volumes {
```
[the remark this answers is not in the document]
mfranczy: Hm, I did this on purpose, few lines below I am updating volume source structure (PVC is changing to HostDisk). My point is that I would have a different memory address and would have to do more to update the volume structure, because of ```range``` is copying values:
```
&vmi.Spec.Volumes[i] != &volume 
```

Although I am not so experienced with golang, so after this explanation do you still think I should change this? Thx.

fabiand:
@mfranczy please add a release note

pkg/host-disk/host-disk_test.go
```diff
@@ -0,0 +1,227 @@
...
+					Expect(err).NotTo(HaveOccurred())
+					Expect(img3.Size()).To(Equal(int64(83886080))) // 80Mi
+				})
+				It("Should stop creating disk images if there is not enough space and should return err", func() {
+					By("Creating a new minimal vmi")
+					vmi := v1.NewMinimalVMI("fake-vmi")
+
+					By("Adding a HostDisk volumes")
+					addHostDisk(vmi, "volume1", v1.HostDiskExistsOrCreate, "64Mi")
+					addHostDisk(vmi, "volume2", v1.HostDiskExistsOrCreate, "1E")
```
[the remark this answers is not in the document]
mfranczy: I am creating a directory in temp for sparse files and after each test, it's removed.

But talking specifically about ```1E```, I took this because it is the biggest unit from ```resource.Quantity``` (https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L46) and I'm sure that nobody has a disk like this, ofc I could take some Peta value... but ```1E``` says "oh man this is huge". So, I am able to test condition if there is enough space for a disk.img.

Even if my check would be wrong, it still shouldn't affect the system because this is a sparse file.

mfranczy:
> overall looks good, but I believe we will have some troubles when block PVC will be introduced.
@cynepco3hahue I agree with you, when block PVC will be introduced we will have to change a few things. I think PVC should not be replaced by HostDisk then.

pkg/virt-handler/vm.go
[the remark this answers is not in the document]
cynepco3hahue: I see do you mean this one `volumeSource.PersistentVolumeClaim = nil`, works for me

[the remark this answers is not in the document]
mfranczy: Yup, and this one: 
```
volumeSource.HostDisk = &v1.HostDisk{
				Path:     hostdisk.GetPVCDiskImgPath(vmi.Spec.Volumes[i].Name),
				Type:     v1.HostDiskExistsOrCreate,
				Capacity: pvcSize,
			}
```

fabiand:
It might be that this PR is fixing the issue, but we should at least have a test which is testing that `disk.img` is getting created on an empty PVC.

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -310,6 +313,34 @@ func validateVolumes(field *k8sfield.Path, volumes []v1.Volume) []metav1.StatusC
 				})
 			}
 		}
+
+		// validate HostDisk data
+		if hostDisk := volume.HostDisk; hostDisk != nil {
```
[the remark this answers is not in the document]
mfranczy: Ok, in progress...

mfranczy:
@fabiand I have added integration tests for PVC initialization,
@cynepco3hahue unit tests are added as well for webhook

pkg/virt-api/validating-webhook/validating-webhook.go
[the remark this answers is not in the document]
mfranczy: Done

rmohr:
> Now it looks like this PR is introducing a volume type which seems to be a combination of a local PV and the disk.img initialization.

Yes. I have the HostDisk on my list for some time and it is a solid base for this feature, it is not necessary to solve the issue but helpful here and for otther things (e.g. placing an image on the node and accessing it without the need to run a full storage provisioner or a data importer). The `hostDisk` is the equivalend of `hostPath`. Local PVs are something different. The relation between `hostDisk` and local PVs is the same as  the relationhip between `hostPath` and local PVs.

pkg/api/v1/schema.go
```diff
@@ -32,6 +32,20 @@ import (
...
+// --
```
rmohr: One `-` is missing. Three `---` will make everything after it not appear in the api docs.

pkg/host-disk/host-disk.go
```diff
@@ -0,0 +1,101 @@
...
+		if hostDisk := volume.VolumeSource.HostDisk; hostDisk != nil && hostDisk.Type == v1.HostDiskExistsOrCreate && hostDisk.Path != "" {
+
+			if _, err := os.Stat(hostDisk.Path); os.IsNotExist(err) {
+				availableSpace, err := dirBytesAvailable(path.Dir(hostDisk.Path))
+				if err != nil {
+					return err
+				}
+				size, _ := hostDisk.Capacity.AsInt64()
+				if uint64(size) > availableSpace {
+					return errors.New(fmt.Sprintf("Unable to create %s with size %s - not enough space on the cluster", hostDisk.Path, hostDisk.Capacity.String()))
```
rmohr: You can use `fmt.Errorf` instead of the `errors.New(fmt.Sprintf(` pattern.

pkg/virt-handler/vm.go
```diff
@@ -441,6 +442,28 @@ func (d *VirtualMachineController) execute(key string) error {
...
+				Capacity: pvcSize,
+			}
+			// PersistenVolumeClaim is replaced by HostDisk
+			volumeSource.PersistentVolumeClaim = nil
+		}
+	}
+	return nil
+}
+
 func (d *VirtualMachineController) injectCloudInitSecrets(vmi *v1.VirtualMachineInstance) error {
```
rmohr: Either we rename this method or we should create a separate method for injecting the additional PVC data.

tests/storage_test.go
```diff
@@ -306,5 +308,68 @@ var _ = Describe("Storage", func() {
...
+
+				By("checking if disk.img exists")
+				vmiPod := tests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
+				output, err := tests.ExecuteCommandOnPod(
+					virtClient,
+					vmiPod,
+					vmiPod.Spec.Containers[0].Name,
+					[]string{"find", "/data", "-name", "disk.img", "-size", "1G"},
+				)
+				Expect(strings.Contains(output, "/data/disk.img")).To(BeTrue())
```
rmohr: Since the path is hardcoded I would expect that after an initial test run the disk.img already exists and subsequent runs would not test exactly the same scenario, is that the case?

mfranczy: Yes, that's the case here, I am testing ```DiskExistsOrCreate``` (first run should create a disk.img) and then ```DiskExists``` (the second run should not receive any error since disk.img has been created)

rmohr: That would mean that if I run `make functest` two times, I am testing different things. We should not do that.

mfranczy: ofc.. you're right, don't know why I expected every ```make functest``` will run on clean cluster..
I will fix that. Thanks.

rmohr: Btw. also expect that tests can be executed in a randomized order (there is an extra ginkgo flag for that).

rmohr: I see different ways:

 * splitting the suite into two contexts, one where a disk is already created in `BeforeEach` or `BeforeAll` and one where it is not
 * Just flattening the table test in this case and describe with `By()` what you do

Not sure if a loop helps you a lot when you want do document the tests

To test the case where the image already exists in a pvc, using the alpine pvc is definitely good. 

In general you could also use random image names per `It` if they should not overlap and simply delete all images in that folder in a `BeforeSuite` (to make it reentrant in error cases) and `AfterSuite`.

pkg/host-disk/host-disk.go
mfranczy: Done

pkg/api/v1/schema.go
mfranczy: Fixed

pkg/virt-handler/vm.go
rmohr: Forget the comment.

tests/storage_test.go
mfranczy: Done.

mfranczy:
@rmohr according to our conversation, tests are fixed, please take a look when you will find some free time, thanks.

mfranczy:
ci test please

mfranczy:
ci test please

mfranczy:
ci test please

mfranczy:
ci test please

tests/utils.go
```diff
@@ -322,6 +326,8 @@ func AfterTestSuitCleanup() {
 	DeletePVC(osAlpineHostPath)
 	DeletePV(osAlpineHostPath)
 
+	cleanupHostDiskData()
```
rmohr: The last thing I saw: The pod created for deleting this is cleaning up asynchronously. That implies a few things:

1) If this line succeeds, the data from the node is not really deleted
2) we would create a lot of cleanup pods

I think that you need to run this cleanup especially for the related tests (before and afterwards) and also wait for the pod to succeed or fail.

mfranczy: Yes you're right, as well I have to handle a situation where there is more than one node, basically, I think that's why my tests are failing on ```check-patch.k8s-1.11.0-dev.el7.x86_64 ```, I am just lucky that others pass.
In progress...

mfranczy:
ci test please

rmohr:
ci test please

tests/storage_test.go
```diff
@@ -306,5 +310,130 @@ var _ = Describe("Storage", func() {
...
+					tests.WaitForVirtualMachineToDisappearWithTimeout(vmi, 120)
+				})
+
+				AfterEach(func() {
+					tests.RemoveHostDiskImage(diskPath, nodeName)
+				})
+			})
+		})
+
+		Context("With a multiple empty PVCs", func() {
```
rmohr: `With multiple empty PVCs`

rmohr:
@mfranczy could you clean up the commit history a little bit?

mfranczy:
Sure I will clean it.

mfranczy:
@rmohr done, commit history looks better now.

rmohr:
@mfranczy great. Let's wait for CI.

PR #1352, opened 2018-07-20
Generalize dockerization

benwatcisco:
This generalizes the "dockerize" script and associated plumbing
to allow different containers for different purposes.  As a proof
of concept, the manifest generation is performed in a simple Python
container, since that's all that it really needs.

N/A

Environment variables for each container are specified in a per-
container file.  This is related to the work in PR #1311, where I tried a couple of ways to get environment variables into the container.

Note that this also obviates all the specification of environment variables in the root Makefile

The existing rsync synchronization stuff is retained for now, although
it's probably not necessary for this particular container (a simple bind
mount would probably suffice for the output).
**Release note**:

```
NONE
```
@rmohr

### DIFF
--- a/hack/dockerized
+++ b/hack/dockerized
@@ -14,9 +14,9 @@ while getopts "c:e:i:" opt; do
     c)
         CONTAINER_BASE=$OPTARG
         ;;
-    # Allow specifying an environment variable file for runtime in the container
+    # Allow specifying one or more environment variable files for runtime in the container
     e)
-        ENV_FILE=$OPTARG
+        ENV_FILE="$ENV_FILE $OPTARG"
         ;;
     # Allow specifying a template for building the container
     i)
@@ -51,12 +51,14 @@ fi
 # Build up arguments for passing environment data to containers
 ENV_ARG=
 if [ -n "$ENV_FILE" ]; then
-    FULL_ENV_FILE=${DOCKER_DIR}/${ENV_FILE}
-    if [ ! -f $FULL_ENV_FILE ]; then
-        echo "Environment file $FULL_ENV_FILE not found"
-        exit 1
-    fi
-    ENV_ARG="--env-file=$FULL_ENV_FILE"
+    for e in $ENV_FILE; do
+        FULL_ENV_FILE=${DOCKER_DIR}/$e
+        if [ ! -f $FULL_ENV_FILE ]; then
+            echo "Environment file $FULL_ENV_FILE not found"
+            exit 1
+        fi
+        ENV_ARG="$ENV_ARG --env-file=$FULL_ENV_FILE"
+    done
 fi
 
 # Build the build container
--- a/hack/dockerized
+++ b/hack/dockerized
@@ -1,9 +1,38 @@
 #!/bin/bash
 set -e
 
+usage() {
+    echo "Usage: $(basename $0) [ -c container name] [ -e environment ] [ -i docker image template ]"
+    exit 1
+}
+
+DOCKER_IMG=docker-builder
+ENV_FILE=
+while getopts "c:e:i:" opt; do
+    case $opt in
+    # Allow specifying a named container
+    c)
+        CONTAINER_BASE=$OPTARG
+        ;;
+    # Allow specifying an environment variable file for runtime in the container
+    e)
+        ENV_FILE=$OPTARG
+        ;;
+    # Allow specifying a template for building the container
+    i)
+        DOCKER_IMG=$OPTARG
+        ;;
+    *)
+        usage
+        ;;
+    esac
+done
+shift $((OPTIND - 1))
+
+# Fill in the environment
 source $(dirname "$0")/common.sh
 
-DOCKER_DIR=${KUBEVIRT_DIR}/hack/docker-builder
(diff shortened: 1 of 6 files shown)


ovirt-infra:

rmohr:
> The existing rsync synchronization stuff is retained for now, although
> it's probably not necessary for this particular container (a simple bind
> mount would probably suffice for the output).

That would have been my choice too. Sadly we had severe issues with that. It is extremely hard to keep the permissions in sync and in general the bindmount performance was extremely bad when doing builds. So the rsync method is the second generation. A pattern which we borrowed from k8s.

For me, right now I see a constant overhead of about 2 seconds compared to running the commands outside of the docker builder. How bad is it for you? What I was trying to do, is having one docker container which I only have to start once per makefile target and use each makefile target as a flat entrypoint (no interdependencies between Makefile targets).

> Environment variables for each container are specified in a per-
> container file.

That alone would be a good improvement. Maybe we could have one env file per "task" instead of different containers?

benwatcisco:
@rmohr : I changed this to go back to using a single container, but allowing task-specific environment files.  I retained the docker generalization, though, since we may find it useful in the future.  Let me know what you think.

benwatcisco:
@rmohr : Any more thoughts on this?  This, or something like it, is needed for the ImagePullPolicy stuff to really work.

I don't really understand the Travis CI build failure.  Maybe a stale build environment?

hack/dockerized
```diff
@@ -77,7 +119,7 @@ _rsync --delete "rsync://root@127.0.0.1:${RSYNCD_PORT}/build/cluster/examples/"
 if [ "$SYNC_VENDOR" = "true" ]; then
     _rsync --delete "rsync://root@127.0.0.1:${RSYNCD_PORT}/vendor" "${VENDOR_DIR}/"
 fi
-# Copy the build output out of the container, make sure that _out exactly matches the build result
+# Copy the build output out of the container
 if [ "$SYNC_OUT" = "true" ]; then
-    _rsync --delete "rsync://root@127.0.0.1:${RSYNCD_PORT}/out" ${OUT_DIR}
```
rmohr: The `delete` here is very important. If you remove it you will see stale artifacts in `_out`. In `_out` we only have compiled or generated things, so we need to keep it in sync with the builder docker-volume.

rmohr:
> @rmohr : Any more thoughts on this? This, or something like it, is needed for the ImagePullPolicy stuff to really work.

Can you explain why it can't really work otherwise? I really think that you can just follow the example on how other providers make their configs to the build system available, like I mentioned in #1314 where you try to add an external provider.

I am pretty sure that we don't need any of these changes here to make the external provider work properly.

hack/dockerized
benwatcisco: I agree that something is needed to avoid having stale artifacts, but the `delete` directive is too broad a brush given the chosen file layout.  For example, if I run `make manifests`, this deletes all the binaries, and vice-versa.  Maybe a more targeted copy would be better, but I agree that maybe this is becoming too complicated.  The important goal here that I see is to get the appropriate environment variables into the containers.

benwatcisco:
@rmohr : The external provider stuff works fine.  The ImagePullPolicy stuff does not, though, because the necessary environment variable (`IMAGE_PULL_POLICY`) does not propagate into the build container.  This is orthogonal to any provider logic.

hack/dockerized
rmohr: > I agree that something is needed to avoid having stale artifacts, but the delete directive is too broad a brush given the chosen file layout. For example, if I run make manifests, this deletes all the binaries, and vice-versa.

Not really. As said, the actual building and generating is happening inside the docker container and the artifacts are stored in a docker-volume: Check `docker volume ls` for a volume called `kubevirt`.

If you run `make`, the binaries are compiled and stored in the `_out` directory inside the volume. Then the `rsync --delete` at the end ensures that `_out`  in the volume is in sync with `_out` outside (there is a little bit more to also get rid of binaries not part of that branch, but don't want to make it too complicated here). 

When we next run `make manifests`, the manifests are also generated inside the container and put in the `_out` folder in the docker volume. Then `rsync --delete` makes again sure that `_out` inside the volume is in sync with `_out` outside. **Note:** This does not clean any of the binaries which were built before. Also they would not even be copied out again by `rsync`. Only stale stuff would disappear from the `_out` directory outside. That is exactly what we want.

Finally, regarding to syncing, building or generating it does not make much difference if you run

  `hack/dockerized "hack/build.sh && hack/manifests.sh `

or
 
  `make build && make manifests`

the only difference is the overhead of starting the container and a very small rsync overhead for detecting changes.

You can also have a look yourself into the volume. Run `make && make manifests`, then run `hack/dockerized.sh /bin/sh` to enter the  container. That will directly drop you inside the volume mirror location. You will find binaries and the artifacts in `_out` there.

benwatcisco:
@rmohr : I've rebased this and removed anything controversial.  I don't believe the TravisCI failure is real, but am unable to start it manually.  What do you think?

ben-skyportsystems:
@cynepco3hahue : Yes.  I will bring this up to speed for some assurance that it will be addressed.

benwatcisco:
This other ID of mine agrees :)

ovirt-infra:

PR #1353, opened 2018-07-20
Added k8s 1.11.0 provider

### LINKED ISSUE #1265: Add support for k8s-1.11.0 in CI
Our CI should run KubeVirt on Kubernetes 1.11.0

slintes:

Added a k8s 1.11.0 provider

Fixes #1265

**Release note**:
```release-note
Added a Kubernetes 1.11.0 provider
```


cynepco3hahue:
@slintes Can you also add it as part of standard-ci and maybe some other changes like:
- change default environment for k8s 1.10 to 1.10.4 instead of 1.10.3 and remove for 1.10 at all
- add 1.11 release and dev

rmohr:
@slintes I don't think that this is already enabled on the ci servers, so you will not see any results for the 3.11 cluster.

slintes:
oh ok, I understood that all configuration which is needed is in the automation directory. Too bad that we also need sth on the server side :/

rmohr:
oh @slintes for standardci you are right. @ifireball do you see something wrong in the config?

ifireball:
@rmohr , @slintes - need to also change the stdci.yaml file to match the filename changes you've made and invoke the new test suit

slintes:
@ifireball I don't see such a file in our automation folder and assumed that we use defaults this way, and substages are "detected" just by the script names. Is that file on the server? If so, would it make sense to put into the source repo too?

ifireball:
@slintes its in the root directory

slintes:
@ifireball ooohhh..... thanks!

slintes:
retest this please

slintes:
retest this please

cynepco3hahue:
I can see that `make cluster-up` fails on 
```
[preflight] Some fatal errors occurred:

	[ERROR FileContent--proc-sys-net-bridge-bridge-nf-call-iptables]: /proc/sys/net/bridge/bridge-nf-call-iptables does not exist

	[ERROR FileContent--proc-sys-net-ipv4-ip_forward]: /proc/sys/net/ipv4/ip_forward contents are not set to 1

[preflight] If you know what you are doing, you can make a check non-fatal with `--ignore-preflight-errors=...`

provisioning node node01 failed

```

automation/test.sh
```diff
@@ -36,12 +36,16 @@ if [[ $TARGET =~ openshift-.* ]]; then
...
-elif [[ $TARGET =~ .*-1.9.3-.* ]]; then
-  export KUBEVIRT_PROVIDER="k8s-1.9.3"
+elif [[ $TARGET =~ .*-1.10.4-.* ]]; then
+  export KUBEVIRT_PROVIDER="k8s-1.10.4"
 else
-  export KUBEVIRT_PROVIDER="k8s-1.10.3"
+  export KUBEVIRT_PROVIDER="k8s-1.11.0"
 fi
 
+# network settings for kubeadm preflight check (ignore failures on old CI)
```
rmohr: Something like this needs to be part of the image itself. What you do here is not even executed inside of the cluster  images. Also since we have full control over the qemu-vms, it should never benecessary to add something like `|| true` to account for different qemu-vms.

rmohr:
> In an earlier version I did that without || true. That worked for standard-ci, tests were executed, but failed on the old kubevirt ci with permission denied.

You were modifying the ci host in the case of standardci.

rmohr:
@ifireball are builds cancelled on standardci if we push new updates to the PR? We do that on the other infra to save some resources.

rmohr:
@slintes I see that the 1.10.X cluster on kubevirtci tries to do this:

```
+ make cluster-up
./cluster/up.sh
Downloading .....
Error: No such image: kubevirtci/k8s-1.11.0@sha256:4f34bda5a290bbc6dc7ed5dd1677e13b356cbe6a5cd00eeda35dbdbd0e000ca1
```

slintes:
retest this please

cynepco3hahue:
@slintes I see both `1.11` runs have a lot of failures, can you please check it(check it on standard CI)?

cynepco3hahue:
retest this please

rmohr:
@slintes: there are about 12 permanent errors in the 1.11 test run. Can you have a look?

slintes:
looking at the storage failures at the moment... something is not working with the PVs on k8s 1.11, but I don't understand yet what's missing. I see the PV with status `available`, but the PVC is `pending` with message `waiting for first consumer to be created before binding`, and virt-launcher complains about `0/1 nodes are available: 1 node(s) didn't find available persistent volumes to bind`...

cynepco3hahue:
I inclined to agree, I just afraid that in this case, we will drop 1.11 issues to the backlog, and will get more unstable CI than usual(it also possible that 1.11 has bugs, in this case we will need to revert changes) :worried:

rmohr:
> I inclined to agree,

I disagree. My assumption was actually that kubevirt was already tested against the built cluster locally and the tests worked locally. Apparently not. It does not make sense at all to test something broken.

> I just afraid that in this case, we will drop 1.11 issues to the backlog, and will get more unstable CI than usual(it also possible that 1.11 has bugs, in this case we will need to revert changes)

Exactly.

@slintes I disable  the 1.11 build for now. We can enable it once it passes locally for you.

slintes:
Fixed the access tests... `kubectl auth ...` returns with an exit code != 0 / error now when the response is `no`. Will continue with the storage tests tomorrow.

tests/access_test.go
```diff
@@ -38,7 +38,7 @@ var _ = Describe("User Access", func() {
 		tests.BeforeTestCleanup()
 	})
 
-	Describe("With default kubevirt service accounts", func() {
+	FDescribe("With default kubevirt service accounts", func() {
```
j-griffith: Left you Focus decorator here

slintes:
> Other than the Focus decorator on your access_test.go it LGTM

ups, thanks! :)

slintes:
storage should also work now
let's see if there are more non flaky tests left...

slintes:
hooray :) `check-patch.k8s-1.11.0-dev.el7.x86_64 — Test is successful`
but why does kubevirtci not start?

slintes:
retest this please

rmohr:
@slintes can it be that the PVC for the windows tests also needs to be updated?

rmohr:
retest this please

PR #1364, opened 2018-07-24
Allow creating VMIs with minimal guest overhead to achieve a higher VMI density per node

### LINKED ISSUE #1305: Support for memory over commitment
The expectation is that all of the hypervisors memory can be allocated to VMs - if wanted.
This obviously ignores that the hypervisor itself has memory requirements.

Thus this is about memory over commitment.

Example: Host has 32GB of ram, thus I can fit 8 VMs with 4GB of RAM each on it.

rmohr:

This PR allows configuring VMs with a small guest overhead. That involves two changes:

1. Headless VMIs (no video ram overhead, but also no VNC)
2. Allow overcommiting for the per-guest overhead.

Ad 1.) Allow headless VirtualMachineInstances:

 * Add autoattachGraphicsDevice with default to true
 * If set to false, don't consider video ram in vm overhead
 * If set to false, don't allow VNC connections to the vm

Ad 2.) Allow overcommiting guest overhead

 * Add overcommitGuestOverhead to the API
 * Let it default to false to keep the current behaviour
 * If set, only add the guest overhead to memory limits (if present)

Fixes #1305

**Release note**:

```release-note

```

### DIFF
--- a/pkg/api/v1/schema.go
+++ b/pkg/api/v1/schema.go
@@ -176,6 +176,9 @@ type Devices struct {
 	Interfaces []Interface `json:"interfaces,omitempty"`
 	// Whether to attach a pod network interface. Defaults to true.
 	AutoattachPodInterface *bool `json:"autoattachPodInterface,omitempty"`
+	// Wheater to attach the default graphics device or not.
+	// VNC will not be available if set to false. Defaults to true.
+	AutoattachGraphicsDevice *bool `json:"autoattachGraphicsDevice,omitempty"`
 }
 
 // ---
--- a/pkg/api/v1/schema.go
+++ b/pkg/api/v1/schema.go
@@ -114,6 +114,10 @@ type ResourceRequirements struct {
 	// Valid resource keys are "memory" and "cpu".
 	// +optional
 	Limits v1.ResourceList `json:"limits,omitempty"`
+	// Don't ask the scheduler to take the guest-management overhead into account. Instead
+	// put the overhead only into the requested memory limits. This can lead to crashes if
+	// all memory is in use on a node. Defaults to false.
+	OvercommitGuestOverhead bool `json:"overcommitGuestOverhead,omitempty"`
 }
 
 // CPU allows specifying the CPU topology.
--- a/pkg/api/v1/schema_swagger_generated.go
+++ b/pkg/api/v1/schema_swagger_generated.go
@@ -39,8 +39,9 @@ func (DomainPresetSpec) SwaggerDoc() map[string]string {
 
 func (ResourceRequirements) SwaggerDoc() map[string]string {
 	return map[string]string{
-		"requests": "Requests is a description of the initial vmi resources.\nValid resource keys are \"memory\" and \"cpu\".\n+optional",
-		"limits":   "Limits describes the maximum amount of compute resources allowed.\nValid resource keys are \"memory\" and \"cpu\".\n+optional",
+		"requests":                "Requests is a description of the initial vmi resources.\nValid resource keys are \"memory\" and \"cpu\".\n+optional",
+		"limits":                  "Limits describes the maximum amount of compute resources allowed.\nValid resource keys are \"memory\" and \"cpu\".\n+optional",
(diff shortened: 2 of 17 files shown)


pkg/api/v1/schema.go
```diff
@@ -114,6 +114,10 @@ type ResourceRequirements struct {
 	// Valid resource keys are "memory" and "cpu".
 	// +optional
 	Limits v1.ResourceList `json:"limits,omitempty"`
+	// Don't ask the scheduler to take the guest-management overhead into account. Instead
+	// put the overhead only into the requested memory limits. This can lead to crashes if
+	// all memory is in use on a node. Defaults to false.
+	OvercommitGuestOverhead bool `json:"overcommitGuestOverhead,omitempty"`
```
cynepco3hahue: What use case does it have?
can you also add `// +optional`

pkg/api/v1/schema.go
```diff
@@ -176,6 +180,9 @@ type Devices struct {
 	Interfaces []Interface `json:"interfaces,omitempty"`
 	// Whether to attach a pod network interface. Defaults to true.
 	AutoattachPodInterface *bool `json:"autoattachPodInterface,omitempty"`
+	// Wheater to attach the default graphics device or not.
+	// VNC will not be available if set to false. Defaults to true.
+	AutoattachGraphicsDevice *bool `json:"autoattachGraphicsDevice,omitempty"`
```
cynepco3hahue: please add `// +optional`

pkg/api/v1/schema_swagger_generated.go
```diff
@@ -80,10 +81,11 @@ func (Firmware) SwaggerDoc() map[string]string {
...
 	return map[string]string{
-		"disks":                  "Disks describes disks, cdroms, floppy and luns which are connected to the vmi.",
-		"watchdog":               "Watchdog describes a watchdog device which can be added to the vmi.",
-		"interfaces":             "Interfaces describe network interfaces which are added to the vm.",
-		"autoattachPodInterface": "Whether to attach a pod network interface. Defaults to true.",
+		"disks":                    "Disks describes disks, cdroms, floppy and luns which are connected to the vmi.",
+		"watchdog":                 "Watchdog describes a watchdog device which can be added to the vmi.",
+		"interfaces":               "Interfaces describe network interfaces which are added to the vm.",
+		"autoattachPodInterface":   "Whether to attach a pod network interface. Defaults to true.",
+		"autoattachGraphicsDevice": "Wheater to attach the default graphics device or not.\nVNC will not be available if set to false. Defaults to true.",
```
stu-gott: s/Wheater/Whether

pkg/virt-launcher/virtwrap/api/converter_test.go
```diff
@@ -745,6 +745,42 @@ var _ = Describe("Converter", func() {
 			Expect(err).To(BeNil())
 		})
 	})
+
+	Context("graphics and video device", func() {
+
+		table.DescribeTable("should check autoattachGraphicsDevicse", func(autoAttach *bool, devices int) {
```
stu-gott: autoAttachGraphicsDevices?

pkg/virt-controller/services/template_test.go
```diff
@@ -630,3 +693,13 @@ func MakeFakeConfigMapWatcher(configMaps []kubev1.ConfigMap) *cache.ListWatch {
 	}
 	return cmListWatch
 }
+
+func True() *bool {
```
stu-gott: These functions are probably useful in general. Is there a reasonable module we could share them from?

pkg/api/v1/schema.go
rmohr: It allows a higher vm density. If you expose 32gb of ram for the k8s scheduler, you can schedule 8x4gb vms to the node. Under the assumptions that not all vms use all their memory at the same time you vms should mostly work normal then.

cynepco3hahue: Thanks for the explanation I forgot that we have `-realtime mlock=off`.

cynepco3hahue:
Merged by mistake wrong PR, @rmohr can you please respond to comments under another PR? Or do I need to revert this one?

PR #1373, opened 2018-07-25
Provide a useful message when VMI's time out

stu-gott:

Provide a user readable message for all variants of `waitForVMIStart`. Without this the only error message is "false != true".

**Release note**:
```release-note
NONE
```


cynepco3hahue:
Looks like currently, we have some problems with our CI, but standard CI looks almost good

PR #1377, opened 2018-07-26
Set reporting style for STDCI to plain HTML

ifireball:
Make the links posted on tested PRs point to a static HTML file that
loads much faster then Blue Ocean.

### DIFF
--- a/stdci.yaml
+++ b/stdci.yaml
@@ -9,3 +9,5 @@ sub-stages:
 runtime_requirements:
   support_nesting_level: 2
   isolation_level: container
+reporting:
+  style: stdci


rmohr:
@ifireball can it be that the test results do not show both `stdout` and `stderr`?

ifireball:
@rmohr given that we're seeing some {{bash -x}} output (which goes to stderr) I don't think that is likely..

The command that runs the script is:

    "$script_path" 2>&1 | tee -a $logdir/stdout_stderr.log

so it seems to me both channels should end up in the log file

rmohr:
@ifireball the command looks like that ,but when I look at https://jenkins.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/1382//artifact/check-patch.k8s-1.10.3-dev.el7.x86_64/mock_logs/script/stdout_stderr.log, I don't see the output from the test binary itself. If you compare it with logs: https://jenkins.ovirt.org/blue/organizations/jenkins/kubevirt_kubevirt_standard-check-pr/detail/kubevirt_kubevirt_standard-check-pr/1382/pipeline, then you see that the test output itself is missing.

ifireball:
@rmohr very strange, is there some FD manipulation going on? Because we are seeing _some_ of the output that goes to stderr (the lines that start with '+' go to stderr too)

rmohr:
Not that I know of. Btw. our jenkins jobs do the same. Also the current jenkins on kubevirt-ci does pretty much what you do:

```
bash automation/test.sh 2>&1 | tee ${WORKSPACE}/k8s-1.10.4-release-console.log
```

ifireball:
@rmohr so far I've not managed to reproduce this - I've tried plain scripts, scripts called from scripts and scripts called from makefiles called from scripts

Also it seems it may not just be stderr, looks like all output gets cutoff shortly after `./cluster/build.sh` is launched.

Could it be that the 'tee' command is killed by one of scripts somewhere?

Any further idea how I can proceed in analysing the issue?

ifireball:
created a tracker ticket because this seems to be taking a lot of my time: https://ovirt-jira.atlassian.net/browse/OVIRT-2386

managed to make it happen when running kubevirt tests on my laptop, but still no idea why it happens, and so far attempts to change the way the log is collected did not solve the issue.

ifireball:
I thing I finally have an Idea what is going on - since the logs are generated inside the mock environment, but collected outside of it, they needed to be in a place that is share between the inside and outside environments. We used a hidden temporary directory under the source tree for that, only is seems that something in the build process deletes any files with open descriptors to them found in the source tree.

(I know its not just deleting all the files since I can see that other log files that we paced there before test.sh started running are still intact when it finishes).

This deletion causes tee to keep writing to the older inode. I'm not sure why some of the data still ends up in a new inode, now that I think of it, this sounds like some log rotation process.

In any case I'm working on a patch to change our code so that the source tree is no longer trusted for storing unfinished logs.

cynepco3hahue:
ci test please

PR #1381, opened 2018-07-27
Basic guest vCPUs to pCPUs pinning

vladikr:
Replying on k8s CPUManager feature, this PR implements a basic guest vCPUs to a dedicated pods' pCPUs pinning.

In general, k8s CPU manager will pin  pods' CPUs under certain conditions:

-  pods' QOS is Guaranteed
   - resources requests and limits are equal
   - all containers in the pod express cpu and memory resources
- requested number of CPUs in integer 

Once pods' cpus are dedicated and pinned, virt-launcher can pin VMs' vCPUS to the pods pCPUS

This PR also provides a mechanism that once `dedicatedCpuPlacement` policy has been requested, it will "enforce" the above conditions on the pod spec, so the k8s cpu manager will pin pods' cpus and reject inconsistent requirements.

The user may either express the vCPU requirements using cpu.cores to resources.[requests, limits].cpu
togethre with cpu.dedicatedCpuPlacement = true
```
spec:
  domain:
    cpu:
      cores: 2
      dedicatedCpuPlacement: true
    resources:                                                                                                                                                                                                     
      limits:
        memory: 2Gi
```

OR

```
spec:
  domain:
    cpu:
      dedicatedCpuPlacement: true
    resources:                                                                                                                                                                                                     
      limits:
        cpu: 2                                                                 
        memory: 2Gi
```



What's currently missing: 
- [x] Functional tests to verify scheduling
- [x] validation webhooks
- [x] Defaults CPU/memory defaults for hooks

Functional tests are dependant on https://github.com/kubevirt/kubevirtci/pull/37

**Release note**:

```release-note
Pinning vCPU to a set of pCPUs is now possible
```


pkg/registry-disk/registry-disk.go
```diff
@@ -111,7 +112,10 @@ func GenerateContainers(vmi *v1.VirtualMachineInstance, podVolumeName string, po
 			volumeMountDir := generateVolumeMountDir(vmi, volume.Name)
 			diskContainerName := fmt.Sprintf("volume%s", volume.Name)
 			diskContainerImage := volume.RegistryDisk.Image
-
+			resources := kubev1.ResourceRequirements{}
+			resources.Limits = make(kubev1.ResourceList)
+			resources.Limits[kubev1.ResourceCPU] = resource.MustParse("50m")
```
yanirq: How was this value decided ?

vladikr:
> I like that we can introduce CPU pinning, but in my opinion it very limited, we still do not give any possibility to a user to control exact CPU pinning.

Thanks for the review!

Yea, it's because we have to rely on the k8s CPU manager, which is limited...
Hope it's improve in the future or we'll come up with a better approach.
There was a survey from k8s that samples which cpu related features people are interested in:
https://docs.google.com/forms/d/e/1FAIpQLSest47hzQyKO_ENEbAsyZz_78aPCDpuToHjsyg7n75fF70ArA/viewanalytics

pkg/registry-disk/registry-disk.go
vladikr: hmm, just needed to start somewhere.
This container only mounts a path. - I'll make it a global default variable

pkg/virt-handler/vm.go
```diff
@@ -776,6 +783,33 @@ func (d *VirtualMachineController) heartBeat(interval time.Duration, stopCh chan
 	}
 }
 
+func (d *VirtualMachineController) addNodeCpuManagerLabel() {
+	entries, err := filepath.Glob("/proc/*/cmdline")
```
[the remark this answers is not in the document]
vladikr: I don't 20k processes running anywhere. on my machine it with ~1k processes it took the following, but that also includes go compilation and getting its' stuck up:
```
time go run  trym22.go 

real    0m0.143s
user    0m0.115s
sys     0m0.058s
```

I don't believe it should be a problem, however, this is just a workaround at the moment, until k8s addresses the issue: https://github.com/kubernetes/kubernetes/issues/66525

pkg/virt-controller/services/template.go
```diff
@@ -251,6 +251,33 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
+		vmi.Spec.NodeSelector[v1.CPUManager] = "true"
+
+		if resources.Limits == nil {
+			resources.Limits = make(k8sv1.ResourceList)
+		}
+		if cpus.Cores != 0 {
+			resources.Limits[k8sv1.ResourceCPU] = *resource.NewQuantity(int64(cpus.Cores), resource.BinarySI)
+			resources.Requests[k8sv1.ResourceCPU] = *resource.NewQuantity(int64(cpus.Cores), resource.BinarySI)
+		} else {
+			if cpuLimit, ok := resources.Limits[k8sv1.ResourceCPU]; ok {
```
cynepco3hahue: I just wonder if we need to check that `resources` can be converted to an integer because a user can to specify
- `cpus.DedicatedCPUPlacement`
- `resources.Limits` equal to something like 200m

In this case I think you will fail in converter, so better to add this check to validation webhook

vladikr: Absolutely. I've added several validations but didn't post yet as I'm still struggling with the functional tests.

Also, I wanted to explicitly check, somewhere in the SyncVM that the pods QOS is `Guaranteed`, but didn't figure out yet how to get to pods informer at that stage.

vladikr:
Added validations, still working on the functional tests.

pkg/virt-api/validating-webhook/validating-webhook_test.go
```diff
@@ -959,6 +959,81 @@ var _ = Describe("Validating Webhook", func() {
...
+			vmi.Spec.Domain.Resources.Limits = k8sv1.ResourceList{
+				k8sv1.ResourceCPU: resource.MustParse("2"),
+			}
+			causes := ValidateVirtualMachineInstanceSpec(k8sfield.NewPath("fake"), &vmi.Spec)
+			Expect(len(causes)).To(Equal(1))
+			Expect(causes[0].Field).To(Equal("fake.domain.cpu.dedicatedCpuPlacement"))
+		})
+		It("should reject specs with non-integer cpu limits values", func() {
+			vmi.Spec.Domain.Resources.Limits = k8sv1.ResourceList{
+				k8sv1.ResourceCPU: resource.MustParse("800m"),
```
yanirq: maybe add another negative test with a fraction value (cpu = 1.5) not in milli format.

yanirq: @vladikr if it does so i do :)

dcritch:
I've been testing CPU manager as is with kubevirt and its working pretty good. Trying to understand what this change will improve. Seems like we can specify particular CPUs? Currently with CPU manager, the VM is pinned on whatever CPU the process happens to start on, but can't be specified.

vladikr:
> I've been testing CPU manager as is with kubevirt and its working pretty good. Trying to understand what this change will improve. Seems like we can specify particular CPUs? Currently with CPU manager, the VM is pinned on whatever CPU the process happens to start on, but can't be specified.

Yes, we are not planning to let users specify on which exact cpu their vCPUs will be pinned. Instead, we will reply on the k8s CPU manager to find the right place where the instance can be pinned and pin the VMs' vCPUs to the dedicated pCPUs of the pod.

I've updated the description of the PR. Hope it'll clarify things.

vladikr:
standard-ci log shows that os-3.10 passed too:
```
12:22:36 [check-patch.openshift-3.10-release.el7.x86_64] • [SLOW TEST:14.652 seconds]
12:22:36 [check-patch.openshift-3.10-release.el7.x86_64] Configurations
12:22:36 [check-patch.openshift-3.10-release.el7.x86_64] /root/go/src/kubevirt.io/kubevirt/tests/vmi_configuration_test.go:47
12:22:36 [check-patch.openshift-3.10-release.el7.x86_64]   VirtualMachineInstance with CPU pinning
12:22:36 [check-patch.openshift-3.10-release.el7.x86_64]   /root/go/src/kubevirt.io/kubevirt/tests/vmi_configuration_test.go:489
12:22:36 [check-patch.openshift-3.10-release.el7.x86_64]     with cpu pinning enabled
12:22:36 [check-patch.openshift-3.10-release.el7.x86_64]     /root/go/src/kubevirt.io/kubevirt/tests/vmi_configuration_test.go:498
12:22:36 [check-patch.openshift-3.10-release.el7.x86_64]       should be scheduled on a node with running cpu manager
12:22:36 [check-patch.openshift-3.10-release.el7.x86_64]       /root/go/src/kubevirt.io/kubevirt/tests/vmi_configuration_test.go:511
```

pkg/virt-controller/watch/vmi.go
```diff
@@ -380,6 +384,15 @@ func (c *VMIController) sync(vmi *virtv1.VirtualMachineInstance, pods []*k8sv1.P
 		return nil
 	} else if isPodReady(pod) && !isPodOwnedByHandler(pod) {
 		pod := pod.DeepCopy()
+
+		// fail vmi creation if CPU pinning has been requested but the Pod QOS is not Guaranteed
+		podQosClass := qos.GetPodQOS(pod)
```
cynepco3hahue: Can we check it on the validation stage?

vladikr:
> I still have some questions, but overall looks good

Thanks for the review :)

pkg/util/hardware/hw_utils.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"strconv"
+	"strings"
+)
+
+const CPUSET_PATH = "/sys/fs/cgroup/cpuset/cpuset.cpus"
+
+// Parse linux cpuset into an array of ints
+// See: http://man7.org/linux/man-pages/man7/cpuset.7.html#FORMATS
+func ParseCPUSetLine(cpusetLine string) (cpusList []int, err error) {
```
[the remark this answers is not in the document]
vladikr: yep, will do.

pkg/virt-controller/services/template.go
```diff
@@ -44,6 +44,7 @@ const KvmDevice = "devices.kubevirt.io/kvm"
 const TunDevice = "devices.kubevirt.io/tun"
 
 const CAP_NET_ADMIN = "NET_ADMIN"
+const CAP_SYS_NICE = "SYS_NICE"
```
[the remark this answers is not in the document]
vladikr: Yes. Libvirt does the pinning of vcpu threads to specific cores. 
CPU manager, pins the pod to specific pcpus. We then let libvirt pin the vcpus only to this set. So no, it cannot pin any other pcpus that were not allocated by the cpu manager for the virt-launcher.

pkg/virt-controller/services/template.go
```diff
@@ -253,6 +254,30 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
+			vmi.Spec.NodeSelector = make(map[string]string)
+		}
+		vmi.Spec.NodeSelector[v1.CPUManager] = "true"
+
+		if resources.Limits == nil {
+			resources.Limits = make(k8sv1.ResourceList)
+		}
+		cores := vmi.Spec.Domain.CPU.Cores
+		if cores != 0 {
+			resources.Limits[k8sv1.ResourceCPU] = *resource.NewQuantity(int64(cores), resource.BinarySI)
```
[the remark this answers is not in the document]
vladikr: Yes, this is a requirement for QOS guaranteed.

pkg/virt-controller/watch/vmi.go
vladikr: I think it's important to the check this when the pod is running.
After the validation there are changes in the template, then there can limitrange added to the namespace and all kind of admission controllers.
I think we shouldn't start the vm if we didn't get the right QOS.

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -509,6 +509,95 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
...
+		if limitsCPU > 0 && limitsCPU*1000 != spec.Domain.Resources.Limits.Cpu().MilliValue() {
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeFieldValueInvalid,
+				Message: "provided resources CPU limits must be an interger",
+				Field:   field.Child("domain", "resources", "limits", "cpu").String(),
+			})
+		}
+
+		// resources requests must be equal to limits
+		if requestsMem > 0 && limitsMem > 0 && requestsMem != limitsMem {
```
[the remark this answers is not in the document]
cynepco3hahue: hm does it have common QoS for the CPU and memory? I expected that it will have separate QoS for each resource.

pkg/virt-controller/watch/vmi.go
cynepco3hahue: will move it to validation after adding mutating webhook :smile:

pkg/virt-api/validating-webhook/validating-webhook.go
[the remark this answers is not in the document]
vladikr: It's a combination, take a look:
https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed

pkg/util/hardware/hw_utils.go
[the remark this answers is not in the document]
vladikr: Done

fabiand:
@rmohr could you please also take a look

pkg/virt-controller/services/template.go
```diff
@@ -253,6 +254,30 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		})
 	}
 
+	// Handle CPU pinning
+	if vmi.IsCPUDedicated() {
+		// schedule only on nodes with a running cpu manager
+		if vmi.Spec.NodeSelector == nil {
+			vmi.Spec.NodeSelector = make(map[string]string)
+		}
+		vmi.Spec.NodeSelector[v1.CPUManager] = "true"
```
fabiand: We need to document this requirement if dedicated placement is requested.

pkg/virt-api/validating-webhook/validating-webhook.go
```diff
@@ -509,6 +509,95 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
...
+				Message: fmt.Sprintf("%s or %s should be provided",
+					field.Child("domain", "resources", "requests", "memory").String(),
+					field.Child("domain", "resources", "limits", "memory").String(),
+				),
+				Field: field.Child("domain", "resources", "limits", "memory").String(),
+			})
+		}
+
+		// provided CPU requests must be an interger
+		if requestsCPU > 0 && requestsCPU*1000 != spec.Domain.Resources.Requests.Cpu().MilliValue() {
```
davidvossel: are we just looking for the mod of the millivalue here to ensure it is divisible by 1000? If so is this logic the same as what I have below?

```
requestsCPU  > 0 && (spec.Domain.Resources.Requests.Cpu().MilliValue() % 1000) > 0
```

pkg/virt-handler/vm.go
```diff
@@ -786,6 +793,33 @@ func (d *VirtualMachineController) heartBeat(interval time.Duration, stopCh chan
...
+	}
+
+	for _, entry := range entries {
+		content, err := ioutil.ReadFile(entry)
+		if err != nil {
+			log.DefaultLogger().Reason(err).Errorf("failed to set a cpu manager label on host %s", d.host)
+			return
+		}
+
+		if strings.Contains(string(content), "kubelet") && strings.Contains(string(content), "cpu-manager-policy=static") {
```
davidvossel: The kubelet can also be configured via a configfile with --config now. It's possible this cpu manager value won't be in the comandline.

https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/#start-a-kubelet-process-configured-via-the-config-file

pkg/registry-disk/registry-disk.go
```diff
@@ -111,7 +112,14 @@ func GenerateContainers(vmi *v1.VirtualMachineInstance, podVolumeName string, po
 			volumeMountDir := generateVolumeMountDir(vmi, volume.Name)
 			diskContainerName := fmt.Sprintf("volume%s", volume.Name)
 			diskContainerImage := volume.RegistryDisk.Image
-
+			resources := kubev1.ResourceRequirements{}
+			if vmi.IsCPUDedicated() {
+				resources.Limits = make(kubev1.ResourceList)
+				// TODO(vladikr): adjust the correct cpu/mem values - this is mainly needed to allow QemuImg to run correctly
```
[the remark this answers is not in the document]
vladikr: This is how the k8s cpu manager works. It only pins pods that have QOS level `Guaranteed`. According to [1] all containers in the pod must express cpu and memory limits, equal to requests. 

[1] https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed

pkg/virt-controller/services/template.go
```diff
@@ -363,10 +387,19 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 	containers = append(containers, container)
 
 	for i, requestedHookSidecar := range requestedHookSidecarList {
+		resources := k8sv1.ResourceRequirements{}
+		// add default cpu and memory limits to enable cpu pinning if requested
```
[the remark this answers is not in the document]
vladikr: This is needed to get POD QOS Guaranteed, as above.

pkg/virt-api/validating-webhook/validating-webhook.go
vladikr: I did this a first and then looked at what cpu_manager does [1]
I think this is needed to make sure that `.Value()`  didn't round the original value.

[1] https://github.com/NVIDIA/kubernetes/blob/master/pkg/kubelet/cm/cpumanager/policy_static.go#L201

pkg/virt-handler/vm.go
vladikr: Maybe, I'm not sure. 
However, in the worst case, this node won't be labeled and no harm will be done.
But I've looked at openshift and they also add the values to the command line, while the configuration of the cpu manager is in a configmap.

In general, we will need to get a more deterministic answer where cpu manager is running or not. I've opened a bug https://github.com/kubernetes/kubernetes/issues/66525

I don't see a risk here. It's not perfect and we should improve it.

vladikr:
> For Item 2, unless there's a reliable way for us to determine static placement is enabled on the kubelet, I'm inclined for us to punt on trying to influence scheduling at the moment. Maybe in the future we can get the kubelet to set a label on the node to indicate the cpu manager policy.

If we won't influence the scheduling we will end up being placed on a node where virt-launcher container cpus are not pinned at all. Then there is a risk we'll pin vcpus to the same pcpus. 
I don't see a benefit it this.

On the other hand - if we can't determine whether the cpu manager is running we simply don't label the node so workloads with dedicated cpu requirement won't be scheduled there. This is much less risky from my pov.

pkg/registry-disk/registry-disk.go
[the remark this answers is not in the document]
davidvossel: umm, this has some implications for how tightly bin VMs can be bin packed on nodes with dedicated placement. 

in OpenStack, if we have a 16 core compute node, is it possible to schedule two 8 core VMs each pinned to a subset of cores on the node while letting the infrastructure processes float between the cores?

In a similar scenario using k8s and cpu manager, if we have a 16 core compute node available to run VMs and tried to schedule two VMs that require 8 cores and dedicated placement, only 1 of those VMs would be able to run if any sidecar containers are present, right? Because the sidecar containers would require cores to remain in the shared pool on that node which makes them ineligible for static placement. 

Wouldn't we rather both 8 core VMs to be capable of being scheduled to the compute node, and let the infrastructure related side car container processes float around between cores?

If so, I don't see a way around this using cpu manager today. We need to bring up some of these advanced use cases k8s so we can hopefully support them in the future.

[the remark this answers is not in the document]
vladikr: So, not exactly.

Yes, sidecars are an overhead, but a quantitative one.

A container that doesn't express an integer in their requests/limits doesn't get pinned.
So, virt-launcher will get pinned, but the sidecars won't and they will still float as they express a certain amount of cycles and not a full core. In that case, for 2 vms requesting 8 cores each we will need to keep one core for the overhead. 
Openstack need to have a larger reservation for all the infrastructure resources - here we can calculate it. When OpenStack has 16 cores for scheduling it means that there is one core (at least) already reserved. 

I have validations that make sure that virt-launchers cpu request is an integer, is a dedicated cpu is required.

vladikr:
> What would you think about feature gating the dedicated cpu placement feature? This would require users to toggle some value in kubevirt-config in order to utilize dedicated cpu placement. We'd be depending on the user to properly setup the static placement policy on the kubelet. I guess the down side to that is that it assumes a homogeneous cluster where every node has static placement enabled... Then again that's kind of the position everyone who wants to use static placement right now is placed in anyway right now.
> 
> With the feature gate, we could eventually have it default to true once we have the detection sorted out.

I understand... How can I configure it per each node?
Isn't it the same as manually labeling the nodes?

I think we can get away with a documentation. I think it's easy enough to get `describe nodes` and see the labels - if one is missing where it supposed to me, the admin can add it.

In parallel, I'll push a PR to kubelet to get the cpu manager label the node when it's running.

fabiand:
> I'm just worried about maintaining this detection logic. We're not really using a supported interface to detect the feature is enabled or not so it's probably going to be something that breaks in the future.

I think we need to do it interim, but as we are not the only ones consuming CPU manager, I think we can file an kubernetes issue to get this somehow addressed and allow consumerst to discover cpu manager availability.

Either directly in k8s or through i.e. node-feature-discovery.

pkg/registry-disk/registry-disk.go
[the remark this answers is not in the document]
rmohr: Since, right now we did not look into mount-propagation for registry-disks, we can also convert the registry disk containers to init containers. As such, they can always request and consume less or equal resources than the final VMI. At least we can right now let the registry disk container exit once it is done, so that the resources are not consumed and other best effort pods can make use of it.

rmohr:
> Don't introduce a dedicatedCpuPlacement and follow the (awkward) way of how the cpu manager wants to specify if you want full cores or if floating is ok.

Forget the part about not having `dedicatedCpuPlacement` where i talk about removing it. It would only make sense if we can always assume that we want to pin cpus if we request cores for the VMI alone, which we don't necessarily want.

michalskrivanek:
not really. it's not even possible to do that in oVirt as when you do pinning and you're specifying the exact CPUs on host to use it's really not much more work to map them to vCPUs. For performance you typically want to match the topology and map 1:1 (also NUMA topology) as otherwise it doesn't make much sense to bother with pinning.

rmohr:
> not really. it's not even possible to do that in oVirt as when you do pinning and you're specifying the exact CPUs on host to use it's really not much more work to map them to vCPUs. For performance you typically want to match the topology and map 1:1 (also NUMA topology) as otherwise it doesn't make much sense to bother with pinning.

In that light it and the limitations of the cpu manager in mind, could it be good to not have the `dedicatedCpuPlacement` and simply assume that you want to pin the cpus if vcpu and core request match? Maybe even by default change the core size to cpu requests if virt-launcher detects that we have dedicated nodes and fulll cpu cores were requested?

michalskrivanek:
@rmohr yes, I think so. 
So you don't want to reject the request if limits: cpu differ from domain: cpu?

IIUC one drawback still is that you do not really know which CPUs you pin to so you can't really make sure the topology is correct (you don't want to pin a single guest NUMA across 2 host's NUMAs)- is that solvable within kubevirt?

fabiand:
Topology awareness is also being addressed by Kubernetes. My take would be
to wait for it.

rmohr:
> Going forward we will probably need to express how do we place the vCPUs on pCPUs. Do we try to pack it all on a socket or select cores from different sockets. dedicatedCpuPlacement may need to turn into a policy.
> 
> It's not clear to me why we shouldn't have a tunable that clearly express the user intention.

We can. Normally I also prefer it. If we need/want to cover the case where a user wants to request exclusive CPUs but not pin, we also need it. What I was trying to find out is, if it makes sense to not pin the CPUs but still request exclusive CPUs from k8s. Keeping the feedback from @michalskrivanek in mind, it seems like you normally only want to request CPUs if you also want to pin them. Since the CPU manager is not capable of doing anything else than that and there is no topology awareness there right now, it looks like a user would have to specify two times that cpu pinning is desired. Once by setting `dedicatedCpuPlacement` to `true` and once by requesting CPUs as non-decimal values.

If we have `dedicatedCpuPlacement`, then we definitely need some kind of a detection mechanism (like you added for that reason), since we would try to guarantee more than k8s, where basically the admin and the users have to agree if and how to find nodes with static cpu assignment configured. We have to fail where k8s right now proceeds (with potentially less performance) by default.

rmohr:
> I understand... How can I configure it per each node?
> Isn't it the same as manually labeling the nodes?
> 
> I think we can get away with a documentation. I think it's easy enough to get describe nodes and see the labels - if one is missing where it supposed to be, the admin can add it.

What if I disable the CPU manager on a node? I  would at least expect that the label goes away.

If we feature-gate the complete cpu detection, it would be ok for me to use it, since users can then work around it if they e.g. use a config map to configure the kubelet. Once there is a reliable way to detect the cpu manager we can remove the feature-gate. But then I would really want it to be fully implemented, including dynamic labelling (add + remove) on the node.

At the end that is the main point why I am hesitating a little bit with the `dedicatedCpuPlacement` flag. For our API flag to really work we will need the CPU detection. I was basically hoping that we can buy us some time by providing exactly the guarantees which the CPU manager provides through the cluster. We can also say that if the feature-gate is disabled, label `X` needs to be added to nodes, or the VMIs will not be schedulable.

rmohr:
> I've added the CPU manager to the feature gates.
> But honestly, I feel like it doesn't solve anything really.
> 

Well you can say that we have limitations X. If you are withing that boundary (e.g. not setting up the kubelet via a config map), then you can enable the feature gate and you get automatic labeling for now. Further, if the cpu manager comes up with a canonical way it is easier to deprecate/remove our way.

> Those who will enable the feature may still face the problem that if CPU manager is removed from kubelet, the label will not be removed from the node.

Yes, in my opinion, if you want to keep `dedicatedCpuPlacement`, you will have to make it a full implementation (that's orthogonal to the feature gate request). It will also have to remove the label.

> 
> Those who didn't enable the feature or didn't request dedicated CPU resources (with the feature enabled) have no effect either way.

Also here, if you want to keep `dedicatedCpuPlacement`, we will have to make it fully consistent. In case the auto-detection of the CPU manger is disabled via the feature-gate, still make the label mandatory. But in this case admins have to set the label mandatory.

pkg/registry-disk/registry-disk.go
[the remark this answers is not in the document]
davidvossel: @rmohr good point. We should move registry disks to be init containers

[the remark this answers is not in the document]
rmohr: Btw. some people started some time ago to implement an oci based volume type (so runtime independent). It would pull and attach containers through the CDI interface. That might be something for us on the long run: https://github.com/kubernetes-csi/drivers/issues/85.

vladikr:
The latest update re-labels the node when cpu manager is not running. Added a functional test to verify.

vladikr:
ci test please

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -140,6 +140,14 @@ func (l *LibvirtDomainManager) SyncVMI(vmi *v1.VirtualMachineInstance, useEmulat
 	// Set defaults which are not coming from the cluster
 	api.SetObjectDefaults_Domain(domain)
 
+	// Adjust guest vcpu config. Currenty will handle vCPUs to pCPUs pinning
```
cynepco3hahue: hm why do you not set VCPU pinning under `converter.go`?

rmohr: Yeah. Strictly speaking  the cpu discovery information should be placed in the converter context and the converter should do the conversion.

pkg/virt-controller/watch/vmi.go
rmohr: @vladikr I think that just checking the `status.qosClass` field on the pod should be good enough.

pkg/util/qos/qos.go
```diff
@@ -0,0 +1,100 @@
...
+
+func isSupportedQoSComputeResource(name v1.ResourceName) bool {
+	return supportedQoSComputeResources.Has(string(name))
+}
+
+// GetPodQOS returns the QoS class of a pod.
+// A pod is besteffort if none of its containers have specified any requests or limits.
+// A pod is guaranteed only when requests and limits are specified for all the containers and they are equal.
+// A pod is burstable if limits and requests do not match across all containers.
+func GetPodQOS(pod *v1.Pod) v1.PodQOSClass {
```
rmohr: I think you can remove the whole file and just checkt `status.qosClass` on the pod.

pkg/virt-controller/services/template.go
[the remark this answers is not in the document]
rmohr: that we need SYS_NICE is kind of bad but good enough for now ...

pkg/virt-launcher/virtwrap/util/cpu_utils.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	defer file.Close()
+	scanner := bufio.NewScanner(file)
+	if scanner.Scan() {
+		cpuset = scanner.Text()
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+	cpusList, err := hardware.ParseCPUSetLine(cpuset)
```
rmohr: In general, to be more secure, I would prefer a smart sparse list here. Since the cpusetline is not something the user can set  this should be fine. Let's make sure that we never expand something like this if it comes from a user.

pkg/virt-handler/vm.go
```diff
@@ -782,10 +786,46 @@ func (d *VirtualMachineController) heartBeat(interval time.Duration, stopCh chan
 				return
 			}
 			log.DefaultLogger().V(4).Infof("Heartbeat sent")
+			// Label the node if cpu manager is running on it
+			// This is a temporary workaround until k8s bug #66525 is resolved
+			featuregates.ParseFeatureGatesFromConfigMap()
+			if featuregates.CPUManagerEnabled() {
+				d.updateNodeCpuManagerLabel()
```
rmohr: Can you do both patch operations in one combined operation? We have to think about scale here too. For instance have a look at https://github.com/kubevirt/kubevirt/pull/1409/files#diff-9b18fe582fd0ecdc06b3fb28c3ad98b2R49. There, for different reasons I do something similar with constructiong a json-patch which conditionally adds or removes labels.

tests/vmi_configuration_test.go
```diff
@@ -530,5 +535,144 @@ var _ = Describe("Configurations", func() {
...
+				By("adding a cpumanger=true lable to a node")
+				nodes, err := virtClient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: v1.CPUManager + "=" + "false"})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(nodes.Items).To(HaveLen(1))
+
+				node := &nodes.Items[0]
+				node, err = virtClient.CoreV1().Nodes().Patch(node.Name, types.StrategicMergePatchType, []byte(fmt.Sprintf(`{"metadata": { "labels": {"%s": "true"}}}`, v1.CPUManager)))
+				Expect(err).ToNot(HaveOccurred())
+				// wait one heartbeat
+				time.Sleep(60 * time.Second)
```
rmohr: Is this a leftover? The `Eventually` should also wait long enough to handle the worst waiting-case ...

pkg/virt-launcher/virtwrap/manager.go
vladikr: > hm why do you not set VCPU pinning under converter.go?
I thought that all of the cpu discovery doesn't belong in the converter that basically should convert from api to domain xml, only setting the CPUTune belongs there, but it was too small. But you are right... it should be there.

tests/vmi_configuration_test.go
vladikr: done

pkg/util/qos/qos.go
vladikr: Done

pkg/virt-handler/vm.go
vladikr: Yea, I've seen it, but wasn't sure as I'm not familiar with it.
Another thing is that maybe we should turn features gate into a list instead of a string.

pkg/virt-controller/services/template.go
```diff
@@ -374,10 +398,19 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
+		if vmi.IsCPUDedicated() {
+			resources.Limits = make(k8sv1.ResourceList)
+			resources.Limits[k8sv1.ResourceCPU] = resource.MustParse("200m")
+			resources.Limits[k8sv1.ResourceMemory] = resource.MustParse("64M")
+		}
 		containers = append(containers, k8sv1.Container{
 			Name:            fmt.Sprintf("hook-sidecar-%d", i),
 			Image:           requestedHookSidecar.Image,
 			ImagePullPolicy: requestedHookSidecar.ImagePullPolicy,
+			Resources:       resources,
```
davidvossel: @vladikr This super minor, but maybe worth thinking about.

Since we're defaulting all the sidecar and registry disk container resource limits to 200m/64M, would it make more sense to just iterate over the Pod's container list at the end of this function and force the 200m/64M limit on any container that isn't the compute container?

That would also enforce that limits are required in the future as more container types are potentially added to the VMI pod.

pkg/virt-handler/vm.go
davidvossel: regardless of the featuregate, it looks like updateNodeCpuManagerLabel is still only setting the label based on if the cpu-manager-policy=static in the kubelet command args.

I was under the impression we'd us the feature flag to automatically set the cpu manager label for now, and then transition to a dynamic detection of whether cpumanager is enabled per a node once we have a consistent way of doing so. Essentially we were forcing the user to both have a homogeneous cluster with respects to cpu manager and explicitly enable the feature gate.

By enabling the feature gate right now, users will be acknowledging that it is up to them to properly enable cpu manager on all nodes in order to use dedicated cpu placement. In the future when we eventually remove the feature gate, it will be because we can make this decision accurately for the user. 

is that reasonable for now?

vladikr:
> Regarding to combining the patch, I would prefer it, as said but maybe I can live with two patches right now. Json patch will fail if the label is already there, so you would need a node informer to check on the cache if the label is already there. Considering that this is behind a feature gate for now, I will not insist on doing it.

ah, ok. I'll remove the last patch. Weird that it didn't fail in the test...

vladikr:
@rmohr , @davidvossel raised a concern here: https://github.com/kubevirt/kubevirt/pull/1381#discussion_r215368217

What's your take on it?

pkg/virt-handler/vm.go
davidvossel: I don't think we can reasonably support the cpu manager auto detection logic today. There are already multiple ways to configure the kubelet outside of the cmdline args [1]. 

The feature gate is a way for us to remove KubeVirt from attempting to perform any sort of detection logic until k8s provides a supported interface for us to use in order to determine cpu manager is enabled. With the feature gate, we aren't taking responsibility for safety. That burden is on the user until we have a supportable way to auto detect cpu manager. The feature gate is saying, **"by enabling this, you understand that cpu manager must be enabled on all nodes in the cluster or bad things will happen"**.  Users willing to accept that condition have to explicitly enable the feature gate to use dedicated cpu placement. 

Once we have the supportable interface for determining cpu manager's policy, we can turn the feature gate on by default in a future kubevirt release and no longer require users to explicitly enable this support on our end. The feature gate is a stop gap for us until we have the k8s support we need to properly detect cpu manager on every node. 

1. https://kubernetes.io/blog/2018/07/11/dynamic-kubelet-configuration/

vladikr: I didn't see anywhere that enabling the cpu manager is supported via dynamic configuration.
All of the existing configuration show that on k8s[1] you have to update kubelet params manually and on Openshift[2] you must update the node-config-compute config map.
However, what this does in reality, is that each node runs a background script that reads the config map and updates kubelet params manually and restarts it.
According to this, we should find these parameters on the kubelet process.

[1] https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/
[2] https://docs.okd.io/latest/scaling_performance/using_cpu_manager.html

vladikr: I'm going to stop arguing. :)
If the consensus is to remove it, I'll remove it.
I'm not attached to much, it's just I think it's safer.

vladikr: > The feature gate is a way for us to remove KubeVirt from attempting to perform any sort of detection logic until k8s provides a supported interface for us to use in order to determine cpu manager is enabled. With the feature gate, we aren't taking responsibility for safety. That burden is on the user until we have a supportable way to auto detect cpu manager. The feature gate is saying, "by enabling this, you understand that cpu manager must be enabled on all nodes in the cluster or bad things will happen". Users willing to accept that condition have to explicitly enable the feature gate to use dedicated cpu placement.

@davidvossel I think I'm missing something. Thinking about this again. 
The feature gate is global to all nodes. 
According to your suggestion, by enabling it we will label all the nodes - even if there is no cpu manager running. There is no way to disable it for a certain host.
In that case, the user will have to run the cpu manager on all the nodes. This is not straightforward at all IMHO.
On the other hand, why do we need the feature gate at all in this case? We can simply let the users label the relevant nodes manually and not to force them to run the cpu manager on all hosts.

However, this is exactly what we have now. If the gate is enabled - we will do auto detection.
If the feature gate is disabled, don't detect but let the user label the nodes where he knows cpu manager is running.

davidvossel: > the user will have to run the cpu manager on all the nodes. 

yep, that's the condition the user would have to accept when enabling the feature gate. The user would be committing to a homogeneous cluster with respects to cpu manager policy.  

> On the other hand, why do we need the feature gate at all in this case? We can simply let the users label the relevant nodes manually and not to force them to run the cpu manager on all hosts.

@vladikr I'd be fine with that. So in this case, we'd remove the feature gate and the autodetect logic. Users then have to set the label on nodes that they have cpu manager enabled on.

If we do this, then we can follow up on a discussion around whether or not to auto detect cpu manager afterwards. This lets us decouple the part we're in disagreement about from this PR so we can merge the pieces we're in agreement about. Essentially it gives us a supportable stop gap that can segue into more sophisticated detection logic later.

fabiand:
So, I see the discussion is now about how we enable this feature.

1. Feature gate only
2. Feature gate and _auto_ node labeling
2.a Feature gate and _manual_ node labeling

Approach 1 is simple, approach 2 more complex, but makes it pretty usable in heterogeneous clusters.
To me both solutiosn have pros and cons.

The thing is that we are going in cycles, and the feature is not landing because of the unclarity about how to enable it. I'm not sure this helps anybody.

We could trust the author - and for surely refine these details over time.

because I don#t have a strong opinion, I'm trusting the author.

fabiand:
Key is that we get into the habit to be willing to change if we see that it turns out to be bad.

davidvossel:
> 1. Feature gate only
> 2. Feature gate and auto node labeling
> 2.a Feature gate and manual node labeling

no, that's not the discussion. I don't want feature gate and auto node labeling or feature gate and manual node labeling.

The discussion is. 

1. auto labeling using the kubectl cmdline pid parsing.
2. feature gate with forced labeling which puts the burden on the user to ensure cpu manager is enabled on all nodes.
3. manual labeling which puts the burden on the user to ensure nodes with cpu manager enabled have a special label. 

> We could trust the author - and for surely refine these details over time.

the argument isn't what's more or less complex, it's about what we can support. My concern is centered around how autodetection works. We have to come to an consensus that the autodetection logic is supportable in order to include this.

davidvossel:
> Key is that we get into the habit to be willing to change if we see that it turns out to be bad.

sure, but lets not prematurely merge something that there's an active discussion going on about. We aren't at a stalemate. This is still evolving.

rmohr:
> the argument isn't what's more or less complex, it's about what we can support. My concern is centered around how autodetection works. We have to come to an consensus that the autodetection logic is supportable in order to include this.

My take is this: If we want autodetection, I want it behind a feature-gate because it is not necessarily supportable (as you said). We can also go without auto-detection, then we don't need the feature-gate. But: Since the user is explicitly saying that the VMI should be pinned, I would still want that admins label the nodes and that we only schedule to nodes with that label. Putting the whole feature (not just the auto-labeling) behind a feature-gate and leave the detection up to a consensus between the users and admins works for me too (e.g. well known label or homogenous cluster).

vladikr:
Tests are failing because of the admission webhook...
No idea what's wrong there, trying to figure it out.
" reason="admission webhook \"virtualmachineinstances-update-validator.kubevirt.io\" denied the request:  update of VMI object is restricted" msg="reenqueuing VirtualMachineInstance default/vmi-ephemeral" "

davidvossel:
> Tests are failing because of the admission webhook...
No idea what's wrong there, trying to figure it out.
" reason="admission webhook "virtualmachineinstances-update-validator.kubevirt.io" denied the request: update of VMI object is restricted" msg="reenqueuing VirtualMachineInstance default/vmi-ephemeral" "

Something in this PR is attempting to update the VMI's DomainSpec. That's only allowed in the mutation webhook.

pkg/virt-controller/services/template.go
```diff
@@ -264,6 +265,30 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		})
 	}
 
+	// Handle CPU pinning
+	if vmi.IsCPUDedicated() {
+		// schedule only on nodes with a running cpu manager
+		if vmi.Spec.NodeSelector == nil {
+			vmi.Spec.NodeSelector = make(map[string]string)
+		}
+		vmi.Spec.NodeSelector[v1.CPUManager] = "true"
```
davidvossel: @vladikr this is probably what the validation code is complaining about.  This needs to be moved to the defaulting code in the new mutation webhook. 

Also, I'm not sure it's safe to mutate the vmi in this "render" function. I think the vmi pointer might be directly from the informer cache

vladikr: Yeah, thanks a lot! I've seen that this is happening because I'm updating the node selector but wasn't sure how does the mutation webhook work and where do I need to update this...

Thanks, I'll move there.

rmohr: ok, that should not be necessary. Simply add the CPUManager label to the pod only. This would be just one of  more cases which we have, where we are helping the scheduler to make some requests which we made by our VMI api (in this  case, we want pinning) schedulable. Or do we think that this needs to have that much transparency?

vladikr: At some point, I thought that it's better to show it, but you're right.
Updated.

vladikr:
ci test please

vladikr:
ci test please

Issue #1375, opened 2018-07-26
[RFC] [WIP] Evolve vm presets into virt profiles


fromanirh:
/kind enhancement

**What happened**:
We would like to have [virt profiles](https://github.com/nertpinx/virt-manager/pull/1) in kubevirt.
This issue maps the gaps between the current implementation of presets in kubevirt (considering commit 35862a65) and the aforementioned virt profiles design document.

**What you expected to happen**:
A much desired property of the implementation of virt profiles is to have the changes implied by profiles reflected back in the vm spec - IOW, the profiles implementation should not make hidden
changes to the vm definition that the user can't see.
To achieve this we can evolve the vm presets to cover all the features of virt profiles

* allow to express all possible settings - this has to be evaluated on a case-by-case basis

* allow to exclude some vm features - e.g. require the absence of some devices, not their presence

* allow to set priority between presets - specify ordering between them: patch proposed in https://github.com/kubevirt/kubevirt/pull/1383

* handle conflicts between presets: allow partial conflicts? RFC patch in https://github.com/kubevirt/kubevirt/pull/1390

* ~~allow to specify what to do in case of conflicts - reject the VM creation? partial application?~~ currently KubeVirt rejects the VM if one or more preset could not be applied, should be good enough.

**Open questions:**

* How to express backend-related settings in the presets/profiles? see [frontend/backend split](https://github.com/kubevirt/kubevirt/blob/master/docs/vm-configuration.md#design-implications) and the [current mapping in kubevirt](https://github.com/kubevirt/kubevirt/blob/master/docs/vm-configuration.md#mapping-rules-for-mapping-libvirt-xml-to-the-kubevirt-api)

**Related Issues:**

https://github.com/kubevirt/kubevirt/issues/749
https://github.com/kubevirt/kubevirt/issues/805 - previous discussion of the same topic of https://github.com/kubevirt/kubevirt/pull/1383
https://github.com/kubevirt/kubevirt/issues/1208
https://github.com/kubevirt/kubevirt/issues/1278
https://github.com/kubevirt/kubevirt/issues/1327

fromanirh:
/cc @michalskrivanek @nertpinx @fabiand @stu-gott for discussion

Implementation wise, we are still looking for a generic solution reusable from other projects, like cockpit, and maybe maybe even oVirt? Thus, I think it's fine to reuse some kubevirt components granted that the solution can be easily deployed standalone, outside kubevirt.

fromanirh:
[...]

PR #1383, opened 2018-07-27
presets: allow to sort by annotations

fromanirh:
We would like to let the operator have control on the order of which
the presets are applied. To do so, we introduce an optional "priority"
annotation, that can be used to enforce the order.

No change is expected, thus the order of application is not specified,
when no annotations are used.

Partially addresses https://github.com/kubevirt/kubevirt/issues/1375

**Release note**:
```release-note
Allow to annotate presets with a priority, so the operator can control the order on which the presets are applied to VirtualMachineInstances.
```

### DIFF
--- a/pkg/virt-controller/watch/preset.go
+++ b/pkg/virt-controller/watch/preset.go
@@ -23,6 +23,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"sort"
+	"strconv"
 	"time"
 
 	k8sv1 "k8s.io/api/core/v1"
@@ -52,6 +54,7 @@ type VirtualMachinePresetController struct {
 
 const initializerMarking = "presets.virtualmachines." + kubev1.GroupName + "/presets-applied"
 const exclusionMarking = "virtualmachineinstancepresets.admission.kubevirt.io/exclude"
+const priorityMarking = "virtualmachineinstancepresets.admission.kubevirt.io/priority"
 
 func NewVirtualMachinePresetController(vmiPresetInformer cache.SharedIndexInformer, vmiInitInformer cache.SharedIndexInformer, queue workqueue.RateLimitingInterface, vmiInitCache cache.Store, clientset kubecli.KubevirtClient, recorder record.EventRecorder) *VirtualMachinePresetController {
 	vmii := VirtualMachinePresetController{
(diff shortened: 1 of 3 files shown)


pkg/virt-controller/watch/preset.go
```diff
@@ -182,7 +185,46 @@ func listPresets(vmiPresetInformer cache.SharedIndexInformer, namespace string)
 		result = append(result, *vmi)
 	})
 
-	return result, err
+	if err != nil {
+		return result, err
+	}
+	return sortPresets(result), nil
```
stu-gott: Unfortunately more logic is needed than just sorting. Please refer to the conversation in https://github.com/kubevirt/kubevirt/issues/805. Logic elsewhere in this module ends up defeating this.

In the function `mergeDomainSpec` the first call is to `checkMergeConflicts`. If two presets conflict with each other's settings, the entire set of presets gets rejected. Because only orthogonal presets can currently even be added, ordering doesn't matter. Because we're changing the rules to allow presets to conflict (implied by the fact that we're asserting priority order), you almost certainly want to remove that check and adjust the rules in all the associated tests.

fromanirh: Thanks @stu-gott for raising this point, this is exactly the discussion I was hoping to have.

In order to support virt-profiles, conflicts (more precisely, overlaps) should be allowed to some extent.

We can re-evaluate how we handle conflicts between presets - the abstract logic is pretty much the same.
Right now just bail out on overlap: the portion of the spec being updated by the preset should either be blank, or look
like the preset wants it to be. From another angle, the portion of the spec should match exactly what the preset wants.

This means that the preset wants to enforce a "equals to" relation.

To have profiles, we want a "at least to" relation. Few examples:

When applying a profile:
- if the existing specs requests 4 cores and the preset says 2, there is no conflict (4 cores >= 2 :)), so we can go ahead
- if the existing specs requests a cpu model and the preset says another, we can either check if the CPUs are compatible (libvirt has APIs for that, and the check is not host-dependent, we are just comparing abstract models) or raise the conflict; or raise anyway the conflict if the CPUs are not compatible
- if the existing specs has one SPICE device and the preset requires one graphic device, there is no conflict

stu-gott: So you're suggesting that what we're going to need is not a one-size-fits all approach to overlaying presets, but more of a fine-grained approach based on what makes sense for each data type. e.g. numeric "minimum requested" sort of logic can apply to number of CPUs but not type.

fromanirh: @stu-gott in a nutshell, yes. IMHO *IFF* we want to allow overlaps, *then* we should check  case-by-case. Some things may be allowed, some things cannot.

fromanirh:
after long discussions, turns out we don't want to go in this direction after all

PR #1385, opened 2018-07-27
Virtiorng

### LINKED ISSUE #1260: Add virtio-rng support
> As a user I want to be able to provide my guest with good random numbers. One way to achieve this is by using a virtio RNG device.

Design: https://docs.google.com/document/d/1G75lPm0dCf3ZcFnG6NJpUDZgTaQnFKKM7_glpXzH9uc/edit#

This issue is about adding a virtio rng device type including it's knbos, and two configuration profiles which reflect the oVirt and OpenStack configuration.

The device can be added by a user on demand (thus is _not_ automatically added to each VM).

The device should expose all the detailed settings you can configure on the the virtio-rng device.
And in addition it must expose a `profile` field which will set the knobs (i.e. backend, rate period and byets) according to profiles derived from oVirt and OpenStack.
A user can eitehr set the knbos himself, or choose from these two options.

The backend should be `/dev/random` if available or `dev/urandom` as a fallback.

# API

```yaml
kind: VMI
spec:
  domain:
    devices:
      rng:
        profile: ovirt  # DEFAULT
        # OR
        profile: openstack
        # OR
        rate: 
          period: 2000
          bytes: 1234
      …
```

libvirt reference. https://libvirt.org/formatdomain.html#elementsRng

Open Questions
- Is hardware RNG slower? Wouldn’t it be better to use /dev/urandom?
- If we know explicitly that a hardware RNG is present, we should request it (device plugins?), otherwise don’t request it and use /dev/random
  - This can be a follow up a) expose hwrng via DP, add virt controller awareness for hwrng devies
- How can we schedule nodes?
- How do we accommodate device plugins (unlimited devices?)
- A daemonset might need to prepare something?

petrkotas:
This PR introduces the virtio RNG device support for the virtual machine. This PR also limits itself purely to implement the API part of the virtio RNG.
The changes are made in convertor to the XML as well as in the kubernetes API. 

This PR does not solve the hardware RNG, support for windows or prepare any presents for the virtio rng. This issues will be tacled in separate PRs to keep everything simple.

Fixes #1260 

**Release note**:

```release-note
Virtio RNG allows exposing random device from host machine to the guest machine.
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/1385)


michalskrivanek:
> or prepare any presents for the virtio rng

sure, let's leave that to Santa

pkg/virt-launcher/virtwrap/api/schema_test.go
```diff
@@ -61,6 +61,10 @@ var exampleXML = `<domain type="kvm" xmlns:qemu="http://libvirt.org/schemas/doma
     <watchdog model="i6300esb" action="poweroff">
       <alias name="mywatchdog"></alias>
     </watchdog>
+    <rng model="virtio">
+      <rate period="2000" bytes="1234"></rate>
+      <backend model="random">/dev/random</backend>
```
michalskrivanek: why using /dev/random when the recommended one is urandom?

petrkotas: Only tests, anything would do really. Even `North Pole` would be a great choice here.

michalskrivanek: I think it's best to use a relevant example

petrkotas: Indeed you are right. Fixed that.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -241,6 +241,36 @@ func Convert_v1_Watchdog_To_api_Watchdog(source *v1.Watchdog, watchdog *Watchdog
...
+	// default rng model for KVM/QEMU virtualization
+	rng.Model = "virtio"
+
+	// default backend model, random for /dev/random and /dev/urandom
+	rng.Backend = &RngBackend{
+		Model: "random",
+	}
+
+	// the source used to get the entropy
+	if source.Source == "" {
```
cynepco3hahue: it must be under validation webhook or we need to set it to some default

pkg/virt-launcher/virtwrap/api/schema.go
```diff
@@ -561,16 +562,29 @@ type Ballooning struct {
...
-}
-
 type Watchdog struct {
 	Model  string `xml:"model,attr"`
 	Action string `xml:"action,attr"`
 	Alias  *Alias `xml:"alias,omitempty"`
 }
 
-// TODO ballooning, rng, cpu ...
+type Rng struct {
```
cynepco3hahue: Please set descriptions at least on structs

cluster/examples/vmi-preset-virtiorng.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: kubevirt.io/v1alpha2
```
cynepco3hahue: I thought we had the plan to add it by default, why do we need preset for it?

pkg/virt-launcher/virtwrap/api/converter.go
petrkotas: I have provided the default in the `converter.go` since other defaults are also present here. 

As for the validation, it does not make any sense to validate the `source` field, it can be any valid device/path. The rates are set as `uint32` and the default validator handle negative case.

cluster/examples/vmi-preset-virtiorng.yaml
petrkotas: This is just example present if you want to create one. I find it handy, but can be removed.

petrkotas:
@cynepco3hahue Thanks for review, done.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -427,6 +457,24 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 		domain.Spec.Devices.Watchdog = newWatchdog
 	}
 
+	if vmi.Spec.Domain.Devices.Rng != nil {
```
michalskrivanek: so how do I explicitely disable virtio-rng if I do not want to see it in VM?

cynepco3hahue: @michalskrivanek I remember we had the discussion about it, what the reason that you do not want rng device at all?

michalskrivanek: @cynepco3hahue we did indeed, and AFAICT the outcome was that we want it by default but opting-out is important as well. There are OSes/arches where this device doesn't work/doesn't make sense

cynepco3hahue: @michalskrivanek Works for me, thanks for clarification.

pkg/api/v1/schema.go
```diff
@@ -758,3 +762,18 @@ type PodNetwork struct {
...
+	// Source sets the host source used to provide entropy
+	Source string `json:"source"`
+	// RateBytes limit the amount of bytes that the VM can read from entropy source
+	// per period
+	// +optional
+	RateBytes *uint32 `json:"rateBytes,omitempty"`
+	// RatePeriod sets the length of a read period
+	// +optional
+	RatePeriod *uint32 `json:"ratePeriod,omitempty"`
+}
```
vladikr: Are you planning to remove the RateBytes and RatePeriod in the next iteration? 
I thought we've agreed not to expose RateBytes and RatePeriod to the users as it introduces a security issue.

petrkotas:
The thing is, I am still unsure whether the YAML is pure API or it is "the
golden way" for the user to communicate.

My take is, YAML is just an API and what is exposed is the user GUI. If
this is not the case, the rate limiting will be removed, for sure.

How should I treat the YAML? Thanks.

senior7515:
@petrkotas  - so question w.r.t RDSEED and RDRAND - https://wiki.qemu.org/index.php/Features/VirtIORNG

so if i passthrough the CPU to the guest and I know my CPU has RDRAND, i'd like to disable virtio-rng device. since qemu will use the `/proc/cpuinfo | grep -i rdrand` effectively.

petrkotas:
@cynepco3hahue Finished the work on this. Would you mind giving me a review once more? Thanks a lot.

vladikr:
@petrkotas I think you need to rebase and call `make generate` again to resolve the conflicts. Thanks!

pkg/api/v1/schema.go
```diff
@@ -758,3 +762,13 @@ type PodNetwork struct {
 	// Default 10.0.2.0/24 if not specified.
 	VMNetworkCIDR string `json:"vmNetworkCIDR,omitempty"`
 }
+
+// Rng represents the random device passed from host
+// ---
+// +k8s:openapi-gen=true
+type Rng struct {
+	// Enabled defines whether use or do not use Virtio Rng
+	Enabled bool `json:"enabled"`
```
vladikr: I wonder whether we need this flag or can we just get away with not setting `Rng`?
If we set it, it becomes enabled otherwise disabled.

petrkotas:
@vladikr @cynepco3hahue I did the rebase.

pkg/api/v1/schema.go
michalskrivanek: I think people need to finally realize that minimal spec is useless. That's why we have OS-based presets, and in future even a better solution - virtuned assisted values/defaults.

Vladik's suggestion is cleaner, and makes most sense once we have virtuned.
For the interim we can as well just add explicit rng: to current presets.

petrkotas: @michalskrivanek the option is still required for the GUI to have the feedback that something changed and reflect in some user click tool.

However, taking into account your feedback, I propose relaxed version, where you do nothing to opt-in. And explicitly define `disabled: true`, when you do not want RNG added.

@michalskrivanek @senior7515 @vladikr Can you agree on that? The PR is already updated.

pkg/api/v1/schema.go
```diff
@@ -765,3 +769,15 @@ type PodNetwork struct {
...
+// Rng represents the random device passed from host
+// ---
+// +k8s:openapi-gen=true
+type Rng struct {
+	// Disabled defines whether use or do not use Virtio Rng
+	// +optional
+	Disabled bool `json:"disabled"`
+	// Source sets the host source used to provide entropy
+	// +optional
+	Source string `json:"source"`
```
cynepco3hahue: I think it must be `omitempty`.

pkg/virt-launcher/virtwrap/api/converter.go
cynepco3hahue: does libvirt can work with any provided path?

pkg/virt-launcher/virtwrap/api/schema.go
```diff
@@ -573,16 +574,37 @@ type Ballooning struct {
...
+// Rng represents the source of entropy from host to VM
+type Rng struct {
+	// Model attribute specifies what type of RNG device is provided
+	Model string `xml:"model,attr"`
+	// Backend specifies the source of entropy to be used
+	Backend *RngBackend `xml:"backend,omitempty"`
+}
+
+// RngRate sets the limiting factor how to read from entropy source
+type RngRate struct {
```
cynepco3hahue: I think you do not specify it, libvirt will set defaults one, so at least for now we do not need to introduce it.

vladikr:
I'm still not convinced why do we need the `disabled` filed, not because of a minimal spec, but simply this is how the rest of the features are getting enabled/disabled.
It should be pretty straightforward if a struct is present - it's enabled otherwise - not, or am I missing something?

manifests/generated/vm-resource.yaml
```diff
@@ -127,6 +127,15 @@ spec:
                                 required:
                                 - name
                               type: array
+                            rng:
+                              properties:
+                                enabled:
```
fabiand: Why do we need the boolean?

manifests/generated/vm-resource.yaml
```diff
@@ -127,6 +127,15 @@ spec:
...
+                                  type: boolean
+                                source:
```
fabiand: This is a security risk.

What would happen if you point this to /etc/my/secret/cert?

petrkotas:
@fabiand ah sorry about that. I saw that the generated files have not been generated. They are now.

To explain the attribute `disabled`. We mutually agreed to have a default of RNG added when not present. This means, that if the section is removed to have it disabled, the defaulter will add it again.
That lead me to introduce the `disabled` switch to have option to disable the RNG.

@vladikr am I handling the defaulter right? I did what every other default section was doing in our code.

michalskrivanek:
> We mutually agreed to have a default of RNG added when not present.

I think that later on we agreed not to do it that way, and rather require the presence (see vladik's comment above)

+1 or not adding it by default. It's really simple to add "rng" to every relevant OS preset. We shouldn't be adding anything silently if we can avoid that, it always looks ugly (see autoattachGraphicsDevice)

petrkotas:
@michalskrivanek @vladikr I see that presets can solve the defaulting. However, as it is still not clearly defined what presets are and what are they doing, we still need some sort of defaulting now. Moreover, in the future we will need the ultimate ability to overpower the preset and disable the virtio rng. For instance think of system wide present, where you want to turn this feature for only one VM.

@michalskrivanek @vladikr What is the issue you are seeing? Why you do not want the `disabled` switch?

pkg/api/v1/defaults.go
```diff
@@ -128,6 +134,9 @@ func SetDefaults_VirtualMachineInstance(obj *VirtualMachineInstance) {
 	if obj.Spec.Domain.Machine.Type == "" {
 		obj.Spec.Domain.Machine.Type = "q35"
 	}
+	if obj.Spec.Domain.Devices.Rng == nil {
```
vladikr: let's remove this please, so it won't get enabled by default.

pkg/api/v1/defaults.go
```diff
@@ -111,6 +111,12 @@ func SetDefaults_Firmware(obj *Firmware) {
 	}
 }
 
+func SetDefaults_Rng(obj *Rng) {
+	if obj.Source == "" && !obj.Disabled {
```
vladikr: I think this should just always set the obj.Source = "/dev/urandom" - no need to check if the obj.Source
And SetDefaults_Rng should be called only if vmi.Spec.Domain.Devices.Rng != nil

vladikr:
> For instance think of system wide present, where you want to turn this feature for only one VM.
Right, but can't we just remove the definition of RNG from the spec for that given VM

> @michalskrivanek @vladikr What is the issue you are seeing? Why you do not want the disabled switch?
@petrkotas I'm not against anything. As I mentioned above, I just see why enabling/disabling of this feature should be different from any other feature we have?
If RNG is not added by default, it can be explicitly enabled or disabled. The defaulter should just set the relevant properties - in our case it's just source I think (maybe libvirt sets the source to /dev/urandom by default, then we don't need anything added at all)

Apologise, I confused you with my previous comments.

petrkotas:
@vladikr I understand. I weighted all pros and cons. 
With everything considered I will remove defaulting in the manner of "alway add the rng" and leave only the defaulting for `/dev/urandom`. This way only the raw logic will remain in the API. 

To ensure the right defaults, KubeVirt will have to have some high level logic which handles it. 

 @michalskrivanek @fabiand @vladikr do we agree on this?

fabiand:
@petrkotas if you are saying that:

1. rng will not be added by default
2. the `enabled` attribute will be removed
3. the `source` attribute will be removed

then yes

petrkotas:
@fabiand that is exactly what I meant. Thanks.

petrkotas:
@fabiand @vladikr I made the changes. Can we merge now?

vladikr:
@petrkotas looks like it needs a rebase.

fabiand:
Rebasing sounds good.

petrkotas:
@vladikr @fabiand did rebase. Can merge now :).

fabiand:
Please add a release note @petrkotas - Reviewing now

petrkotas:
@fabiand @vladikr Added the release note.

fabiand:
retest this please

cynepco3hahue:
ci test please

fabiand:
CI fails on your code:

```
[Fail] Configurations VirtualMachineInstance definition with rng [It] should not have the virtio rng device when not present 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_configuration_test.go:367
```

Should I ask to take a look? ;)

petrkotas:
No need, fixed.
Thanks :)
--
s pozdravem,
Petr Kotas

petrkotas:
@fabiand I see the test passing, at least for k8s. Unfortunately openshift timeouts (not due to virtio rng tests). The tests fail finally in later stage.

```
[check-patch.k8s-1.10.4-release.el7.x86_64] SUCCESS! -- 145 Passed | 0 Failed | 0 Pending | 14 Skipped
[check-patch.k8s-1.11.0-release.el7.x86_64] SUCCESS! -- 145 Passed | 0 Failed | 0 Pending | 14 Skipped
[check-patch.windows2016-release.el7.x86_64] SUCCESS! -- 6 Passed | 0 Failed | 0 Pending | 153 Skipped
```

Can we merge now?

petrkotas:
@fabiand ping?

fabiand:
A user guide update?

petrkotas:
@fabiand Yes of course. The user guide is separate repository belonging to kubevirt site, so I it will be separate PR.

petrkotas:
@fabiand @rmohr the user guide was already merged. Is this the right time to merge? The PR is approved, CI is passing and user guide is there.

fabiand:
The failing and not-debuggable CI prvents me from merging this RP.

stu-gott:
ci test please

petrkotas:
@fabiand I went through the CI logs. It seems that the server is simply failing randomly.
```
[check-patch.k8s-1.11.0-release.el7.x86_64] Traceback (most recent call last):
[check-patch.k8s-1.11.0-release.el7.x86_64]   File "/var/lib/stdci/container-slaves/3/home/jenkins/workspace/kubevirt_kubevirt_standard-check-pr/jenkins/scripts/docker_cleanup.py", line 299, in <module>
[check-patch.k8s-1.11.0-release.el7.x86_64]     main()
[check-patch.k8s-1.11.0-release.el7.x86_64]   File "/var/lib/stdci/container-slaves/3/home/jenkins/workspace/kubevirt_kubevirt_standard-check-pr/jenkins/scripts/docker_cleanup.py", line 38, in main
[check-patch.k8s-1.11.0-release.el7.x86_64]     stop_all_running_containers(client)
[check-patch.k8s-1.11.0-release.el7.x86_64]   File "/var/lib/stdci/container-slaves/3/home/jenkins/workspace/kubevirt_kubevirt_standard-check-pr/jenkins/scripts/docker_cleanup.py", line 129, in stop_all_running_containers
[check-patch.k8s-1.11.0-release.el7.x86_64]     _remove_container(client, container)
[check-patch.k8s-1.11.0-release.el7.x86_64]   File "/var/lib/stdci/container-slaves/3/home/jenkins/workspace/kubevirt_kubevirt_standard-check-pr/jenkins/scripts/docker_cleanup.py", line 157, in _remove_container
[check-patch.k8s-1.11.0-release.el7.x86_64]     client.stop(container)
[check-patch.k8s-1.11.0-release.el7.x86_64]   File "/usr/lib/python2.7/site-packages/docker/utils/decorators.py", line 21, in wrapped
[check-patch.k8s-1.11.0-release.el7.x86_64]     return f(self, resource_id, *args, **kwargs)
[check-patch.k8s-1.11.0-release.el7.x86_64]   File "/usr/lib/python2.7/site-packages/docker/api/container.py", line 403, in stop
[check-patch.k8s-1.11.0-release.el7.x86_64]     self._raise_for_status(res)
[check-patch.k8s-1.11.0-release.el7.x86_64]   File "/usr/lib/python2.7/site-packages/docker/client.py", line 174, in _raise_for_status
[check-patch.k8s-1.11.0-release.el7.x86_64]     raise errors.APIError(e, response, explanation=explanation)
[check-patch.k8s-1.11.0-release.el7.x86_64] docker.errors.APIError: 500 Server Error: Internal Server Error ("{"message":"Cannot stop container 920bb0fe50cfb032129d964bdf5320564f85bb9006b2da0c668ff70028b9557d: Cannot kill container 920bb0fe50cfb032129d964bdf5320564f85bb9006b2da0c668ff70028b9557d: rpc error: code = 14 desc = grpc: the connection is unavailable"}")
[check-patch.k8s-1.11.0-release.el7.x86_64] Aborting.
```

Fortunately is passes at some point. The openshift passed this time, at least partially. The list of failing tests:
```
[check-patch.openshift-3.10-release.el7.x86_64] [Fail] VMIlifecycle Delete a VirtualMachineInstance with grace period greater than 0 [It] should run graceful shutdown 
[check-patch.openshift-3.10-release.el7.x86_64] /root/go/src/kubevirt.io/kubevirt/tests/vmi_lifecycle_test.go:1046
[check-patch.openshift-3.10-release.el7.x86_64] 
[check-patch.openshift-3.10-release.el7.x86_64] [Fail] VMIlifecycle Killed VirtualMachineInstance [It] should be in Failed phase 
[check-patch.openshift-3.10-release.el7.x86_64] /root/go/src/kubevirt.io/kubevirt/tests/vmi_lifecycle_test.go:1077
[check-patch.openshift-3.10-release.el7.x86_64] 
[check-patch.openshift-3.10-release.el7.x86_64] [Fail] VMIlifecycle Killed VirtualMachineInstance [It] should be left alone by virt-handler 
[check-patch.openshift-3.10-release.el7.x86_64] /root/go/src/kubevirt.io/kubevirt/tests/vmi_lifecycle_test.go:1104
[check-patch.openshift-3.10-release.el7.x86_64] 
[check-patch.openshift-3.10-release.el7.x86_64] [Fail] Expose Expose service on a VM [BeforeEach] Expose ClusterIP service Should expose a Cluster IP service on a VMI and connect to it 
[check-patch.openshift-3.10-release.el7.x86_64] /root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:27
[check-patch.openshift-3.10-release.el7.x86_64] 
[check-patch.openshift-3.10-release.el7.x86_64] [Fail] Expose Expose service on a VM [BeforeEach] Expose ClusterIP service with string target-port Should expose a ClusterIP service and connect to the vm on port 80 
[check-patch.openshift-3.10-release.el7.x86_64] /root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:27
[check-patch.openshift-3.10-release.el7.x86_64] 
[check-patch.openshift-3.10-release.el7.x86_64] [Fail] Expose Expose service on a VM [BeforeEach] Expose NodePort service Should expose a NodePort service on a VMI and connect to it 
[check-patch.openshift-3.10-release.el7.x86_64] /root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:27
[check-patch.openshift-3.10-release.el7.x86_64] 
[check-patch.openshift-3.10-release.el7.x86_64] Ran 148 of 159 Specs in 3875.137 seconds
[check-patch.openshift-3.10-release.el7.x86_64] FAIL! -- 142 Passed | 6 Failed | 0 Pending | 11 Skipped
```
Luckily none if related to virtioRNG. Can we merge?

fabiand:
The failures don't look virtio rng related and rather like flakiness.

@petrkotas will oyu take care of the CI issues?

PR #1390, opened 2018-07-30
RFC: preset: partially allow overrides

fromanirh:
Change the way how we handle conflcits among presets, parially allowing some overrides.
Some portions of the spec can be overridden - easy to detect compatible or incompatible updates, some others should not be overriden.

Partially addresses https://github.com/kubevirt/kubevirt/issues/1375
Related to https://github.com/kubevirt/kubevirt/pull/1383 - spawned from initial comments
This is a RFC. The PR is reviewable but not complete: I want to see if I can add more tests.

**Release note**:
```release-note
Change the way we detect conflicts between presets, partially allowing overlaps between profiles.
```


fromanirh:
after long discussions, turns out we don't want to go in this direction after all

PR #1392, opened 2018-07-30
1.10.3 prodider doesn't exist anymore so make 1.10.4 default?

mhenriks:
If KUBEVIRT_PROVIDER was not defined, `make cluster-up` failed.  1.10.3 was the default provider.  But it got removed when we added support for Kubernetes 1.11.0.

Fixes #1353

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -19,7 +19,7 @@ function build_func_tests() {
     mv ${KUBEVIRT_DIR}/tests/tests.test ${TESTS_OUT_DIR}/
 }
 
-KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-k8s-1.11.0}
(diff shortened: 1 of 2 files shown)


slintes:
oh, there are 2 places where we define the default provider, in `common.sh` and `config.sh`. In `config.sh` I updated the default to k8s 1.10.4. [0]. Not sure what policy we have for the default version, but it would make sense to use the same in those 2 files (or probably better, prevent having this in 2 files at all).

[0] https://github.com/kubevirt/kubevirt/pull/1353/files#diff-f6bcfd6045e29f8a81030bac5211374aL5

mhenriks:
Yeah, at a minimum, the values should be the same in both files.  No duplication would be ideal if possible.  I guess the biggest question is what the default value should be?

mhenriks:
Looks like everywhere that includes `config.sh` includes `common.sh` before it so it should be safe to remove from `config.sh`

Issue #1325, opened 2018-07-16
Add support for stable (PCI) addresses and introspection


fabiand:
> As a user I expect that my guest devices (and their location) only change if I modify the VM definition.

IOW: The guest ABI should have stable addresses.

This is a mandatory feature for VM identity, and avoiding that certain guests need reactivatio, i.e. Microsoft Windows

ronf:
I need this capability as well, but for disk devices, as the OS I'm running in the virtual machine expects content to appear on specific PCI addresses. More specifically, it's a custom boot loader and it's having trouble finding the images to boot.

When I start up a VM with libvirt directly, it works when I specify something like:

    <disk type='file' device='disk'>
      <driver name='qemu' type='qcow2'/>
      <source file='/var/lib/libvirt/images/my_image.qcow2'/>
      <target dev='vda' bus='virtio'/>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x05' function='0x0'/>
    </disk>

However, with kubevirt the PCI addresses are automatically selected and I end up with something like:

    <disk type='file' device='disk'>
      <driver name='qemu' type='qcow2'/>
      <source file='/var/run/libvirt/kubevirt-ephemeral-disk/registry-disk-data/default/sgva/disk_boot/disk-image.qcow2'/>
      <target dev='vda' bus='virtio'/>
      <address type='pci' domain='0x0000' bus='0x03' slot='0x00' function='0x0'/>
    </disk>

fabiand:
That's interesting - we wondered about the disk case. So far we knew that the serials were the primary guest sided identifier. Interesting to see the use-case where the PCI address is relevant.

ronf:
Yeah - I know it's a bit of an edge case (particularly for disks), but it'd be great if there was a way to specify the address information for any PCI device. I'm not sure yet if I'll run into a similar issue with the NICs, but I suspect I'll need it there too.

fabiand:
@vladikr we also need pciAddresses for network devices, as they influence the interface name.

vladikr:
> @vladikr we also need pciAddresses for network devices, as they influence the interface name.

@fabiand did you mean pciAddresses for disks? #1393 adds a support for setting pciAddresses for network devices.

vladikr:
I'm working on allowing setting pciAddresses for disks.

MarSik:
[...]

PR #1393, opened 2018-07-31
Allow setting vNIC pci address

vladikr:
This will allow users to specify an optional PCI address for network interfaces.
This is particularly needed for users who are concerned with PCI addresses stability and would like to set their own.

Related #1325

**Release note**:
```
Allow users or higher level applications to set a specific PCI address for guest interfaces to maintain stability.

```

### DIFF
--- a/pkg/util/hw_helper.go
+++ b/pkg/util/hw_helper.go
@@ -26,6 +26,7 @@ import (
 
 const PCI_ADDRESS_PATTERN = `^([\da-fA-F]{4}):([\da-fA-F]{2}):([\da-fA-F]{2}).([0-7]{1})$`
 
+// ParsePciAddress returns an array of PCI DBSF fields (domain, bus, slot, function)
 func ParsePciAddress(pciAddress string) ([]string, error) {
 	pciAddrRegx, err := regexp.Compile(PCI_ADDRESS_PATTERN)
 	if err != nil {
--- a/pkg/util/hw_helper.go
+++ b/pkg/util/hw_helper.go
@@ -35,5 +35,5 @@ func ParsePciAddress(pciAddress string) ([]string, error) {
 	if len(res) == 0 {
 		return nil, fmt.Errorf("failed to parse pci address %s", pciAddress)
 	}
-	return res, nil
+	return res[1:], nil
 }
--- a/pkg/util/hw_helper.go
+++ b/pkg/util/hw_helper.go
@@ -1,0 +1,39 @@
+/*
+ * This file is part of the KubeVirt project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2018 Red Hat, Inc.
+ *
+ */
+
+package util
+
+import (
+	"fmt"
+	"regexp"
+)
+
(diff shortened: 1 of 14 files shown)


vladikr:
need to add a functional test.

pkg/util/hw_helper.go
```diff
@@ -0,0 +1,39 @@
...
+package util
+
+import (
+	"fmt"
+	"regexp"
+)
+
+const PCI_ADDRESS_PATTERN = `^([\da-fA-F]{4}):([\da-fA-F]{2}):([\da-fA-F]{2}).([0-7]{1})$`
+
+func ParsePciAddress(pciAddress string) ([]string, error) {
```
phoracek: Please add a documentation for this function, what does it return? Maybe it would be better to return (string, string, string, string, error) and give it proper names.

pkg/util/hw_helper.go
```diff
@@ -0,0 +1,39 @@
...
+	pciAddrRegx, err := regexp.Compile(PCI_ADDRESS_PATTERN)
```
phoracek: Cannot we do it during compilation?

tests/vmi_networking_test.go
```diff
@@ -457,4 +467,18 @@ var _ = Describe("Networking", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 	})
+
+	Context("VirtualMachineInstance with custom PCI address", func() {
+		It("should configure custom Pci address", func() {
+			By("checking eth0 Pci address")
+			testVMI := tests.NewRandomVMIWithEphemeralDiskAndUserdata(tests.RegistryDiskFor(tests.RegistryDiskCirros), "#!/bin/bash\necho 'hello'\n")
```
phoracek: the `echo` is a custom command for the VM? Doesn't that mean that it will immediately exit?

vladikr: I don't think so. Look at other tests that we are running in this file.

pkg/util/hw_helper.go
vladikr: I'd prefer to return the virt-launcher Address struct here, but that creates a circular dependency. If it's not a struct then I'd prefer a list of PCI DSBF fields. I'll add a comment.

pkg/util/hw_helper.go
vladikr: Do what? :D

phoracek: In python we Compile regex beforehand and just use it when a function is called. Not sure if something like that exists here, I guess it does not matter.

tests/vmi_networking_test.go
phoracek: Yea, looks ok then.

vladikr:
/cc @stu-gott @cynepco3hahue @davidvossel @fabiand can you guys please review it? :) 
@phoracek comments were addressed. Don't like to see this code rot. Please let me know if you have any concerns. Thanks!

tests/vmi_networking_test.go
cynepco3hahue: This command only needed for cloud-init, to avoid waiting for the cloud-init source on the VMI start(that can take more than 1 minute)

cynepco3hahue:
ci test please

vladikr:
> I just wonder if the plan is to add PCI Address support to every device class individually, or if you considered to generalize this - to make it applicable to other device classes as well?

My plan was only to allow setting PCI addresses for nics (now) and disks (in another PR) and maybe vgpu in the future. Libvirt also sets PCI addresses for controllers. Hope we are not planning to allow changing this via API.

@fabiand , I think I'm missing something - What did you have in mind? How do you see this generalization? Different addresses need to be set on each device, it's important to control it on a device based level - so I don't see how it can be generalized.

vladikr:
> Maybe we can create additional struct PCIDevices move all relevant devices under it and add it under Devices struct.

Ok, maybe we should. I'm just not sure what's the purpose. Did you think to use it for validations or anything else?

cynepco3hahue:
> Ok, maybe we should. I'm just not sure what's the purpose. Did you think to use it for validations or anything else?

I checked your response to @fabiand and it works for me, so just ignore my comment :smile:

PR #1394, opened 2018-07-31
fix expose example unclear

shiywang:

```release-note
NONE
```

### DIFF
--- a/pkg/virtctl/expose/expose.go
+++ b/pkg/virtctl/expose/expose.go
@@ -76,8 +76,8 @@ virtualmachineinstance (vmi), virtualmachine (vm), virtualmachineinstancereplica
 }
 
 func usage() string {
-	usage := `  # Expose SSH to a virtual machine instance called 'myvm' as a node port (5555) of the cluster:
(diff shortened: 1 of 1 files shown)


shiywang:
@slintes @fabiand PTAL

pkg/virtctl/expose/expose.go
```diff
@@ -77,7 +77,7 @@ virtualmachineinstance (vmi), virtualmachine (vm), virtualmachineinstancereplica
 
 func usage() string {
 	usage := `  # Expose SSH to a virtual machine instance called 'myvm' as a node port (5555) of the cluster:
-  virtctl expose vmi myvm --port=5555 --target-port=22 --name=myvm-ssh --type=NodePort")`
+  virtctl expose vmi myvm --node-port=5555 --target-port=22 --name=myvm-ssh --type=NodePort")`
```
slintes: `--node-port` is an additional parameter for a NodePort service, you still need `--port`. And please use a port >= 30000 for it, because typically lower node ports are not allowed by k8s.

shiywang: oh, I see, so the unclear part was description actually

PR #1407, opened 2018-08-04
[POC] Bridge attachment rethought

rmohr:

This PR depends on #1404. Only the last two commits contain code related to this PR.

Add API support for a `bridge` network source. The network source is very low-level. It understands on how to attach a bridge from the node to the vm, or if the bridge is already inside then pod (e.g. via OCI hook, device plugin), how directly connect to it.

Connect a VMI to a bridge called `br100` on the node:

```yaml
  networks:
  - name: testPod
    pod: {}
  - bridge:
      TargetName: br100
      type: BridgeOrCreate
    name: test1
```

Connect the VMI to a bridge called `br100` which must already be inside the container:

```yaml
  networks:
  - name: testPod
    pod: {}
  - bridge:
      targetName: br100
      type: Bridge
    name: test1
```

Connect a VMI to a bridge called `br99` on the node and call the bridge `br98` inside the container:

```yaml
  networks:
  - name: testPod
    pod: {}
  - bridge:
      targetName: br98
      nodeTargetName: br99
      type: BridgeOrCreate
    name: test1
```

The purpose of this PR is to provide a base for directly attaching bridges to the VMI on one hand, but on the other hand to teach kubevirt in general and **virt-handler** specifically on the api level about one of the most low-level ways on how to connect devices to the VMI.

Based on such primitive types it is then trivial to e.g. implement PV/PVC equivalent for networking. No matter if later network devices may be added via OCI hooks or device plugins, virt-handler can act as the integrating point and do the wiring. Once it is done, it can translate the result to such primitive connection types and forward it to virt-launcher.

**Note**: The goal is not that KubeVirt implements networking all itself. The goal is only to have a base for different attachments from different systems. However KubeVirt may have to have its own translation later between different types of plugins from external parties. There is for instance right now no way on finding out which devices k8s device plugins are creating and which the k8s device plugin invocation order is. As a consequence KubeVirt may have at least a simple translation layer.

**Illustrative example of a basic translation layer**

Let's suppose we add a network source called `resource` like in #1342 and add two devices from different device plugins :

```yaml
  networks:
  - name: testPod
    pod: {}
  - resource:
      resourceName: very.alien.io/red
    name: test1
  - resource:
      resourceName: stranger.io/green
    name: test2
```

We have no way to tell what the very alien `very.alien.io/green` resource will do inside the container. The same for `stranger.io/green`. Also we have no influence on the execution order of the two plugins. 

To overcome this, there may be some kind of a registration mechanism needed in KubeVirt, which can provide the missing information. Here a simple example on how something like this may look like based on a ConfigMap.

The first example would map `very.alien.io/red` to nodes with the `label` `very.alien.io/red`:

```yaml
bridges:
  - name: very.alien.io/red
    type: label
   network:
     targetName: br100
     type: Bridge  
```

The second example would map `very.alien.io/green` to a node resource `very.alien.io/green` based on a k8s device plugin:

```yaml
bridges:
  - name: stranger.io/green
    type: resource
   network:
     targetName: br100
     type: Bridge  
```

`virt-controller` would watch this map and only add labelSelectors or resources to the pod. No more knowledge would ever be needed in `virt-controller`. `virt-handler` can do the wiring and all privileged operations based on the network definition attached in the ConfigMap and then forward one of the very simple network source types to `virt-launcher` which also only needs to know about the ever equal base types.

Fixes #

**Release note**:

```release-note
NONE
```


rmohr:
/hold
still POC but attaching bridges works if one wants to try it out.

rmohr:
@booxter @vladikr I would want to move the whole device moving and IP/MAC reconfiguration from `virt-launcher` to  `virt-handler`, to transition everything privileged as much as possible out of virt-launcher. What do you think?

pkg/virt-handler/devices/bridge.go
```diff
@@ -0,0 +1,263 @@
...
+				return err
+			}
+
+			// Set defaults
+			if net.Bridge.NodeTargetName == "" && net.Bridge.Type == v1.BridgeOrCreate {
+				net.Bridge.NodeTargetName = net.Bridge.TargetName
+			}
+
+			// First let's create the veth pair and move one part into the host namespace
+			// Note: It's important to create the veth pair inside the container, to inherit automatic cleanup for the veth pairs in case of errors.
```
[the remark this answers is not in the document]
yuvalif: @rmohr thanks, now is see the issue (https://github.com/kubevirt/kubernetes-device-plugins/blob/master/pkg/network/bridge/plugin.go#L258) we created the veth on the host and them moved one side to the pod's namespace. Will try that.

vladikr:
> @booxter @vladikr I would want to move the whole device moving and IP/MAC reconfiguration from virt-launcher to virt-handler, to transition everything privileged as much as possible out of virt-launcher. What do you think?

Well, I suppose it's possible. But, besides the network setup will become less atomic, how will we get the original MAC/IP back to virt-launcher to create the domain config?

rmohr:
@vladikr by entering the namespace of the container like  it is done here: https://github.com/kubevirt/kubevirt/pull/1407/files#diff-5b80ff518958e93de348b14803f57c34R258

rmohr:
@yuvalif restructured the structs to make it more obvious. What do you think?

phoracek:
@rmohr 

> The second example
Network Device Plugin should create an interface inside Pod netns, not a bridge. It should be our responsibility to create a bridge on provided interface or pass it directly to VM. How would that work with the proposed API?

api/openapi-spec/swagger.json
```diff
@@ -3767,6 +3767,22 @@
      }
     }
    },
+   "v1.HostBridge": {
+    "description": "Represents a network which the vmi should connect to via a bridge",
+    "required": [
+     "bridgeName"
+    ],
+    "properties": {
+     "bridgeName": {
```
phoracek: Correct me if I'm wrong. In the currently proposed Network API we split the part how to get interface to pod and how to connect VM to it. With your proposed API, we always create the bridge inside Pod? I agree with the nodeBridgeName part, but I think we should just get a veth pair and let the other part of API do whatever user wants (L2 passtrough or bridge with DHCP proxy).

I haven't read actual code yet, if we actually expose just the veth, the name `bridgeName` is confusing.

rmohr: In my opinion `podBridge` (removed for now) and `hostBridge` are primitives out of more thich have stand-alone value but also can act as primitives for connecting. So these two are not supposed to support everything.

rmohr:
@yuvalif @phoracek removed the `podBridge`.

phoracek:
I tested this patch rebased on boot-order one. PXE on top of the bridge works good, however, there was one funny side-effect, once i created VMI connected to the bridge, it connected flannel.1 to the bridge too.

rmohr:
> I tested this patch rebased on boot-order one. PXE on top of the bridge works good, however, there was one funny side-effect, once i created VMI connected to the bridge, it connected flannel.1 to the bridge too.

@phoracek can it be that you passed in a bridge called `br1`?

phoracek:
@rmohr bridge on the host is called `red` and is connected to `eth0`

This is VMI spec I used:

```yaml
kind: VirtualMachineInstance
metadata:
  creationTimestamp: null
  labels:
    special: vmi-bridge
  name: vmi-bridge
spec:
  domain:
    devices:
      disks:
      - disk:
          bus: virtio
        name: registrydisk
        volumeName: registryvolume
      - disk:
          bus: virtio
        name: cloudinitdisk
        volumeName: cloudinitvolume
      interfaces:
      - bridge: {}
        name: red 
        bootOrder: 1
    machine:
      type: ""
    resources:
      requests:
        memory: 1024M
  networks:
  - name: red 
    hostBridge:
      bridgeName: red
  volumes:
  - name: registryvolume
    registryDisk:
      image: kubevirt/fedora-cloud-registry-disk-demo
  - cloudInitNoCloud:
      userData: |-
        #cloud-config
        password: fedora
        chpasswd: { expire: False }
    name: cloudinitvolume
status: {}
```

phoracek:
@rmohr I removed the VMI and created it again, veth pair was created, but its side on host stays DOWN and is not attached to the bridge.

rmohr:
rebase on bootOrder changes on master

pkg/virt-handler/devices/bridge.go
```diff
@@ -0,0 +1,278 @@
...
+					bridge = &netlink.Bridge{
+						LinkAttrs: netlink.LinkAttrs{Name: net.HostBridge.BridgeName},
+					}
+					err := netlink.LinkAdd(bridge)
+					if err != nil {
+						return fmt.Errorf("failed to create bridge %s in the container: %v", net.HostBridge.BridgeName, err)
+					}
+					bridge, err = netlink.LinkByName(net.HostBridge.BridgeName)
+					if err != nil {
+						return fmt.Errorf("failed to get bridge %s on the node namespace: %v", net.HostBridge.NodeBridgeName, err)
```
SchSeba: node->pod

pkg/api/v1/schema.go
```diff
@@ -697,7 +697,7 @@ type Interface struct {
 	// TODO:(ihar) switch to enums once opengen-api supports them. See: https://github.com/kubernetes/kube-openapi/issues/51
 	Model string `json:"model,omitempty"`
 	// BindingMethod specifies the method which will be used to connect the interface to the guest.
-	// Defaults to Bridge.
+	// Defaults to HostBridge.
```
booxter: What does this change in defaults imply? For what I see, `DefaultNetworkInterface` still creates a `bridge` network. Is it something else?

dankenigsberg:
This pull request is clean and tidy, but it does not meet the demands of "The KubeVirt Razor". We know that Pods need to access host bridges, so we should not implement this in KubeVirt. It should be implemented on a lower level, so plain Pods could benefit, too.

PR #1411, opened 2018-08-06
add boot order to network interface

yuvalif:
This PR add boot order to network interfaces

Doing integration test id dependent with exposing L2 network.

```release-note
It will be now possible to boot the virtual machine from one of its network interfaces. 
To achieve that, a numeric boot order value will have to be provided to the interface with following format:
interfaces:
    - bridge: {}
      bootOrder: 2
These values used to determine ordering of boot devices.
Lower values take precedence.
Each interface or disk that has a boot order must have a unique value.
Interfaces without a boot order are not tried.
The network connected to the interface must support booting (e.g. run PXE server there). 
This is not yet supported by the networks available for the virtual machine (the "pod" network).
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3475,7 +3475,7 @@
     ],
     "properties": {
      "bootOrder": {
-      "description": "BootOrder is an integer value \u003e 0, used to determine ordering of boot devices.\nLower values take precedence.\nDisks without a boot order are not tried if a disk with a boot order exists.\n+optional",
+      "description": "BootOrder is an integer value \u003e 0, used to determine ordering of boot devices.\nLower values take precedence.\nEach disk or interface that has a boot order must have a unique value.\nDisks without a boot order are not tried if a disk with a boot order exists.\n+optional",
       "type": "integer",
       "format": "integer"
      },
@@ -3829,9 +3829,10 @@
      "name"
     ],
     "properties": {
-     "bootorder": {
-      "description": "Boot order for this interface.\nany number larger than zero gives the order in which this interface will be used\nfor bootting among all other devices (interfaces and disks)",
-      "type": "string"
+     "bootOrder": {
+      "description": "BootOrder is an integer value \u003e 0, used to determine ordering of boot devices.\nLower values take precedence.\nEach interface or disk that has a boot order must have a unique value.\nInterfaces without a boot order are not tried.\n+optional",
+      "type": "integer",
+      "format": "integer"
      },
      "bridge": {
       "$ref": "#/definitions/v1.InterfaceBridge"
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3829,6 +3829,10 @@
      "name"
     ],
     "properties": {
+     "bootorder": {
+      "description": "Boot order for this interface.\nany number larger than zero gives the order in which this interface will be used\nfor bootting among all other devices (interfaces and disks)",
+      "type": "string"
+     },
      "bridge": {
       "$ref": "#/definitions/v1.InterfaceBridge"
      },
--- a/pkg/api/v1/schema.go
+++ b/pkg/api/v1/schema.go
@@ -198,6 +198,7 @@ type Disk struct {
 	DiskDevice `json:",inline"`
 	// BootOrder is an integer value > 0, used to determine ordering of boot devices.
 	// Lower values take precedence.
+	// Each disk or interface that has a boot order must have a unique value.
 	// Disks without a boot order are not tried if a disk with a boot order exists.
 	// +optional
 	BootOrder *uint `json:"bootOrder,omitempty"`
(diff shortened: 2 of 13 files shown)


api/openapi-spec/swagger.json
```diff
@@ -3829,6 +3829,10 @@
      "name"
     ],
     "properties": {
+     "bootorder": {
+      "description": "Boot order for this interface.\nany number larger than zero gives the order in which this interface will be used\nfor bootting among all other devices (interfaces and disks)",
```
phoracek: s/any/Any/

phoracek: Please add `.` to the end of the sentence, keep it consistent with previous descriptions.

api/openapi-spec/swagger.json
```diff
@@ -3829,6 +3829,10 @@
...
+      "type": "string"
```
phoracek: Why is it a string?

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -543,6 +543,15 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 		return "virtio"
 	}
 
+	getBootOrder := func(iface *v1.Interface, network *v1.Network) *BootOrder {
+		// TODO: should take network into account deciding whether boot order is suported
+		order, err := strconv.ParseUint(iface.BootOrder, 10, 0)
+		if err != nil || order < 1 {
+			return nil
+		}
+		return &BootOrder{Order: uint(order)}
```
phoracek: Do you need to convert it to uint? I'd expect uint is returned from ParseUint.

pkg/api/v1/schema.go
```diff
@@ -697,6 +697,10 @@ type Interface struct {
 	Ports []Port `json:"ports,omitempty"`
 	// Interface MAC address. For example: de:ad:00:00:be:af or DE-AD-00-00-BE-AF.
 	MacAddress string `json:"macAddress,omitempty"`
+	// Boot order for this interface.
+	// any number larger than zero gives the order in which this interface will be used
+	// for bootting among all other devices (interfaces and disks)
+	BootOrder string `json:"bootorder,omitempty"`
```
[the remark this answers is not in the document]
yuvalif: going to make it identical to the way it is used with disks - including description

api/openapi-spec/swagger.json
yuvalif: will make it the same as the disk boot order description

api/openapi-spec/swagger.json
yuvalif: changed to *uint

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -572,6 +581,7 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 				Alias: &Alias{
 					Name: iface.Name,
 				},
+				BootOrder: getBootOrder(&iface, net),
```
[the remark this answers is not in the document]
yuvalif: this would allow for the value "0", do we want to prevent that (in both cases), or let libvirt fail on it?

yuvalif:
@SchSeba i added to the description the requirement for uniqueness of the values (for both interfaces and disks). Do you think this is sufficient, or should we add actual code that checks that?

fabiand:
This is worth a release note!

pkg/api/v1/schema.go
```diff
@@ -697,6 +698,12 @@ type Interface struct {
 	Ports []Port `json:"ports,omitempty"`
 	// Interface MAC address. For example: de:ad:00:00:be:af or DE-AD-00-00-BE-AF.
 	MacAddress string `json:"macAddress,omitempty"`
+	// BootOrder is an integer value > 0, used to determine ordering of boot devices.
+	// Lower values take precedence.
+	// Each interface or disk that has a boot order must have a unique value.
+	// Interfaces without a boot order are not tried.
+	// +optional
+	BootOrder *uint `json:"bootOrder,omitempty"`
```
rmohr: Please add a validation check to the validation webhook to make sure that it is unset or greater than 0.

yuvalif: what about the validation that all orders are unique among interfaces and disks?

rmohr: Is this a libvirt requirement?

michalskrivanek: I think it's not, but it's also quite useless/unpredictable otherwise. The boot order functionality is wonky on its own, perhaps best to ensure some basic sanity

yuvalif: libivirt would fail if we have duplicate values

rmohr: Then let's catch this early.

rmohr:
@ifireball can it be that there are some machines which do not have nested virt enabled in the pool?

E.g. see https://jenkins.ovirt.org/blue/organizations/jenkins/kubevirt_kubevirt_standard-check-pr/detail/kubevirt_kubevirt_standard-check-pr/1511/pipeline/122

```
  KVM devices are required for testing, but are not present on cluster nodes
  Expected
      <bool>: false
  to be true
```

rmohr:
Have passing lanes for k8s and ocp where kvm seems to be running with nesting.

cynepco3hahue:
@danielBelenky I saw the same under https://jenkins.ovirt.org/blue/organizations/jenkins/kubevirt_kubevirt_standard-check-pr/detail/kubevirt_kubevirt_standard-check-pr/1516/pipeline/122

PR #1414, opened 2018-08-07
Allow separating guest memory from resource request

rmohr:

To allow real memory overcommit for VMIs, it is necessary to allow giving VMIs more memory than we actually request from the KubeVirt scheduler. Alongside `spec.domain.resources.memory.requests`, which requests reserved memory from the scheduler, `spec.domain.memory.guest` can contain a differen value for the guest.

In case a VMI with `100M` wants to overcommit 10%, it can set `spec.domain.resources.memory.requests` to `90M` and `spec.domain.memory.guest` to `100M`: 

```yaml
kind: VirtualMachineInstance
metadata:
  name: vmi-ephemeral
spec:
  domain:
      memory:
        guest: 90M
    devices:
      disks:
      - disk:
          bus: virtio
        name: registrydisk
        volumeName: registryvolume
    machine:
      type: ""
    resources:
      requests:
        memory: 100M
  volumes:
  - name: registryvolume
    registryDisk:
      image: registry:5000/kubevirt/cirros-registry-disk-demo:devel
```

Soft-eviction and Hard-eviction rules from the kubelet will apply as configured on the node to these VMIs.

In case that a k8s cluster with kubevirt is run, mostly just for the purpose to run Virtual Machines, it is possible to reconfigure the cluster to use `swap`.

```release-note
Allow real memory overcommit on a node by adding a field spec.domain.memory.guest, which sets the visible memory for the guest and which can be higher than the amount of memory than requested from the k8s scheduler.
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3961,6 +3961,10 @@
    "v1.Memory": {
     "description": "Memory allows specifying the VirtualMachineInstance memory features.",
     "properties": {
+     "guest": {
+      "description": "Guest allows to specifying the amount of memory which is visible inside the Guest OS.\nThe Guest must lie between Requests and Limits from the resources section.\nDefaults to the requested memory in the resources section if not specified.\n+ optional",
+      "type": "string"
+     },
      "hugepages": {
       "description": "Hugepages allow to use hugepages for the VirtualMachineInstance instead of regular memory.\n+optional",
       "$ref": "#/definitions/v1.Hugepages"
--- a/manifests/generated/vm-resource.yaml
+++ b/manifests/generated/vm-resource.yaml
@@ -204,6 +204,7 @@ spec:
                           - type
                         memory:
                           properties:
+                            guest: {}
                             hugepages:
                               properties:
                                 pageSize:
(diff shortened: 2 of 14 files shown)


rmohr:
ci test please

cynepco3hahue:
@rmohr Can you please add information, why does it relevant to add this field?

vladikr:
Over the direction make sense to me, but feels like there are too many tunables. (as they can set requests and limits and domain.memory)
Maybe we shouldn't let the users set  spec.domain.resources.requests.memory when spec.domain.memory.guest - Instead we can introduce an overCommitRatio field and calculate it our selves. This overCommitRatio can be global for the host even.

Another point, we can also let the users disable the overhead calculation altogether in this case and let decide their destiny if they know how much overcommit they want to allow.

rmohr:
@vladikr in general I agree that we want much more on a global level configurable. 

Still I think it is good to start on the VMI directly. Since `limits` does not have to be set, it is possible for users in k8s to use as much memory as is available (or set on the namespace limits) until someone with higher priority comes in. This option allows users to decide themselves if their VMIs can be evicted from time-to-time but have access to more memory, or if it is a no-go.

fabiand:
@rmohr I think the description yaml above is wrong. (90 and 100) should be swapped

PR #1419, opened 2018-08-08
Use kube-openapi to generate swagger schema

### LINKED ISSUE #1227: Completely switch to openapi-gen from k8s

/kind enhancement

**What happened**:

Right now we still use `https://github.com/rmohr/go-swagger-utils` to generate go-restful structs for swagger.

 Instead of that, we can now completely switch to `https://github.com/kubernetes/code-generator/tree/master/cmd/openapi-gen`.

In order to make this work, do:

1) Use `builder.BuildOpenAPISpec` to build the spec based on go-restful (part of apimachinery)
2) this function requires a config. This config needs to contain the object definitions of our types (they are already generated with openapi-gen).
3) During code-generation, dump the resulting spec as json
4) Install the generated docs to a go-restful endpoint (wiring code exists in vendoring too)

cynepco3hahue:
Will generate more correct from k8s perspective swagger schema, that probably will fix part of python auto-generated client issues.

Fixes #1227 

- [x] Use builder.BuildOpenAPISpec to build the spec based on go-restful 
- [x] During code-generation, dump the resulting spec as json
- [ ] Install the generated docs to a go-restful endpoint (wiring code exists in vendoring too)

**Release note**:

```release-note
Use kube-openapi to generate swagger schema
```


cynepco3hahue:
@rmohr I still do not understand the last point(Install the generated docs to a go-restful endpoint), can you please clarify it, and also I do not sure if we still need swagger-ui endpoint.

cynepco3hahue:
@pkliczewski FYI

pkg/virt-api/api.go
```diff
@@ -277,10 +282,10 @@ func (app *virtAPIApp) composeSubresources(ctx context.Context) {
 		To(func(request *restful.Request, response *restful.Response) {
 			response.WriteAsJson(subresourceAPIGroup())
 		}).
-		Operation("getAPIGroup").
+		Operation("getSubresourceAPIGroup").
```
lukas-bednar: On line 270 you have SubResources instead of Subresources (R/r) ... I believe it should be consistent also on line 299. I don't mind what variant you pick.

cynepco3hahue: Thx will fix it

PR #1429, opened 2018-08-10
Bridge integration for 0.6

rmohr:

This PR depends on #1404. **Only the last commit contains code related to this PR.**

Add API support for a `hostBridge` network source. The network source is very low-level. It understands on how to attach a bridge from the node to the vm, or if the bridge is already inside then pod (e.g. via OCI hook, device plugin), how directly connect to it.

Connect a VMI to a bridge called `br100` on the node:

```yaml
  networks:
  - name: testPod
    pod: {}
  - hostBridge:
      bridgeName: br100
    name: test1
```

Connect a VMI to a bridge called `br99` on the node and call the bridge `br98` inside the container:

```yaml
  networks:
  - name: testPod
    pod: {}
  - hostBridge:
      bridgeName: br98
      nodeBridgeName: br99
    name: test1
```

The purpose of this PR is to provide a base for directly attaching bridges to the VMI on one hand, but on the other hand to teach kubevirt in general and **virt-handler** specifically on the api level about one of the most low-level ways on how to connect devices to the VMI.

Based on such primitive types it is then trivial to e.g. implement PV/PVC equivalent for networking. No matter if later network devices may be added via OCI hooks or device plugins, virt-handler can act as the integrating point for vm-attachment and do the wiring. Once it is done, it can translate the result to such primitive connection types and forward it to virt-launcher.

**Note**: The goal is not that KubeVirt implements networking all itself. The goal is only to have a base for different attachments from different systems. However KubeVirt may have to have its own translation later between different types of plugins from external parties. There is for instance right now no way on finding out which devices k8s device plugins are creating and which the k8s device plugin invocation order is. As a consequence KubeVirt may have at least a simple translation layer.

Fixes #

**Release note**:

```release-note
Allow specifying a hostBridge as a network source.
```


cynepco3hahue:
ok looks like we lack std-ci configuration for release-0.6 branch, will create a PR to make sure that CI passes for all relevant PR's under 0.6

cynepco3hahue:
ci test please

cynepco3hahue:
I see a lot of errors
```
level=info timestamp=2018-08-12T15:59:57.273830Z pos=utils.go:238 component=tests msg="Created virtual machine pod virt-launcher-testvmi8dtv9-9thk9"
level=info timestamp=2018-08-12T16:00:06.664434Z pos=utils.go:238 component=tests msg="Pod owner ship transferred to the node virt-launcher-testvmi8dtv9-9thk9"
level=error timestamp=2018-08-12T16:00:06.698296Z pos=utils.go:236 component=tests reason="unexpected warning event received" msg="could not prepare root namespace part for network red: failed to get bridge red on the node namespace: Link not found"
• Failure in Spec Setup (BeforeEach) [240.030 seconds]
Bridge
/root/go/src/kubevirt.io/kubevirt/tests/vmi_bridge_test.go:41
  Exposing a network to the VM via bridge device plugin
  /root/go/src/kubevirt.io/kubevirt/tests/vmi_bridge_test.go:105
    VM should be able to connect to the outside world over the default interface [BeforeEach]
    /root/go/src/kubevirt.io/kubevirt/tests/vmi_bridge_test.go:136

    Unexpected Warning event received.
    Expected
        <string>: Warning
    not to equal
        <string>: Warning

    /root/go/src/kubevirt.io/kubevirt/tests/utils.go:240
```

tests/vmi_bridge_test.go
```diff
@@ -0,0 +1,390 @@
...
+
+		// add red and blue bridges to host
+		addBridgeToHost("red")
+		addBridgeToHost("blue")
+	})
+
+	Context("Exposing a network to the VM via bridge device plugin", func() {
+		var vmi *v1.VirtualMachineInstance
+		const networkName = "red"
+		const macAddress = "de:ad:00:00:be:af"
```
phoracek: This is used to assertion, but never set as an parameter, maybe one of the problems is here? I guess this patch must be rebased on the #1430

cynepco3hahue:
We decided to continue with `cnv-1` so I close this one.

tests/vmi_bridge_test.go
cynepco3hahue: It has additional problems, but all except MAC address must be fixed under https://github.com/kubevirt/kubevirt/pull/1437

PR #1452, opened 2018-08-15
Fix python client SDK generator 

SchSeba:
**What this PR does / why we need it**:
Added hard coded python files for both slirp and bridge interface .

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1181 

[client-python 21](https://github.com/kubevirt/client-python/issues/21)

**Special notes for your reviewer**:
Swagger code-gen doesn't generate code for empty structs. So we need to hard coded the python class files.

**Release note**:

```release-note
Add slirp and bridge implementation for python client sdk
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/1452)

PR #1457, opened 2018-08-17
api: missing V1InterfaceBridge type

pkliczewski:
After python client generation using current api definition there is
type missing which makes it impossible to manage VMIs. The missing type
seems to be placeholder and it do not contain any definition. It seems
like it is not used and content for it is unknown so changing it to be
a string fixes the issue.

Fixes: https://github.com/kubevirt/client-python/issues/21


SchSeba:
@pkliczewski I open a [PR](https://github.com/kubevirt/kubevirt/pull/1452) to fix this issue.

I am not sure if making it a string is the right way.
It is good to have a struct that handle the bridge object because in the future we may want to add optional parameter to it.

cynepco3hahue:
@pkliczewski Does this PR still relevant or can I close it?

booxter:
I think it makes sense to reconsider the PR despite a short-term fix for the client issue was merged already.

In the past, we assumed that we will need a lot of additional options per binding type, so we designed the API so that we pass a map of attributes as a value for the per-binding attributes. Now that we spent some time thinking about and building the API, we discovered that we may not have many of these options if at all. Which resulted in usually-empty maps as values to these attributes (`bridge: {}`; `slirp: {}`; etc.) This raises the question whether the initial decision of making these attributes map-typed is correct. I don't think so anymore.

So in this regard, I agree that this API deficiency should be fixed.

In your PR, you convert each of attributes (`bridge`, `slirp`) to string. It's not exactly clear how the value of the string is supposed to be interpreted though. If it's just a flag where a non-empty value means that the particular binding type is to be used then I don't think we need a per-type attribute in the first place. Instead, we could introduce a new attribute (perhaps called `binding` or just `type`) of string type, and allow it to accept values from the list of supported binding types (meaning, one of `slirp`, `bridge`, `direct`, ...) Then also remove `bridge` and `slirp` attributes from API schema since they are superseded by the unified `type` attribute.

What do you think?

fabiand:
ping?

fabiand:
ping?

pkliczewski:
We can work with the code as it is for now.

PR #1480, opened 2018-08-28
Iothreads rethought

stu-gott:
This is a follow up to https://github.com/kubevirt/kubevirt/pull/1421 as an alternate approach. Instead of blindly passing IOThreads through to libvirt, we are a little more intelligent in terms of figuring out how many are needed. based on disks that requested dedicated IO.

There's still a little more to be done in the domain XML conversion, but it's complete enough to get feedback as to which approach is preferred in general.

```release-note
Allow users to set the number of IOThreads used in a VirtualMachineInstance.
```


pkg/api/v1/schema.go
```diff
@@ -73,6 +73,11 @@ type DomainSpec struct {
 	Features *Features `json:"features,omitempty"`
 	// Devices allows adding disks, network interfaces, ...
 	Devices Devices `json:"devices"`
+	// Setting useIOThreads to true will enable IOThreads and
```
fromanirh: UseIOThreads"? (case consistency with the option)

MarSik:
Yeah, I also think that a per disk booleans is the better approach. We can then use other means to make it easier for the user (something related to presets probably) to just say use one common iothread for all my disks.

stu-gott:
@fromanirh Thank you for that insightful feedback! I totally agree I left that use case out. My opinion (and that's all it is--my opinion) is that using label strings instead of booleans makes this approach less understandable as a whole. In the interests of making progress, I'd propose we move forward without implementing that for now.

MarSik:
@stu-gott What about allowing rational numbers for the number of ioThreads in the disk spec? A disk with 0.5 ioThread allocation might express the wish to share the ioThread with some other disk.

stu-gott:
ci test please

stu-gott:
@vladikr @fromanirh This has been re-worked a bit. Instead of "UseIOThreads" as a boolean, we now use "IOThreadsPolicy" which can currently be `shared` or `dedicated`. (If we implement CPU pinning for IOThreads in the future, we can add `pinned` to that list.)

fromanirh:
@stu-gott I agree that we can skip the use case I presented for now. That scenario may not even be important enough for users.

Will review again the PR soon.

vladikr:
@stu-gott Great work! 
I only have one concern about the current approach wrt the "dedicated" policy.
I'm worried that when the number of disks exceeds the number of
vcpu, iothreads will compete with vcpu threads on the cpu time and the
usefulness of this feature will go to zero. This also will create and cpu overhead.

I think that we might want to automatically limit it by the number of vcpus. However, I suggest changing the name of the policy to "preferred" - so it will imply that it's up to kubevirt to figure out the right setting.
Once the cpu pinning PR is merged when users will request dedicated vpcus and also
will have an iothread policy set to "preferred" we will simply pack and pin the
iothreads to the allocated pcpus. 

In the future, we could also re-introduce the "dedicated" policy that will allocate additional pCPUs and pin the iothreads to it.

Here is a similar discussion from Openstack: https://review.openstack.org/#/c/230968/11/specs/ocata/approved/libvirt-iothreads.rst

stu-gott:
ci test please

stu-gott:
Thanks @vladikr ... I've been thinking about that. I wonder if having a policy based approach makes it too easy to "do the wrong thing"... when we had just a boolean and required the user to mark any given disk as dedicated--it takes a certain amount of malice of forethought to request any given disk to have it's own IOThread.

But either way, there's still the issue of VCPU's to address. There's two general approaches I see:

- Limit IOThreads to the number of available CPU's (as you mentioned)
- Request additional CPUs for the pod that belong to dedicated IOThreads. These would not be added to the VM itself, so would otherwise be invisible overall.

stu-gott:
@vladikr oh, actually another great approach might be to have a validating webhook. The system could simply reject the manifest if an unsafe number of IOThreads was requested. This has the advantage of educating users on the performance issues associated with asking for too many dedicated threads.

vladikr:
> Thanks @vladikr ... I've been thinking about that. I wonder if having a policy based approach makes it too easy to "do the wrong thing"... when we had just a boolean and required the user to mark any given disk as dedicated--it takes a certain amount of malice of forethought to request any given disk to have it's own IOThread.

Right, I understand this approach too. I would advocate implementing what seems to be the right thing to do for the user in most of the cases. From my POV, without cpu pinning, this is to enable IOThreads per each disk and limit it by the number of vcpus.

Overall, enabling iothreads allows i/o handling to be done outside of the qemu mutex. So, enabling it per each disk makes the most sense to me. However,  when the user knows that his application doesn't produce much disk i/o, it'll be enough to share a single thread between all disks - That's why you proposed the "shared" policy. 
On the other hand, I think in most cases, the user would benefit from a thread per disk, but it won't be useful to have more threads than vcpus, hence I think it should be limited by the number of vcpus - that's the reason I'd suggest the "preferred" policy - we just need to make it clear that kubevirt calculates the setting.
Using the same policy, with CPU pinning, we will simply pack the i/o threads on the allocated pCPUs and won't go beyond as it could happen without the pinning.

As for the requesting more cpus, I think it's a viable approach, but the user can control the number of vcpus. He can simply request more knowing the limitations, I can't see a reason at the moment to prefer this approach.

Overall, a policy based approach allows us to extend the behavior without introducing more tuneables.

It's just my view of this feature. Doesn't mean I'm right or anything. 

   

> 
> But either way, there's still the issue of VCPU's to address. There's two general approaches I see:
> 
> * Limit IOThreads to the number of available CPU's (as you mentioned)
> * Request additional CPUs for the pod that belong to dedicated IOThreads. These would not be added to the VM itself, so would otherwise be invisible overall.

MarSik:
> @MarSik It's not entirely clear to me what you're suggesting. How would that work and what is it trying to address? As a user, if I was instructed to "assign half an IOThread to this disk" I'd be pretty confused as to what that implied. :)

Well the resources section already uses fractions when specifying the CPU allocation so I was just wondering if we could use that here as well. We could then later compute the cpu id mapping for separate threads to not over-commit based on the fractions. The number would basically be a hint about how much will that disk utilize the thread.

But shared vs. dedicated is probably both simple and good enough.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -460,6 +461,52 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+	useIOThreads := false
+	threadPoolLimit := 1
+
+	if vmi.Spec.Domain.IOThreadsPolicy != nil {
+		useIOThreads = true
+
+		if (*vmi.Spec.Domain.IOThreadsPolicy) == "auto" {
+			threadPoolLimit = 2
+			if vmi.Spec.Domain.CPU != nil {
+				threadPoolLimit = 2 * int(vmi.Spec.Domain.CPU.Cores)
```
vladikr: I think we should be looking at `spec.domain.resources.requests.cpu` and `spec.domain.resources.limits.cpu` to get the number of vcpu. `spec.domain.cpu.cores` just sets the guest topology which doesn't have to match actual vcpus

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -476,6 +523,40 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+
+				logger := log.DefaultLogger()
+				logger.Infof("Bus for this disk: %s", newDisk.Target.Bus)
+
+				if newDisk.Target.Bus == "scsi" {
+					newController := Controller{
+						Type:  "scsi",
+						Model: "virtio-scsi",
+						Driver: &ControllerDriver{
+							IOThread: &ioThreadId,
```
vladikr: Are we adding a controller for each scsi disk? IIRC, each virtio-scsi controller should support up to 256 disks.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -460,6 +461,61 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+			if cpuRequests, ok := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceCPU]; ok {
+				numCPUs = int(cpuRequests.Value())
+			} else if cpuLimit, ok := vmi.Spec.Domain.Resources.Limits[k8sv1.ResourceCPU]; ok {
+				numCPUs = int(cpuLimit.Value())
+			}
+
+			threadPoolLimit = numCPUs * 2
+
+			if vmi.Spec.Domain.CPU != nil {
+				threadPoolLimit = 2 * int(vmi.Spec.Domain.CPU.Cores)
```
vladikr: @stu-gott I don't think we need this at all. Without the cpu pinning in the picture, vmi.Spec.Domain.CPU.Cores only affect the guest topology and it shouldn't affect the calculation of vcpus.

vladikr:
seems like make generate is needed

stu-gott:
ci test please

stu-gott:
ci test please

pkg/api/v1/schema.go
```diff
@@ -111,6 +111,11 @@ type DomainSpec struct {
 	Features *Features `json:"features,omitempty"`
 	// Devices allows adding disks, network interfaces, ...
 	Devices Devices `json:"devices"`
+	// Controls whether or not disks will share IOThreads.
+	// Omitting IOThreadsPolicy disables use of IOThreads.
+	// One of: shared, auto
```
rmohr: Can you introduce constants with their own string subtype?

stu-gott: My bad, I had misread that the openapi spec explicitly did not support enums, but that was old information. fixed in next push :)

vladikr:
ci test please

PR #1482, opened 2018-08-29
Include 'git describe' result in all virt-* containers

booxter:
This helps with determining which exact git tree was used to build
containers.

Just do: docker exec -it <...> cat /gitinfo

**What this PR does / why we need it**: simplifies determination of git hash used to build an image.

**Special notes for your reviewer**: maybe the name of the file inside container should be different from `/gitinfo` (perhaps `/.gitinfo`?)

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build-docker.sh
+++ b/hack/build-docker.sh
@@ -21,6 +21,7 @@ set -e
 
 source hack/common.sh
 source hack/config.sh
+source hack/version.sh
 
 if [ -z "$1" ]; then
     target="build"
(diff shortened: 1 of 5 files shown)


hack/build-docker.sh
```diff
@@ -40,6 +40,7 @@ for arg in $args; do
     if [ "${target}" = "build" ]; then
         (
             cd ${CMD_OUT_DIR}/${BIN_NAME}/
+            git describe > .gitinfo
```
rmohr: Can you change the PR to get the needed info out of `hack/version.sh`? Maybe adding an extra function there. This file is currently used to get such info from git, or alternatively get it from environment variables to allow building outside of the git repo.

booxter:
I don't need to add the file to `.gitignore` because it's inside _out subdirectory that is ignored already.

PR #1484, opened 2018-08-31
An optional custom PCI address for VirtIO disks

### LINKED ISSUE #1325: Add support for stable (PCI) addresses and introspection

vladikr:

This PR allows the API users to set an optional PCI address on VirtIO disks
Note: The first 3 PCI slots (0, 1, 2) are reserved in any custom address 

```
spec:
  domain:
    devices:
      disks:
        name: disk0
        volumeName: volume0
        disk:
          bus: virtio
          pciAddress: 0000:04:10.0
```

Fixes #1325 

**Release note**:
```release-note
NONE
```


vladikr:
> I actually was always thinking of the address fields https://libvirt.org/formatdomain.html#elementsAddress to set device addresses.
@fabiand I think I don't follow. Isn't it what I'm using?

fabiand:
Right.
What confused me that we can set addresses on other disk types as well - they#d just not be PCi addresses.

vladikr:
> This way the address section is extensible

I intentionally allowed to set only PCI addresses. Do we really want to allow users to manually adjust the disks on an arbitrary bus!?

fabiand:
Update: We actually want to investiagte if setting PCi addresses is actually relevant for recetn Windows versions.
After all Windows reactivation is one of the main use-cases of fixating devices.

PR #1485, opened 2018-08-31
Introduce an optional subprotocol for browser websocket connections

rmohr:

Browsers need to use subprotocols to pass credentials to kubernetes to
open websocket connections. Kubernetes will remove the authentication
subprotocol from the protocols. In order to let the browser continue the
handshake, the server needs to return another subprotocol. The
subprotocol is named "plain.kubevirt.io" and is an alias for a plain
websocket connection.

A valid connection attempt from the browser may look like this:
```
Sec-WebSocket-Protocol: base64url.bearer.authorization.k8s.io.bXl0b2tlbg, plain.kubevirt.io
```
The server will then respond with
```
Sec-WebSocket-Protocol: plain.kubevirt.io
```
If no subprotocols are given, the connection is upgraded regardless, for
backward compatibility.

**Release note**:

```release-note
To allow browsers to connect to KubeVirt websocket subresources, the "plain.kubevirt.io" subprotocol is introduced.
```

### DIFF
--- a/tests/vnc_test.go
+++ b/tests/vnc_test.go
@@ -40,6 +40,7 @@ import (
 	"kubevirt.io/kubevirt/pkg/api/v1"
 	"kubevirt.io/kubevirt/pkg/kubecli"
 	"kubevirt.io/kubevirt/pkg/log"
+	"kubevirt.io/kubevirt/pkg/util/subresources"
 	"kubevirt.io/kubevirt/tests"
 )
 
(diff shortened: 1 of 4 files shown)


rmohr:
@mareklibra FYI.

tests/vnc_test.go
```diff
@@ -120,10 +121,14 @@ var _ = Describe("VNC", func() {
 			})
 		})
 
-		table.DescribeTable("should upgrade subresource connections if an origin header is given", func(subresource string) {
+		table.DescribeTable("should upgrade websocket connectsion which look like comming from a browser", func(subresource string) {
```
michalskrivanek: typo

tests/vnc_test.go
```diff
@@ -120,10 +121,14 @@ var _ = Describe("VNC", func() {
...
 			config, err := kubecli.GetKubevirtClientConfig()
 			Expect(err).ToNot(HaveOccurred())
-			rt, err := upgradeCheckRoundTripperFromConfig(config)
+			// Browsers need a subprotocol, since they will have to use the subprotocol mechanism to forward the bearere token.
```
michalskrivanek: "bearer"

PR #1495, opened 2018-09-05
Simplify the template_test logic

shiywang:

quoting ginkgo doc here: https://onsi.github.io/ginkgo/#separating-creation-and-configuration-justbeforeeach

> You can have multiple JustBeforeEaches at different levels of nesting. Ginkgo will first run all the BeforeEaches from the outside in, then it will run the JustBeforeEaches from the outside in. While powerful, this can lead to confusing test suites – so use nested JustBeforeEaches judiciously.

> Some parting words: JustBeforeEach is a powerful tool that can be easily abused. Use it well.

reduce unnecessary nested logic make code more readable for others.

```release-note
NONE
```

### DIFF
--- a/tests/template_test.go
+++ b/tests/template_test.go
@@ -59,123 +59,86 @@ var _ = Describe("Templates", func() {
 
 	Describe("Launching VMI from VM Template", func() {
 
-		assertGeneratedVMJson := func() func() {
-			return func() {
-				By("Generating VM JSON from the Template via oc-process command")
-				_, err := runOcProcessCommand(templateJsonFile, templateParams, vmJsonFile)
+		assertGeneratedVMJson := func() {
+			By("Generating VM JSON from the Template via oc-process command")
+			_, err := runOcProcessCommand(templateJsonFile, templateParams, vmJsonFile)
+			ExpectWithOffset(1, err).ToNot(HaveOccurred())
+			ExpectWithOffset(1, vmJsonFile).To(BeAnExistingFile())
+		}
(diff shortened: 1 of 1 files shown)


shiywang:
@gszasz @cynepco3hahue PTAL

shiywang:
@gbenhaim @fabiand  compare to this
```
BeforeEach()
JustBeforeEach()
AfterEach()
Context("with given Fedora Template", func() {
        BeforEach()..
	It("....")	
 			Context("", func() {	
				JustBeforeEach()	
				AfterEach()	
 				It("....")	
 				Context("...", func() {	
					JustBeforeEach()	
 					It("....")	
 					Context("", func() {	
						JustBeforeEach()	
 						It(")	
					})	
				})
}
```

I found this is more readable 
```
BeforeEach()
JustBeforeEach()
AfterEach()
Context("with given Fedora Template", func() {
             BeforEach(){}
             It("....")
}
```	
and I think it did the same thing, could you tell me which point my code didnt cover without using nested block ?

gszasz:
@shiywang The nested context structure is a standard way how to properly handle a chain of mutually dependent test cases in Ginkgo. The goal here is to define a clear dependency relationship between individual test cases while minimizing code redundancy. The implementation was pre-prepared to cover the whole test plan currently defined by QE, where every `It` block maps to a defined test case. The simplification you propose violates the original intent and it introduces conflicts to the consequent PRs which take advantage of the existing code logic.

PR #1499, opened 2018-09-06
Add more error handling/logging to device plugin startup

stu-gott:

Device Plugin server needs more logging on startup. This was noted while debugging an issue with minishift where the device plugin framework appeared to start, but the cluster nodes never had any allocated tun/kvm devices.

In particular, errors during grpc server registration were not caught/handled.

**Release note**:
```release-note
NONE
```


stu-gott:
@xsgordon This PR should help us understand what is happening on your minishift cluster

rmohr:
@stu-gott any reason why we can't follow the route outlined in #1404?

stu-gott:
@rmohr I fully support 1404 and we should do that. this was more an immediate-short-term approach as things have been moving slowly on that front.

PR #1504, opened 2018-09-11
api: return empty array in resources for subresources endpoint

pkliczewski:
Dynamic client is complaining about null value in resources. This
PR fixes this issue by returning empty array.

Fixes: https://github.com/kubevirt/kubevirt/issues/1500


pkliczewski:
@rmohr @davidvossel please review

pkliczewski:
@rmohr Yes, verified with the client and it works fine

PR #1506, opened 2018-09-12
Add default value for oc/virtctl/kubectl

gouyang:
Signed-off-by: Guohua Ouyang <gouyang@redhat.com>

Set default value for these binaries so our tests won't be failed if not specify them. Convenient for run tests manually.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -71,19 +71,17 @@ import (
 	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/datavolumecontroller/v1alpha1"
 )
 
-var KubeVirtVersionTag = "latest"
-var KubeVirtRepoPrefix = "kubevirt"
-var KubeVirtKubectlPath = ""
-var KubeVirtOcPath = ""
-var KubeVirtVirtctlPath = ""
-var KubeVirtInstallNamespace = "kube-system"
+var KubeVirtVersionTag string
+var KubeVirtRepoPrefix string
+var KubeVirtInstallNamespace string
+var KubeVirtKubectlPath, KubeVirtOcPath, KubeVirtVirtctlPath string
 
 func init() {
 	flag.StringVar(&KubeVirtVersionTag, "tag", "latest", "Set the image tag or digest to use")
 	flag.StringVar(&KubeVirtRepoPrefix, "prefix", "kubevirt", "Set the repository prefix for all images")
-	flag.StringVar(&KubeVirtKubectlPath, "kubectl-path", "", "Set path to kubectl binary")
-	flag.StringVar(&KubeVirtOcPath, "oc-path", "", "Set path to oc binary")
-	flag.StringVar(&KubeVirtVirtctlPath, "virtctl-path", "", "Set path to virtctl binary")
+	flag.StringVar(&KubeVirtKubectlPath, "kubectl-path", "/usr/bin/kubectl", "Set path to kubectl binary")
+	flag.StringVar(&KubeVirtOcPath, "oc-path", "/usr/bin/oc", "Set path to oc binary")
+	flag.StringVar(&KubeVirtVirtctlPath, "virtctl-path", "/usr/bin/virtctl", "Set path to virtctl binary")
 	flag.StringVar(&KubeVirtInstallNamespace, "installed-namespace", "kube-system", "Set the namespace KubeVirt is installed in")
 }
 


ovirt-infra:

gouyang:
@rmohr The tests can be skipped or failed, it's tests writing strategy. But it's not the purpose of this PR.
Currently, it need to pass `-oc-path=${OC_PATH} -kubectl-path=${KUBECTL_PATH} -virtctl-path=${VIRTCTL_PATH}` to `go test` in order to run tests with function `RunCommand`. By adding default values of these binaries, user can just run tests like `go test -kubeconfig=config .`.

rmohr:
> The tests can be skipped or failed, it's tests writing strategy.

Could you explain what you mean by that?

> But it's not the purpose of this PR.

I know. Tests can be run as ordinary user. If admins have install arbitrary oc/kubeadm/virtctl packages, that is not at all related to the test run. I understand that maybe in your case you installed oc/kubectl just for running the tests against it.

I just think that it should be a conscious decision which binaries should be selected.

gouyang:
> > The tests can be skipped or failed, it's tests writing strategy.
> 
> Could you explain what you mean by that?

eg: By default, It will get below error if just assert the error is not occurred. As you said the tests can be skipped based on this kind of error, but not failed. So it's a strategy about how to handle the error.
```
    Timed out after 90.022s.
    Timed out waiting for VMI to enter Running phase
    Expected
        <bool>: false
    to equal
        <bool>: true
```

> I know. Tests can be run as ordinary user. If admins have install arbitrary oc/kubeadm/virtctl packages, that is not at all related to the test run. I understand that maybe in your case you installed oc/kubectl just for running the tests against it.

It seems there is no tests related to these binaries in kubevirt. But there are lots of tests in kubevirt-ansible. From QE's perspective, use these binaries for tests are more likely as end user will do.

Since these flags are there, why don't set default values for them so they can be called more conveniently.

rmohr:
> eg: By default, It will get below error if just assert the error is not occurred. As you said the tests can be skipped based on this kind of error, but not failed. So it's a strategy about how to handle the error.

They can be skipped if the binaries are not set. You can add a `Skip("No kubectl binary specified")` in `RunCommand` if the `*Path` is empty. You don't have to detect errors or something like that. Right now it looks like RunCommand returns an error instead of skipping the test.

> It seems there is no tests related to these binaries in kubevirt. But there are lots of tests in kubevirt-ansible.

We have such tests too. Using arbitrary binaries will result in unreproducable errors, which makes the tests pretty useless. If you want to run the tests in kubevirt-ansible you should be aware of the kubectl/oc/virtctl versions you want to test/support and explicitly choose the binary. If you want to use the version which is installed on the node you can specify it on the test run and you should be fine.

If your QE runs are not reproducible then you will have issues like, oh on node A stuff fails, on node B not. Oh yesterday I had an error, but today not, because *maybe* centos pushed  a new kubectl version. As said, you can for sure use the node installed binaries, but it should be a conscious decision and the versions of the binaries need to be recorded or know.

As a rule of thumb, the test-suite in kubevirt is self-containing to be reproducible. Only by specifying specific external things to use it should start using stuff from its surrounding.

>  From QE's perspective, use these binaries for tests are more likely as end user will do.

As said before, if you want to use them from the node, simply configure the test suit to use them. It is then up to you to pre-record the exact configuration on the node to have reproducible results.

gouyang:
I usually invoke the tests locally before push the PR, so I run tests manually.

rmohr:
> It's a safe change and make life easy.

It is against the principle of the goal of reproducability and self-containment of the tests. So I am not sure I can help you there.

rmohr:
> This is not the point of this PR. And to me, I don't like to put `Skip` to a function in `utils.go`, instead, I'd like to return the error and put`Skip` it in the test file, which is more readable. Just personal opinions.

You will then have to distinguish between `Skip` condition errors and not-`Skip` condition errors.

rmohr:
> I usually invoke the tests locally before push the PR, so I run tests manually.

Great. You can specify the binary if you want  to run the tests against your locally available binary.

gouyang:
> It is against the principle of the goal of reproducability and self-containment of the tests. So I am not sure I can help you there.

I don't understand what it is? why and how?

rmohr:
> I don't understand what it is? why and how?

If I invoke the test suite with the same command-arguments on on different machines I end up with a different set of tests which are run, against different versions of the binaries.

As said before:

As a rule of thumb, the test-suite in kubevirt is self-containing to be reproducible. Only by specifying specific external things to use it should start using stuff from its surrounding.

rmohr:
> It's a safe change and make life easy.

By the way, with your change people would have to make sure that they either have the binaries available on these locations or would need to make it clear that they don't want to run these tests because they don't have these binaries.

gouyang:
> By the way, with your change people would have to make sure that they either have the binaries available on these locations or would need to make it clear that they don't want to run these tests because they don't have these binaries.

If people don't  have the binaries on the location, people specify them on their own.

rmohr:
> If people don't have the binaries on the location, people specify them on their own.

I think you miss the point. Right now the binaries should not have to be present at all (therefore the skip, that's a bug in the testsuit that  it fails the tests instead of skipping them).

gouyang:
> I think you miss the point. Right now the binaries should not have to be present at all

Yes, it does not need the binaries to be present. And the change does no harmful things to that.

rmohr:
can we resolve that offline? We go in circles here.

PR #1513, opened 2018-09-16
Limit emulated machine values to the q35 family

### LINKED ISSUE #1230: Limit machine type to q35
Today we can set the machine type.

In order to strive for stability and avoid complexity we should limit ourselfs to a future proof (ha! q35 is sooo odl, but still the most recent ..) machine type

/kind enhancement

In order to achieve this we should ~either remove the field from the API or~ limit it to a single value (in order to still give visibility of the machine type in the API)

ahadas:
Limit emulated machine values to the q35 family at the API-level.

Fixes #1230 

**Release note**:
```release-note
Machine type is now limited to 'q35' or one of its derivatives.
```

### DIFF
--- a/tests/console_test.go
+++ b/tests/console_test.go
@@ -69,6 +69,13 @@ var _ = Describe("Console", func() {
 		Expect(err).ToNot(HaveOccurred())
 	}
 
+	OpenConsole := func(vmi *v1.VirtualMachineInstance) (expect.Expecter, <-chan error) {
+		By("Expecting the VirtualMachineInstance console")
+		expecter, errChan, err := tests.NewConsoleExpecter(virtClient, vmi, 30*time.Second)
+		Expect(err).ToNot(HaveOccurred())
+		return expecter, errChan
+	}
+
 	Describe("A new VirtualMachineInstance", func() {
 		Context("with a serial console", func() {
 			Context("with a cirros image", func() {
@@ -101,6 +108,31 @@ var _ = Describe("Console", func() {
 				for i := 0; i < 5; i++ {
 					ExpectConsoleOutput(vmi, "login")
 				}
+			}, 220)
+
+			It("should close console connection when new console connection is opened", func() {
+				vmi := tests.NewRandomVMIWithEphemeralDisk(tests.RegistryDiskFor(tests.RegistryDiskAlpine))
+
+				RunVMIAndWaitForStart(vmi)
+
+				By("opening 1st console connection")
+				expecter, errChan := OpenConsole(vmi)
+				defer expecter.Close()
+
+				By("expecting error on 1st console connection")
+				go func() {
+					defer GinkgoRecover()
+					select {
+					case receivedErr := <-errChan:
+						Expect(receivedErr.Error()).To(ContainSubstring("closed"))
+					case <-time.After(60 * time.Second):
+						Fail("timed out waiting for closed 1st connection")
+					}
+				}()
+
+				By("opening 2nd console connection")
+				ExpectConsoleOutput(vmi, "login")
+
 			}, 220)
 
 			It("should wait until the virtual machine is in running state and return a stream interface", func() {


ahadas:
CI failures seem unrelated:
`--- PASS: TestValidatingWebhook (0.05s)`

ahadas:
@cynepco3hahue I missed that somehow, thanks!

ahadas:
Right, this is something that was discussed in the context of supporting q35 in oVirt as well - the ability to use SATA instead of IDE should enable that.

cynepco3hahue:
@rmohr IIRC, windows image already have virtio drivers, so it will not cover our case(but worth to double check it).

rmohr:
@cynepco3hahue so they use sata disks right now, I saw it in the tests. If you did not install extra sata drivers (which I would find unbelievable) we should be good.

cynepco3hahue:
ci test please

cynepco3hahue:
@ahadas I see you need to run `make generate` and commit again.

ahadas:
@cynepco3hahue oh, right, I didn't do that after adding the last two commits - done.

rmohr:
ci test please

fabiand:
I got the request to make this filter configurable.

Could this be extended to take the filter pattern - or a list of allowed values from the kubevirt-config configmap?

michalskrivanek:
why do we even want to validate anything? It should be a configuration, not even changeable normally.
Also q35 is not valid for non x86 architectures

fabiand:
Currently we make silent assumptions about the primary machine typ, i.e. we set e1000e by default, which is only applicable to q35.

Validating this field is only the safety net for these kind implicit assumptions.

We need to know the baseline we operate on.

I'm not saying that we should prevent any other machine type altogether, I am merely saying that we should add some validation here and there to give the user some guarantees. And to also know what users are looking for.

IOW: A whitelist of machine types instead of full freedom.
We will need this moving forward in order to give guarantees about guest support.

ahadas:
> Could this be extended to take the filter pattern - or a list of allowed values from the kubevirt-config configmap?

Sure, looks like an over-specification at the moment but may be handy in the future.

> With https://github.com/fabiand/common-templates handling that I wonder how far you want to go in kubevirt. It doesn't make sense to maintain it in two places

Would all VMs be based on those templates? What about VMs that are converted from VMware or oVirt? VMs that are created from scratch?

>  (just use this on ppc64le and you're screwed - there's no q35* machine type, no e1000e at all)

Right, we would need to extend that filtering. We would probably need to code more things in order to support additional architectures though.

> (as much as we want Windows to work as well, I believe we should favor performance and compatibility with Linux)

So shouldn't it be OS-specific setting?

> It should allow whatever we can theoretically support because sooner or later someone always finds some use case for it

But is there a reason to think that we would be able to support a new chipset type by just allow setting it instead of q35? Our past experience in oVirt shows that supporting q35 while you only support i440fx (on x86_64) was not trivial to achieve. Even if we won't have that many things that are incompatible with the new chipset as we had in oVirt, the fact that we need to convert some VM configuration to the domain XML suggests that some coding would probably be needed.

fabiand:
The reason why we should have this is the testing matrix and implicit assumptions.

Right now our default machien type is q35. And for simplicity we want to focus around this machine type.

It is true that this will (initially!) prevent certain VMs which require non-q35 machien type to not be able to run on KubeVirt. however, there is work going on in v2v to allow migrating between machien types.

After all, we have the opportunity to start with a rather fresh (but 10yrs old) machien type (q35) and avoid to drag all legacy from i440 into KubeVirt.

Thus - For now let's limit to q35, this will help us to focus on this and make running VMs on this solid.
We'll need to update test cases to ensure that q35 is used as well.

> > It should allow whatever we can theoretically support because sooner or later someone always finds some use case for it
> 
> But is there a reason to think that we would be able to support a new chipset type by just allow setting it instead of q35? Our past experience in oVirt shows that supporting q35 while you only support i440fx (on x86_64) was not trivial to achieve. Even if we won't have that many things that are incompatible with the new chipset as we had in oVirt, the fact that we need to convert some VM configuration to the domain XML suggests that some coding would probably be needed.

I'm fully on your side. The only reason why I am not totally against making this configurable is that we can _try_ stuff and allow people to gather experience.

fabiand:
The change itself actually shows how it impacts certain configurations of the VM
If we do not limit us to q35, then the testing matrix is much larger.

Creating device topologies will be much larger. etc.

Btw - I consider this a API breaking change, thus something for #1576.

rmohr:
See above.

/hold

ahadas:
The supported emulated machines are now configurable via the kubevirt-config configmap. I guess we should consider unifying the module introduced in this PR with feature-gaps as they both read that configmap at some point

ahadas:
Not sure what happened but I probably updated my master branch (yeah, shouldn't use the master branch for this PR..) - I'll post a new PR

ahadas:
Posted #1718

PR #1515, opened 2018-09-17
Expose Prometheus metrics for kubevirt

### LINKED ISSUE #1077: virt-handler must not be ready if it can not list and/or watch VMs
There are cases where virt-handler is not ready or working but the very simple heartbeat mechanism still report virt-handler as ready and working.

One example where we encountered this situation is where virt-handler could not list and watch virtual machines but it could still update the nodes: https://github.com/kubevirt/kubevirt/issues/1071#issuecomment-392054031

rmohr:

Start exposing metrics for virt-handler, virt-controller and virt-api.
The metrics are accessible on the `/metrics` endpoint of every component.
The rest-client, workqueues and caches will be monitored.

This allows cluster monitoring and detecting issues, like if virt-handler on a specific node has issues updating its caches.

Errors like in #1077 will be counted like this:

```
rest_client_requests_total{code="<error>",host="127.0.0.1:33229",method="GET"} 81
```

***How to integrate with the prometheus-operator***

First deploying the prometheus-operator, then create a prometheus instance:

```yaml
apiVersion: monitoring.coreos.com/v1
kind: Prometheus
metadata:
  name: prometheus
spec:
  serviceAccountName: prometheus
  serviceMonitorSelector:
    matchLabels:
      prometheus.kubevirt.io: ""
  resources:
    requests:
      memory: 400Mi
```

Afterwards, it is sufficient to drop this `ServiceMonitor` definition into the cluster:

```yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: kubevirt
  labels:
    openshift.io/cluster-monitoring: ""
    prometheus.kubevirt.io: ""
spec:
  selector:
    matchLabels:
      prometheus.kubevirt.io: ""
  endpoints:
  - port: metrics
    scheme: https
    tlsConfig:
      insecureSkipVerify: true
```

This PR contains a `Service` called `kubevirt-prometheus-metrics` which will be referenced by the `ServiceMonitor` and all KubeVirt components which expose metrics have a registered `Endpoint` by that service. The prometheus-operator looks that up and generates the appropriate scrape config.

***How to integrate with the openshift cluster-monitoring-operator***

After the cluster-monitoring-operator is up and running deploy the `ServiceMonitor` definition from above. Because the definition contains the `openshift.io/cluster-monitoring` label, it should automatically be picket up by the cluster monitor.

***Why self-generated certificates***

In order to make integration as easy as possible all metrics are now exposed on the same port and via the same scheme in kubevirt. Since virt-pai requires TLS to register itself to the apiserver, TLS is used for all components.

CSRs would provide a solution to get signed certificates. However, right now getting them automatically signed would require us to our own auto-approver. In order to not explode the PR and to give some room to discuss this on its own, self-signed certificates are now used for all components. The metrics do not contain sensitive data.

Fixes #1077

This fixes #1077 and superseeds #1356. Instead of trying to detect errors and go in virt-handler to a non-responsive state, we simply expose issues and let the cluster interpret the data.

**Release note**:

```release-note
Add prometheus endpoints to virt-handler, virt-controller and virt-api. Workqueues, caches and REST clients are monitored.
```


fabiand:
Nice one!

rmohr:
@MarSik @yanirq FYI

rmohr:
Will add some tests which ensure that the metrics are really served and that our components are really registered to the mentioned service, but ready for a first review round.

rmohr:
Tests added. Ready for review. Should be done if CI is happy.

rmohr:
Documentation is here: https://github.com/kubevirt/user-guide/pull/122

fromanirh:
Hey, I'm late to this party, but no big deal because I like this PR a lot and I wanted it merged :)

Point is, for the Scale & Performance side of things, we would like to have even more metrics, for example we would like to know how much CPU and memory (PSS/RSS/VIRT) are the processes eating inside the "compute" container of the "virt-launcher" PODs. I have crafted something to obtain those metrics and to report to prometheus servers running inside the cluster, I think we can possibly evolve what Roman implemented here to achieve that.

What's the best place to discuss this possible future extension?

fabiand:
To me the foundations are now there, so if you got a PR which shows hwo you
would do this for other metrics, then please use a RP :)

PR #1516, opened 2018-09-17
Don't use tty in remote exec

slintes:

virtctl vnc didn't work with CRI-O, because the remote exec command in
virt-api converted LFs into CR + LF, which broke the VNC network protocol.
The root cause for this is that this seems to be a default behaviour
of tty, see https://github.com/moby/moby/issues/8513#issuecomment-216191236.
The solution is to not use tty, we don't need it.
(Still unclear why this only happened with CRI-O)

**Release note**:
```release-note
Fixed virtctl vnc for Openshift CRI-O
```

### DIFF
--- a/cmd/virt-launcher/sock-connector
+++ b/cmd/virt-launcher/sock-connector
@@ -29,5 +29,4 @@ if ! [ -S "$SOCKET" ]; then
 	exit 1
 fi
 
-stty -echo
-socat unix-connect:/$SOCKET stdio,cfmakeraw
+socat unix-connect:/$SOCKET stdio
--- a/pkg/virt-api/rest/subresource.go
+++ b/pkg/virt-api/rest/subresource.go
@@ -48,7 +48,36 @@ type SubresourceAPIApp struct {
 	VirtCli kubecli.KubevirtClient
 }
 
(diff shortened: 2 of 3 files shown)


slintes:
also see https://bugzilla.redhat.com/show_bug.cgi?id=1619218

slintes:
looking into the failed tests...

slintes:
@fabiand do we want do backport this to the `release-0.8` branch?

fabiand:
Yes, please prepare a patch.

slintes:
@fabiand done: https://github.com/kubevirt/kubevirt/pull/1518

gageorsburn:
This might have caused issues with something we were doing @slintes. I'm going to do some more looking into it but now when I VNC into a VMI I've been able to VNC into forever I get an error.

slintes:
@gageorsburn  That's interesting, because those bytes which your client (is it Chicken VNC on Mac?) is trying to interpret as auth type, is the Server protocol version "RFB 003.008". And this PR indeed fixed sending this server version on Openshift with CRI-O (though I don't fully understand yet why this was an issue with CRI-O only). But it was always send this way on Openshift and Kubernetes with Docker. Which type of cluster are you using? Is there an option in your VNC client to disable auth?

PR #1524, opened 2018-09-19
Handle the (unlikely) event of spoofed MAC address being same

booxter:
When we spoof a MAC address, we keep the first 3 octets the same (to keep the
vendor) but randomize the last 3 octets. We never guard against the (unlikely,
1 in 256^3 chance) event of the new address to be the same as permanent. This
may result in pod interface and the bridge carrying the same MAC address as the
VMI, which will break traffic flow.

This patch makes us try 10 times if the MAC address is for some reason returned
the same, and gracefully fail if we are so unlucky that we hit the jackpot of
256^(3*10). Perhaps at this point we should also buy a lottery ticket after
returning the error, but I will leave this for a follow-up patch... ;)

Related to bug 1501

**What this PR does / why we need it**: closing a tiny unlikely gap in podinterface MAC spoofing logic.

**Special notes for your reviewer**: this is more of a theoretical excercise, we haven't seen this in the wild. The patch is just a result of code inspection.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/network/common.go
+++ b/pkg/virt-launcher/virtwrap/network/common.go
@@ -37,6 +37,8 @@ import (
 	lmf "github.com/subgraph/libmacouflage"
 )
 
+const randomMacGenerationAttempts = 10
+
 type VIF struct {
 	Name    string
 	IP      netlink.Addr
@@ -103,7 +105,7 @@ func (h *NetworkUtilsHandler) GetMacDetails(iface string) (net.HardwareAddr, err
 	return currentMac, nil
 }
 
-// SetRandomMac changes the MAC address for a agiven interface to a randomly generated, preserving the vendor prefix
+// SetRandomMac changes the MAC address for a given interface to a randomly generated, preserving the vendor prefix
 func (h *NetworkUtilsHandler) SetRandomMac(iface string) (net.HardwareAddr, error) {
 	var mac net.HardwareAddr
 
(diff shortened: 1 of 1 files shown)


pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -112,18 +114,28 @@ func (h *NetworkUtilsHandler) SetRandomMac(iface string) (net.HardwareAddr, erro
...
-	if err != nil {
-		log.Log.Reason(err).Errorf("failed to spoof MAC for an interface: %s", iface)
-		return nil, err
-	}
+	changed := false
 
-	if changed {
-		mac, err = Handler.GetMacDetails(iface)
+	for i := 0; i < randomMacGenerationAttempts; i++ {
+		changed, err = lmf.SpoofMacSameVendor(iface, false)
```
vladikr: I can't remember.. does SpoofMacSameVendor guarantees that we are getting an address that doesn't equal to the old one?

booxter: My reading of it suggests that no, we don't. Hence the PR.

booxter: Oh wait you are asking *if* it returns "changed" that it's indeed changed? Yes, if changed is true, then we have a guarantee that the current MAC address is different from "persistent" MAC address which I think is enough.

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -112,18 +114,28 @@ func (h *NetworkUtilsHandler) SetRandomMac(iface string) (net.HardwareAddr, erro
...
 			return nil, err
 		}
-		log.Log.Reason(err).Errorf("updated MAC for interface: %s - %s", iface, mac)
+
+		if changed {
+			mac, err = Handler.GetMacDetails(iface)
+			if err != nil {
+				return nil, err
+			}
+			log.Log.Reason(err).Errorf("updated MAC for interface: %s - %s", iface, mac)
```
gageorsburn: Is this really an error? If I'm understanding correctly if we've made it to this point everything has gone as planned.

booxter: Yes. This code is present in previous version though.

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -37,6 +37,8 @@ import (
 	lmf "github.com/subgraph/libmacouflage"
 )
 
+const randomMacGenerationAttempts = 10
```
yuvalif: it is 1:16M chance. probably 2 attempts are enough?

PR #899, opened 2018-04-11
[WIP] [BLOCKED] Allow VM's to use block PVC's

stu-gott:
This feature is incomplete, but it's worth pushing for review early as there were some design related discoveries noted along the way -- that might need further review.
- virt-launcher cannot be responsible for mapping block devices (it's too late! the pod already exists)
- virt-launcher cannot contact Kubernetes to look up PVC information (it doesn't have permission to, by design)
- How do we communicate the proper device path from virt-controller to virt-launcher?
- We'll need another cache/informer to track PVC's in virt-controller

PR #1540, opened 2018-09-25
Block volume

### LINKED ISSUE #816: PVC disks fail if they are block PVs

slintes:

Support Blockvolume PVCs
Replaces https://github.com/kubevirt/kubevirt/pull/899, but credits for the initial version still go to @stu-gott :)

Fixes #816

This PR has 3 parts:
- modified disks-images-provider, for providing a block device backed by a cirros raw image for testing
- a PV and PVC consuming this block device via local-volume-provider for testing
- the actual code changes for checking if a PVC is in BlockVolume mode, and applying the needed configuration for virt-launcher and the domain

**Release note**:
```release-note
Support BlockVolume PVCs
```


pkg/virt-controller/services/template.go
```diff
@@ -97,6 +99,30 @@ func GetImagePullPolicy(store cache.Store) (policy k8sv1.PullPolicy, err error)
...
+// return true if a PersistentVolumeClaim uses VolumeMode = Block
+func isPVCBlock(namespace string, claimName string) (bool, error) {
+	// FIXME: using a client to make synchronous calls is suboptimal
+	// ideally this would be using a cache/informer
+	client, err := kubecli.GetKubevirtClient()
+	if err != nil {
+		return false, err
+	}
+	opts := metav1.GetOptions{}
+	pvc, err := client.CoreV1().PersistentVolumeClaims(namespace).Get(claimName, opts)
```
rmohr: Can you introduce an informer which watches for all PersistentVolumeClaims in the cluster? Everyone of us is just writing that we should do it properly. Someone has to start doing it properly.

slintes: ok

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -153,6 +153,10 @@ func Convert_v1_Volume_To_api_Disk(source *v1.Volume, disk *Disk, c *ConverterCo
 		return Convert_v1_HostDisk_To_api_Disk(source.HostDisk.Path, disk, c)
 	}
 
+	if source.PersistentVolumeClaim != nil {
```
mfranczy: Shouldn't we use HostDisk as a vehicle between virt-handler and virt-launcher? 
@slintes look at https://github.com/kubevirt/kubevirt/pull/1349#discussion_r205722177
@rmohr do you agree?

I think with the current state of this PR we still have an issue https://github.com/kubevirt/kubevirt/issues/991

pkg/virt-handler/vm.go
```diff
@@ -457,11 +457,19 @@ func (d *VirtualMachineController) execute(key string) error {
...
+	// Filesystem PersistenVolumeClaim is mounted into pod as directory from node filesystem
 	for i := range vmi.Spec.Volumes {
 		if volumeSource := &vmi.Spec.Volumes[i].VolumeSource; volumeSource.PersistentVolumeClaim != nil {
+
+			isBlockVolumePVC, err := hostdisk.IsBlockVolumePVC(volumeSource.PersistentVolumeClaim.ClaimName, vmi.Namespace, d.clientset)
+			if err != nil {
+				return err
+			}
+			if isBlockVolumePVC {
+				continue
```
mfranczy: If the ```BlockVolumePVC``` is empty then virt-launcher will not create a disk.img, a user will have to do it by himself.
Instead of skipping, maybe it's a good idea to get a real size of PVC (not over API but https://github.com/kubevirt/kubevirt/pull/1349#discussion_r205513560) and let it be converted to HostDisk?

stu-gott: I'm not sure we can introspect the contents of the volume in this case. Since it's block mode, it's not mounted/we can't tell if it's empty or not.

pkg/virt-controller/services/template.go
stu-gott: I'll take the blame for the synchronous call here. I had intended to switch to an informer once something end-to-end was working, but never got to that point.

pkg/virt-handler/vm.go
rmohr: Yeah a VMI can always write to the block-device without initialization from our side. The block-device is the "empty disk" already in this case.

pkg/virt-launcher/virtwrap/api/converter.go
rmohr: @mfranczy in this case we are fine. We don't need a size or something like that. The block device is ready to use.

pkg/virt-handler/vm.go
mfranczy: Ok, understood.

pkg/virt-launcher/virtwrap/api/converter.go
mfranczy: @slintes forget about the comment above.

pkg/virt-controller/services/template.go
slintes: PS: done

pkg/host-disk/host-disk.go
```diff
@@ -60,6 +60,19 @@ func GetPVCDiskImgPath(volumeName string) string {
 	return path.Join(pvcBaseDir, volumeName, "disk.img")
 }
 
+func IsBlockVolumePVC(pvcName string, namespace string, clientset kubecli.KubevirtClient) (bool, error) {
+	precond.CheckNotNil(pvcName)
+	precond.CheckNotEmpty(namespace)
+	precond.CheckNotNil(clientset)
+
+	pvc, err := clientset.CoreV1().PersistentVolumeClaims(namespace).Get(pvcName, metav1.GetOptions{})
+	if err != nil {
```
slintes: will use a informer / cache here too...

pkg/virt-controller/services/template.go
```diff
@@ -97,6 +99,36 @@ func GetImagePullPolicy(store cache.Store) (policy k8sv1.PullPolicy, err error)
...
+	// We do not need to consider the data in a PersistentVolume (as of Kubernetes 1.9)
+	// If a PVC does not specify VolumeMode and the PV specifies VolumeMode = Block
+	// the claim will not be bound. So for the sake of a boolean answer, if the PVC's
+	// VolumeMode is Block, that unambiguously answers the question
+	isBlock := pvc.Spec.VolumeMode != nil && *pvc.Spec.VolumeMode == k8sv1.PersistentVolumeBlock
+	return isBlock, nil
+}
+
+func (t *templateService) getPersistentVolumeClaim(namespace string, name string) (*k8sv1.PersistentVolumeClaim, error) {
+	if obj, exists, err := t.persistentVolumeClaimStore.GetByKey(namespace + "/" + name); err != nil {
```
stu-gott: +1

slintes:
Added PVC informer/cache on virt-controller and virt-launcher, fixed and improved tests, and rebased. Ready for another round of review.
@stu-gott @rmohr

pkg/host-disk/host-disk.go
slintes: done

cmd/virt-handler/virt-handler.go
```diff
@@ -130,6 +131,11 @@ func (app *virtHandlerApp) Run() {
 		cache.Indexers{},
 	)
 
+	pvcSharedInformer := cache.NewSharedIndexInformer(
```
rmohr: Can you revert that? We need a different solution for virt-handler. That would effectively mean that we have `N` watchers which will watch all PVCs for `N` nodes.

slintes: done

slintes:
these were the failing tests:
- openshift-3.10-release: 2nd node didn come up
- k8s-multus-1.11.1:
```
[Fail] Networking Multiple virtual machines connectivity should be able to reach [It] the Inbound VirtualMachineInstance 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:234

[Fail] Networking Multiple virtual machines connectivity should be able to reach [It] the Inbound VirtualMachineInstance with pod network connectivity explicitly set 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:234

[Fail] Networking Multiple virtual machines connectivity should be able to reach [It] the Inbound VirtualMachineInstance with custom MAC address 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:234

[Fail] Networking Multiple virtual machines connectivity should be able to reach [It] the internet 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:234

[Fail] Multus Networking VirtualMachineInstance with cni ptp plugin interface [It] should create a virtual machine with two interfaces 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_multus_test.go:113

[Fail] Expose Expose service on a VM Expose ClusterIP service [It] Should expose a Cluster IP service on a VMI and connect to it 
/root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:94

[Fail] Expose Expose UDP service on a VMI Expose ClusterIP UDP service [It] Should expose a ClusterIP service on a VMI and connect to it 
/root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:201

[Fail] Expose Expose UDP service on a VMI Expose NodePort UDP service [It] Should expose a NodePort service on a VMI and connect to it 
/root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:247

[Fail] Expose Expose service on a VMI replica set Expose ClusterIP service [It] Should create a ClusterIP service on VMRS and connect to it 
/root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:313

[Fail] Expose Expose service on an VM Expose ClusterIP service [It] Connect to ClusterIP services that was set when VM was offline 
/root/go/src/kubevirt.io/kubevirt/tests/expose_test.go:381
```
- k8s-1.10.4-release:
```
[Fail] Configurations VirtualMachineInstance definition with namespace memory limits above VMI required memory [It] should failed to start the VMI 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_configuration_test.go:169
```
- openshift-3.10-crio:
```
[Fail] DataVolume Integration Starting a VirtualMachineInstance with a DataVolume as a volume source using Alpine import [It] should be successfully started and stopped multiple times 
/root/go/src/kubevirt.io/kubevirt/tests/utils.go:263

[Fail] Console A new VirtualMachineInstance with a serial console [It] should wait until the virtual machine is in running state and return a stream interface 
/root/go/src/kubevirt.io/kubevirt/tests/console_test.go:113

[Fail] Configurations VirtualMachineInstance definition with namespace memory limits above VMI required memory [It] should failed to start the VMI 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_configuration_test.go:169
```
- openshift-3.10-multus:
```
[Fail] DataVolume Integration Starting a VirtualMachineInstance with a DataVolume as a volume source using Alpine import [It] should be successfully started and stopped multiple times 
/root/go/src/kubevirt.io/kubevirt/tests/utils.go:263

[Fail] Storage Starting a VirtualMachineInstance With VirtualMachineInstance with two PVCs [It] should start vmi multiple times 
/root/go/src/kubevirt.io/kubevirt/tests/utils.go:1381

[Fail] Configurations VirtualMachineInstance definition with namespace memory limits above VMI required memory [It] should failed to start the VMI 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_configuration_test.go:169
```
let's try again

slintes:
ci test please

pkg/util/types/pvc_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	pvcCache := cache.NewIndexer(cache.DeletionHandlingMetaNamespaceKeyFunc, nil)
+	pvcCache.Add(&filePvc1)
+	pvcCache.Add(&filePvc2)
+	pvcCache.Add(&blockPvc)
+
+	Context("PVC block device test", func() {
+		It("should detect filesystem device for empty VolumeMode", func() {
+			isBlock, err := IsPVCBlockFromStore(pvcCache, namespace, file1Name)
+			Expect(err).ToNot(HaveOccurred())
```
petrkotas: @slintes can you please provide labels for Expects? Without the labels, the fail messages are similar to 
`expected 3 to be nil` which is usually not helpful and does not provide a clue what failed.

I have been working on improving this in VM and VMI tests and doing it retrospectively is much harder than doing it when writing the code.

Thank you.

pkg/util/types/pvc_test.go
```diff
@@ -0,0 +1,85 @@
...
+		})
+
+		It("should detect filesystem device for filesystem VolumeMode", func() {
+			isBlock, err := IsPVCBlockFromStore(pvcCache, namespace, file2Name)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isBlock).To(Equal(false))
+		})
+
+		It("should detect block device for block VolumeMode", func() {
+			isBlock, err := IsPVCBlockFromStore(pvcCache, namespace, blockName)
```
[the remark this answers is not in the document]
slintes: done

pkg/virt-launcher/virtwrap/api/schema.go
```diff
@@ -305,6 +305,7 @@ type DiskSecret struct {
 type ReadOnly struct{}
 
 type DiskSource struct {
+	Dev           string          `xml:"dev,attr,omitempty"`
```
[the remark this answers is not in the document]
slintes: done

cmd/virt-handler/virt-handler.go
```diff
@@ -177,6 +178,12 @@ func (app *virtHandlerApp) Run() {
 	}
 }
 
+// resyncPeriod computes the time interval a shared informer waits before resyncing with the api server
```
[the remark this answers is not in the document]
slintes: deleted

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -153,6 +153,10 @@ func Convert_v1_Volume_To_api_Disk(source *v1.Volume, disk *Disk, c *ConverterCo
...
+		return Convert_v1_PersistentVolumeClaim_To_api_Disk(source.Name, source.PersistentVolumeClaim.ClaimName, disk, c)
```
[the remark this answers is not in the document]
slintes: done

pkg/util/types/pvc_test.go
slintes: done

tests/storage_test.go
```diff
@@ -431,7 +431,7 @@ var _ = Describe("Storage", func() {
 
 				By("Checking that the VirtualMachineInstance console has expected output")
 				expecter, err := tests.LoggedInCirrosExpecter(vmi)
-				Expect(err).To(BeNil())
+				Expect(err).To(BeNil(), "Cirros login successfully")
```
stu-gott: To be clear, Expect messages are only ever displayed if the expected condition was not met, so something along the lines of "expected to be able to log in to cirros" might be more clear.

By the way, I think we generally try to express error tests specifically:

`Expect(err).ToNot(HaveOccurred())`

It just changes the test output from "expected <random error> to equal nil" to something along the lines of "we didn't expect an error, but we got one. here it is....".

slintes: mh, about the expect message: isn't it convention that the message expresses what we expect? That's at least what @petrkotas does, and he suggested to add these messages. See e.g. https://github.com/kubevirt/kubevirt/pull/1531/files

Will change the err expectations, thanks for the heads up!

slintes:
@rmohr can you please have another look? Test failures seem to be random and unrelated to me

pkg/virt-handler/vm.go
```diff
@@ -466,11 +468,19 @@ func (d *VirtualMachineController) execute(key string) error {
...
 func (d *VirtualMachineController) replacePVCByHostDisk(vmi *v1.VirtualMachineInstance) error {
-	// if PVC is defined then it is replaced by HostDisk
-	// PersistenVolumeClaim is mounted into pod as directory from node filesystem
-	// it is not attached as independent block device
+	// if PVC is defined and it's not a BlockMode PVC, then it is replaced by HostDisk
+	// Filesystem PersistenVolumeClaim is mounted into pod as directory from node filesystem
 	for i := range vmi.Spec.Volumes {
 		if volumeSource := &vmi.Spec.Volumes[i].VolumeSource; volumeSource.PersistentVolumeClaim != nil {
+
+			isBlockVolumePVC, err := util.IsPVCBlockFromClient(d.clientset, vmi.Namespace, volumeSource.PersistentVolumeClaim.ClaimName)
```
rmohr: Please fetch the PV once, then check for the `volumeMode` on the PV (not PVC). It is the task of the storage provider to create a PV with the right `volumeMode` which matches the `PVC`, so it is ok to check on the PV.

pkg/virt-handler/vm.go
```diff
@@ -466,11 +468,19 @@ func (d *VirtualMachineController) execute(key string) error {
...
+			if err != nil {
+				return err
+			}
+			if isBlockVolumePVC {
+				continue
+			}
+
 			pvcSize, err := hostdisk.GetPVCSize(volumeSource.PersistentVolumeClaim.ClaimName, vmi.Namespace, d.clientset)
```
rmohr: Change this to pass in the PV directly, so that we don't fetch the PV again.

pkg/virt-handler/vm.go
rmohr: Edit: PVC is fine, misread the code. But still fetch it only once and pass it to both functions.

pkg/virt-controller/watch/vmi_test.go
```diff
@@ -153,9 +154,10 @@ var _ = Describe("VirtualMachineInstance watcher", func() {
 		dataVolumeInformer, dataVolumeSource = testutils.NewFakeInformerFor(&cdiv1.DataVolume{})
 		recorder = record.NewFakeRecorder(100)
 
-		configMapInformer, _ = testutils.NewFakeInformerFor(&k8sv1.Pod{})
+		configMapInformer, _ = testutils.NewFakeInformerFor(&k8sv1.ConfigMap{})
+		pvcInformer, _ = testutils.NewFakeInformerFor(&k8sv1.PersistentVolumeClaim{})
 		controller = NewVMIController(
-			services.NewTemplateService("a", "b", "c", configMapInformer.GetStore()),
+			services.NewTemplateService("a", "b", "c", configMapInformer.GetStore(), pvcInformer.GetStore()),
```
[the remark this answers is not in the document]
slintes: I have no idea on how to write a useful test for this over here...?
I will add a test in `template_test.go` for not existing PVC. Here we already have a test for adding the pod condition to the VMI when it can't be scheduled. I will also add a func test to `storage_test.go`, which tests if a non existing PVC will result in the correct VMI condition.

pkg/virt-controller/services/template.go
```diff
@@ -134,7 +138,22 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 			MountPath: filepath.Join("/var/run/kubevirt-private", "vmi-disks", volume.Name),
 		}
 		if volume.PersistentVolumeClaim != nil {
-			volumesMounts = append(volumesMounts, volumeMount)
+			logger := log.DefaultLogger()
+			isBlock, err := types.IsPVCBlockFromStore(t.persistentVolumeClaimStore, namespace, volume.PersistentVolumeClaim.ClaimName)
```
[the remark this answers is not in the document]
slintes: done

pkg/virt-handler/vm.go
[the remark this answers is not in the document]
slintes: I moved the `ReplacePVCByHostDisk` function to the `host-disk` package and added tests there

slintes: oh, good catch, fixed (see `ReplacePVCByHostDisk` in `host-disk` package now)

pkg/virt-handler/vm.go
slintes: see comment above

pkg/virt-controller/services/template.go
```diff
@@ -134,7 +138,25 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
 		if volume.PersistentVolumeClaim != nil {
-			volumesMounts = append(volumesMounts, volumeMount)
+			logger := log.DefaultLogger()
+			claimName := volume.PersistentVolumeClaim.ClaimName
+			_, exists, isBlock, err := types.IsPVCBlockFromStore(t.persistentVolumeClaimStore, namespace, claimName)
+			if err != nil {
+				logger.Errorf("error getting PVC: %v", claimName)
+				return nil, err
+			} else if !exists {
+				logger.Warningf("didn't find PVC %v", claimName)
```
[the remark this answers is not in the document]
slintes: Ouch, yes, that's bad indeed.
But should I create the `PodScheduled` condition manually? Feels wrong without having a pod.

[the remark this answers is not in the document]
rmohr: hm, at least that is what we would get if the pod would be created. Introducing a new condition does also not sound too good. Could you go with `PodScheduled` right now? We may need to consider our own `Scheduled` condition, not sure yet.

pkg/controller/virtinformers.go
```diff
@@ -210,6 +213,14 @@ func (f *kubeInformerFactory) ConfigMap() cache.SharedIndexInformer {
 	})
 }
 
+func (f *kubeInformerFactory) PersistentVolumeClaim() cache.SharedIndexInformer {
```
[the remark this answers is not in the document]
slintes: added to the application's leaderelection callback

pkg/virt-controller/services/template.go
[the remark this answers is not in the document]
slintes: done

slintes:
travis fails with an error which don't get locally.... will rebase and check

/hold

slintes:
fixed

/unhold

slintes:
hey prow, I edited my last comment... ;)

/hold cancel

fabiand:
Woot!

PR #1544, opened 2018-09-25
Improve reaction speed if virt-handler is unresponsive

### LINKED ISSUE #1077: virt-handler must not be ready if it can not list and/or watch VMs

rmohr:

* If virt-handler is unresponsive and pods are already in a final
  state, let the node controller immediately move the VMIs to failed
  state.
* Send events when node gets marked as unhealthy and send events for each
  VMI which gets moved to failed.
* Add an API field `status.reason`, similar to the one on pods and set it to `NodeUnresponsive` to indicate why the VMI was moved to Failed.

Fixes #1077 

**Release note**:

```release-note
Add the field status.reason which indicates the reason why the cluster may unexpectedly have stopped VMIs and improve the reaction speed of the node controller, in case virt-handler becomes unresponsive.
```

### DIFF
--- a/pkg/virt-controller/watch/node.go
+++ b/pkg/virt-controller/watch/node.go
@@ -23,6 +23,12 @@ import (
 	"kubevirt.io/kubevirt/pkg/log"
 )
 
+const (
+	// NodeUnresponsiveReason is in various places as reason to indicate that
+	// an action was taken because virt-handler became unresponsive.
+	NodeUnresponsiveReason = "NodeUnresponsive"
+)
+
 // NodeController is the main NodeController struct.
 type NodeController struct {
 	clientset        kubecli.KubevirtClient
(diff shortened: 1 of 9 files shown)


pkg/virt-controller/watch/node.go
```diff
@@ -23,6 +23,10 @@ import (
 	"kubevirt.io/kubevirt/pkg/log"
 )
 
+const (
+	NodeUnresponsiveVirtualMachineInstanceReason = "NodeUnresponsive"
```
slintes: Just a very little thing, but it catched my eyes: I think it's quite common to start these kind of const names with WHERE it's used (e.g. `VirtualMachineInstanceStatusReason`) and then the actual reason (`NodeUnresponsive`), so this would be `VirtualMachineInstanceStatusReasonNodeUnresponsive`. And because this could potentially be a reasonable status reason for other resources too, I would be fine with skipping `VirtualMachineInstance`, which also makes the name shorter. See e.g. https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L613

rmohr: Done.

PR #1567, opened 2018-09-30
Add vhost-net support

lkpdn:

For performance reasons, some want to get vhost-net to work for VMI.

Fixes #

I'd like some early feedback with opening this PR. If opening an Issue is more appropriate, please let me know.

**Release note**:

```release-note
Support vhost-net backend for virtio-net interface
```


ovirt-infra:

rmohr:
ok to test

rmohr:
ci test please

rmohr:
for reference: https://jenkins.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/2042/artifact/ci_build_summary.html

rmohr:
ci test please

pkg/virt-launcher/virtwrap/util/libvirt_helper.go
```diff
@@ -284,6 +284,11 @@ func SetupLibvirt() error {
 		return err
 	}
 
+	_, err = os.Stat("/dev/vhost-net")
```
rmohr: Why is that check here? If `/dev/vhost-net` is present on the node or not should be determined by the device plugin and not by virt-launcher. I think that you can remove that.

pkg/virt-launcher/virtwrap/api/converter_test.go
```diff
@@ -803,6 +803,38 @@ var _ = Describe("Converter", func() {
...
+				if driver == "" {
+					// If empty, then the domain defaults to 'vhost' if vhost device is present,
+					// but silently falls back to 'qemu' without error. (Since 0.8.8)
+					Expect(domain.Spec.Devices.Interfaces[i].Driver).To(BeNil())
+				} else {
+					Expect(domain.Spec.Devices.Interfaces[i].Driver.Name).To(Equal(driver))
+				}
+			}
+		},
+			table.Entry("one interface backed by default driver", []string{""}),
```
rmohr: Would it make sense for us to set the default explicitly? Just for transparency and just a nit.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -702,6 +702,13 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 				return err
 			}
 		}
+		if iface.Driver != "" {
+			if iface.Driver != "qemu" && iface.Driver != "vhost" {
+				log.Log.Infof("The network interface driver type of %s is not supported. Silently ignored and fallback to default one.", iface.Name)
```
rmohr: We should return an error. If the unsupported interface type gets that far we would from my opinion just mask a software bug.

pkg/virt-controller/services/template.go
```diff
@@ -561,6 +562,11 @@ func getRequiredResources(vmi *v1.VirtualMachineInstance) k8sv1.ResourceList {
 	if (vmi.Spec.Domain.Devices.AutoattachPodInterface == nil) || (*vmi.Spec.Domain.Devices.AutoattachPodInterface == true) {
 		res[TunDevice] = resource.MustParse("1")
 	}
+	for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
+		if iface.Driver == "vhost" {
```
rmohr: Do we want to introduce constants like `NetDriverVHost` and `NetDriverQEMU` which we can use in the converter and on the cluster side?

rmohr:
> ok.. so many failures, though I guess most of these are not relevant to this one, as far as I see in recent results for kubevirt_kubevirt_standard-check-pr, right?

Yes they are unrelated.

> 
> If noone has time to deal with these, I'm happy to do that. Or, is there some reason not to tackle these?
> E.g. about `Tests Suite.Configurations VirtualMachineInstance definition with namespace memory limits above VMI required memory should failed to start the VMI`, I guess we have to make a trade-off between introducing dirty polling(*1) in the test code and just ignoring the failure with its test code cleanness being kept, and the latter has been chosen for now. Just guessing. If I missed relevant discussion in the past, sorry about that.

That would be great. Others are looking into other tests already but that one should also be taken care of. I think that we should retry creating the VMI like you suggested. Retrying 5 times with one second delay should be good enough. If it then still fails we can be sure that there is an  issue in the cluster, because the informers should update *almost* immediately. It should normally always succeed on the second poll at last. 
> 
> (*1): to get expected result, we have to wait for limitrange to be filled in in the indexer on virt-api side, so to keep the code abstraction clean, IMHO there are no other choice than `wait.Poll(...` with its timeout set to say 30 seconds.

pkg/api/v1/schema.go
```diff
@@ -751,6 +751,9 @@ type Interface struct {
 	// If specified, the virtual network interface will be placed on the guests pci address with the specifed PCI address. For example: 0000:81:01.10
 	// +optional
 	PciAddress string `json:"pciAddress,omitempty"`
+	// Backend driver name.
+	// One of: qemu, vhost
+	Driver string `json:"driver,omitempty"`
```
yuvalif: should mark that "optional" defaulting to "qumu" for backward compatibility

vladikr:
Thank for the contribution. However, I don't think we should introduce the `driver` option to the api.
From my point of view we should always add the `/dev/vhost-net` to the virt-launcher container but then go with the libvirt default or __always__ set the domain interface driver to `vhost`.

Realistically, we want to use vhost-net with virtio, when the kernel supports, but in our case, we are always running on a kernel that does.

fabiand:
How would we add `/dev/vhost-net` to the launcher?

Do we care about heterogeneous clusters were some nodes support vhost and
some don't? (I don't think we should)

rmohr:
> Thank for the contribution. However, I don't think we should introduce the `driver` option to the api.
> From my point of view we should always add the `/dev/vhost-net` to the virt-launcher container but then go with the libvirt default or **always** set the domain interface driver to `vhost`.
> 
> Realistically, we want to use vhost-net with virtio, when the kernel supports, but in our case, we are always running on a kernel that does.

Ok, so this basically means that we can remove the user-facing changes on the api and keep the part which moves in the device. If I understand you correct we may even always ask for that device because we expect it always to be there ...

vladikr:
> Ok, so this basically means that we can remove the user-facing changes on the api and keep the part which moves in the device. If I understand you correct we may even always ask for that device because we expect it always to be there ...

Yep, exactly. I prefer to always ask for the device.

vladikr:
> I think we can not assume this. After all there are many kernel configurations, and after all this is an optional feature which we should not block on if not present.

All modern kernels support vhost. Other projects like opnestack assume exactly this.

lkpdn:
@rmohr @fabiand thanks for the review. Whether or not we expose choices between vhost/qemu to users, I will push fixes reflecting the review.

@vladikr thanks for your comment. I cannot remember exactly but I thought vhost-net had some characteristic where some flow could starve another in some situation. Have you got any thought about it?

That being said, even if that were the case, fairness between flows should be taken care on CNI side bandwidth limitation, I think. Utilising vhost-net "where possible" sounds reasonable, I don't have any "insightful" objection. I'm not sure things like `--use-emulation` for vhost is useful as well.

pkg/virt-launcher/virtwrap/util/libvirt_helper.go
lkpdn: My bad, thanks.

pkg/virt-launcher/virtwrap/api/converter_test.go
lkpdn: Seems not to me. With introducing const driver names in api side, I will remove these. If I misunderstood your point, please correct me later in a patch.

pkg/virt-launcher/virtwrap/api/converter.go
lkpdn: Totally true, will fix it.

pkg/virt-controller/services/template.go
lkpdn: That's reasonable. For me, `InterfaceDriverVhost` and `InterfaceDriverQEMU` seem more natural as enum members for `Interface`'s `Driver` if used on the cluster side.

lkpdn:
@rmohr 

> > ok.. so many failures, though I guess most of these are not relevant to this one, as far as I see in recent results for kubevirt_kubevirt_standard-check-pr, right?
> 
> Yes they are unrelated.
> 
> > If noone has time to deal with these, I'm happy to do that. Or, is there some reason not to tackle these?
> > E.g. about Tests Suite.Configurations VirtualMachineInstance definition with namespace memory limits above VMI required memory should failed to start the VMI, I guess we have to make a trade-off between introducing dirty polling(*1) in the test code and just ignoring the failure with its test code cleanness being kept, and the latter has been chosen for now. Just guessing. If I missed relevant discussion in the past, sorry about that.
> 
> That would be great. Others are looking into other tests already but that one should also be taken care of. I think that we should retry creating the VMI like you suggested. Retrying 5 times with one second delay should be good enough. If it then still fails we can be sure that there is an issue in the cluster, because the informers should update almost immediately. It should normally always succeed on the second poll at last.

All right, thanks.

pkg/api/v1/schema.go
lkpdn: Thanks, will fix it.

vladikr:
> @vladikr thanks for your comment. I cannot remember exactly but I thought vhost-net had some characteristic where some flow could starve another in some situation. Have you got any thought about it?

I'm not aware of anything like that. using vhost-net with virtio is default in other projects as well.

vladikr:
> That being said, even if that were the case, fairness between flows should be taken care on CNI side bandwidth limitation, I think. Utilising vhost-net "where possible" sounds reasonable, I don't have any "insightful" objection. I'm not sure things like `--use-emulation` for vhost is useful as well.

Yes, you are definitely right about  `--use-emulation`. We shouldn't ask for vhost-net with `--use-emulation` , same way as we don't ask for `/dev/kvm` and we should not request it on the pod spec.

lkpdn:
I don't mean to change the subject but about NET performance, I guess you are also watching for libvirt-side action on macvtap one (option 1 on https://www.redhat.com/archives/libvir-list/2018-July/msg00323.html) as Roman commented on it, so maybe both this vhost-net one and macvtap could be linked in one Issue. Just a side note.

michalskrivanek:
>  I guess that you normally have virtio present in the kernel if you do virtualization with qemu. @michalskrivanek do you have some insights?

surprisingly I don't:)

IMHO your assumption is reasonable. We will always have some limits on what we can support, and this is a small one to ask for.

rmohr:
> IMHO your assumption is reasonable. We will always have some limits on what we can support, and this is a small one to ask for.

If we add it as a requirement, it might make sense to allow disabling via the emulation flag (exists for deoming and development only) in our config map, to still make demoing in various environments simple: https://github.com/kubevirt/user-guide/blob/806368487ea23ca400d034806db4605be815d23a/installation/README.md#software-emulation. @vladikr @lkpdn

lkpdn:
> If we add it as a requirement, it might make sense to allow disabling via the emulation flag (exists for deoming and development only) in our config map, to still make demoing in various environments simple: 

That makes sense, I'm all for this. Covering the rare case where `/dev/vhost-net` is unavailable with existing `debug.useEmulation` in kubevirt-config seems the best option to me. I'll push a commit later.
Thanks.

lkpdn:
PTAL: https://github.com/kubevirt/kubevirt/pull/1567/commits/d5f9e2098cc6b7fbc30e165961d8715c0c02d25c
If rebase + fixup is better at the moment, please let me know.
Thanks for the review.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -642,15 +654,22 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
 			},
 			Alias: &Alias{
 				Name: iface.Name,
 			},
 		}
 
+		// if UseEmulation unset and at least one NIC model is virtio,
+		// /dev/vhost-net must be present as we should have asked for it.
+		if ifaceType == "virtio" && virtioNetProhibited {
+			return fmt.Errorf("In-kernel virtio-net device emulation '/dev/vhost-net' not present")
```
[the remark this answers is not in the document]
lkpdn: It will, but to make it consistent with the fact that without useEmulation we should have asked for it as a requirement if virtio-net interface is needed, I thought returning error is better in that corner case.

Sent from mobile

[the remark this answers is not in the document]
vladikr: I understand. I just don't understand how can we get into this situation. This code is running in a virt-launcher container. A device plugin had to add `/dev/vhost-net` to it prior to the launch. So to get here the device has to disappear somehow right after the container launch. Or am I missing something? 

I wouldn't do anything special in the converter at this stage and let libvirt do the right thing.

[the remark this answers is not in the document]
lkpdn: I get your point. It would be quite a peculiar case e.g. kernel module unloaded on the host side in the quite short window. Still possible, but we might be able to ignore it. OTOH I thought we covered similar situation for /dev/kvm in current code base. I'll check it later...

Sent from mobile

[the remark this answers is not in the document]
lkpdn: Sorry for late comment, I meant this part:
https://github.com/kubevirt/kubevirt/blob/d5f9e2098cc6b7fbc30e165961d8715c0c02d25c/pkg/virt-launcher/virtwrap/api/converter.go#L417
I think it's better to make what `useEmulation` indicates consistent. The assumption of "the device is always there when not specifying `useEmulation`" seems reasonable for both `/dev/kvm` and `/dev/vhost-net`.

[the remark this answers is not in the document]
rmohr: As I see it: It can happen, although it is unlikely without a coding error and it can definitely happen if we have an error in our cluster components. Therefor I prefer to be defensive.

rmohr:
ci test please

lkpdn:
@rmohr @vladikr @fabiand @yuvalif thanks for reviewing guys.

@rmohr 
> Now I just wonder if we shouldn't completely remove the model field on the interface on the cluster. We should discuss that, but it does not need to be done in this PR.

Sounds good. :)

fabiand:
Awesome to see this landing.

Does this actually enable DPDK usage? (me is a noob)

fabiand:
> Looks good. Now I just wonder if we shouldn't completely remove the `model` field on the interface on the cluster. We should discuss that, but it does not need to be done in this PR.

The model is required because we need to allow a user (or system) to set the right NIC model for a given guest.
There is no way for the system to identify a guest and infer the right model.

Well defined templates like https://github.com/kubevirt/common-templates will however "know" what guest will be running and can set the correct NIC model using our API.

lkpdn:
@fabiand this vhost-net support is not relevant to DPDK at the moment. For now I'm not sure DPDK could be beneficial because I think accelerating VMI-to-VMI communication (both on the same node) with e.g. DPDK + vhost-user while conforming to the policy enforced by CNI plugin would be needlessly complicated. In my opinion, improving latency + throughput between NIC inside VMI to veth exposed to virt-launcher seems more of importance, e.g. "direct" network mode (macvtap) to bypass a trip to/from a bridge device, this also meaning that no DNS inside launcher pod.

lkpdn:
> >  Looks good. Now I just wonder if we shouldn't completely remove the model field on the interface on the cluster. We should discuss that, but it does not need to be done in this PR.
> 
> The model is required because we need to allow a user (or system) to set the right NIC model for a given guest.
> There is no way for the system to identify a guest and infer the right model.
> 
> Well defined templates like https://github.com/kubevirt/common-templates will however "know" what guest will be running and can set the correct NIC model using our API.

I guess the intention was to suggest an option to make kubevirt UX more simple as possible while limiting support in a sensible way, e.g. _supporting virtio-net or e1000 capable VM only_.

To be honest, when I first met kubevirt I was thinking that kubevirt would cover the case where some old complicated system could not be easily converted to containers while having a great possibility that only k8s orchestration/scheduling part could improve resource utilization much, by virtue of kubevirt. So supporting ESXi (!) + **vmxnet3** would boost kubevirt as a solution for big enterprise companies. However 
michalskrivanek's comment above
> We will always have some limits on what we can support, and this is a small one to ask for.

makes sense to me now.

Issue #1478, opened 2018-08-28
Libvirt logs in virt-launcher are not structured


rmohr:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>

/kind bug

> /kind enhancement

**What happened**:

We stream libvirt logs to stdout of our `virt-launcher` container. The logs look like this:

```
2018-08-28 15:50:46.780+0000: 41: error : virNodeSuspendSupportsTarget:336 : internal error: Cannot probe for supported suspend types
2018-08-28 15:50:46.781+0000: 41: warning : virQEMUCapsInit:1229 : Failed to get host power management capabilities
```

**What you expected to happen**:

They should look somewhat like this:

```
level=error timestamp=2018-08-28T15:51:07.040751Z pos= virNodeSuspendSupportsTarget:336 component=virt-launcher msg="internal error: Cannot probe for supported suspend types"
```

**How to reproduce it (as minimally and precisely as possible)**:

Start a VMI and look at the virt-launcher logs.

**Anything else we need to know?**:

**Environment**:
- KubeVirt version (use `virtctl version`):
- Kubernetes version (use `kubectl version`):
- VM or VMI specifications:
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #1571, opened 2018-10-01
Collect output from all different loggers and log it  to our structured logger

### LINKED ISSUE #1479: logs from client-go are not logged to stdout

>
/kind bug
> /kind enhancement

**What happened**:

Increasing the verbosity level of virt-launcher does not increase the log verbosity for our structured logger and for client-go but the client-go logs are not written to stdout unless `--logtostderr` is not supplied too.

That we wrap glog in our codbase and only write logs via our structured logger.

rmohr:

 * Collect logs from the classic golang `log` package
 * Collect logs from `glog`
 * Collect libvirtd logs
 * Collect our structured logs

All sources are now written to stderr by all our components in a consistend structured log format.

Fixes #1479, #1478 

In order to make this work, this PR contains a re-implementation of the `glog` interface.

**Release note**:

```release-note
NONE
```


rmohr:
@cynepco3hahue could you have another look?

hack/dep-prune.sh
```diff
@@ -1,2 +1,5 @@
 #!/bin/bash
 for file in $(find vendor/ -name "*_test.go"); do rm ${file}; done
+rm -rf vendor/github.com/golang/glog
+mkdir -p vendor/github.com/golang/
+ln -s ../../../pkg/staging/glog/ vendor/github.com/golang/glog
```
slintes: Mh, interesting idea, using a symlink in the vendor dir. But it's also a bit hacky, isn't it? :)
What do you think about this:

- put the glog impl into its own github repository
- configure glog in glide.yaml with original `package` name, but our `repo` name

I didn't test this, but I think it should work

pkg/staging/glog/glog.go
```diff
@@ -0,0 +1,496 @@
...
+var LogLevelNames = map[LogLevel]string{
+	INFO:    "info",
+	WARNING: "warning",
+	ERROR:   "error",
+	FATAL:   "fatal",
+}
+
+var logger log2.Logger
+var v int
+var c string
```
slintes: why not using more meaningful var names?

hack/dep-prune.sh
rmohr: It would. However I would prefer if we keep it together. Note that k8s does the same. For instance `apimachinery` is symlinked into the `vendor` folder. There exists a repo for it too, but it is synced out of `kubevirt/kubevirt` and not used by it. I configured `glide` to ignored `glog`. I can actually remove the whole extra logic.

slintes: oh, I knew k8s does sth special with apimachinery etc., but wasn't aware that it's exactly this. That also explains the `staging` directory name you use. In that case I'm fine with it.
But would you mind adding a readme to the staging dir, which explains what's happening here (a short version of https://github.com/kubernetes/kubernetes/blob/master/staging/README.md)?

rmohr: Ok, can't remove it. Since glide does not track the package, it would delete the  symlink.

pkg/staging/glog/glog.go
rmohr: Improved.

hack/dep-prune.sh
rmohr: > But would you mind adding a readme to the staging dir, which explains what's happening here (a short version of

Done

cynepco3hahue:
Checked jobs, failures not relevant to PR.

PR #1575, opened 2018-10-02
Wait in virt-api for cache sync before reporting it as ready

rmohr:

virt-api should have an almost consistent view of the cluster-state
before it starts listening for connections and therefore marking intself
as ready. Otherwise we may see weird inconsistencies on bigger clusters where virt-api may need some time to do an initial sync.

Further remove all traces of the no longer used VMIInformer in virt-api.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-api/api.go
+++ b/pkg/virt-api/api.go
@@ -39,6 +39,7 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/cert"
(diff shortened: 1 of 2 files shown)


pkg/virt-api/api.go
```diff
@@ -952,10 +953,11 @@ func (app *virtAPIApp) Run() {
 
 	stopChan := make(chan struct{}, 1)
 	defer close(stopChan)
-	go webhookInformers.VMIInformer.Run(stopChan)
```
slintes: maybe I'm missing something obvious, but why is this informer not started anymore?

rmohr: Because it is a left-over. We don't use it in virt-api. Updated the PR description.

slintes: ah, its use in the webhook was removed with a recent PR, didn't have that locally yet, sorry.

PR #1580, opened 2018-10-03
[WIP] Added custom dhcp options to vmi interface spec.

### LINKED ISSUE #1579: Custom DHCP Options
As a virtual network operator, I want to be able to pass custom dhcp arguments such as `dhcp.OptionTFTPServerName` so that virtual network appliances will be able to load configuration. An extendable solution should be implemented so that any dhcp options that do not interfere with what kubevirt interface setup requires.

As an example the api could look like,

```
spec:
  domain:
    devices:
      disks:
      - disk:
          bus: virtio
        name: registrydisk
        volumeName: registryvolume
      interfaces:
      - bridge: {}
        name: default
        dhcpOptions: 
        - dhcpCode: 150
           value: tftp.default.svc.cluster.local
```

gageorsburn:
This PR adds custom DHCP options to interface config to set options that might be needed. 

Fixes #1579

This PR still needs:

- Documentation
- Tests
- Checks to make sure important options set by kubevirt aren't being overriden
- Example?

**Release note**:

```release-note

```

### DIFF
--- a/pkg/api/v1/schema.go
+++ b/pkg/api/v1/schema.go
@@ -768,6 +768,15 @@ type Interface struct {
 	// If specified, the virtual network interface will be placed on the guests pci address with the specifed PCI address. For example: 0000:81:01.10
 	// +optional
 	PciAddress string `json:"pciAddress,omitempty"`
+	// Custom DHCP options that may be needed.
+	// +optional
+	DHCPOptions []DHCPOption `json:"dhcpOptions,omitempty"`
+}
+
+// +k8s:openapi-gen=true
+type DHCPOption struct {
+	OptionCode  byte   `json:"optionCode"`
+	OptionValue string `json:"optionValue"`
 }
 
 // Represents the method which will be used to connect the interface to the guest.
--- a/pkg/virt-launcher/virtwrap/api/schema.go
+++ b/pkg/virt-launcher/virtwrap/api/schema.go
@@ -406,6 +406,12 @@ type Interface struct {
 	LinkState           *LinkState       `xml:"link,omitempty"`
 	FilterRef           *FilterRef       `xml:"filterref,omitempty"`
 	Alias               *Alias           `xml:"alias,omitempty"`
+	DHCPOptions         []DHCPOption     `xml:"dhcpOptions,omitempty"`
+}
+
+type DHCPOption struct {
+	OptionCode  byte   `xml:"optionCode"`
+	OptionValue string `xml:"optionValue"`
 }
 
 type LinkState struct {
--- a/pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
+++ b/pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
@@ -35,6 +35,7 @@ import (
 	"os"
 
 	"kubevirt.io/kubevirt/pkg/log"
+	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
 )
 
 const (
@@ -57,7 +58,8 @@ func SingleClientDHCPServer(
 	dnsIPs [][]byte,
 	routes *[]netlink.Route,
 	searchDomains []string,
-	mtu uint16) error {
+	mtu uint16,
+	customDHCPOptions []api.DHCPOption) error {
 
 	log.Log.Info("Starting SingleClientDHCPServer")
 
@@ -90,6 +92,12 @@ func SingleClientDHCPServer(
 		return fmt.Errorf("reading the pods hostname failed: %v", err)
 	}
 	dhcpOptions[dhcp.OptionHostName] = []byte(hostname)
+
+	for _, customDHCPOption := range customDHCPOptions {
+		log.Log.Infof("Setting dhcp option %s to %s", customDHCPOption.OptionCode, customDHCPOption.OptionValue)
+		optionCode := dhcp.OptionCode(customDHCPOption.OptionCode)
+		dhcpOptions[optionCode] = []byte(customDHCPOption.OptionValue)
+	}
 
 	handler := &DHCPHandler{
 		clientIP:      clientIP,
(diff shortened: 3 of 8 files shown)


ovirt-infra:

pkg/api/v1/schema.go
```diff
@@ -768,6 +768,15 @@ type Interface struct {
...
 	PciAddress string `json:"pciAddress,omitempty"`
+	// Custom DHCP options that may be needed.
+	// +optional
+	DHCPOptions []DHCPOption `json:"dhcpOptions,omitempty"`
+}
+
+// +k8s:openapi-gen=true
+type DHCPOption struct {
+	OptionCode  byte   `json:"optionCode"`
+	OptionValue string `json:"optionValue"`
```
rmohr: Would `OptionValue` have to bye of type `[]byte`?

pkg/virt-launcher/virtwrap/api/schema.go
```diff
@@ -406,6 +406,12 @@ type Interface struct {
 	LinkState           *LinkState       `xml:"link,omitempty"`
 	FilterRef           *FilterRef       `xml:"filterref,omitempty"`
 	Alias               *Alias           `xml:"alias,omitempty"`
+	DHCPOptions         []DHCPOption     `xml:"dhcpOptions,omitempty"`
```
rmohr: Do we not have access to the vmi definition where we start the dhcp server?

pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
```diff
@@ -91,6 +93,12 @@ func SingleClientDHCPServer(
 	}
 	dhcpOptions[dhcp.OptionHostName] = []byte(hostname)
 
+	for _, customDHCPOption := range customDHCPOptions {
+		log.Log.Infof("Setting dhcp option %s to %s", customDHCPOption.OptionCode, customDHCPOption.OptionValue)
+		optionCode := dhcp.OptionCode(customDHCPOption.OptionCode)
+		dhcpOptions[optionCode] = []byte(customDHCPOption.OptionValue)
```
rmohr: I guess that works for a lot of options. At least for the search domains array it would not work.

pkg/virt-launcher/virtwrap/api/schema.go
gageorsburn: I don't think we do or at least I didn't see it thats why I used the xml schema as kind of a holding area a bit weird I know. If you know of better ways would love to change it.

gageorsburn:
@rmohr The only option I'm interested in modifying at this time is TFTP Address or option 66. I figured it would be more flexible to add in something like this rather than the one option.

pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
gageorsburn: I plan on adding a conditional that doesn't allow any options kubevirt set to be overriden.

pkg/virt-launcher/virtwrap/api/schema.go
rmohr: I think it makes sense to extend `StartDHCP` at https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/network/common.go#L61 to also pass in extra options, or pass in the VMI. Something along that line.

rmohr:
> @rmohr The only option I'm interested in modifying at this time is TFTP Address or option 66. I figured it would be more flexible to add in something like this rather than the one option.

I can see a few potential complications with the completely generic case:

 * We don't want to allow overriding all options
 * Some fields may need to be encoded in a special way
 * Probably only a small subset of the options makes sense to be modified per VMI

wouldn't it make sense to make the types explicit? That would have the following advantages:

 * Specifying options which are not allowed to be overwritten can not even be specified
 * Since we know what option can come in, we can ensure especially for them that we encode them right (instead of having to go through the whole spec)
 * We can whitelist them on a use-case basis, adding more fields would still be a compatible API change. We could do that also in the generic case via the validation webhook, but it might be confusing.

What do you think?

pkg/api/v1/schema.go
```diff
@@ -768,6 +768,15 @@ type Interface struct {
 	// If specified, the virtual network interface will be placed on the guests pci address with the specifed PCI address. For example: 0000:81:01.10
 	// +optional
 	PciAddress string `json:"pciAddress,omitempty"`
+	// Custom DHCP options that may be needed.
+	// +optional
+	DHCPOptions []DHCPOption `json:"dhcpOptions,omitempty"`
```
rmohr: The DHCPOptions seems to be valid only for the `PodNetwork` (bridge and maybe slirp). Would it make sense to move this to the `PodNetwork` struct?

pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
gageorsburn: Search domain is an option kubevirt sets so it would be excluded from this.

pkg/api/v1/schema.go
gageorsburn: It would make more sense for it to be `[]byte` just not sure how/where to do that translation from yaml to []byte.

gageorsburn:
More than happy to change this to be just TftpAddress instead of the generic use case. Either way will suite my use case. What does the general community think? I can bring this PR up at the next community meeting to get broader thoughts on approach,

rmohr:
That topic was not discussed before. Bringing it up at the next community meeting sounds great.

gageorsburn:
Going to close this and add in TftpAddress option specifically

gageorsburn:
@justinbarrick I think it would be fine. This isn't the PR that ended up getting merged so it doesn't follow this style.

Issue #906, opened 2018-04-12
Missing search domains in windows vm (Windows does not support DHCP option 119) (not a bug)


karmab:
when i deploy a windows vm, i note that my network config misses search domains (checking with ipconfig /all).
this is present on linux boxes on the same cluster.

mlsorensen:
Are we sure windows DHCP client supports option 119?

fabiand:
> In short, there is no way to assign a DNS suffix search list to a Windows client using DHCP.
> A DNS suffix search list can be assigned to domain-joined Windows machines via Group Policy by following the procedure in Configure the DNS Suffix Search List for a Disjoint Namespace.

Source: http://www.dell.com/support/article/de/de/debsdt1/sln291195/unable-to-configure-dns-suffix-search-list-via-dhcp?lang=en

Hum.

- Documentation for this
- Manual workaround
- Implementing Group Policies in KubeVirt - I think this is to stretched :)

fabiand:
Summarizing the impact again:

- Windows guests are not able to lookup _cluster_ service (and other entities with a DNS entry) by their "short name" i.e. `myservice`. Instead they need to use the full fqdn: `myservice.default.svc.cluster` (IIRC)
- DNS resolution of the broader internet is not affected

fabiand:
Side note - this problem is not specific to kubevirt, it applies to any windows guest.
Although it might affect guests on kubevirt more, as kubernetes is using DNS records extensively for service discovery.

fabiand:
@karmab you can not resolve any service name into an ip?

fabiand:
@duyanyan are you able to reproduce the issue @karmab is encountering?

rmohr:
Are we sending option 015 right now? If we send e.g. svc.local there,  it should be picket up and fqdn should work for k8s services.

vladikr:
> Are we sending option 015 right now?

Nope, we are sending a list via option 119 (RFC 3397) and it looks like windows doesn't support it at all...
However, I think option 015 is only intended to be a single domain name

duyanyan:
Yes, I could reproduce it. when I check the vms, seems there are no search domain in both windows and cirros vm:
```
# ssh cirros@172.30.234.1
cirros@172.30.234.1's password: 
$ cat /etc/resolv.conf 
nameserver 172.30.0.1
$ ping http
ping: bad address 'http'
```
But exist inside a normal pod at the same cluster
```
# oc rsh hello-pod
[...]

PR #1582, opened 2018-10-04
Set option 15 for windows machines on DHCP requests

rmohr:

In order to allow service-name DNS resolution not only via the FQDN, it
is necessary to put the VMI into a domain. This domain equals to the
search domain entry of the namespace the VMI runs in.

As a consequence, services in the namespace of the VMI can be resolved

 * via the FQDN: `myservice.mynamespace.svc.cluster.local`
 * relative to the namespaces: `myservice`

Headless services can also be accessed

 * via the FQDN: `myservice.mydomain.mynamespace.svc.cluster.local`
 * relative to the namespace: `myservice.mydomain`

Before only FQDN was possible.

Related to https://github.com/kubevirt/user-guide/issues/126 and #906 

**Release note**:

```release-note
Allows Windows VMIs DNS resolution of services without FQDN in their own namespace
```

### DIFF
--- a/pkg/util/net/dns/dns_suite_test.go
+++ b/pkg/util/net/dns/dns_suite_test.go
@@ -1,0 +1,13 @@
+package dns_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"testing"
+)
+
+func TestDns(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Dns Suite")
+}
--- a/pkg/util/net/dns/resolveconf.go
+++ b/pkg/util/net/dns/resolveconf.go
@@ -1,0 +1,73 @@
+package dns
+
+import (
+	"bufio"
+	"net"
+	"regexp"
+	"strings"
+)
+
+const (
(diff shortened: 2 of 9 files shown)


rmohr:
ci test please

rmohr:
ci test please

rmohr:
should be ok now if CI passes

/hold cancel

cynepco3hahue:
@rmohr do you still some work here or can we merge it? I just see `debug` as commit message :smile:

rmohr:
@cynepco3hahue there seems to be an issue with my tests. Sometims I see that the command returns the exact value, sometimes it never manages to run the winrm command at all and sometimes the command returns not the expected root domain.

rmohr:
@cynepco3hahue turns out that it can take some time until the domain is set, where it is already possible to connect to the vmi via winrm. Changed the test to cope with that. Let's see.

rmohr:
@cynepco3hahue looks good. Could you do another review?

rmohr:
> Looks great. Just weird that all k8s multus networking tests failed. Still doesn't seem to be related to the change, I think.

Yes they should be back in working state afer #1583 is merged.

PR #1592, opened 2018-10-09
Extend and refactor template test

gszasz:
This PR adds the following functionality:
- [x] Refactor and extend `tests.RunCommand()` function to be more verbose, to allow separate handling of `stdout` and `stderr`, and to confine the executed external command to clearly defined namespace.
- [x] Implement new function `test.RunCommandPipe()` function that allows executing a set of commands chained by stdout pipe. It follows the logic of the new `tests.RunCommand()`.
- [x] Extend and refactor `tests/template_test.go`, simplify code structure and add more test case to comply actual test plan, optimize test time, add the proper annotation to all Gomega matcher calls to simplify debugging.

We need to this PR to optimize test execution and to extend test coverage. This PR also sets the ground for future test automation. The Template test was also modified to be backward compatible with the older KubeVirt releases; thus it is recommendable to include this test into cherry-picked backports.

I provide a detailed description of the partial contributions in the corresponding commit messages. All commits were rebased on top of `master` branch.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/template_test.go
+++ b/tests/template_test.go
@@ -248,7 +248,7 @@ var _ = Describe("Templates", func() {
 		AfterEach(AssertTestCleanupSuccess())
 
 		Context("with Fedora Template", func() {
-			BeforeEach(AssertTemplateSetupSuccess(vmsgen.GetTestTemplateFedora(), nil))
+			BeforeEach(AssertTemplateSetupSuccess(vmsgen.GetTemplateFedora(), nil))
 
 			AssertTemplateTestSuccess()
 		})
--- a/tests/template_test.go
+++ b/tests/template_test.go
@@ -22,17 +22,29 @@ package tests_test
 import (
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"os"
+	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+
 	"kubevirt.io/kubevirt/pkg/kubecli"
 	"kubevirt.io/kubevirt/tests"
 	vmsgen "kubevirt.io/kubevirt/tools/vms-generator/utils"
+)
+
+const (
+	defaultNamePrefix = "testvm-"
+	defaultCPUCores   = "2"
+	defaultMemory     = "2Gi"
 )
 
 var _ = Describe("Templates", func() {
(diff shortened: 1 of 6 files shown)


tests/template_test.go
```diff
@@ -22,221 +22,244 @@ package tests_test
...
+					"CPU_CORES": defaultCPUCores,
+					"MEMORY":    defaultMemory,
+				}
+				templateFile = ""
+				ExpectWithOffset(1, templateParams).To(HaveKeyWithValue("NAME", Not(BeEmpty())), "invalid NAME parameter: VirtualMachine name cannot be empty string")
+				ExpectWithOffset(1, templateParams).To(HaveKeyWithValue("CPU_CORES", MatchRegexp(`^[0-9]+$`)), "invalid CPU_CORES parameter: %q is not unsigned integer", templateParams["CPU_CORES"])
+				ExpectWithOffset(1, templateParams).To(HaveKeyWithValue("MEMORY", MatchRegexp(`^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$`)), "invalid MEMORY parameter: %q is not valid quantity", templateParams["MEMORY"])
+				vmName = templateParams["NAME"]
+				vm, err := virtClient.VirtualMachine(tests.NamespaceTestDefault).Get(vmName, &metav1.GetOptions{})
+				ExpectWithOffset(1, errors.IsNotFound(err) || vm.ObjectMeta.DeletionTimestamp != nil).To(BeTrue(), "invalid NAME parameter: VirtualMachine %q already exists", vmName)
```
cynepco3hahue: If you want to be sure that VM does not exist, you need to check only `errors.IsNotFound(err)`, when object has `vm.ObjectMeta.DeletionTimestamp != nil` it still exists under the `etcd` and creation of the object with the same name will fail, on the other side I can see that you use `rand.String(12)`, so I do not sure if you need this check at all.

tests/template_test.go
```diff
@@ -22,221 +22,244 @@ package tests_test
...
-					out, err := runOcGetCommand("vms")
-					ExpectWithOffset(1, err).ToNot(HaveOccurred())
-					return out == "No resources found.\n"
-				}, time.Duration(60)*time.Second).Should(BeTrue(), "Timed out waiting for VM to disappear")
+				if vm, err := virtClient.VirtualMachine(tests.NamespaceTestDefault).Get(vmName, &metav1.GetOptions{}); err == nil && vm.ObjectMeta.DeletionTimestamp == nil {
+					By("Deleting the VirtualMachine")
+					ExpectWithOffset(1, virtClient.VirtualMachine(tests.NamespaceTestDefault).Delete(vmName, &metav1.DeleteOptions{})).To(Succeed(), "failed to delete VirtualMachine %q: %v", vmName, err)
+					EventuallyWithOffset(1, func() bool {
+						obj, err := virtClient.VirtualMachine(tests.NamespaceTestDefault).Get(vmName, &metav1.GetOptions{})
+						return errors.IsNotFound(err) || obj.ObjectMeta.DeletionTimestamp != nil
```
cynepco3hahue: above true also for this check

tests/template_test.go
gszasz: This idea for this check comes from the stress test we have designed with @petrkotas in the past 

```go
Context("without waiting for an active pod", func() {
    It("should be able to delete VirtualMachineInstance promptly over 100 attempts", func() {
        for i := 0; i < 100; i++ {
            By(fmt.Sprintf("Starting attempt %d", i+1))
            By("Creating the VirtualMachineInstance")
            obj, err := virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
            Expect(err).ToNot(HaveOccurred(), "Failed to create VirtualMachineInstance")
            podSelector := tests.UnfinishedVMIPodSelector(vmi)

            By("Deleting the VirtualMachineInstance")
            Expect(virtClient.VirtualMachineInstance(vmi.Namespace).Delete(obj.Name, &metav1.DeleteOptions{})).To(Succeed())
 
            By("Verifying that VirtualMachineInstance's pod is deleted")
            Eventually(func() int {
                pods, err := virtClient.CoreV1().Pods(tests.NamespaceTestDefault).List(podSelector)
                return len(pods.Items)
            }, 75, 0.5).Should(Equal(0), "VirtualMachineInstance's pod is still active")

            By("Checking that the VirtualMachineInstance is either deleted or has received deletion timestamp")
            Eventually(func() bool {
                _, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(obj.Name, &metav1.GetOptions{})
                return err != nil
            }, 30, 0.5).Should(BeTrue(), "VirtualMachineInstance still exists and has not received deletion timestamp")
        }
    })
})
```

We found out that if you try to delete a VMI promptly over 100 attempts *without waiting for an active pod*, you can experience a conflict with having a VMI object still hanging in the `etcd` with deletion timestamp set, but still waiting for the garbage collector. The race condition was resolved by modifying the return value to:

```go
`return err != nil || obj.ObjectMeta.DeletionTimestamp != nil`
```

Although this evidence is purely empirical, I have decided to implement this check into the refactored Template test. You might be right, but I need to do some research to figure this out.

gszasz:
@cynepco3hahue I got your point, and I will most likely refactor the `RunCommandPipeWithNS()` function within the scope a consequent PR. Thank you for your suggestion.

tests/template_test.go
gszasz: I have discussed this issue with @davidvossel and it turned out you were right. The check for deletion timestamp is completely useless because the API server ignores deletion timestamp anyway. Thanks a lot for your input. I will open a new PR to fix it.

PR #1595, opened 2018-10-09
Report mac addresses in vmi status interfaces section

mmirecki:
Report mac addresses in vmi status interfaces section
Preserve the ip address given to eth0 by dhcp
The patch recognizes the 'primary' interface (with ip set by our dhcp code) by it's name "k6t-eth0"

```release-note
VMI status is updated with a list of all network interfaces added to the vmi.
Initially each interface consists of the interface name, and the IP for the pod interface.
After the vmi is started and additional data is received from libvirt, the interfaces in the
vmi status are populated with mac addresses.
```

### DIFF
--- a/pkg/api/v1/types.go
+++ b/pkg/api/v1/types.go
@@ -293,6 +293,9 @@ type VirtualMachineInstanceNetworkInterface struct {
 	IP string `json:"ipAddress,omitempty"`
 	// Hardware address of a Virtual Machine interface
 	MAC string `json:"mac,omitempty"`
+	// Name of the interface, corresponds to name of the network assigned to the interface
+	// TODO: remove omitempty, when api breaking changes are allowed
+	Name string `json:"name,omitempty"`
 }
 
 type VirtualMachineInstanceMigrationState struct {
--- a/pkg/virt-controller/watch/vmi.go
+++ b/pkg/virt-controller/watch/vmi.go
@@ -310,11 +310,15 @@ func (c *VMIController) updateStatus(vmi *virtv1.VirtualMachineInstance, pod *k8
 			if isPodOwnedByHandler(pod) {
 				// vmi is still owned by the controller but pod is already handed over,
 				// so let's hand over the vmi too
-				vmiCopy.Status.Interfaces = []virtv1.VirtualMachineInstanceNetworkInterface{
-					{
-						IP: pod.Status.PodIP,
-					},
+				interfaces := make([]virtv1.VirtualMachineInstanceNetworkInterface, 0)
+				for _, network := range vmi.Spec.Networks {
+					if network.NetworkSource.Pod != nil {
+						ifc := virtv1.VirtualMachineInstanceNetworkInterface{Name: network.Name, IP: pod.Status.PodIP}
+						interfaces = append(interfaces, ifc)
+					}
 				}
+				vmiCopy.Status.Interfaces = interfaces
+
 				vmiCopy.Status.Phase = virtv1.Scheduled
 				if vmiCopy.Labels == nil {
 					vmiCopy.Labels = map[string]string{}
--- a/pkg/virt-handler/vm.go
+++ b/pkg/virt-handler/vm.go
@@ -237,6 +237,32 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 
 	oldStatus := vmi.DeepCopy().Status
 
+	if domain != nil {
+
+		// This is needed to be backwards compatible with vmi's which have status interfaces
+		// with the name not being set
+		if len(vmi.Status.Interfaces) == 1 && vmi.Status.Interfaces[0].Name == "" {
+			for _, network := range vmi.Spec.Networks {
+				if network.NetworkSource.Pod != nil {
+					vmi.Status.Interfaces[0].Name = network.Name
+				}
+			}
+		}
+
+		interfacesByName := make(map[string]int)
+		for i, existingInterface := range vmi.Status.Interfaces {
+			interfacesByName[existingInterface.Name] = i
+		}
+
+		for _, domainInterface := range domain.Spec.Devices.Interfaces {
+			if i, exists := interfacesByName[domainInterface.Alias.Name]; exists {
+				vmi.Status.Interfaces[i].MAC = domainInterface.MAC.MAC
+			} else {
+				vmi.Status.Interfaces = append(vmi.Status.Interfaces, v1.VirtualMachineInstanceNetworkInterface{MAC: domainInterface.MAC.MAC, Name: domainInterface.Alias.Name})
+			}
+		}
+	}
+
 	// Only update the VMI's phase if this node owns the VMI.
 	if vmi.Status.NodeName != "" && vmi.Status.NodeName != d.host {
 		// not owned by this host, likely the result of a migration
--- a/tests/vmi_multus_test.go
+++ b/tests/vmi_multus_test.go
@@ -29,6 +29,7 @@ import (
 	. "github.com/onsi/gomega"
 	k8sv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v13 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"kubevirt.io/kubevirt/pkg/api/v1"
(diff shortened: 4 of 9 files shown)


ovirt-infra:

pkg/virt-handler/vm.go
```diff
@@ -181,6 +181,31 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 
 	oldStatus := vmi.DeepCopy().Status
 
+	if domain != nil {
+		var interfaces = vmi.DeepCopy().Status.Interfaces
```
rmohr: I think the `VirtualMachineInstanceNetworkInterface` is missing a `Name` field. It should look like this:
```yaml
type VirtualMachineInstanceNetworkInterface struct {
        // Name of the interface this status belongs to
        Name string `json:"name"`
	// IP address of a Virtual Machine interface
	IP string `json:"ipAddress,omitempty"`
	// Hardware address of a Virtual Machine interface
	MAC string `json:"mac,omitempty"`
}
```

As a consequence users and our application can reliably detect which interface status belongs to which interface. Since also the interface name matches the qemu device alias on the domain specification, we don't have to guess on the bridge name then.

@mmirecki could you add that field?

mmirecki: sure

pkg/api/v1/types.go
```diff
@@ -295,6 +295,8 @@ type VirtualMachineInstanceNetworkInterface struct {
 	IP string `json:"ipAddress,omitempty"`
 	// Hardware address of a Virtual Machine interface
 	MAC string `json:"mac,omitempty"`
+	// Name of the interface, corresponds to the pod bridge the interface is
+	Name string `json:"name,omitempty"`
```
mmirecki: The here is actually the name of the bridge to which the vnic is connected on the pod.
Should we name it BridgeName?

pkg/virt-controller/watch/vmi.go
```diff
@@ -324,7 +326,8 @@ func (c *VMIController) updateStatus(vmi *virtv1.VirtualMachineInstance, pods []
 				// so let's hand over the vmi too
 				vmiCopy.Status.Interfaces = []virtv1.VirtualMachineInstanceNetworkInterface{
 					{
-						IP: pod.Status.PodIP,
+						IP:   pod.Status.PodIP,
+						Name: podInterface0_name,
```
rmohr: You should use here `volumes[x].Name` or `devices.interfaces[x].Name` of the volume/interface which is connected to the pod volume. So if you consider this Spec:

```yaml
spec:
  domain:
    devices:
      interfaces:
      - name: testnet
        bridge: {}
  networks:
  - name: testnet
    pod: {}
```

The status should then look like this:

```yaml
status:
  interfaces:
    - name testnet
      ip: ...
      mac: ...
```

`testnet` will also be set as alias on the domain xml, so you can always rely on that identifier throughout the code.

pkg/api/v1/types.go
rmohr: Yeah, that's a misunderstanding. I meant the name of the volume/interface, and not the name of the bridge. I hope my review clarifies that. Let me know if you have further questions.

mmirecki:
```release-note
VMI status is updated with a list of all network interfaces added to the vmi.
Initially each interface consists of the interface name, and the IP for the pod interface.

After the vmi is started and additional data is received from libvirt, the interfaces in the
vmi status are populated with mac addresses.
```

rmohr:
ci test please

pkg/virt-controller/watch/vmi.go
```diff
@@ -310,11 +310,17 @@ func (c *VMIController) updateStatus(vmi *virtv1.VirtualMachineInstance, pod *k8
...
-						IP: pod.Status.PodIP,
-					},
+
+				interfaces := make([]virtv1.VirtualMachineInstanceNetworkInterface, 0)
+				for _, network := range vmi.Spec.Networks {
+					ifc := virtv1.VirtualMachineInstanceNetworkInterface{Name: network.Name}
+					if network.NetworkSource.Pod != nil {
+						ifc.IP = pod.Status.PodIP
+					}
+					interfaces = append(interfaces, ifc)
```
rmohr: Can we move the append inside the `if`? I think that the controller can only add a status for pod interfaces ever. Most of the k8s objects normally only add something if they give additional value. I am not sure if we want a list consisting mostly of names alone. `virt-handler` can later on fill in any missing entries independent of the controller, if it has additional information (e.g. via the guest-agent). @mmirecki what do you think?

pkg/api/v1/types.go
```diff
@@ -317,6 +317,8 @@ type VirtualMachineInstanceNetworkInterface struct {
...
```
rmohr: Does not matter much since the status is only manipulated by the controller, but since the `Name` is mandatory now, we should remove the `omitempty`.

pkg/virt-handler/vm.go
```diff
@@ -237,6 +237,25 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
...
+		var domain_interfaces []api.Interface = domain.Spec.Devices.Interfaces
+		for _, domain_interface := range domain_interfaces {
```
rmohr: Can you pre-fill a map of vmi interfaces based on their name and then iterate through the list of domain interfaces and look them up in the map?

pkg/virt-controller/watch/vmi.go
rmohr: Can you add a test to `vmi_test.go`?

pkg/virt-handler/vm.go
rmohr: Can you add some unit-tests to `vm_test.go`

pkg/virt-controller/watch/vmi.go
mmirecki: There is some value of having a complete list of interfaces in the status, even if we just display the names. It shows we know about the interface, but not about it's mac/ip (not assiged yet, which is also some kind of info).
But having no entry in status could also mean that the ifc has not been created yet, which I'm also fine with.
Moving to if.

pkg/virt-handler/vm.go
```diff
@@ -237,6 +237,26 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 
 	oldStatus := vmi.DeepCopy().Status
 
+	if domain != nil {
+		var interfaces = vmi.DeepCopy().Status.Interfaces
```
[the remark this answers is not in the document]
SchSeba: You can also just drop this and make 
```golang
for idx, iface := range vmi.Status.Interfaces
```

pkg/virt-handler/vm.go
```diff
@@ -237,6 +237,26 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
...
+		interfaces_by_name := make(map[string]int)
+		for i, existing_iface := range interfaces {
+			interfaces_by_name[existing_iface.Name] = i
+		}
+
+		var domain_interfaces []api.Interface = domain.Spec.Devices.Interfaces
+		for _, domain_interface := range domain_interfaces {
+			var is_present bool = false
+			i, is_present := interfaces_by_name[domain_interface.Alias.Name]
```
SchSeba: You can do something like this 
```golang
if i, isPresent := interfacesByName[domainInterface.Alias.Name]; isPresent {
// Exist
} else {
// Not exist
}
```

pkg/virt-handler/vm.go
```diff
@@ -237,6 +237,26 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 
 	oldStatus := vmi.DeepCopy().Status
 
+	if domain != nil {
+		var interfaces = vmi.DeepCopy().Status.Interfaces
+		interfaces_by_name := make(map[string]int)
```
SchSeba: interfaces_by_name -> interfacesByName

pkg/virt-handler/vm.go
```diff
@@ -237,6 +237,26 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
...
+		for i, existing_iface := range interfaces {
```
SchSeba: existing_iface -> existingInterface

pkg/virt-handler/vm.go
```diff
@@ -237,6 +237,26 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
...
+			interfaces_by_name[existing_iface.Name] = i
+		}
+
+		var domain_interfaces []api.Interface = domain.Spec.Devices.Interfaces
+		for _, domain_interface := range domain_interfaces {
```
SchSeba: No need to create a new variable you can do something like this
```golang
for _,domainInterface := range domain.Spec.Devices.Interfaces
```

pkg/api/v1/types.go
```diff
@@ -317,6 +317,8 @@ type VirtualMachineInstanceNetworkInterface struct {
 	IP string `json:"ipAddress,omitempty"`
 	// Hardware address of a Virtual Machine interface
 	MAC string `json:"mac,omitempty"`
+	// Name of the interface, corresponds to the pod bridge the interface is
```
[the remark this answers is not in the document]
mmirecki: --> // Name of the interface, corresponds to name of the network assigned to the interface

rmohr:
ci test please

rmohr:
Travis complains that you have to run "make generate". Probably related to the rebase.

rmohr:
ci test please

rmohr:
To sum up the offline discussion with @mmirecki: we make the `name` for now optional and let virt-handler add any missing names, in case it encounters one without a name. Otherwise updating running VMIs from before an update may become impossible.

rmohr:
ci test please

tests/vmi_multus_test.go
```diff
@@ -219,6 +220,47 @@ var _ = Describe("Multus Networking", func() {
...
+			Expect(len(updatedVmi.Status.Interfaces)).To(Equal(2))
+			interfacesByName := make(map[string]v1.VirtualMachineInstanceNetworkInterface)
+			for _, ifc := range updatedVmi.Status.Interfaces {
+				interfacesByName[ifc.Name] = ifc
+			}
+
+			for _, network := range networks {
+				ifc, is_present := interfacesByName[network.Name]
+				Expect(is_present).To(BeTrue())
+				Expect(ifc.MAC).To(Not(BeZero()))
```
rmohr: Let's add a check that the two mac addresses are not equal (easy to do and makes regressions a little bit less likely).

rmohr:
ci test please

tests/vmi_multus_test.go
```diff
@@ -219,6 +220,48 @@ var _ = Describe("Multus Networking", func() {
...
+			}
+			Expect(interfacesByName["default"].MAC).To(Not(Equal(interfacesByName["ovs"].MAC)))
+
+			err = tests.CheckForTextExpecter(updatedVmi, []expect.Batcher{
+				&expect.BSnd{S: fmt.Sprintf("ip addr show eth0 | grep %s | wc -l", interfacesByName["default"].MAC)},
```
zakkg3: in ubuntu interfaces are named enp1s0 :(

PR #1596, opened 2018-10-09
[RFC][POC] Add collectd node agent to export VMI metrics

fromanirh:
In order to monitor, troubleshoot and tune VMs, we would like
to expose detailed metrics about how the VM instance is behaving.
We are concerned about the infrastructure here, so the infrastructure
processes running inside the virt-launcher pod.
For example, we want to assess the overhead of the aformenentioned
infrastructure, allowing to have answers to queries like

- how much memory is libvirtd taking?
- is that memory growing?
- how much memory QEMU takes?
- is libvirtd leaking threads?

This patch provide the initial support to report those metrics.

Signed-off-by: Francesco Romani <fromani@redhat.com>

This PR is a POC implementation to expose VMI detailed metrics

Fixes https://github.com/kubevirt/kubevirt/issues/1594

Still pending:
- polishing: need to review permissions and reduce even more the privileges needed, if possible
- ServiceMonitor integration, still WIP

**Release note**:

```release-note
Add detailed VMI metric reporting, for tuning and troubleshooting.
```

### DIFF
--- a/manifests/cluster-monitoring-collectd-config.yaml
+++ b/manifests/cluster-monitoring-collectd-config.yaml
@@ -1,0 +1,22 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: collectd-config
+  namespace: kube-system
+data:
+  node-collectd.conf: |-
+    LoadPlugin syslog
+    <Plugin syslog>
+       LogLevel info
+    </Plugin>
+    LoadPlugin write_prometheus
+    <Plugin "write_prometheus">
+      Port "9090"
+    </Plugin>
+    LoadPlugin unixsock
+    <Plugin unixsock>
+      SocketFile "/var/run/collectd.sock"
+      SocketPerms "0700"
+      DeleteSocket true
+    </Plugin>
+
--- a/manifests/cluster-monitoring-collectd-node-agent.yaml
+++ b/manifests/cluster-monitoring-collectd-node-agent.yaml
@@ -1,0 +1,39 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: collectd-node-agent
+  namespace: kube-system
+  labels:
+    app: collectd-node-agent
+spec:
+  template:
+    metadata:
+      labels:
+        name: collectd-node-agent
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "9090"
+    spec:
+      serviceAccountName: kubevirt-privileged
+      hostNetwork: true
+      hostPID: true
+      nodeSelector:
+        node-role.kubernetes.io/infra: "true"
+      containers:
+      - name: collectd
+        ports:
+        - containerPort: 9091
+          protocol: TCP
+        image: fromanirh/collectd:0.3.5
+        imagePullPolicy: IfNotPresent
+        volumeMounts:
+        - name: collectd-config
+          mountPath: /etc/collectd
+      volumes:
+      - name: collectd-config
+        configMap:
+          name: collectd-config
+          items:
+          - key: node-collectd.conf
+            path: collectd.conf
+


fromanirh:
I agree with Roman that in-code solution is better in the long run in a number of ways. However, in the short (and maybe mid) term, this approach with collectd is cheap enough, is minimally invasive (no changes to kubernetes core are expected!) and gives us room to explore and tune while the in-code solution is developed.

For these reasons I believe this approach has merits worth discussing

rmohr:
> I agree with Roman that in-code solution is better in the long run in a number of ways. However, in the short (and maybe mid) term, this approach with collectd is cheap enough

Could you explain on what this is collectiong, how it is expressing all the different libvirtds and qemus running and how the processes and metrics are matched with running VMIs?

fromanirh:
> > I agree with Roman that in-code solution is better in the long run in a number of ways. However, in the short (and maybe mid) term, this approach with collectd is cheap enough
> 
> Could you explain on what this is collectiong, how it is expressing all the different libvirtds and qemus running and how the processes and metrics are matched with running VMIs?

Sure. This content should probably copied and expanded in a separate document, but let's start from here.

We have a daemonset, so one collectd-per-host exposing a metrics endpoint.
The daemonset should run on each host which run VMs (I may have got nodeSelector wrong, if so it's an artifact of my test setup, please suggest fixes)

The collectd instance uses just two plugins: the prometheus one (https://github.com/kubevirt/kubevirt/pull/1596#issuecomment-428509514) and a procfs collector.
The procfs collector does two things:
- gather *separate* metric for each qemu/libvirtd process running in the host
- maps the PIDs of the aforementioned qemu/libvirtd process to PODs.
Collectd has a processes plugin (https://collectd.org/wiki/index.php/Plugin:Processes) which is not completely up to the task because it aggregates the stats if multiple instances of the same process are found, and we want separate stats

The procfs collector:
- scans /proc in read only mode
- find all the processes whose commandline matches any glob expression (qemu and libvirt at the moment)
- collect VM, RSS and CPU (user/sys..) metrics from procfs for the matching processes
- if the basic approach is validated we can extend the collection of other interesting metrics (PSS?)
- we use the "gopsutil" package to extract the information from procfs

to match PIDs to PODs/VMs:
- we reverse-engineer the container id from the cgroupfs names. AFAIK this is pretty much what other monitoring soluion do
- we query the runtime (docker) using the CRI interface to learn which container belong to which pod
- we query the runtime (docker) using the CRI interface to get the details of the PODs
- if a POD has a "kubevirt.io/domain" annotation, we use this value for the name of the metric, otherwise we use the pod name

There are more implementation details but this should be it

rmohr:
> to match PIDs to PODs/VMs:
> 
> * we reverse-engineer the container id from the cgroupfs names. AFAIK this is pretty much what other monitoring soluion do
> * we query the runtime (docker) using the CRI interface to learn which container belong to which pod
> * we query the runtime (docker) using the CRI interface to get the details of the PODs
> * if a POD has a "kubevirt.io/domain" annotation, we use this value for the name of the metric, otherwise we use the pod name

Hm, I have the impression that this is even more complicated than directly collecting the metrics via virt-handler and virt-launcher.

fromanirh:
> > to match PIDs to PODs/VMs:
> > 
> > * we reverse-engineer the container id from the cgroupfs names. AFAIK this is pretty much what other monitoring soluion do
> > * we query the runtime (docker) using the CRI interface to learn which container belong to which pod
> > * we query the runtime (docker) using the CRI interface to get the details of the PODs
> > * if a POD has a "kubevirt.io/domain" annotation, we use this value for the name of the metric, otherwise we use the pod name
> 
> Hm, I have the impression that this is even more complicated than directly collecting the metrics via virt-handler and virt-launcher.

Here's the POC: https://github.com/fromanirh/procwatch/blob/master/podfind/podfind.go

michalskrivanek:
> Hm, I have the impression that this is even more complicated than directly collecting the metrics via virt-handler and virt-launcher.

OTOH, I think we better keep virt-launcher small if possible. It's a critical piece and we don't want to bring VM down when it crashes for some reason, right?

rmohr:
@fromanirh considering the following things:

 * collectd and the used plugin are written, maintained and improved with e.g. new metrics outside of kubevirt
 * we have releases which provide a fixed set of container versions and manifests
 * the general concerns raised in #1594 regarding to this solution
 * and the fact that there is nothing this needs right now from KubeVirt itself
 * that you can develop and use this independent of KubeVirt if you decide to use it

I think it is best if you do your own releases of the manifests outside of KubeVirt. What do you think?

fromanirh:
> @fromanirh considering the following things:
> 
>     * collectd and the used plugin are written, maintained and improved with e.g. new metrics outside of kubevirt
> 
>     * we have releases which provide a fixed set of container versions and manifests
> 
>     * the general concerns raised in #1594 regarding to this solution
> 
>     * and the fact that there is nothing this needs right now from KubeVirt itself
> 
>     * that you can develop and use this independent of KubeVirt if you decide to use it
> 
> 
> I think it is best if you do your own releases of the manifests outside of KubeVirt. What do you think?

We are going this direction for the time being. Thanks everyone for the comments, closing this PR.

PR #1614, opened 2018-10-12
Ensure that transitive builds work with kubevirt after glog shadowing was introduced

### LINKED ISSUE #1612: glog shadow breaks transitive builds.

>
/kind bug
> /kind enhancement

**What happened**:
the glog build shadowing glog -> ../../../pkg/staging/glog//

breaks transitive builds, if they have libs that actually depend on glog. 

The whole dependency closure doesn't work. 

```
vendor/kubevirt.io/kubevirt/pkg/log/log.go:58:24: undefined: glog.LogLevel
vendor/kubevirt.io/kubevirt/pkg/log/log.go:59:24: undefined: glog.LogLevel
Makefile:39: recipe for target 'vet' failed

```

For kubevirt to not shadow glog.

Copy the  cmd/example-hook-sidecar folder into an empty folder. 

copy this Gopkg.toml 

https://gist.github.com/43f0fddf5876eb26f510d16c71923ead

and try building 

**Environment**:

Latest head as of today - see Gopkg.toml for reference

rmohr:

Ensure that our glog package only exposes what the original glog package exposes. This way, if kubevirt is vendored, the other project can pull in the original glog package and does not even notice that we shadow it. The long term solution is to synchronize our public APIs into a separate repo.

Fixes #1612 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/fake-qemu-process/fake-qemu.go
+++ b/cmd/fake-qemu-process/fake-qemu.go
@@ -20,7 +20,6 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	"os"
 	"os/signal"
@@ -31,13 +30,12 @@ import (
 )
 
 func main() {
-	uuid := flag.String("uuid", "", "some fake arg")
+	uuid := pflag.String("uuid", "", "some fake arg")
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt,
 		syscall.SIGTERM,
 	)
 
-	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
 	fmt.Printf("Started fake qemu process with uuid %s\n", *uuid)
 
--- a/cmd/virt-handler/virt-handler.go
+++ b/cmd/virt-handler/virt-handler.go
@@ -99,7 +99,7 @@ func (app *virtHandlerApp) Run() {
 	}
 
 	logger := log.Log
-	logger.V(1).Level(glog.INFO).Log("hostname", app.HostOverride)
+	logger.V(1).Level(log.INFO).Log("hostname", app.HostOverride)
 
 	// Create event recorder
 	virtCli, err := kubecli.GetKubevirtClient()
--- a/cmd/virt-launcher/virt-launcher.go
+++ b/cmd/virt-launcher/virt-launcher.go
@@ -20,7 +20,7 @@
 package main
 
 import (
-	"flag"
+	goflag "flag"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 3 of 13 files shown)


senior7515:
@rmohr  

I see issues still after `make dep-install`

```
➜  upstream-kubevirt git:(master) md5sum pkg/staging/glog/glog.go      
6de3d30bca7b109124143307d64f317c  pkg/staging/glog/glog.go

➜  upstream-kubevirt git:(master) md5sum vendor/github.com/golang/glog/glog.go 
6de3d30bca7b109124143307d64f317c  vendor/github.com/golang/glog/glog.go

```

There is really no changes. (i.e.: via git status)

Building from w/ this change still doesn't fix 

```
vendor/kubevirt.io/kubevirt/pkg/log/log.go:58:24: undefined: glog.LogLevel
vendor/kubevirt.io/kubevirt/pkg/log/log.go:59:24: undefined: glog.LogLevel
Makefile:39: recipe for target 'vet' failed
```

However, it doesn't make sense since I see this

```
// This package reimplements github.com/golang/glog and logs to a structured logger.
package glog

import (
	"bytes"
	"flag"
	"fmt"
	"io"
	stdLog "log"
	"os"
	"path/filepath"
	"runtime"
	"strconv"
	"strings"
	"sync/atomic"
	"time"

	log2 "github.com/go-kit/kit/log"
)

type LogLevel int32

const (
	INFO LogLevel = iota
	WARNING
	ERROR
	FATAL
)


```

senior7515:
Note that the package manager is correct. 

https://github.com/golang/glog/blob/master/glog.go

has no loglevel. 

The problem is this. 

**Kubevirt's** shadow package **contains** glog.LogLevel  and upstream does.

**my** app uses proper upstream glog.

The dependency solver will fetch the correct glog `github.com/golang/glog` from github. 

When it sees that dependency the build system will look in the right directory - upstream. 

**not** in the kubevirt's shadow package. 

I think the real change is to create a kvirtlogger and to not shadow upstream packages.

senior7515:
Looking at the code, the functionality of `"github.com/go-kit/kit/log"` that kubevirt uses is rather minimal. 

I.e.: not sure it's worth the shadow package. 

If you look at the dependency of the build system, kubevirt explicitly excludes a bunch of packages from go-kit.

senior7515:
I think to unblock me and to leave the discussion, we should backport the `DiskSecret` patch to 0.9.1 -> 0.9.2 while we make the bigger change of fixing the logger

@fabiand  - thoughts?

senior7515:
Also, i think if you use `glog.Level` vs `glog.LogLevel`  it might work. let me give this a shot.

senior7515:
@rmohr  - so i tested the idea of rolling back to 0.9.0 and porting my disksecret changes and it works. 

It's a hack: https://github.com/senior7515/kubevirt/releases/tag/v0.9.0.2

but will work for this. thanks.

fabiand:
@rmohr any further comment?

rmohr:
> I think the real change is to create a kvirtlogger and to not shadow upstream packages.

That's what we did so far. The main issue is that we have now different sources for our logs, also that the glog logs are non-structured, whereas we are using a structured logger.

The structured logger makes it pretty easy to track objects across the clusters in logs via things like e.g.:

```golang
logger.Object(vmi).Reason(err).Error("this or that failed")
```

So at least from my perspective having a structured looger is worth it. 

`glog` also directly uses `os.Stderr`. An alternative would probably be to replace os.Stderr with our own stream, process the glog output and then log to the original  Stderr. If we also take the glog `flag` dependencies into account, it might work.

> Looking at the code, the functionality of `"github.com/go-kit/kit/log"` that kubevirt uses is rather minimal.

Yes it is rather minimal and it could  for instance as well have been [logrus](https://github.com/sirupsen/logrus) instead.

senior7515:
ill submit a PR for backporting the disk secret type to the 0.9.1 tag.  i
tested that and it seems to work. doing clustwr testing this week.

rmohr:
@senior7515 I updated the PR. It should now work again with transitive build. Would be happy if you can give it a try.

rmohr:
> @senior7515 I updated the PR. It should now work again with transitive build. Would be happy if you can give it a try.

@senior7515 needed it merged but it would still be great if you could give it a try.

PR #1616, opened 2018-10-15
api: return empty array in resources for subresources endpoint

pkliczewski:
Dynamic client is complaining about null value in resources. This
PR fixes this issue by returning empty array.

Fixes: https://bugzilla.redhat.com/1639162


rmohr:
@fabiand will there be further 0.8 releases?

fabiand:
I don't necessarily see the need for a release - but I am open to do a tag if there is a request.
Besides that we can continue backporting patches to release-0.8, as I expect that there are parties usign this branch for their stable streams.

PR #1625, opened 2018-10-15
Added sriov interface type support

booxter:
The type extracts PCI ID numbers from SRIOV-VF-PCI-ADDR environment variable
set by SR-IOV CNI plugin; and configures hostdev libvirt devices that
corresponds to the extracted IDs.

The change also mounts additional host mounts for /sys and /dev to allow
containerized libvirt / qemu to plug pci devices using vfio kernel interface.

User guide: https://github.com/kubevirt/user-guide/pull/134

Fixes #

**Release note**:

```release-note
Support new `sriov` interface type. This type should be used in conjunction with Intel SR-IOV device plugin, SR-IOV CNI plugin, and Multus. When used, the SR-IOV device is directly passed through into VMI guest using `vfio` userspace API.
```


pkg/virt-controller/services/template.go
```diff
@@ -578,6 +618,10 @@ func getRequiredCapabilities(vmi *v1.VirtualMachineInstance) []k8sv1.Capability
 	if vmi.IsCPUDedicated() {
 		res = append(res, CAP_SYS_NICE)
 	}
+	// todo: check that SYS_ADMIN is indeed needed
```
stu-gott: We should definitely double check this before merging.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -506,6 +506,28 @@ func Convert_v1_FeatureHyperv_To_api_FeatureHyperv(source *v1.FeatureHyperv, hyp
 	return nil
 }
 
+func getSriovPciAddresses() []string {
```
SchSeba: Can you please add a comment to explain how the `SRIOV-VF-PCI-ADDR` env variable looks like?

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -861,83 +883,114 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
-					Bridge: DefaultBridgeName,
-				}
+			dbsfFields, err := util.ParsePciAddress(pciAddr)
+			if err != nil {
+				return err
 			}
 
-			if iface.BootOrder != nil {
-				domainIface.BootOrder = &BootOrder{Order: *iface.BootOrder}
+			hostDev := HostDevice{
```
SchSeba: Can you add  the bootOrder also ?
https://libvirt.org/formatdomain.html#elementsHostDevSubsys

booxter:
@stu-gott for the record, functional tests for this code are not possible because they would require SR-IOV hardware and external components set to allocate them. Perhaps in the future when we fully automate the deployment, and have special hardware for CI, we could reconsider it.

Do you have specific tests in mind that would not require access to the hardware?

As for testing in general, the latest upload contains a bunch of unit tests.

pkg/virt-launcher/virtwrap/api/converter.go
booxter: Thanks for pointing out I missed it, added it in the latest upload.

booxter:
Documentation patch: https://github.com/kubevirt/user-guide/pull/134

pkg/virt-controller/services/template.go
```diff
@@ -149,6 +159,60 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
+			},
+		})
+
+		// todo: revisit when SR-IOV DP registers /dev/vfio/NN with pod
+		// device group:
+		// https://github.com/intel/sriov-network-device-plugin/pull/26
+		//
+		// Run virt-launcher compute container privileged to allow qemu
+		// to open /dev/vfio/NN for PCI passthrough
+		privileged = true
```
rmohr: Making the pods privileged and implementing logic which requires privileges inside virt-launcher will be problematic on the long run:

 * users can't use sidecars anymore since we can't give them access to the pods (e.g. istio can't be used by users)
 * if privileged logic and logic in general is done in virt-launcher, we have more likely issues when updating kubevirt. We can't, for instance, add new features which apply to running VMIs, because the old logic is still sitting in virt-launcher.

Having said that, I don't know of any better solution for the time being if we need sriov support right now.

pkg/api/v1/schema.go
```diff
@@ -803,6 +803,7 @@ type Interface struct {
 type InterfaceBindingMethod struct {
 	Bridge *InterfaceBridge `json:"bridge,omitempty"`
 	Slirp  *InterfaceSlirp  `json:"slirp,omitempty"`
+	Sriov  *InterfaceSriov  `json:"sriov,omitempty"`
```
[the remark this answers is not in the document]
booxter: @rmohr Isn't Slirp also an abbreviation?

vladikr:
Are we planning to let the user specify which device (vedor:product) is he asking for?

pkg/virt-controller/services/template.go
booxter: My plan is to address this issue ASAP, it's not going to stay like that for too long. Consider the current impl experimental first peak into the future.

pkg/api/v1/schema.go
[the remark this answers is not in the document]
rmohr: Maybe but at least on the main page it is written as `Slirp`: http://slirp.sourceforge.net/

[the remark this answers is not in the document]
vladikr: I assume, SRIOV means a SRIOV VF? Maybe we should specify that we are asking for a VF.

booxter:
@vladikr my understanding is that it's against DP design where it's DP job to allocate.

vladikr:
> @vladikr my understanding is that it's against DP design where it's DP job to allocate.

I see. It's a bit odd as a node may have several devices, connected to completely different networks.

booxter:
@vladikr valid comment but this should be handled on DP side. There is work in progress patches for the DP to allow to configure different resourceNames for each PCI pool which will then allow us to have multiple sriov network CRDs, one per resourceName.

pkg/api/v1/schema.go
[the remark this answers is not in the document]
booxter: I can use SRIOV, no problem with that; it just looks a bit odd in the context, but ok.

@vladikr AFAIU the DP also allows to configure CRD so that it allocates the whole PF. (I haven't tested it but there is pfOnly attribute in CNI config map.)

booxter:
I will hold off the change to rename Sriov attribute into SRIOV, in hope there are other comments to address before a respin.

pkg/virt-controller/services/template.go
```diff
@@ -149,6 +159,60 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		MountPath: "/var/run/libvirt",
 	})
 
+	if isSriovVmi(vmi) {
+		// libvirt needs this volume to unbind the device from kernel
```
phoracek: Don't we need a TODO here as well? With these mounts, pod can break the host, can't it?

booxter: @phoracek I may miss something, but I don't think we are able to get rid of these mounts because libvirt assumes access to them to calculate PCI IDs / IOMMU groups / detach devices from hosts. Maybe I am able to mark some of them read-only to reduce the scope, but I believe at least one of them is needed r/w to unbind the device; and /dev/vfio/ is needed to register the device with vfio subsystem.

booxter:
The latest upload merely renames Sriov -> SRIOV as requested above. No other changes.

pkg/virt-controller/services/template.go
booxter: @phoracek tangentially, what is a valid reason for virt-launcher to stay owned by VMI owner? It seems to me this pod is a service pod and should really be isolated from user meddling.

phoracek: @booxter i guess you are right there

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -491,6 +491,35 @@ func Convert_v1_FeatureHyperv_To_api_FeatureHyperv(source *v1.FeatureHyperv, hyp
...
 }
 
+// This function parses the SRIOV-VF-PCI-ADDR variable that is set by SR-IOV
+// device plugin listing PCI IDs for devices allocated to the pod. The format
+// is as follows:
+//
+// "": for no allocated devices
+// "0000:81:11.1,": for a single device
+// "0000:81:11.1,0000:81:11.2[,...]": for multiple devices
+func getSRIOVPciAddresses() []string {
```
vladikr: How do we know to correlated the interface to the allocated device? Is the order guaranteed here?

pkg/virt-controller/services/template.go
rmohr: > @phoracek tangentially, what is a valid reason for virt-launcher to stay owned by VMI owner? It seems to me this pod is a service pod and should really be isolated from user meddling.

Here are some reasons why, just to linke this discussion: https://github.com/kubevirt/kubevirt/pull/1625/files#r227701675

rmohr: @booxter: I have a couple more questions here (some of them may be naive):

> @phoracek I may miss something, but I don't think we are able to get rid of these mounts because libvirt assumes access to them to calculate PCI IDs / IOMMU groups / detach devices from hosts. 

So I guess libvirt has to unbind all related devices (which were made ready for consumption by the device plugin) from the host and then create a group out of it?

> Maybe I am able to mark some of them read-only to reduce the scope, but I believe at least one of them is needed r/w to unbind the device;

Do we have to give the device back if the container goes down dirty? Is unbind here related to unbinding it from the host?

> and /dev/vfio/ is needed to register the device with vfio subsystem.

If i understand https://www.kernel.org/doc/Documentation/vfio.txt correctly (not an expert here), it looks like at the end qemu will just open the device group in `/dev/vfio/` and then ask with a system call for the right device via the PCI address. Since we are dealing with a device here, it should at the end get a major (probably 196) and minor number (sequential?) which we can then recreate via `mknod` inside the container. That might be a path for the future with less things mounted.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -491,6 +491,35 @@ func Convert_v1_FeatureHyperv_To_api_FeatureHyperv(source *v1.FeatureHyperv, hyp
...
+	pciAddrString, isSet := os.LookupEnv("SRIOV-VF-PCI-ADDR")
```
rmohr: Do we know why this variable is exposed in the first place? Good for our use-case, but it looks like the device plugin provides a proper setup interface already ...

pkg/virt-launcher/virtwrap/api/converter_test.go
```diff
@@ -1330,6 +1330,40 @@ var _ = Describe("Converter", func() {
 	})
 })
 
+var _ = Describe("getSRIOVPciAddresses", func() {
```
rmohr: I miss a full conversion from VMI to Domain which uses some SRIOV devices. Also, I would prefer if we put the SRIOV-VF-PCI-ADDR into the converter context and look it up from outside the converter.

pkg/virt-controller/services/template.go
booxter: @rmohr you raise an interesting question. Indeed libvirt handles the unbinding / registration with vfio itself. I think it will operate just fine if for some reason the group is already present.

Now, SR-IOV DP folks assume that the groups are pre-registered with vfio / uio before the first device is allocated. It's just that libvirt already does it for us, so we don't need to introduce a separate component / instructions to set vfio devices up. I think that's ok and simplifies setup.

> Do we have to give the device back if the container goes down dirty

I don't think so, unless we are going to allocate the same devices to both vfio and netlink users. (Not supported or planned for SR-IOV DP as of now.) Libvirt should detect the group is already registered and do nothing.

> That might be a path for the future with less things mounted.

I think with multiendpoints branch for SR-IOV DP, we can mount just /dev/vfio/vfio; and the appropriate /dev/vfio/<num> will be registered by DP itself. (We may need to create it, yes; for that, we can crawl the other /sys/* directories exposed. Not convenient but doable.)

pkg/virt-launcher/virtwrap/api/converter.go
booxter: We don't with current SR-IOV DP. With multiendpoints branch merged, each device pool will get its own variable: https://github.com/intel/sriov-network-device-plugin/pull/26

pkg/virt-launcher/virtwrap/api/converter_test.go
vladikr: I think @rmohr suggested to store SRIOV-VF-PCI-ADDR in the ConverterContext [1] instead of using a env var.

[1] https://github.com/kubevirt/kubevirt/blob/900c672c57cb2b3cdf7d516b90110d64bc45b17c/pkg/virt-launcher/virtwrap/api/converter.go#L54

pkg/virt-launcher/virtwrap/api/converter.go
vladikr: I see. Feels like everything is up in the air. 
However, I do understand that this is a moving target...

pkg/virt-launcher/virtwrap/api/converter_test.go
booxter: @vladikr thanks!

The suggestion makes sense in scope of the current DP implementation where we deal with a single variable. My only concern is that when we eventually move to multiple variables as was just merged to sriovdp master today, we may want to extract information for specific networks that are actually attached to the VMI which implies a bit more knowledge about the VMI spec than perhaps is fitting for the caller.

Anyway, we can revisit it later.

booxter:
@rmohr @vladikr @phoracek see the following concerns addressed in the latest upload (you can focus on last three patches, the first in the series is identical to the previous upload).

1. Introduced a new unit test case for converter when acting on a sriov attached vmi.
2. Introduced a new functional test for sriov vmi.
3. Moved environment variable extraction to syncVMI to make converter more pure.

I hope this addresses most of the issues.

tests/utils.go
```diff
@@ -1760,6 +1760,31 @@ func LoggedInFedoraExpecter(vmi *v1.VirtualMachineInstance) (expect.Expecter, er
 	return expecter, err
 }
 
+func LoggedInFedoraExpecter(vmi *v1.VirtualMachineInstance) (expect.Expecter, error) {
```
rmohr: I think that such a function got merged a few days  ago in another PR: #1543. Can we merge them?

tests/utils.go
```diff
@@ -1902,6 +1927,13 @@ func SkipIfNoRhelImage(virtClient kubecli.KubevirtClient) {
 	}
 }
 
+func SkipIfNoSriovDevicePlugin(virtClient kubecli.KubevirtClient) {
```
rmohr: @dankenigsberg @eedri @booxter do you think that we can organize something for CI, so that we can run sriov tests? If not we have the risk of breaking sriov on any commit without noticing.

dankenigsberg: I did not manage to find a solution for this question just yet. I wish qemu/libvirt add support for virtual sriov, but until then - I don't know how we can possibly test it on CI. Another crazy option is to provide Lago with bare metal support, but this is not reasonable anytime soon.

rmohr: If we have a node with a sriov-device, we could pass it through to our qemu processes.

booxter: @rmohr in theory yes, we could pass it through. I haven't got to a point where we could do it in scope of kubevirt. Are we sure this is even possible? It may be a similar case to how you can run a layered virtualization but only down to L2 but not L3. I am not an expert.

Another option, and we are looking into it, is having a separate bare metal node with the card directly deploying kubevirt without additional virtualization layer. This will require quite some work to have local based provider that deploys all components; alternatively, we build it based on kubevirt-ansible that should get support for SR-IOV anyway. Regardless, it would be a bare metal deployment which obviously has its issues (for one, it's hard or impossible to completely clean it up).

tests/utils.go
booxter: Nice. Adopted.

tests/utils.go
rmohr: Let's suppose we have a dedicated machine for that. We  could use the `local-provider`. Before and after each test-run we could simply destroy all containers, including the etcd data. We should then get a completely fresh cluster without much effort.

booxter: @rmohr Yes. Stackpath folks are going to donate a machine in their lab for the cause.

We'll need to run cluster-up in a separate thread since it blocks execution. We also need to write automation for sriov components' configuration and deployment. I don't see unbearable issues with the approach.

I would like to proceed with merging the code without CI set up. We can document setup in a README file in the repo if that helps. Are you open to this approach (merging without CI set up)?

rmohr: > We'll need to run cluster-up in a separate thread since it blocks execution.

We might be able to change that.

> We also need to write automation for sriov components' configuration and deployment. I don't see unbearable issues with the approach.

Sounds great.

> 
> I would like to proceed with merging the code without CI set up. We can document setup in a README file in the repo if that helps. Are you open to this approach (merging without CI set up)?

Do we have a rough idea when we can have the testing infra to be ready? Will there be a commitment to deliver it? I would want to avoid a situation where it looks promising right now that we will get CI but then we loose focus and other things look more urgent.

And thank you for looking into it.

booxter: I can't give exact estimates but let's just say that I am committed to it and I will push back de-prioritizing it; we also want stackpath donation to move forward irrespective of particular technical challenges with CI.

It's on my todo list. I am going to look into it in ~two weeks. (There are several other high priority SR-IOV pieces like support for multiple SR-IOV networks and kubevirt-ansible integration that I can't ignore right now.)

Also, I've just added a documentation page for SR-IOV deployment while we wait on installer and CI automation, if that helps.

dankenigsberg: FWIW, the commitment is not only @booxter's but also mine.
Life is too short to solve the same bug twice. We need CI for SR-IOV.

rmohr: > FWIW, the commitment is not only @booxter's but also mine.
> Life is too short to solve the same bug twice. We need CI for SR-IOV.

ok, then let's go ahead.

rmohr: It would be great that in the future, we understand CI as part of the PR. It should not be an after-thought.

booxter:
@rmohr @SchSeba @phoracek you've approved the PR. Can we merge it now? I don't think I am allowed to do it myself, am I?

rmohr:
ci test please

rmohr:
@booxter stdci test results are missing. Once they are there I will merge it.

booxter:
The "standard-ci" job results are available. The job failed. My understanding is that this is a meta-job that combines results of other checks.

Some jobs fails due to infrastructure issues.

```
Error: No such image: kubevirtci/k8s-1.10.4@sha256:b60a61ca03a1a6c504481020709a04f65e4dd9c929a8bcad18821c5f80d1b2b6
```

Some fail because of timeouts waiting for VMIs to become ready: https://jenkins.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/2518//artifact/check-patch.k8s-1.11.0-release.el7.x86_64/mock_logs/script/stdout_stderr.log

```
DataVolume Integration
  /root/go/src/kubevirt.io/kubevirt/tests/datavolume_test.go:67
    using Alpine import
    /root/go/src/kubevirt.io/kubevirt/tests/datavolume_test.go:68
      /root/go/src/kubevirt.io/kubevirt/tests/datavolume_test.go:69

      Timed out after 120.000s.
      Timed out waiting for VMI to enter Running phase
      Expected
          <bool>: false
      to equal
          <bool>: true
```
It seems each job shows different failures. AFAIU majority of these failures are known / expected though I have no guidance as to what we can ignore for the goal of merging of this PR.

I would like to be helpful here, so @rmohr could you please tell me what you would like to see from these results? How do I know that it's not just bad but bad-bad to the point where we should block the merge here?

rmohr:
@booxter yes test results look ok so far. There are only the usual candidates which is separetly looked into :/

booxter:
@rmohr thanks for merging it.

Is CI documented anywhere? If we are going to experience CI instability for the foreseeable future it would perhaps be less confusing if we document somewhere which jobs are set up, which are "expected" to fail, and in which particular way; perhaps with a link to the CI tracker issue. A short description somewhere in kubevirt[ci] repo could be helpful (as part of CONTRIBUTING.md?)

I would be happy to fill in details but I don't really know all of them. Maybe people responsible for CI setup (do we have anyone dedicated to it?) could chime in here.

Regardless of that, it's confusing we still run 1.10.* jobs. I thought we stopped supporting the release. Where do we store our jenkins configuration?

PR #1626, opened 2018-10-16
[release-0.8] goimports works now better, fix imports

rmohr:

Backprot of #1617 for 0.8 releases. A new version of goimports better sorts our dependencies. Fixes travis builds.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/virt-handler/virt-handler.go
+++ b/cmd/virt-handler/virt-handler.go
@@ -22,7 +22,6 @@ package main
 import (
 	"fmt"
 	"os"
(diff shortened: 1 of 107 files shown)


rmohr:
> don't we need the change in `hack/docker-builder/Dockerfile` (`go get **-u** golang.org/x/tools/cmd/goimports`) too?

No, not necessary. If the container is rebuilt goimports is not installed, so it does not make a difference.

PR #1636, opened 2018-10-17
Improved block volume PVC test

slintes:
Use a new PV and PVC for the test, it's not possible to run multiple tests with the same PV

**Release note**:
```release-note
NONE
```


slintes:
Thx for review @mfranczy and @stu-gott 
Failing tests are unrelated, will merge.

PR #1645, opened 2018-10-22
Don't deploy test infra if using external provider

### LINKED ISSUE #1644: Missing required test infrastructure for tests in some external provider cases
/kind enhancement

**What happened**:
When using the external provider option, we loose control of our testing environment.  In some cases a user can set things up appropriately for everything to work, but in other cases the cluster may already be deployed or it may be something like a standard GKE cluster where things can't be added/modified.

When setting `KUBEVIRT_PROVIDER=external` and using a standard GKE cluster I expect to run `make cluster-up and make cluster-sync` and have my KubeVirt build deployed on my GKE cluster ready to use.

What actually happens however is that we fail cluster-sync when attempting to deploy the tests manifest.  This fails because we already created a kubevirt configmap to enable emulation mode, then the tests deploy attemtps to  create a kubevirt configmap again and fails.  Even if we changed this to patch or add to the existing configmap, we'll be missing some components expected by the functional tests.

Configure a standard GKE cluster using the latest version avaialble.
Set up a configmap to use emulation
export KUBEVIRT_PROVIDER=external
run make cluster-up && make cluster-sync

The sync will fail when attempting to deploy tests.

j-griffith:

This patch modifies cluster/deploy.sh to NOT attempt deploying test
infra when using an external KubeVirt Provider.

Since we don't have control over the target Cluster in this case there's
a high probability that it won't have all of the expected configuration
settings and avaialble resources we expect for our tests.  Rather than
just running it and having it fail for various reasons, this patch will
skip that portion of the deployment for external providers.

In the case of a user that knows what they're doing and has everything
he/she needs on their cluster, they can override this behavior with an
enviornment variable.

`KUBEVIRT_DEPLOY_TESTS_TO_EXTERNAL_PROVIDER=true` will override the
provider check and attempt to deploy the additional test infra.

Fixes #1644 

**Release note**:

```release-note
Modifies make cluster-sync to skip deploying test infra when using an external provider.  To override this user must export KUBEVIRT_DEPLOY_TESTS_TO_EXTERNAL_PROVIDER=true

```

### DIFF
--- a/cluster/deploy.sh
+++ b/cluster/deploy.sh
@@ -47,7 +47,12 @@ elif [[ $TARGET =~ .*-release ]] || [[ $TARGET =~ windows.* ]]; then
 fi
 
 # Deploy additional infra for testing
-_kubectl create -f ${MANIFESTS_OUT_DIR}/testing -R $i
+# In the case of external providers we don't control all of the various requriements here
+# rather than try and call them all out and check let's just explicitly skip deployment
+# we'll provide a new varialbe to over ride this for folks that know what they're doing
+if [[ "$KUBEVIRT_PROVIDER" != "external" ]] || [[ "$KUBEVIRT_DEPLOY_TESTS_TO_EXTERNAL_PROVIDER" == true ]]; then
+    _kubectl create -f ${MANIFESTS_OUT_DIR}/testing -R $i
+fi
 
 if [[ "$KUBEVIRT_PROVIDER" =~ os-* ]]; then
     _kubectl adm policy add-scc-to-user privileged -z kubevirt-controller -n ${namespace}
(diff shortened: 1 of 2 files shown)


j-griffith:
ci test please

rmohr:
@j-griffith test results look good. Did you have a chance to go through the comments on #1644?

rmohr:
Let's wait for the outcome of the discussion in #1644 

/hold

j-griffith:
> @j-griffith test results look good. Did you have a chance to go through the comments on #1644?

Yeah, after reading those comments and going back through some things I'm not sure this is going to be necessary.  The deploy works fine if I do things in the "correct order".  Let me play with it a little more but I think we can probably close this and I can just fix things up in my own documentation.

Thanks!

j-griffith:
Thanks to folks for pointing out some things here; this PR isn't necessary so I'm going to close it.

PR #1640, opened 2018-10-19
Patch/sock connector exec

steigr:
**What this PR does / why we need it**:

Safe resources when `virtctl [vnc|console] vm` by `exec`ing into socat and remove dependency on PID-file for serial connection tracking.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #1650, opened 2018-10-24
Improved console test

slintes:
Ensure that a console connection is closed as soon as a new one is opened
See discussion on https://github.com/kubevirt/kubevirt/pull/1640

I cross checked that this test fails if we don't close console connections in virt-launcher's sock-connector

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/console_test.go
+++ b/tests/console_test.go
@@ -69,6 +69,13 @@ var _ = Describe("Console", func() {
 		Expect(err).ToNot(HaveOccurred())
 	}
 
+	OpenConsole := func(vmi *v1.VirtualMachineInstance) (expect.Expecter, <-chan error) {
+		By("Expecting the VirtualMachineInstance console")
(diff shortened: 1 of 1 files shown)


tests/console_test.go
```diff
@@ -103,6 +110,33 @@ var _ = Describe("Console", func() {
...
 			}, 220)
 
+			It("should close console connection when new console connection is opened", func() {
+				vmi := tests.NewRandomVMIWithEphemeralDisk(tests.RegistryDiskFor(tests.RegistryDiskAlpine))
+
+				RunVMIAndWaitForStart(vmi)
+
+				By("opening 1st console connection")
+				expecter, errChan := OpenConsole(vmi)
+				defer func() {
```
cynepco3hahue: hm why `defer func() ...` and not just `defer expecter.Close()`

slintes: very good question ;) fixed

PR #1651, opened 2018-10-25
Bugfix - disk device cache

mfranczy:
It moves function responsible for setting a driver cache mode to preStartHook,
because of that I have already prepared disk images and can easily check if fs supports direct I/O.
I don't have to create temporary files in disk location anymore, that was causing some problems if dir was not created yet (my bad).

Fixes #

In progress:
- [x] figure out why tests didn't show that and fix them

**Release note**:

```release-note
VM high performance - add cache options
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/api/converter.go
+++ b/pkg/virt-launcher/virtwrap/api/converter.go
@@ -100,7 +100,8 @@ func Convert_v1_Disk_To_api_Disk(diskDevice *v1.Disk, disk *Disk, devicePerBus m
 		}
 	}
 	disk.Driver = &DiskDriver{
-		Name: "qemu",
+		Name:  "qemu",
+		Cache: string(diskDevice.Cache),
 	}
 	if numQueues != nil {
 		disk.Driver.Queues = numQueues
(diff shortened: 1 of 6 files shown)


mfranczy:
@rmohr @vladikr @slintes when you have some free time can you take a look? Thanks.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -681,11 +666,6 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 			newDisk.Driver.IOThread = &ioThreadId
 		}
 
-		err = setDriverCacheMode(&newDisk, disk.Cache)
```
mfranczy: this has been moved to `preStartHook` [link](https://github.com/kubevirt/kubevirt/pull/1651/files#diff-9b1d64950edb57271a1a2fc36192d464R359) after disks initialization.

rmohr:
@mfranczy under which conditions could it happen that the directory was not created yet?

mfranczy:
@rmohr good example here will be `ephemeral-disk`.
It creates a volume directory when is creating a disk.image [[1]](https://github.com/kubevirt/kubevirt/blob/master/pkg/ephemeral-disk/ephemeral-disk.go#L86) and function `CreateEphemeralImages` is called from `preStartHook` [[2]](https://github.com/kubevirt/kubevirt/blob/v0.9.2/pkg/virt-launcher/virtwrap/manager.go#L337) so at this stage I had only existing base path where ephemeral disks are stored.

In converter (`Convert_v1_VirtualMachine_To_api_Domain`), `disk.Source.File` gives me a full path to the image, I thought that directories are created but disk image might be created later, that's why I was executing open with O_CREAT flag, but this couldn't work because of missing dirs in the path.

Instead of creating dirs by myself I think it's better to move function to `preStartHook`, this allows me just to open images with O_DIRECT flag without creating any temporary files, and it's gonna be executed only once, I think it's simpler.

Functional tests didn't show me that which means they are bad, so I have to figure out what's wrong there.

mfranczy:
Functional tests didn't show errors because checks there are not good enough, it should check all disks, not only if libvirt xml contains some cache options. Fix on the way.

mfranczy:
Tests are fixed, checks are now correct. Tested as well on the master branch, as expected they were failing.

mfranczy:
@rmohr yes good idea, I am going to change that.

mfranczy:
@rmohr @slintes disks tests are simplified to one test.

mfranczy:
Now, failing tests are not related.

PR #1655, opened 2018-10-26
Allow static networking in cloudinit metadata

karmab:
```release-note
Allow static networking in cloudinit metadata
```

We parse the interface section of a vm and if ip, address and gateway are set, populate this so that the static networking configuration is injected through cloudinit.
a sample interface section looks like this

```
interfaces:
  - name: default
     bridge: {}
     ip: 192.168.122.4
     netmask: 255.255.255.0
     gateway: 192.168.122.1
```

### DIFF
--- a/pkg/api/v1/schema.go
+++ b/pkg/api/v1/schema.go
@@ -794,6 +794,14 @@ type Interface struct {
 	// If specified, the virtual network interface will be placed on the guests pci address with the specifed PCI address. For example: 0000:81:01.10
 	// +optional
 	PciAddress string `json:"pciAddress,omitempty"`
+	// Boot Protocol of a Virtual Machine interface
+	BootProto string `json:"bootProto,omitempty"`
+	// Ip of a Virtual Machine interface
+	Ip string `json:"ip,omitempty"`
+	// Netmask of a Virtual Machine interface
+	Netmask string `json:"netmask,omitempty"`
+	// Gateway of a Virtual Machine interface
+	Gateway string `json:"gateway,omitempty"`
 }
 
 // Represents the method which will be used to connect the interface to the guest.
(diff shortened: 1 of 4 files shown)


karmab:
@rmohr updated PR using network config v1 instead of deprecated ENI.
let me know if this is worth merging :)

SchSeba:
@karmab Maybe allow this configuration only to additional networks (not the pod network) because we don't have control on that network allocation.

We can also add a bootproto (dhcp/static) for every interface.(default can be the dhcp one we already have)

BTW if we want to merge the PR we will need to add some validation (ip,subnet,gateway) and functional tests

gbock:
I've been working on something similar for the #1692 pr as sriov devices can't be dhcp'd from the launcher container as it no longer has access to the device. It operates in a manner similar to the network discovery for the default bridge dhcp setup. It also handles the case of passing cloud-init network data when no userData is passed. I'll see about getting my changes up on branch for comments. It is still pretty raw and doesn't have tests but it works against the #1692 pr and _should_ support multiple sriov devices. It could also handle the default pod network configuration into the VM with a few changes if desired.

greg-bock:
First pass at supporting cloud-init network v1 for sriov interfaces: https://github.com/greg-bock/kubevirt/commit/eb8f00ac58b6196626072ec75440d1eb73263558

greg-bock:
With some of my lastest changes you could actually just pass in network data yaml (version 1) along with userdata in the spec.

greg-bock:
@karmab can you take a look at #1806? This should take care of both of our use cases.

booxter:
I feel like https://github.com/kubevirt/kubevirt/pull/1806 is perhaps a better foundation for the feature at this point.

PR #1660, opened 2018-10-30
Fixed virtctl release artifacts

slintes:
Skip virtctl-windows.exe and virtctl-darwin, only release virtctl-v*

**Release note**:
```release-note
NONE
```


fabiand:
Nice!

PR #1665, opened 2018-10-31
Deprecating CreatedByLabel in favor of OwnerReferences

### LINKED ISSUE #1369: annotation "kubevirt.io/created-by" for VMI pod is not readable

ahadas:
We wish to replace CreatedByLabel, that virt-launcher is set with, with the a controller reference.
Unfortunately, we cannot do it right now due to backward compatibility issue (we might find running VMIs that are not set with a controller in their OwnerReferences).
This PR does the following:
1. Replace our implementations of GetControllerOf and IsControlledBy with those in Kubernetes.
2. Encapsulates the retrieval of the controller of virt-launcher in a way that it would be clear that it overrides that in Kubernetes. We first try to use the Kubernetes way and fall back to the existing logic that relies  on CreatedByLabel.
3. Deprecates CreatedByLabel (using a comment).

Fixes #1369

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/api/v1/types.go
+++ b/pkg/api/v1/types.go
@@ -362,6 +362,7 @@ const (
 	MigrationJobNameAnnotation string = "kubevirt.io/migrationJobName"
 	// This label is used to match virtual machine instance IDs with pods.
 	// Similar to kubevirt.io/domain. Used on Pod.
+	// Deprecated: would be replaced by a Controller Reference in a future release.
 	CreatedByLabel string = "kubevirt.io/created-by"
 	// This label is used to indicate that this pod is the target of a migration job.
 	MigrationJobLabel string = "kubevirt.io/migrationJobUID"
--- a/pkg/controller/controller_ref.go
+++ b/pkg/controller/controller_ref.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+Copyright 2017 The KubeVirt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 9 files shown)


rmohr:
@ahadas on a first glance, I don't see issues in the code but the tests show that something is not right. Could you have a look? I would suggest to pick one or two of the tests which fail in all lanes, focus on them locally and check what's wrong ...

For instance this test:

```
[Fail] Storage Starting a VirtualMachineInstance With a HostDisk defined With 'DiskExists' type [It] Should use existing disk image and start 
/root/go/src/kubevirt.io/kubevirt/tests/utils.go:1496
```

fails across all lanes.

ahadas:
@rmohr ack, thanks, looking

ahadas:
ci test please

ahadas:
ci test please

ahadas:
@rmohr good catch, you were right - the controller crashed due to improper handling in case of controllerRef==nil so I modified `resolveControllerRef` in virt-controller/watch/vmi.go to handle that correctly. the funtional tests look ok now when I run them locally, but seems there is some problem when they ran in jenkins. I'll trigger them again.

ahadas:
ci test please

rmohr:
@ahadas looks great!

PR #1670, opened 2018-11-02
[POC] Hotplug disk

stu-gott:
POC implementation of disk hotplugging in KubeVirt. This implementation uses Network Block Devices from a complementary pod to make dynamically introduced disks available to already-running VMI's.

**Release note**:
```release-note
Disk hotplugging on the virtoio bus is now implemented through Network Block Devices.
```

### DIFF
--- a/cluster/examples/hotplug-disk.yaml
+++ b/cluster/examples/hotplug-disk.yaml
@@ -1,0 +1,37 @@
+apiVersion: v1
+kind: Pod
+metadata:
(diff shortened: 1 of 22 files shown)


itamarh:
wouldn't this make live migration much more complex?

fabiand:
Thanks.

As it was discussed offline, we wait until Kubernetes provides better ways to do hotplug.

PR #1216, opened 2018-06-25
add vnc script support for MAC OSX

shiywang:
Signed-off-by: Shiyang Wang <shiywang@redhat.com>

**What this PR does / why we need it**:
I just finished the code and make sure compile success, but since I'm a poor kid, haven't got any chance to borrow a macbook to test it yet : (   I'll try to ask my colleague to borrow them a macbook to test.  

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/kubevirt/kubevirt/issues/812 https://github.com/kubevirt/kubevirt/issues/1084

**Special notes for your reviewer**:
@fabiand @rmohr @slintes 

```release-note
Add virtctl vnc support for MAC OSX
```

Issue #1084, opened 2018-05-28
Add virtctl logic for mac viewers


fabiand:
On mac we need to try a few viewers, because the stock `screen sharing` app does not support a "fluent connection", as it always requires the user to input credentials.

The script from PR #1076 can be used for reference

shiywang:
I can take a look of this one

fabiand:
That would be awesome!

Feel assigned, @shiywang and let me know if we can give a hand (#kubevirt on freenode and #virtualization on k8s.slack.com)

slintes:
fixed by https://github.com/kubevirt/kubevirt/pull/1216

PR #1673, opened 2018-11-05
Add vnc script support for MAC OSX

### LINKED ISSUE #812: Can't get "virtctl vnc..." running on OSX
Hi,

I installed the kubevirt demo on a remote cluster. Everything seems to run fine, `virtctl console ...` works, but I can't get `virtctl vnc ...` running. A window is opening and showing these errors: first "Unknown graphic type for the guest vnc://127.0.0.1:55073", and after a click on `ok` "Couldn't create a session for this type: vnc".

For getting so far I installed the Spice OSX client (https://www.spice-space.org/osx-client.html), and added a script named `remote-viewer` to my PATH, which just calls the OSX binary with `/Applications/RemoteViewer.app/Contents/MacOS/RemoteViewer "$@"`.

Do I need to install more?

mlsorensen:
Completes/addresses comments on  (seemingly abandoned) pull #1216, adding 'virtctl vnc' capability for OSX with installed third party VNC viewers.

Fixes #812 #1084

@fabiand @rmohr @slintes

**Release note**:

```release-note
Allows 'virtctl vnc' command to work on OSX, provided one of Tiger VNC, Chicken, or Real VNC are installed.

```

### DIFF
--- a/pkg/virtctl/vnc/vnc.go
+++ b/pkg/virtctl/vnc/vnc.go
@@ -159,14 +159,11 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
 		listenResChan <- err
 	}()
 
-	// execute remote viewer
+	// execute VNC
 	go func() {
+		defer close(doneChan)
 		port := ln.Addr().(*net.TCPAddr).Port
-		args := []string{fmt.Sprintf("vnc://127.0.0.1:%d", port)}
-		if glog.V(4) {
-			args = append(args, "--debug")
-			glog.Infof("remote-viewer commandline: %v", args)
-		}
+		args := []string{}
 
 		vncBin := ""
 		osType := runtime.GOOS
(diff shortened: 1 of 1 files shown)


slintes:
Thanks, looks good to me. The only question is: how to test this... would be nice if at least someone else could verify this manually imho. @karmab do you have a mac, and can you give it a try please?

karmab:
@slintes testing it

karmab:
hello, tested it and it worked fine,
concretely, in my case, i dont have any of the proposed vnc packages so i got an error
actually i dot have remote-viewer in my path, so i suggest adding the same patch i used in the code ( in `pkg/virtcl/vnc/vnc.go` in the darwin case section


```
            } else if _, err := exec.LookPath("remote-viewer"); err == nil {
                vncBin = "remote-viewer"
                args = remoteViewerArgs(port)
            } else if !os.IsNotExist(err) {
                viewResChan <- err
                return
            }
```

karmab:
probably should mentioned that my remote-viewer comes from this spice package `http://people.freedesktop.org/~teuf/spice-gtk-osx/dmg/0.5.7/RemoteViewer-0.5.7-1.dmg`

pkg/virtctl/vnc/vnc.go
```diff
@@ -146,32 +159,74 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
+			} else if !os.IsNotExist(err) {
+				viewResChan <- err
+				return
+			} else if _, err := os.Stat(REAL_VNC); err == nil {
+				vncBin = REAL_VNC
+				args = realVncArgs(port)
+			} else if !os.IsNotExist(err) {
+				viewResChan <- err
+				return
+			}
```
karmab: we should also handle case when remote-viewer was installed in osx and can be found in path

mlsorensen: The thing about the RemoteViewer package is that it seems to require a supplemental script on the user's side to get it to launch cleanly via CLI, I can't just 'ln -s /Applications/RemoteViewer.app/Contents/MacOS/RemoteViewer /usr/local/bin/remote-viewer' for example.

I do like the idea though of having a fallback of just looking for a 'remote-viewer' in the path as a fallback, people are free to script that to what they desire. I'll add the suggestion.

mlsorensen: I've pushed the suggested changes and re-tested with all four 'if' cases on MacOS now.  For the remote-viewer case I still get the error 'Unknown graphic type for the guest vnc://127.0.0.1:59476' as described in #812 but I'm just going to assume that is an issue with my quick attempt to get a custom remote-viewer working, and that if someone has a working remote-viewer setup in their path that it will work.

mlsorensen:
I see there are some failed checks, I don't believe they have anything to do with this change. Let me know if there's anything left to do.

slintes:
@mlsorensen can you please add looking for "remote-viewer" binary for windows? See https://github.com/kubevirt/kubevirt/pull/1694. Thanks!

mlsorensen:
> @mlsorensen can you please add looking for "remote-viewer" binary for windows? See #1694. Thanks!

Sure. I don't have a windows system to test on, would it be as simple as:

```
@@ -198,7 +198,7 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
                                viewResChan <- err
                                return
                        }
-               case "linux":
+               case "linux", "windows":
                        _, err := exec.LookPath(REMOTE_VIEWER)
                        if exec.ErrNotFound == err {
                                viewResChan <- fmt.Errorf("could not find the remote-viewer binary in $PATH")
```

Or does it need custom handling/args? Is it actually called 'virt-viewer', or does 'remote-viewer' work? If "virtctl vnc" on windows isn't officially supported yet, perhaps that can be added by someone who can test it and make the change along with the documentation change in a subsequent pull.

slintes:
> perhaps that can be added by someone who can test it and make the change along with the documentation change in a subsequent pull

fair point

rmohr:
@mlsorensen thank you!

PR #1675, opened 2018-11-06
backport of virtctl image-upload to release-0.9

mhenriks:

This is a backport of #1605 which did not make the cut for release-0.9

Fixes #

**Release note**:

```release-note
image-upload command added to virtctl
```


rmohr:
ci test please

rmohr:
For reference: http://jenkins.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/2434/artifact/ci_build_summary.html

Test results look pretty much ok. One upload test fails in two lanes but that seems to be the issue which we seem to see on master too (let's put something together to improve that after this PR, since as far as I remember there was really the issue that the upload fails for unknown reasons).

Running tests once more to be really sure.

rmohr:
ci test please

mhenriks:
@rmohr in my debugging of the upload tests, I've seen two issues that come up.

1.  The most frequent issue is that virtctl times out waiting for the upload pod to be "Running."  Seems to be a deadlock in the the local volume provisioner.  Basically, the PVC is waiting for a consumer and the pod is waiting for the PVC.  This usually only happens the first time the local volume provisioner creates a PVC.  I initially had the functional test handle this but took the code out.

2.  The other failure I see (much more infrequently) occurs with the openshift tests.  Sometimes the upload fails because the upload server pod doesn't have permission to write to the PVC. 

Any ideas for how to deal with these issues would be appreciated!

aglitke:
> > * The most frequent issue is that virtctl times out waiting for the upload pod to be "Running."  Seems to be a deadlock in the the local volume provisioner.  Basically, the PVC is waiting for a consumer and the pod is waiting for the PVC.  This usually only happens the first time the local volume provisioner creates a PVC.  I initially had the functional test handle this but took the code out.
> 
> @awels do you have an idea here? Maybe you can help @mhenriks finding the root cause?

@mhenriks You follow the pattern of other cdi controllers and create the uploadserver pod without waiting for any particular PVC status though, right?  My feeling is that this is due to some level of local storage flakiness.  I agree we should look to solve this but that should really initiate in a new issue/PR against master.  @mhenriks perhaps you could create that so we can focus the discussion here on whether what is currently in master should also be in release-0.9.

> > * The other failure I see (much more infrequently) occurs with the openshift tests.  Sometimes the upload fails because the upload server pod doesn't have permission to write to the PVC.
> 
> Hm, can as well be local storage related, but could also be a bug in in the importer CDI-pod or openshift.

mhenriks:
@aglitke yes, don't wait for any special PVC status before creating the pod.

I created https://github.com/kubevirt/kubevirt/issues/1679 to track the functional test failures.  Will address in separate PR.

PR #1678, opened 2018-11-07
add variable for setting memory size for node

### LINKED ISSUE #1677: Wrong node memory size, when windows TARGET env variable is set

 /kind bug

**What happened**:

when user starts kubevirt cluster with command 
`make cluster-up`
and sets env variable:
`export TARGET="windows2016"` 
it creates node with only 4GB memory. But according to this [code](https://github.com/kubevirt/kubevirt/blob/master/cluster/ephemeral-provider-common.sh#L34), it should creates node with 8GB of memory.

It should creates node with 8GB memory
set env variable:
`export TARGET="windows2016"`
make cluster-up
cluster/kubectl.sh describe nodes
see wrong memory size

ksimon1:
This PR adds possibility to set memory size when node is created by setting special env variable
Fixes #1677

**Release note**:
```release-note
added new env variable for setting node memory size
```

### DIFF
--- a/cluster/ephemeral-provider-common.sh
+++ b/cluster/ephemeral-provider-common.sh
@@ -29,11 +29,11 @@ function _registry_volume() {
 }
 
 function _add_common_params() {
-    local params="--nodes ${KUBEVIRT_NUM_NODES} --memory 4096M --cpu 5 --random-ports --background --prefix $provider_prefix --registry-volume $(_registry_volume) kubevirtci/${image} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
+    local params="--nodes ${KUBEVIRT_NUM_NODES} --memory ${KUBEVIRT_MEMORY_SIZE} --cpu 5 --random-ports --background --prefix $provider_prefix --registry-volume $(_registry_volume) kubevirtci/${image} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
     if [[ $TARGET =~ windows.* ]]; then
-        params="--memory 8192M --nfs-data $WINDOWS_NFS_DIR $params"
+        params=" --nfs-data $WINDOWS_NFS_DIR $params"
     elif [[ $TARGET =~ openshift.* ]]; then
-        params="--memory 4096M --nfs-data $RHEL_NFS_DIR $params"
+        params=" --nfs-data $RHEL_NFS_DIR $params"
     fi
     echo $params
 }
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -21,6 +21,7 @@ function build_func_tests() {
 
 KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-k8s-1.10.4}
 KUBEVIRT_NUM_NODES=${KUBEVIRT_NUM_NODES:-1}
+KUBEVIRT_MEMORY_SIZE=${KUBEVIRT_MEMORY_SIZE:-5120M}
 
 # Use this environment variable to set a custom pkgdir path
 # Useful for cross-compilation where the default -pkdir for cross-builds may not be writable
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -46,6 +46,15 @@ export KUBEVIRT_NUM_NODES=2 # schedulable master + one additional node
 make cluster-up
 ```
 
+You can use the `KUBEVIRT_MEMORY_SIZE` environment 
+variable to increase memory size per node. Normally you don't need it, 
+because default node memory size is set.
+
+```bash
+export KUBEVIRT_MEMORY_SIZE=8192M # node has 8GB memory size
+make cluster-up
+```
+
 You could also run some build steps individually:
 
 ```bash


ovirt-infra:

cluster/k8s-1.11.0/README.md
```diff
@@ -11,6 +11,7 @@ local machine and are the pushed to a registry which is exposed at
 ```bash
 export KUBEVIRT_PROVIDER=k8s-1.11.0
 export KUBEVIRT_NUM_NODES=2 # master + one nodes
+export KUBEVIRT_MEMORY_SIZE=8196
```
michalskrivanek: why would you add it to all example...they're fine with the default 4GB arent' they? If you need to change that for all (but I think you don't) then it's better to change the default.

cluster/ephemeral-provider-common.sh
```diff
@@ -29,11 +29,18 @@ function _registry_volume() {
 }
 
 function _add_common_params() {
-    local params="--nodes ${KUBEVIRT_NUM_NODES} --memory 4096M --cpu 4 --random-ports --background --prefix $provider_prefix --registry-volume $(_registry_volume) kubevirtci/${image} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
+    local params="--nodes ${KUBEVIRT_NUM_NODES} --cpu 4 --random-ports --background --prefix $provider_prefix --registry-volume $(_registry_volume) kubevirtci/${image} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
```
michalskrivanek: why not keeping it similar to original and just use --memory ${KUBEVIRT_MEMORY_SIZE}

cluster/ephemeral-provider-common.sh
```diff
@@ -29,11 +29,18 @@ function _registry_volume() {
...
     if [[ $TARGET =~ windows.* ]]; then
-        params="--memory 8192M --nfs-data $WINDOWS_NFS_DIR $params"
```
michalskrivanek: ...and then you probably doesn't have to mess with the size here at all, and just set it to 8GB in automation/test.sh for windows target

cluster/ephemeral-provider-common.sh
petrkotas: I would suggest stay with the 4096M as sane default. We have allocated 12GB by default in stdci and We are running 2 nodes, with 8GB it would go over the limit.

Also it would be better as @michalskrivanek noted, direct use of variable similar as {KUBEVIRT_NUM_NODES} is cleaner approach and avoids later use of ifs.

cluster/ephemeral-provider-common.sh
```diff
@@ -29,11 +29,18 @@ function _registry_volume() {
...
+        if [[ ${KUBEVIRT_MEMORY_SIZE} == 4096 ]]; then
+            params="--memory 8192M $params"
```
petrkotas: In my opinion, brute forcing difference value, when user explicitly defined some is very bad practise. If user wants to set 4GB he knows what he is doing. 

When the defaults are set as mentioned in previous comment, this if is no longer needed.

cluster/ephemeral-provider-common.sh
petrkotas: From my point, setting the variable in [test.sh](https://github.com/kubevirt/kubevirt/blob/00321ac43633411eec02c36a1aa5f4870acc4a16/automation/test.sh#L99), where defaults for windows are already set, is more readable for future changes.

cluster/k8s-1.11.0/README.md
ksimon1: Sure we can keep there default 4GB. But at least, there is some notice, that user can change memory size.

docs/env-providers.md
```diff
@@ -28,6 +28,7 @@ Usage:
 ```bash
 export KUBEVIRT_PROVIDER=k8s-1.10.4 # choose this provider
 export KUBEVIRT_NUM_NODES=3 # master + two nodes
+export KUBEVIRT_MEMORY_SIZE=4096
```
rmohr: Why do I have to extra export the default value for every provider?

hack/common.sh
```diff
@@ -21,6 +21,7 @@ function build_func_tests() {
 
 KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-k8s-1.10.4}
 KUBEVIRT_NUM_NODES=${KUBEVIRT_NUM_NODES:-1}
+KUBEVIRT_MEMORY_SIZE=${KUBEVIRT_MEMORY_SIZE:-4096}
```
rmohr: Please make the `M` part of the default value.

cluster/ephemeral-provider-common.sh
```diff
@@ -29,11 +29,11 @@ function _registry_volume() {
 }
 
 function _add_common_params() {
-    local params="--nodes ${KUBEVIRT_NUM_NODES} --memory 4096M --cpu 4 --random-ports --background --prefix $provider_prefix --registry-volume $(_registry_volume) kubevirtci/${image} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
+    local params="--nodes ${KUBEVIRT_NUM_NODES} --memory ${KUBEVIRT_MEMORY_SIZE}M --cpu 4 --random-ports --background --prefix $provider_prefix --registry-volume $(_registry_volume) kubevirtci/${image} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
     if [[ $TARGET =~ windows.* ]]; then
-        params="--memory 8192M --nfs-data $WINDOWS_NFS_DIR $params"
```
rmohr: With the new location only stdci would set the correct memory size. The value needs to be either part of `automation/test.sh` or part of this file.

docs/env-providers.md
ksimon1: You don't have to export default value. It is for user to show him, he can change the memory size.

cluster/ephemeral-provider-common.sh
michalskrivanek: +1 on test.sh, line 55 looks like a best place

rmohr: Overall, I think that just fixing the if statement to not include `--memory` two times here would have been the best thing to do. The build environment is not a general-purpose tool and the scripts *support* the `TARGET` env vars explicitly. I don't see much of a reason to play with the memory values, since CI should test over- and over that the memory size is sufficient. Thoughts?

ksimon1: but then all tests which runs this script will have 8GB of memory/node. I think line 100 will be better.

rmohr: If 8GB are not needed,we can reduce the value. The only thing we want to achieve is this: "Every node of the test environment should be capable of running a windows machine".

rmohr: @cynepco3hahue I remember we had to choose 8gb to make some tests pass, but it looks like we by accident always used 4gb and the tests work fine. Can that be?

rmohr: I wonder if some of our flaky tests in openshift will go away once we fix the `if` statement.

automation/test.sh
```diff
@@ -97,6 +97,8 @@ if [[ $TARGET =~ openshift.* ]]; then
 fi
 
 if [[ $TARGET =~ windows.* ]]; then
+  # Set memory size for node
+  export KUBEVIRT_MEMORY_SIZE=8192M
```
cynepco3hahue: Please remove

hack/common.sh
```diff
@@ -21,6 +21,7 @@ function build_func_tests() {
 
 KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-k8s-1.10.4}
 KUBEVIRT_NUM_NODES=${KUBEVIRT_NUM_NODES:-1}
+KUBEVIRT_MEMORY_SIZE=${KUBEVIRT_MEMORY_SIZE:-4096M}
```
cynepco3hahue: Can you please change it to `5120`

mfranczy:
I think it would be nice if information about ```KUBEVIRT_MEMORY_SIZE``` with a short description would appear only in one place, like ```KUBEVIRT_NUM_NODES``` [link](https://github.com/kubevirt/kubevirt/blob/master/docs/getting-started.md#user-content-compile-and-run-it), maybe it's worth to extend the same section. 

The reason why I think in that way...  this env is not always required to modify, let's imagine that I have to add another optional env then should I add it everywhere as well? However, If a specific provider requires a bigger memory size then it should be mentioned in a provider doc.

cluster/k8s-1.10.4/README.md
```diff
@@ -11,6 +11,7 @@ local machine and are the pushed to a registry which is exposed at
 ```bash
 export KUBEVIRT_PROVIDER=k8s-1.10.4
 export KUBEVIRT_NUM_NODES=2 # master + one nodes
+#export KUBEVIRT_MEMORY_SIZE=8192M # default memory size is 4096M, uncomment this variable only if node needs more ram
```
mfranczy: I think it would be nice if information about ```KUBEVIRT_MEMORY_SIZE``` with a short description would appear only in one place, like ```KUBEVIRT_NUM_NODES``` [link](https://github.com/kubevirt/kubevirt/blob/master/docs/getting-started.md#user-content-compile-and-run-it), maybe it's worth to extend the same section. 

The reason why I think in that way...  this env is not always required to modify, let's imagine that I have to add another optional env then should I add it everywhere as well? However, If a specific provider requires a bigger memory size then it should be mentioned in a provider doc.

ksimon1:
@cynepco3hahue For our windows template test in [common-templates](https://github.com/kubevirt/common-templates) repo we would appreciate some variable, with which we could set memory size for node. Because windows templates needs 4 and 6GB memory size and we would like to test them without any memory modification, like it is now.

rmohr:
ci test please

rmohr:
ci test please

PR #1681, opened 2018-11-08
Add http to scheme of image-upload if not there

karmab:

Fixes #

**Release note**:

```release-note
automatically adds http to image uploader proxy url if not present
```

### DIFF
--- a/pkg/virtctl/imageupload/imageupload.go
+++ b/pkg/virtctl/imageupload/imageupload.go
@@ -23,6 +23,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"net/url"
 	"os"
 	"time"
 
@@ -171,7 +172,12 @@ func (c *command) run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-
+	u, err := url.Parse(uploadProxyURL)
+	if err != nil {
+		return err
+	} else if u.Scheme == "" {
+		uploadProxyURL = fmt.Sprintf("http://%s", uploadProxyURL)
+	}
 	fmt.Printf("Uploading data to %s\n", uploadProxyURL)
 
 	err = uploadData(uploadProxyURL, token, file, insecure)


pkg/virtctl/imageupload/imageupload.go
```diff
@@ -172,6 +172,9 @@ func (c *command) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if !strings.HasPrefix(uploadProxyURL, "http") {
+		uploadProxyURL = fmt.Printf("http://$%s", uploadProxyURL)
```
rmohr: Maybe dumb question, but wouldn't we want to sue `https` as default nowadays?

rmohr: ok, then let's go with `http`. Can you check for `http://`?

rmohr: Maybe you can do something like `url.Parse(str)` and then checke if the `Scheme` is set.

karmab: done

rmohr: Almost there. You will have to check if we see an error. For instance `htt@p://test` would fail and then we would try to access `u` which is `nil`.

karmab: right, addressed

pkg/virtctl/imageupload/imageupload.go
```diff
@@ -171,7 +171,12 @@ func (c *command) run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-
+	u, err := url.Parse(uploadProxyURL)
+	if err != nil {
+		return err
+	} else if u.Scheme == "" {
+		uploadProxyURL = fmt.Printf("http://$%s", uploadProxyURL)
```
rmohr: I guess that the `$` in `fmt.Printf` was not intentional?

karmab: indeed :) fixed

pkg/virtctl/imageupload/imageupload.go
```diff
@@ -171,7 +171,12 @@ func (c *command) run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-
+	u, err := url.Parse(uploadProxyURL)
+	if err != nil {
+		return err
+	} else if u.Scheme == "" {
+		uploadProxyURL = fmt.Printf("http://%s", uploadProxyURL)
```
rmohr: Also you want `fmt.Sprintf` :dagger:

rmohr:
@karmab could you run "make" and commit the difference, so that the code is formatted? Travis complains.

karmab:
> @karmab could you run "make" and commit the difference, so that the code is formatted? Travis complains.

done

PR #1682, opened 2018-11-08
fix virtctl image-upload tests

### LINKED ISSUE #1679: virtctl image-upload functional test failures

/kind bug

**What happened**:

Occasionally the functional tests for "virtctl image-upload" fails for one of the following reasons:

1.  The most frequent issue is that virtctl times out waiting for the upload pod to be "Running." Seems to be a deadlock in the the local volume provisioner. Basically, the PVC is waiting for a consumer and the pod is waiting for the PVC. This usually only happens the first time the local volume provisioner creates a PVC. I initially had the functional test handle this but took the code out.

2.  The other failure I see (much more infrequently) occurs with the openshift tests. Sometimes the upload fails because the upload server pod doesn't have permission to write to the PVC.

Functional tests should pass

Check master functional test results

mhenriks:

Often the virtctl image upload functional times out.  The system gets in a weird state.  The upload pvc is waiting for a pod and the pod is waiting for a pvc.  Deadlock occurs.

Fixes #1679 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -120,7 +120,13 @@ const (
 	NamespaceTestAlternative = "kubevirt-test-alternative"
 )
 
-const LocalStorageClass = "local"
+const (
+	LocalStorageClass = "local"
+
+	HostPathStorageClass = "host-path"
+
+	BlockVolumeStorageClass = "block-volume"
+)
 
 var testNamespaces = []string{NamespaceTestDefault, NamespaceTestAlternative}
 
@@ -364,10 +370,10 @@ func BeforeTestSuitSetup() {
 	createServiceAccounts()
 
 	CreateHostPathPv(osAlpineHostPath, HostPathAlpine)
-	CreatePVC(osAlpineHostPath, defaultDiskSize)
+	CreateHostPathPVC(osAlpineHostPath, defaultDiskSize)
 
-	CreatePVC(osWindows, defaultWindowsDiskSize)
-	CreatePVC(osRhel, defaultRhelDiskSize)
+	CreateLocalPVC(osWindows, defaultWindowsDiskSize)
+	CreateLocalPVC(osRhel, defaultRhelDiskSize)
 
 	EnsureKVMPresent()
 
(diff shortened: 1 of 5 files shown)


mhenriks:
ci test please

cynepco3hahue:
@mhenriks I do not see any upload image test failures, can you please explain what the flag `"--storage-class", "local"` is doing?

tests/utils.go
```diff
@@ -120,7 +120,13 @@ const (
 	NamespaceTestAlternative = "kubevirt-test-alternative"
 )
 
-const LocalStorageClass = "local"
+const (
+	LocalStorageClass = "local"
+
+	HostPathStorageClass = "host-path"
+
+	BlockVolumeStorageClass = "block-volume"
```
rmohr: should this be `local-block`?

mhenriks: Good question!  Not sure it matters.  But I think it may actually be better to use a different storage class.  

The PVC in local-block-storage is referenced by one of our example manifests (vmi-block-pvc.yaml) and is not used in any functional tests.  The PVs created in the functional tests share the same block device as vmi-block-pvc.yaml, though.  And I think if you are going to have two PVs share the same device it would be better to have them in different storage classes.  Although it's probably still a bad situation.  Perhaps I should allocate another block device for functional tests?

mhenriks:
@cynepco3hahue it shouldn't matter because `local` is the default storage class.  Just being more explicit.  Although based on test results it may be having some effect.

mhenriks:
ci test please

tests/utils.go
slintes: Hi, the storage class has to be "local" for the block volume PVs, because they are handled by the local storage provider, which is responsible for the "local" storage class, and has access to the actual block device which is used (see https://github.com/kubevirt/kubevirt/blob/master/images/disks-images-provider/entrypoint.sh#L25)

slintes: and this is the mentioned local storage provisioner: https://github.com/kubevirt/kubevirtci/blob/master/manifests/local-volume.yaml

mhenriks: @slintes the local volume provisioner (called "local") that is created by kubevirtci does not do anything with block volumes.  Look at the code here:  https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume/provisioner.  It creates PVs from directories on a filesystem and only deals with the PVs it creates.
 
And having block volumes share the same storage class name was causing errors with upload tests because a block volume was being assigned to the upload server pod instead of a filesystem PV.

Let's look at the results of the tests (if they ever complete) to see if changing the storage class name for block volumes causes a test failure.  I don't think it does.

mhenriks: @slintes whoops!  it does handle block PVs.  I apologize.  But in my testing it appears that having a storage class that has both block and filesystem PVs, weirdness occurs.  So maybe we should think about about having separate provisioners.

slintes: I'm not familiar yet with the upload, but I assume you create a PVC or a PV? Did you try setting volumeMode=Filesystem on it?

slintes: mh, but probably you don't want to care about the mode, and it's only an issue in this special environment. So, maybe using 2 provisioners is the way to go indeed.

fabiand:
Any update?

mhenriks:
@fabiand @slintes @rmohr - been on vacation last two weeks so just getting back around to this.  Want to get a clean test run to see where we're at.

mhenriks:
ci test please

mhenriks:
@fabiand @rmohr @slintes last two test runs had no image upload failures.  Will try one more.

mhenriks:
ci test please

PR #1683, opened 2018-11-09
Update workflow document

rthallisey:
Document major and minor upgrades for kubevirt.

```release-note
NONE
```


docs/upgrade.md
```diff
@@ -0,0 +1,42 @@
...
+ - Rolling update of virt-controllers pods
+ - Test virt-controller functionality with kubevirt-1.1 data
+ - Rolling update of virt-api pods
+ - Test virt-api functionality with kubevirt-1.1 data
+ - Register kubevirt-1.2 CRD webhooks
+ - Update to kubevirt-1.2 CRDs
+ - Test virt-controller and virt-api functionality with kubevirt- 1.1 and kubevirt-1.2 data
+
+### Virt-launcher Upgrade
+The virt-launcher will be able to function with n-1 and n+1 releases.
```
rmohr: Let's specify this a little bit more. I think something like this is less ambiguous:

```
virt-launcher, which is part of release x.y will be compatible with:
 * all minor kubevirt versions >= y
 * at least the next major version (x + 1)
```

Let's also mention that we may introduce further restrictions once migrations are production ready. Something like

```
Once migrations are production-ready additional compatibility restrictions may be introduced. For example: virt-launcher from x.y can only migrate to other virt-launcher instances of a new major release x +1 and minor release 1. 
```

fabiand:
ping?

rthallisey:
hey @fabiand.  I'll get back to this next week.  Some other tasks had to take priority.

dcritch:
@rthallisey I've been testing some upgrade scenarios and here's what I've got so far.

Major updates, we're kind of screwed right now with the namespace change.

For minor, I did the following and everything seemed to work:
1/ installed the latest kubevirt-manifests RPM
2/ oc apply'ed the new manifest
3/ patched the various daemonsets to always pull the image, e.g.:
`oc patch ds virt-handler -p '{"spec":{"template":{"spec":{"$setElementOrder/containers":[{"name":"virt-handler"}],"containers":[{"imagePullPolicy":"Always","name":"virt-handler"}]}}}}'`
(repeat for all daemonsets and deployment.apps e.g. kube-multus-ds-amd64, ovs-cni-plugin-amd64 and ovs-vsctl-amd64, etc)
4/ delete all the pods to force a fresh pull
`for ds in virt-handler kube-multus-ds-amd64 ovs-cni-plugin-amd64 ovs-vsctl-amd64; do
  echo killing $ds pods
  oc get pods | grep $ds | awk '{print $1}' | xargs oc delete pod
done
`
5/ patch the daemonsets and deploymentapps to revert to default image pull policy, e.g.:
oc patch ds virt-handler -p '{"spec":{"template":{"spec":{"$setElementOrder/containers":[{"name":"virt-handler"}],"containers":[{"imagePullPolicy":"IfNotPresent","name":"virt-handler"}]}}}}'

During the process, running VMs are unaffected and remain available. As you mentioned, they will need to be restarted to pick up the new image. Once a VM has been started one each node with the new image, or after being manually pulled to each node (more cumbersome) the virt-controller deployment can then be patched back to IfNotPresent.

rthallisey:
Moving to https://github.com/kubevirt/kubevirt/pull/1922

PR #1685, opened 2018-11-12
openapi-gen removed from kubernetes/code-generator

mhenriks:

docker-builder container is broken because openapi-gen was removed from kubernetes/code generator in this commit: https://github.com/kubernetes/code-generator/commit/384879d8d5e96ba89dec26f061cd4a0c533f9de4

So, build openapi-gen from kubernetes/kube-openapi.  Unfortunately, that repo is not tagged by k8s release, so use current latest commit.

Fixes #

**Release note**:

```release-note
NONE
```


rmohr:
@mhenriks thanks! :+1:  I guess, first checking out the branch and then  going into the directory would have worked too.

mhenriks:
@rmohr - yeah, I considered checking out the entire kubernetes/code-generator tree then switching to the appropriate release branch once to build all the generators.  But that won't work long term.  Only up to 1.12.  Hopefully they start tagging the kube-openapi repo, though

PR #1688, opened 2018-11-12
Ensure that clusters have enough CPU cores to run multi-core tests

### LINKED ISSUE #1690: "VirtualMachineInstance definition with 3 CPU cores should map cores to virtio net queues" fails
/kind bug

**What happened**:
CI fails on:
```
 (root).Tests Suite.Invoking jobs / check-patch.k8s-multus-1.11.1-release.el7.x86_64 / Configurations VirtualMachineInstance definition with 3 CPU cores should map cores to virtio block queues
5
2472, 2471, 2469, 2467, 2466
(root).Tests Suite.Invoking jobs / check-patch.k8s-multus-1.11.1-release.el7.x86_64 / Configurations VirtualMachineInstance definition with 3 CPU cores should map cores to virtio net queues
```

https://jenkins.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/test_results_analyzer/

**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

rmohr:

Under certain circumstances our test-infra consumed more than 1.0 cpu cores
which could make nodes non-fit for our cpu core tests. Increasing from 4
to 5 cpu cores to give the infrastrucure 2.0 cores.

Makes it possible to run the mult-core tests under more reliable conditions in the 1.11.1-multus clusters:

```
Configurations VirtualMachineInstance definition with 3 CPU cores should map cores to virtio net queues
Configurations VirtualMachineInstance definition with 3 CPU cores should map cores to virtio block queues
```

Further fixes

```
[Fail] IOThreads IOThreads Policies [It] Should honor shared ioThreadsPolicy 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_iothreads_test.go:65
```

Fixes #1690

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster/ephemeral-provider-common.sh
+++ b/cluster/ephemeral-provider-common.sh
@@ -29,7 +29,7 @@ function _registry_volume() {
 }
 
 function _add_common_params() {
-    local params="--nodes ${KUBEVIRT_NUM_NODES} --memory 4096M --cpu 4 --random-ports --background --prefix $provider_prefix --registry-volume $(_registry_volume) kubevirtci/${image} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
+    local params="--nodes ${KUBEVIRT_NUM_NODES} --memory 4096M --cpu 5 --random-ports --background --prefix $provider_prefix --registry-volume $(_registry_volume) kubevirtci/${image} ${KUBEVIRT_PROVIDER_EXTRA_ARGS}"
     if [[ $TARGET =~ windows.* ]]; then
         params="--memory 8192M --nfs-data $WINDOWS_NFS_DIR $params"
     elif [[ $TARGET =~ openshift.* ]]; then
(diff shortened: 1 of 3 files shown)


cynepco3hahue:
Can we guarantee that it will consume only 1-2 cores? Maybe we can set CPU limit on the `gocli run` command.

rmohr:
@cynepco3hahue I think that we could only do that easily if we e.g. move all multus components in their own namespace and set a namespace limit. But I guess that then kubernetes would evict pods there if they go over the limits ... Other thoughts?

cynepco3hahue:
@rmohr Or we can just skip this tests for multus environment :smile: I think it does not really need, because we have other jobs that run this test, and multus should not change anything in this case.

rmohr:
For reference: https://jenkins.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/2486/artifact/ci_build_summary.html

rmohr:
@vladikr @stu-gott, now it should finally be ok. Maybe you want to do another review.

rmohr:
All three tests pass now across all lanes.

PR #1691, opened 2018-11-12
Kubevirt namespace

stu-gott:
Switch default namespace to 'kubevirt'

```release-note
Default namespace for KubeVirt components is now kubevirt
```

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -144,7 +144,7 @@ if [ -n "${JOB_NAME}" ]; then
 namespace=${NAMESPACE}
 EOF
 else
-  export NAMESPACE="${NAMESPACE:-kube-system}"
+  export NAMESPACE="${NAMESPACE:-kubevirt}"
 fi
 
 
@@ -179,8 +179,8 @@ kubectl get nodes
 make cluster-sync
 
 # OpenShift is running important containers under default namespace
-namespaces=(kube-system default)
-if [[ $NAMESPACE != "kube-system" ]]; then
+namespaces=(kubevirt default)
+if [[ $NAMESPACE != "kubevirt" ]]; then
   namespaces+=($NAMESPACE)
 fi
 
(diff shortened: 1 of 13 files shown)


fabiand:
/cc @aglitke @mhenriks please note the mentioned carshes.

@stu-gott nice patch :) Could you please also cross-reference the user-guide updates with this bug, to see when they are in place as well?

aglitke:
See https://github.com/kubevirt/containerized-data-importer/issues/531 .  To unblock this, please keep CDI in the kube-system namespace until it can also be moved.

fabiand:
Ping?

rmohr:
@stu-gott @aglitke I don't understand what is blocking us here for moving KubeVirt. On CI we run KubeVirt already completely in random namespaces. We  don't have to install CDI in a special namespace.

fabiand:
@aglitke please provide a feedback today, otherwise I'd suggest to move forward as is.

stu-gott:
@rmohr the problem with this PR as written is that the RBAC rules we create for KubeVirt moved to the new namespace. CDI was depending on them still being in the `kube-system` namespace.

Two possible approaches:

1. KubeVirt keeps an RBAC rule around allowing read-write to ConfigMaps in `kube-system`
2. CDI creates their own RBAC rules for namespaces those components depend on, so that the projects can be independent.

The better approach from my POV is the second path. Is that reasonable @aglitke @mhenriks ?

fabiand:
Ah, so it sounds as if the rule is about this specific configmap, then I'm
good.

stu-gott:
Please note I made the namespace CDI components use configurable separately.

fabiand:
@stu-gott do you know if the CI passes on your local host?

stu-gott:
ci test please

stu-gott:
ci test please

stu-gott:
I believe the one remaining failing test might be spurious.

fabiand:
ci test please

fabiand:
All k8s tests pass, thus looks good to me.

fabiand:
The remaining kube-system namespaces are related to CDI and vendored code.

PR #1716, opened 2018-11-16
Make virtctl expose more compatible with kubectl expose

rmohr:

I discovered that we introduced a lot of confusion with our `--node-port` and with making `--port` a required field on `virtctl expose`. We wanted to have `virtctl expose` to be have similar to `kubectl expose`. `kubectl` does not have  a `--node-port` flag. If `--port` is not specified, `kubectl` introspects the objects for the ports to expose.

This PR

 * removes the `--node-port` flag
 * adds port introspection, in case `--port` is omitted

Fixes #

**Release note**:

```release-note
* Remove "--node-port" from "virtctl expose"
* Add port auto-discovery in case "--port" is omitted
```


pkg/virtctl/expose/expose.go
```diff
@@ -76,8 +75,11 @@ virtualmachineinstance (vmi), virtualmachine (vm), virtualmachineinstancereplica
 }
 
 func usage() string {
-	usage := `  # Expose SSH to a virtual machine instance called 'myvm' as a port (5555) and specify each node open up port 30001 on the cluster:
-  virtctl expose vmi myvm --port=5555 --node-port=30001 --target-port=22 --name=myvm-ssh --type=NodePort")`
+	usage := `  # Expose SSH to a virtual machine instance called 'myvm' on each node via a NodePort service:
+  virtctl expose vmi myvm --port=22 --name=myvm-ssh --type=NodePort
```
yuvalif: would be good to keep the port vs. target-port distinction in the example, so it is clear they can be different

pkg/virtctl/expose/expose.go
```diff
@@ -201,3 +215,25 @@ func (o *Command) RunE(cmd *cobra.Command, args []string) error {
...
+			podNetworkName = network.Name
+			break
+		}
+	}
+	if podNetworkName != "" {
+		for _, device := range vmiSpec.Domain.Devices.Interfaces {
+			if device.Name == podNetworkName {
+				ports := []v1.ServicePort{}
+				for i, port := range device.Ports {
+					ports = append(ports, v1.ServicePort{Name: fmt.Sprintf("port-%d", i+1), Protocol: v1.Protocol(port.Protocol), Port: port.Port})
```
yuvalif: maybe "k6t-port-..." ?

rmohr: Not sure. I just copied what the kubelet does. What would be the benefit?

pkg/virtctl/expose/expose.go
rmohr: Let me check again, but I think that in the case of `NodePort`, kubectl only uses `--port` and ignores `--target-port`.

pkg/virtctl/expose/expose.go
yuvalif: prevent conflicts with someone calling "kubectl..." on the pod?

rmohr: Ah I see what you want to get at. So that command creates a service with ports where the ports in the service are named `port-${x}`. That is not related to the port name on the Pod. At the end `kubectl` and `virtctl` just create a final service object. If they happen to have the same service-name then only one can succeed. The same like with two parallel kubectl calls.

slintes:
ci test please

PR #1717, opened 2018-11-16
Fix CPU placement test

rmohr:

In case that no core count was set, but dedicated CPUs were requested
via the cpu resource, only one cpu was set inside the VMI, where it
should have been equal to the number of requested dedicated cpu cores.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/api/converter.go
+++ b/pkg/virt-launcher/virtwrap/api/converter.go
@@ -742,16 +742,14 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 
 	if vmi.Spec.Domain.CPU != nil {
 		// Set VM CPU cores
-		if vmi.Spec.Domain.CPU.Cores != 0 {
-			domain.Spec.CPU.Topology = &CPUTopology{
-				Sockets: 1,
-				Cores:   vmi.Spec.Domain.CPU.Cores,
-				Threads: 1,
-			}
-			domain.Spec.VCPU = &VCPU{
-				Placement: "static",
-				CPUs:      vmi.Spec.Domain.CPU.Cores,
-			}
+		domain.Spec.CPU.Topology = &CPUTopology{
+			Sockets: 1,
+			Cores:   calculateRequestedVCPUs(vmi),
+			Threads: 1,
+		}
+		domain.Spec.VCPU = &VCPU{
+			Placement: "static",
+			CPUs:      calculateRequestedVCPUs(vmi),
 		}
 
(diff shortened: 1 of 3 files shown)


pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -992,6 +990,9 @@ func calculateRequestedVCPUs(vmi *v1.VirtualMachineInstance) uint32 {
 	if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Cores != 0 {
 		return vmi.Spec.Domain.CPU.Cores
 	}
+	if !vmi.IsCPUDedicated() {
```
cynepco3hahue: Hm I believe we must set defaults under mutating webhook, so we should not get the situation when `vmi.Spec.Domain.CPU == nil || vmi.Spec.Domain.CPU.Cores == 0`

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -992,6 +990,9 @@ func calculateRequestedVCPUs(vmi *v1.VirtualMachineInstance) uint32 {
...
+		return uint32(1)
+	}
 	if cpuRequests, ok := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceCPU]; ok {
```
cynepco3hahue: I believe I missed it, but do we have the check that `uint32(cpuRequests.Value()) != 0`?

tests/vmi_configuration_test.go
```diff
@@ -816,7 +816,7 @@ var _ = Describe("Configurations", func() {
 				}
 				cpuVmi.Spec.Domain.Resources = v1.ResourceRequirements{
 					Requests: kubev1.ResourceList{
-						kubev1.ResourceMemory: resource.MustParse("64M"),
+						kubev1.ResourceMemory: resource.MustParse("128M"),
```
cynepco3hahue: hm From where coming memory change?

rmohr: I changed it back, since that is not a proper fix. It looks like we don't request enough overhead. Our dedicated CPU VMIs have a QoS class of guaranteed, which also means that they are not allowed to use more memory than the requested memory + overhead. I have seen that they get sometimes killed because of the OOM killer. Might be that our overhead calculation is not good enough. @MarSik FYI. Do we already have some libvirt/qemu numbers in order to see if we should add more overhead?

rmohr:
> was passing somehow re-affirmed my thoughts.

Yep the issue with the tests was, that the expecter timed out, but the returned error was not checked ...

PR #1720, opened 2018-11-19
Add virtio-win registry disk

petrkotas:
Package virtio win drivers as a registry disk
to provide easy distribution for kubevirt users

Fixes #

**Release note**:

```release-note
Add virtio win registry disk
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/1720)

### DIFF
--- a/images/virtio-container-disk/Dockerfile
+++ b/images/virtio-container-disk/Dockerfile
@@ -1,0 +1,34 @@
+#
+# This file is part of the KubeVirt project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2017 Red Hat, Inc.
+#
+
+FROM fedora:28 as builder
+
+RUN curl https://fedorapeople.org/groups/virt/virtio-win/virtio-win.repo -o /etc/yum.repos.d/virtio-win.repo \
+&& dnf install -y virtio-win && mkdir -p /disk 
+
+# the virtio-win package is shipped with version number, move it to standardized location
+# with standardized name, docker copy does not follow links
+RUN cp -L /usr/share/virtio-win/virtio-win.iso /disk/virtio-win.iso
+
+
+FROM kubevirt/registry-disk-v1alpha 
+
+LABEL maintainer="The KubeVirt Project <kubevirt-dev@googlegroups.com>"
+
+WORKDIR /disk
+COPY --from=builder /disk/virtio-win.iso ./virtio-win.iso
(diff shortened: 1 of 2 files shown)


fabiand:
@michalskrivanek can you please doa  review?

images/virtio-container-disk/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2017 Red Hat, Inc.
+#
+
+FROM fedora:28 as builder
```
michalskrivanek: could we use something more stable? e.g. centos. Or anything small/minimal/stable

images/virtio-container-disk/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+RUN curl https://fedorapeople.org/groups/virt/virtio-win/virtio-win.repo -o /etc/yum.repos.d/virtio-win.repo \
+&& dnf install -y virtio-win && mkdir -p /disk 
+
+# the virtio-win package is shipped with version number, move it to standardized location
+# with standardized name
+RUN mv /usr/share/virtio-win/virtio-win-*.iso /disk/virtio-win.iso
```
michalskrivanek: what? there is a symlink right there

images/virtio-container-disk/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+
+FROM kubevirt/registry-disk-v1alpha 
+
+LABEL maintainer="The KubeVirt Project <kubevirt-dev@googlegroups.com>"
+
+WORKDIR /disk
+COPY --from=builder /disk/virtio-win.iso ./virtio-win.iso
```
michalskrivanek: what about the guest agent binaries?

images/virtio-container-disk/Dockerfile
petrkotas: Every image we use for upstream is based on fedora. Therefore it make sense to base even this one on Fedora.

michalskrivanek: there are 2 F27 already in the code base, F28 will slowly get outdated too. I see in 2 cases "FROM fedora" - would that work? There's nothing specific to a Fedora version here

We should use something better than a random Fedora from the time the code was written originally. But that's a common problem for all the containers in kubevirt, so indeed doesn't have to be solved here

petrkotas: Would you kindly please point to specific code locations where do you see it is broken? Thanks in advance.

images/virtio-container-disk/Dockerfile
petrkotas: Actually good point. Changed it to cp -L

images/virtio-container-disk/Dockerfile
petrkotas: We do not need them. This image is container disk providing ISO image. ISO contains all binaries required. There are both windows drivers and guest agent install files.

petrkotas:
@slintes @fabiand what do you think? Can we merge?

petrkotas:
I am working on it currently. It cannot be easily tested, since it has to
be done manually.
The best I can think of is mount the disk a and verify the files are there
as designed.
--
s pozdravem,
Petr Kotas

fabiand:
That would be a great start!

Issue #1247, opened 2018-06-29
Cluster level CPU model


fabiand:
In order to

a. be able to set CPU models on VMs
b. Make this transparaent to the user

the following **rough** plan is proposed

1. Add an VM API field to set the CPU family (virsh cpu-models x86_64) to one of the known ones by libvirt. If it is not supported by a node, then we fail there. End of story.
2. Expose the CPU families per node as labels
3. Modify the controller to set the correct selector to schedule VMs with a specific family only on nodes which support this family.

4. Investigate K8s compatible abstraction mechanism to make this less fine granular (comparable to ResourceClasses and Resources)

Step 4 is further out.

fabiand:
@berrange @vladikr @michalskrivanek How does this look with your OS/oVirt hat's on?

berrange:
It is not sufficient to only set a named CPU model. it is required to be able to also set various extra named flags to protect against security flaws.  You don't necessarily need to expose ability to set CPU flags to the user, but if you choose not to, then you must have code that knows how & when to set the required extra flags correctly.

NB, in openstack there are three settings:

cpu_mode=host-passthrough|host-model|custom
cpu_model=name of a CPU model when cpu_mode==custom
cpu_extra_features=list of extra features, when cpu_mode=host-model or custom

michalskrivanek:
oVirt focuses on named models. We now have the ability to add or remove individual flags, but currently use a "friendly" name for end users in all cases
Also support host passthrough, but that's a minor use case since we then require host pinning and block migration for simplicity.

so far we added a new friendly name for each security flaw, but we're considering to rather use something like secure/insecure variant of each cpu family as those new flags seem to keep coming...

fabiand:
@MarSik I think this has been addressed by the cpu labels on nodes and selectors on the kubevirt side?

PR #1345, opened 2018-07-19
Add possibility to use CPU passthrough for VM's

cynepco3hahue:
**What this PR does / why we need it**:
You can specify `passthrough` as the CPU model and it will
automatically passthrough CPU model from the node to the VM.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1340 

**Special notes for your reviewer**:

**Release note**:

```release-note
add the possibility to pass-through the node CPU to the VM
```

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

PR #1721, opened 2018-11-19
Schedule VMs on nodes which support the requested CPU model

yanirq:
The goal is to set a nodeSelector on the pod that virt-controller creates for vmis, which selects only nodes for scheduling that provide the CPU model set on the VM.
The CPU model is taken from the VirtualMachine spec and should match the CPU family label on nodes.
The CPU family labels on nodes will be created using a [node feature discovery](https://github.com/kubernetes-incubator/node-feature-discovery) plugin.

This feature is disabled by default and can be enabled using the feature-gate : "CPUNodeDiscovery" 

Handles #1247 

The final format of the label is not determined yet (also depends on [Plugin mechanism for customer-specific labels](https://github.com/kubernetes-incubator/node-feature-discovery/issues/144))
Currently its in the form of:
feature.node.kubernetes.io/cpu-model-\<cpu family\> = "true" 

**Release note**:

```release-note
To enable scheduling virtual machines on nodes by matching cpu model on the virtual machine spec
and cpu model labels on nodes (generated and added to node using node feature discovery plugin) :  
Enable the feature gate CPUNodeDiscovery - add it to the kubevirt-config ConfigMap.
The feature is disabled by default.   

```

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -25,6 +25,8 @@ import (
 	"strconv"
 	"strings"
 
+	"kubevirt.io/kubevirt/pkg/feature-gates"
+
 	k8sv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -58,6 +60,9 @@ const CAP_SYS_NICE = "SYS_NICE"
 // LibvirtStartupDelay is added to custom liveness and readiness probes initial delay value.
 // Libvirt needs roughly 10 seconds to start.
 const LibvirtStartupDelay = 10
+//This is a perfix for node feature discovery, used in a NodeSelector on the pod
+//to match a VirtualMachineInstance CPU model(Family) to nodes that support this model.
+const NFD_CPU_FAMILY_PREFIX = "feature.node.kubernetes.io/nfd-cpu-family-"
 
 const MULTUS_RESOURCE_NAME_ANNOTATION = "k8s.v1.cni.cncf.io/resourceName"
 
@@ -612,6 +617,12 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		nodeSelector[k] = v
 
 	}
+	if featuregates.CPUNodeDiscoveryEnabled() {
+		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" {
+			nodeSelector[NFD_CPU_FAMILY_PREFIX+vmi.Spec.Domain.CPU.Model] = "true"
+		}
+	}
+
 	nodeSelector[v1.NodeSchedulable] = "true"
 
 	podLabels := map[string]string{}
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -25,8 +25,6 @@ import (
 	"strconv"
 	"strings"
 
-	"kubevirt.io/kubevirt/pkg/feature-gates"
-
 	k8sv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -37,6 +35,7 @@ import (
 	v1 "kubevirt.io/kubevirt/pkg/api/v1"
 	"kubevirt.io/kubevirt/pkg/config"
 	containerdisk "kubevirt.io/kubevirt/pkg/container-disk"
+	"kubevirt.io/kubevirt/pkg/feature-gates"
 	"kubevirt.io/kubevirt/pkg/hooks"
 	"kubevirt.io/kubevirt/pkg/kubecli"
 	"kubevirt.io/kubevirt/pkg/log"
@@ -60,6 +59,7 @@ const CAP_SYS_NICE = "SYS_NICE"
 // LibvirtStartupDelay is added to custom liveness and readiness probes initial delay value.
 // Libvirt needs roughly 10 seconds to start.
 const LibvirtStartupDelay = 10
+
 //This is a perfix for node feature discovery, used in a NodeSelector on the pod
 //to match a VirtualMachineInstance CPU model(Family) to nodes that support this model.
 const NFD_CPU_FAMILY_PREFIX = "feature.node.kubernetes.io/cpu-model-"
@@ -145,6 +145,15 @@ func isSRIOVVmi(vmi *v1.VirtualMachineInstance) bool {
 		if iface.SRIOV != nil {
 			return true
 		}
+	}
+	return false
+}
+
+func IsCPUNodeDiscoveryEnabled(store cache.Store) bool {
+	var value string
+	value, _ = getConfigMapEntry(store, featuregates.FEATURE_GATES_KEY)
+	if strings.Contains(value, featuregates.CPUNodeDiscoveryGate) {
+		return true
 	}
 	return false
 }
@@ -617,7 +626,7 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		nodeSelector[k] = v
 
 	}
-	if featuregates.CPUNodeDiscoveryEnabled() {
+	if IsCPUNodeDiscoveryEnabled(t.configMapStore) {
 		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" {
 			if vmi.Spec.Domain.CPU.Model != v1.CPUModeHostModel && vmi.Spec.Domain.CPU.Model != v1.CPUModeHostPassthrough {
 				nodeSelector[NFD_CPU_FAMILY_PREFIX+vmi.Spec.Domain.CPU.Model] = "true"
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -618,9 +618,10 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 
 	}
 	if featuregates.CPUNodeDiscoveryEnabled() {
-		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" &&
-			vmi.Spec.Domain.CPU.Model != "host-model" && vmi.Spec.Domain.CPU.Model != "host-passthrough" {
-			nodeSelector[NFD_CPU_FAMILY_PREFIX+vmi.Spec.Domain.CPU.Model] = "true"
+		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" {
+			if vmi.Spec.Domain.CPU.Model != v1.CPUModeHostModel && vmi.Spec.Domain.CPU.Model != v1.CPUModeHostPassthrough {
+				nodeSelector[NFD_CPU_FAMILY_PREFIX+vmi.Spec.Domain.CPU.Model] = "true"
+			}
 		}
 	}
 
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -62,7 +62,7 @@ const CAP_SYS_NICE = "SYS_NICE"
 const LibvirtStartupDelay = 10
 //This is a perfix for node feature discovery, used in a NodeSelector on the pod
 //to match a VirtualMachineInstance CPU model(Family) to nodes that support this model.
-const NFD_CPU_FAMILY_PREFIX = "feature.node.kubernetes.io/nfd-cpu-family-"
+const NFD_CPU_FAMILY_PREFIX = "feature.node.kubernetes.io/cpu-family-"
 
 const MULTUS_RESOURCE_NAME_ANNOTATION = "k8s.v1.cni.cncf.io/resourceName"
 
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -618,7 +618,8 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 
 	}
 	if featuregates.CPUNodeDiscoveryEnabled() {
-		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" {
+		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" &&
+			vmi.Spec.Domain.CPU.Model != "host-model" && vmi.Spec.Domain.CPU.Model != "host-passthrough" {
 			nodeSelector[NFD_CPU_FAMILY_PREFIX+vmi.Spec.Domain.CPU.Model] = "true"
 		}
 	}
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -35,7 +35,6 @@ import (
 	v1 "kubevirt.io/kubevirt/pkg/api/v1"
 	"kubevirt.io/kubevirt/pkg/config"
 	containerdisk "kubevirt.io/kubevirt/pkg/container-disk"
-	"kubevirt.io/kubevirt/pkg/feature-gates"
 	"kubevirt.io/kubevirt/pkg/hooks"
 	"kubevirt.io/kubevirt/pkg/kubecli"
 	"kubevirt.io/kubevirt/pkg/log"
@@ -43,6 +42,7 @@ import (
 	"kubevirt.io/kubevirt/pkg/util"
 	"kubevirt.io/kubevirt/pkg/util/net/dns"
 	"kubevirt.io/kubevirt/pkg/util/types"
+	"kubevirt.io/kubevirt/pkg/virt-config"
 )
 
 const configMapName = "kubevirt-config"
@@ -62,7 +62,7 @@ const LibvirtStartupDelay = 10
 
 //This is a perfix for node feature discovery, used in a NodeSelector on the pod
 //to match a VirtualMachineInstance CPU model(Family) to nodes that support this model.
-const NFD_CPU_FAMILY_PREFIX = "feature.node.kubernetes.io/cpu-model-"
+const NFD_CPU_MODEL_PREFIX = "feature.node.kubernetes.io/cpu-model-"
 
 const MULTUS_RESOURCE_NAME_ANNOTATION = "k8s.v1.cni.cncf.io/resourceName"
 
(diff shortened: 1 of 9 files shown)


yanirq:
[(8)[epic-node-compatibility-1.4] Ensure to schedule VMs on nodes which support the requested CPU model [initiative-vm-high-availability]](https://trello.com/c/9TOt73VW/76-8epic-node-compatibility-14-ensure-to-schedule-vms-on-nodes-which-support-the-requested-cpu-model-initiative-vm-high-availabilit)

yanirq:
When no CPU family is selected the CPU mode is set to `host-model` Which doesn't require a model(family) to be specified. The virt-contorller will not set nodeselector in that case. 
Same goes for host-model and host-passthrough
for more info see : [Add possibility to use CPU passthrough for VM's](https://github.com/kubevirt/kubevirt/pull/1345 )

yanirq:
@MarSik @ksimon1 @ahadas  @fromanirh   please review.

pkg/virt-controller/services/template.go
```diff
@@ -561,6 +567,12 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		nodeSelector[k] = v
 
 	}
+	if featuregates.CPUNodeDiscoveryEnabled() {
+		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" {
+			nodeSelector[NFD_CPU_FAMILY_PREFIX+vmi.Spec.Domain.CPU.Model] = "true"
```
fromanirh: do we want (or need) to sanitize the CPU.Model to make sure the label is well formed?
Example (not sure it applies): is the label name case sensitive? The CPU.Model? What if they are and the name doesn't match?

Furthermore: do we have a common database of cpu models?
I mean: is the cpu label essentially a glorified free string or do we want to validate the labels?

pkg/virt-controller/services/template_test.go
```diff
@@ -210,6 +211,30 @@ var _ = Describe("Template", func() {
...
+				vmi := v1.VirtualMachineInstance{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "testvmi",
+						Namespace: "default",
+						UID:       "1234",
+					},
+					Spec: v1.VirtualMachineInstanceSpec{
+						Domain: v1.DomainSpec{
+							CPU: &v1.CPU{
+								Model: "Conore",
```
fromanirh: typo: Conroe (probably irrilevant for the test as long as the string matches - but this kind stress the point I made above)

pkg/virt-controller/services/template.go
```diff
@@ -568,7 +568,8 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 
 	}
 	if featuregates.CPUNodeDiscoveryEnabled() {
-		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" {
+		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" &&
+			vmi.Spec.Domain.CPU.Model != "host-model" && vmi.Spec.Domain.CPU.Model != "host-passthrough" {
```
fromanirh: do we want to use constants for these?

fromanirh: minor/style: I'd put the host-model check in a different if, or even in a helper function, to make this one simpler to read and to stress that this is a separate check.

pkg/virt-controller/services/template.go
yanirq: > do we want (or need) to sanitize the CPU.Model to make sure the label is well formed?
Example (not sure it applies): is the label name case sensitive? The CPU.Model? What if they are and the name doesn't match?

The cpu model is provided from VM/VMI spec (CPU.Model) and will be used as a label in the node selector we add to virt-launcher:
1. If the model name doesn't comply with the [labels and selectors syntax](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) it will fail during the creation of virt-launcher with a proper error message. This syntax seems to be compatible with [libvirt cpu models](https://github.com/libvirt/libvirt/tree/master/src/cpu_map) (correct me if im wrong)
2. If the model name comply with labels syntax but somehow was written incorrectly (doesn't match  existing cpu families) it wont be scheduled : 
`Message:               0/1 nodes are available: 1 node(s) didn't match node selector.`
labels are case sensitive so if there is a mismatch -  no scheduling happens. 
btw, i see in that case an added value to the existing behaviour we had so far. non existing model names wont be mapped to libvirt's domain spec (converter stage after virt-launcher is up)

All of the above is under the assumption that the node-feature-discovery plugin will set the correct model names on the nodes.

> Furthermore: do we have a common database of cpu models?
I mean: is the cpu label essentially a glorified free string or do we want to validate the labels?

No, we do not have a common database of cpu models nor we check cpu models against a list (in webhook validations for example). 
Conceptually we can maybe gather a list of all existing cpu models from all nodes using the node-feature-discovery plugin and then use it for verification, but  it seems robust.

pkg/virt-controller/services/template.go
MarSik: I would actually encode cpu passthrough differently, a separate boolean maybe. Because we might want to request a machine with certain or better cpu type and then use host-passthrough. But I think it is out of the scope of this patch too.

pkg/virt-controller/services/template.go
MarSik: I think we should check (on VM level) for syntax compatibility and we might want to limit the values to whatever libvirt supports in the future (the value is also passed to libvirt xml). But, we do not have access to libvirt at the VM level and so there is no easy way to do this atm.

pkg/feature-gates/feature-gates.go
```diff
@@ -36,9 +36,10 @@ import (
 const featureGateEnvVar = "FEATURE_GATES"
 
 const (
-	dataVolumesGate   = "DataVolumes"
-	cpuManager        = "CPUManager"
-	liveMigrationGate = "LiveMigration"
+	dataVolumesGate      = "DataVolumes"
+	cpuManager           = "CPUManager"
+	liveMigrationGate    = "LiveMigration"
+	cpuNodeDiscoveryGate = "CPUNodeDiscovery"
```
fabiand: +1

pkg/virt-controller/services/template.go
fabiand: > Conceptually we can maybe gather a list of all existing cpu models from all nodes using the node-feature-discovery plugin and then use it for verification,

I wouldn't do this.


But what we need is a coordination between the CPU models we use in VMI, which are libvirt cpu models. And the models exposed by nfd.

I can imagine  that the names differ between what libvirt calls cpu families, and what nfd calls cpu families.

How do we solve this issue?

MarSik: Fabian, nfd does not expose any families atm. We will be writing the flags + cpuid -> families logic.

Also, for scheduling reasons, the node has to expose all supported families labels (the cpu one + all older that are compatible), because the POD only has one and should run on newer machines as well.

michalskrivanek: +1

MarSik: But if you were talking about validation only, then explicit list of supported families is of course better.

fabiand: > We will be writing the flags + cpuid -> families logic.

Hopefully we can leverage libvirt here?



> Also, for scheduling reasons, the node has to expose all supported families labels (the cpu one + all older that are compatible)

Absolutely



> But if you were talking about validation only, then explicit list of supported families is of course better.

Well, not _only_, but mainly yes.

yanirq:
@ksimon1 @MarSik  please take a look at the format of the label and see if this will be its final form.

yanirq:
@fabiand 
> Please check how far the currently supported CPU families are documented or not.

Under [CPU Model](https://kubevirt.io/user-guide/#/workloads/virtual-machines/virtualized-hardware-configuration?id=model) section we have link to libvirt's list of cpu models. The link to the model list was broken so I opened a PR: https://github.com/kubevirt/user-guide/pull/154 .

> In general - but I need more arguments for this - does the current apprach (we have some labels on the virt side, and magically nfd supports the same lables) pretty "unformatlized". I wonder if there are ideas to make this relationship more formal.

The current approach is to have a NFD plugin that uses libvirt to create the required labels (models/flags) on nodes: [NFD Libvirt](https://github.com/ksimon1/node-feature-discovery-libvirt) [WIP].
The labels we have here should match the output of the NFD libvirt plugin.

fabiand:
> Under [CPU Model](https://kubevirt.io/user-guide/#/workloads/virtual-machines/virtualized-hardware-configuration?id=model) section we have link to libvirt's list of cpu models. The link to the model list was broken so I opened a PR: [kubevirt/user-guide#154](https://github.com/kubevirt/user-guide/pull/154) .

Thank you for fixing it right away.

> The labels we have here should match the output of the NFD libvirt plugin.

Agreed!


Now - I think what we want is a real validation of the permitted CPU models.
We should enforce - on the cluster level - that only CPU model values that are currently supported.
I'd suggest to validate the cpu models which can be set on a VMI spec according to the CPU map provided by libvirt.
In summary: Now that we require CPU models on the cluster level, and we cooperate with other services (nfd) we should start enforcing the (welld efined set of) values on the cluster level.

fabiand:
Regardless of the NFD progress, we probably want to keep the featuregate
disabled by default.
It then is up to the deploying component to either enable or disable it, as
the deploying component knows if NFD is deployed or not.

yanirq:
> Regardless of the NFD progress, we probably want to keep the featuregate disabled by default. It then is up to the deploying component to either enable or disable it, as the deploying component knows if NFD is deployed or not.

Currently it is disabled by default. 
I reckon that once we are done with NFD, we can add the appropriate enablement for the feature gate in deployment (us/deployment team) and when we get to documentation we will add the appropriate instructions.

MarSik:
@fabiand Anything you are still missing? Can you please merge if not?

MarSik:
Btw, the CPU model validation is being discussed in https://github.com/kubevirt/kubevirt/issues/1760

tests/vmi_lifecycle_test.go
```diff
@@ -630,6 +631,61 @@ var _ = Describe("VMIlifecycle", func() {
...
+				ExpectWithOffset(1, err).ToNot(HaveOccurred())
+				originalData = cfgMap.Data[featuregates.FEATURE_GATES_KEY]
+				cfgMap.Data[featuregates.FEATURE_GATES_KEY] = featuregates.CPUNodeDiscoveryGate
+				_, err = virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Update(cfgMap)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred())
+
+			})
+
+			AfterEach(func() {
+				cfgMap, err = virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Get("kubevirt-config", options)
```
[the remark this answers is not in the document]
yanirq: We need to update the configMap (kubevirt-config) to its original state so it wont affect other tests:
`_, err = virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Update(cfgMap)`
the variables are just used to retrieve the latest configMap and update it to its original state.

yanirq:
ci test please

pkg/feature-gates/feature-gates.go
```diff
@@ -100,3 +102,7 @@ func LiveMigrationEnabled() bool {
 func SRIOVEnabled() bool {
 	return strings.Contains(os.Getenv(featureGateEnvVar), SRIOVGate)
 }
+
+func CPUNodeDiscoveryEnabled() bool {
```
rmohr: That function does not seem to be used. Can it be removed?

pkg/virt-controller/services/template.go
```diff
@@ -144,6 +149,15 @@ func isSRIOVVmi(vmi *v1.VirtualMachineInstance) bool {
 	return false
 }
 
+func IsCPUNodeDiscoveryEnabled(store cache.Store) bool {
+	var value string
+	value, _ = getConfigMapEntry(store, featuregates.FEATURE_GATES_KEY)
```
rmohr: Good question what to do when an error occurs here, but we should handle it explicitly in here.

tests/vmi_lifecycle_test.go
```diff
@@ -630,6 +631,61 @@ var _ = Describe("VMIlifecycle", func() {
...
+			var cfgMap *k8sv1.ConfigMap
+			var originalData string
+
+			BeforeEach(func() {
+				options = metav1.GetOptions{}
+				cfgMap, err = virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Get("kubevirt-config", options)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred())
+				originalData = cfgMap.Data[featuregates.FEATURE_GATES_KEY]
+				cfgMap.Data[featuregates.FEATURE_GATES_KEY] = featuregates.CPUNodeDiscoveryGate
+				_, err = virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Update(cfgMap)
```
rmohr: The tests are racy. We will have to improve the detection if virt-controller already received the config-map change. I don't want to make this too complicated for you right now, but one proper solution would be to expose the config-map feature gates on the already existing healthz endpoint of virt-controller. Then this endpoint can be queried and as soon as it reports the right flags, we know that all subsequent calls inside the virt-controller code will see the exact same values.

tests/vmi_lifecycle_test.go
```diff
@@ -630,6 +631,61 @@ var _ = Describe("VMIlifecycle", func() {
...
+
+			AfterEach(func() {
+				cfgMap, err = virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Get("kubevirt-config", options)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred())
+				cfgMap.Data[featuregates.FEATURE_GATES_KEY] = originalData
+				_, err = virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Update(cfgMap)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred())
+			})
+
+			It("the vmi with cpu.model matching a nfd label on a node should be scheduled", func() {
```
rmohr: We need a negative test too. The simplest thing which we do on other scheduling tests as well, is not adding the wanted label to a node but add a custom node selector on the vmi too. This way we would only allow node A but we can then check that it does not gescheduled because node A misses a required label.

pkg/virt-controller/services/template.go
```diff
@@ -59,6 +60,10 @@ const CAP_SYS_NICE = "SYS_NICE"
 // Libvirt needs roughly 10 seconds to start.
 const LibvirtStartupDelay = 10
 
+//This is a perfix for node feature discovery, used in a NodeSelector on the pod
+//to match a VirtualMachineInstance CPU model(Family) to nodes that support this model.
+const NFD_CPU_FAMILY_PREFIX = "feature.node.kubernetes.io/cpu-model-"
```
rmohr: Is it common to do `feature.node.kubernetes.io/cpu-model-mymodel: true` over `feature.node.kubernetes.io/cpu-model: mymodel`?

tests/vmi_lifecycle_test.go
rmohr: If you e.g. wait half a second, put a fixme here and add the negative test, then this should work for now good enough. Otherwise you will have to change a lot. @davidvossel thoughts ?

pkg/virt-controller/services/template.go
yanirq: This is how NFD will label the nodes.  The question is logical, If the labelling method will change then we will adapt.

rmohr: :+1:

fabiand:
Independent of NFD - What are the pros and cons of

```
<key>: <model>
vs
<key>-<model>: true
```

?

rmohr:
> Independent of NFD - What are the pros and cons of ``` <key>: <model> vs <key>-<model>: true ``` ?

Depends on what structure you need. If a node can have multiple models then you will want the `true` form. If you exclusively allow just one of them, then the `key: value` form is better.  The `true` form is probably the most general form and allows expressing more use-cases.

MarSik:
@fabiand 
> Independent of NFD - What are the pros and cons of ` <key>: <model>` vs `<key>-<model>: true ` ?

One vs. many possible values for the same key.

@rmohr 
NFD exports all supported CPU families and features, so we have to use the `true` form.

yanirq:
> > Independent of NFD - What are the pros and cons of `<key>: <model> vs <key>-<model>: true` ?
> 
> Depends on what structure you need. If a node can have multiple models then you will want the `true` form. If you exclusively allow just one of them, then the `key: value` form is better. The `true` form is probably the most general form and allows expressing more use-cases.

We do need to support multiple models and the next step is to support multiple features/flags

fabiand:
… and as a node can support many families `<key>-<value>: true` is the way to go.

yanirq:
ci test please

jekader:
CI system maintenance caused some tests for this PR to be left in "Pending" state. To re-run them please add a comment saying "ci test please".

yanirq:
ci please test

yanirq:
ci test please

yanirq:
ci test please

yanirq:
ci test please

rmohr:
@yanirq in case you are waiting for the openshift lanes, there is a bug in openshift-ansible (a rpm package got renamed). Our clusters can't start. Work is going on here to fix it #1804. The other lanes should be green, in case they were not.

tests/vmi_lifecycle_test.go
davidvossel: add a sleep.

MarSik:
ci please test

MarSik:
@yanirq I think the test is failing, because you forgot to regenerate the files after the API change. At least the Travis log seems to point in that direction at couple of places

https://travis-ci.org/kubevirt/kubevirt/builds/464081250#L664

yanirq:
@rmohr  @fabiand  can you give a final review here ?  (patch re-based and all necessary tests were added)

pkg/virt-controller/services/template.go
```diff
@@ -612,6 +626,14 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		nodeSelector[k] = v
 
 	}
+	if IsCPUNodeDiscoveryEnabled(t.configMapStore) {
+		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" {
+			if vmi.Spec.Domain.CPU.Model != v1.CPUModeHostModel && vmi.Spec.Domain.CPU.Model != v1.CPUModeHostPassthrough {
+				nodeSelector[NFD_CPU_FAMILY_PREFIX+vmi.Spec.Domain.CPU.Model] = "true"
```
fabiand: `NFD_CPU_FAMILY_PREFIX+vmi.Spec.Domain.CPU.Model` this should be a function somewhere. We want this important part ot be isolated somehwere. i.e. `CPUFamilyLabelFromCPUFamily`

tests/vmi_lifecycle_test.go
```diff
@@ -630,6 +631,99 @@ var _ = Describe("VMIlifecycle", func() {
...
+
+			var options metav1.GetOptions
+			var cfgMap *k8sv1.ConfigMap
+			var originalData string
+			NFD_CPU_LABEL := services.NFD_CPU_FAMILY_PREFIX + "Conroe"
+
+			BeforeEach(func() {
+				options = metav1.GetOptions{}
+				cfgMap, err = virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Get("kubevirt-config", options)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred())
```
rmohr: That would just be `Expect`.

tests/vmi_lifecycle_test.go
```diff
@@ -630,6 +631,99 @@ var _ = Describe("VMIlifecycle", func() {
...
+				originalData = cfgMap.Data[virtconfig.FeatureGatesKey]
+				cfgMap.Data[virtconfig.FeatureGatesKey] = virtconfig.CPUNodeDiscoveryGate
+				_, err = virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Update(cfgMap)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred())
```
rmohr: `Expect`

PR #1727, opened 2018-11-19
Tolerate smaller PVs

slintes:

The actual filesystem size of a PV can be slightly smaller than the
requested size of the PV, because of filesystem overhead which was not
taken into account of the storage provisioner.
With this PR a new configuration parameter is introduced, which defines
how big the difference between requested and available size can be,
before KubeVirt will fail with an error when creating disk images on
empty PVs.
The configuration parameter's name is "dev.pvc-tolerate-less-space-up-to-percent",
it's unit is percent, and it can be set in the "kubevirt-config" ConfigMap.
The default value is 10%.

I was asked to report via an event that we accepted a PV with smaller size because of the new toleration. But I have idea how to do this in virt-launcher, because it shouldn't do API server requests, and communication between virt-handler and launcher is one-way only afaik.
@rmohr do you have an idea how to create an event without adding a reverse communication channel from launcher to handler?  
@fabiand or is logging enough?

I also don't have an idea yet how to create a functional test for this. But I hope the unit tests are covering what needs to be covered already.

**Release note**:
```release-note
Accept smaller than expected PVs (due to filesystem overhead) when creating disk images.
```


slintes:
oh, we already have a launcher -> handler communication channel of course (notify-client/server), I can use that one for the event. Will add that tomorrow.

slintes:
note to self: close IntelliJ when running `make generate`, it reverts changes faster than you commit and push them :astonished:

pkg/host-disk/host-disk.go
```diff
@@ -85,19 +90,67 @@ func getPVCDiskImgPath(volumeName string) string {
...
+						// For toleration related tests we explicitly set a value, so let's keep this at 0 for the other tests
+						t = "0"
+					}
+					toleration, err := strconv.Atoi(t)
+					if err != nil {
+						return fmt.Errorf("Unable to create %s, invalid toleration value %v", hostDisk.Path, toleration)
+					}
+					toleratedSize := requestedSize * (100 - int64(toleration)) / 100
+					if uint64(toleratedSize) > availableSize {
+						return fmt.Errorf("Unable to create %s, not enough space, demanded size %d B is bigger than available space %d B, also after taking %d %% toleration into account",
```
stu-gott: Small detail, KubeVirt project generally wants to follow the Kubernetes standard of not capitalizing error messages. This is due to the fact that they're often chained together.

pkg/virt-controller/services/template.go
```diff
@@ -533,6 +554,9 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		},
 		Resources: resources,
 		Ports:     ports,
+		Env: []k8sv1.EnvVar{
```
stu-gott: So the idea here is that virt-controller looks up the space toleration, because it has access to cluster resources (e.g. ConfigMaps), but passes this value down the chain as an Env var because the volume hasn't been create yet thus can't be verified?

slintes:
Hi @rmohr @mfranczy @stu-gott, while functional tests are still missing, I just verified manually that this works:

failing case:
`2018-11-20T15:32:28Z - default - Warning - VirtualMachineInstance vmi-pvc SyncFailed - server error. command Launcher.Sync failed: Unable to create /var/run/kubevirt-private/vmi-disks/pvcvolume2/disk.img, not enough space, demanded size 15728640 B is bigger than available space 12756992 B, also after taking 10 %!t(MISSING)oleration into account` (with a small formatting issue)
tolerated case:
`2018-11-20T15:33:39Z - default - Normal - VirtualMachineInstance vmi-pvc Tolerated too small PV - PV size too small: expected 15728640 B, found 15387648 B. Using it anyway, it is within 10 % toleration`

Could you please start reviewing while I try to get my manual steps into functional tests?

@fabiand fyi

pkg/host-disk/host-disk.go
slintes: thanks, wasn't aware of this, will change it

pkg/virt-controller/services/template.go
slintes: yes, only virt-launcher has the volumes mounted and can verify the available space, when creating  disk images on empty PVs, but it can't look up the configuration.

pkg/virt-controller/services/template.go
```diff
@@ -44,6 +44,8 @@ import (
 const configMapName = "kubevirt-config"
 const UseEmulationKey = "debug.useEmulation"
 const ImagePullPolicyKey = "dev.imagePullPolicy"
+const LessPvcSpaceTolerationKey = "dev.pvc-tolerate-less-space-up-to-percent"
```
slintes: I think the `dev` prefix isn't correct here, will remove it

stu-gott:
Failing tests appear spurious/unrelated. @slintes You're still writing functional tests, correct?

slintes:
> @slintes You're still writing functional tests, correct?
yes...

pkg/virt-controller/services/template.go
```diff
@@ -533,6 +554,9 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		},
 		Resources: resources,
 		Ports:     ports,
+		Env: []k8sv1.EnvVar{
+			{Name: LessPvcSpaceTolerationEnvName, Value: strconv.Itoa(lessPvcSpaceToleration)},
```
rmohr: Can we make it an argument instead of an env var?

pkg/util/notifier/types.go
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2018 Red Hat, Inc.
+ *
+ */
+
+package notifier
+
+type K8sEventArgs struct {
```
rmohr: Would it make sense to use `Event` in `vendor/k8s.io/api/core/v1/types.go`?

pkg/host-disk/host-disk.go
slintes: done

pkg/virt-controller/services/template.go
slintes: done

pkg/virt-controller/services/template.go
slintes: very good idea, done

pkg/util/notifier/types.go
slintes: switched type, but postponed reimplementation using k8s-broadcaster as discussed

pkg/host-disk/host-disk.go
```diff
@@ -85,19 +92,58 @@ func getPVCDiskImgPath(volumeName string) string {
 	return path.Join(pvcBaseDir, volumeName, "disk.img")
 }
 
-func CreateHostDisks(vmi *v1.VirtualMachineInstance) error {
+type DiskImgCreator struct {
+	dirBytesAvailableFunc  func(path string) (uint64, error)
+	notifier               k8sNotifier
+	lessPvcSpaceToleration int
```
rmohr: nit: `lessPVCSpaceToleration`

pkg/virt-controller/services/template.go
```diff
@@ -44,6 +44,8 @@ import (
 const configMapName = "kubevirt-config"
 const UseEmulationKey = "debug.useEmulation"
 const ImagePullPolicyKey = "dev.imagePullPolicy"
+const LessPvcSpaceTolerationKey = "pvc-tolerate-less-space-up-to-percent"
+const LessPvcSpaceTolerationEnvName = "LESS_PVC_SPACE_TOLERATION"
```
rmohr: I think this is unused now?

pkg/host-disk/host-disk.go
```diff
@@ -25,14 +25,21 @@ import (
...
 
 	"kubevirt.io/kubevirt/pkg/api/v1"
 	"kubevirt.io/kubevirt/pkg/kubecli"
 	"kubevirt.io/kubevirt/pkg/util/types"
 )
 
-const pvcBaseDir = "/var/run/kubevirt-private/vmi-disks"
+const (
+	pvcBaseDir                    = "/var/run/kubevirt-private/vmi-disks"
+	lessPvcSpaceTolerationEnvName = "LESS_PVC_SPACE_TOLERATION"
```
rmohr: I think this is unused now?

pkg/host-disk/host-disk.go
```diff
@@ -25,14 +25,21 @@ import (
...
+	EventReasonToleratedSmallPV   = "ToleratedSmallPV"
+	EventTypeToleratedSmallPV     = "Normal"
```
rmohr: There exists a `k8sv1.EventTypeNormal` constant.

pkg/host-disk/host-disk.go
```diff
@@ -85,19 +92,58 @@ func getPVCDiskImgPath(volumeName string) string {
...
+	dirBytesAvailableFunc  func(path string) (uint64, error)
+	notifier               k8sNotifier
+	lessPvcSpaceToleration int
+}
+
+type k8sNotifier interface {
+	SendK8sEvent(vmi *v1.VirtualMachineInstance, severity string, reason string, message string) error
+}
+
+func NewHostDiskCreator(notifier k8sNotifier, lessPvcSpaceToleration int) DiskImgCreator {
```
rmohr: nit: `lessPVCSpaceToleration`.

pkg/host-disk/host-disk.go
```diff
@@ -85,19 +92,58 @@ func getPVCDiskImgPath(volumeName string) string {
...
+						return fmt.Errorf("unable to create %s, not enough space, demanded size %d B is bigger than available space %d B, also after taking %v %% toleration into account",
+							hostDisk.Path, uint64(requestedSize), availableSize, hdc.lessPvcSpaceToleration)
+					}
+					diskSize = int64(availableSize)
+
+					msg := fmt.Sprintf("PV size too small: expected %v B, found %v B. Using it anyway, it is within %v %% toleration", requestedSize, availableSize, hdc.lessPvcSpaceToleration)
+					err = hdc.notifier.SendK8sEvent(vmi, EventTypeToleratedSmallPV, EventReasonToleratedSmallPV, msg)
+					if err != nil {
+						log.Log.Reason(err).Warningf("Couldn't send k8s event for tolerated PV size: %v", err)
+						log.Log.Info(msg)
```
rmohr: I think we want to log the message always.

pkg/virt-launcher/notify-client/client.go
```diff
@@ -219,3 +209,28 @@ func StartNotifier(virtShareDir string, domainConn cli.Connection, deleteNotific
 	log.Log.Infof("Registered libvirt event notify callback")
 	return nil
 }
+
+func (c *NotifyClient) SendK8sEvent(vmi *v1.VirtualMachineInstance, severity string, reason string, message string) error {
+
+	reply := &notifyserver.Reply{}
+
+	event := k8sv1.Event{
+		InvolvedObject: k8sv1.ObjectReference{
```
rmohr: Can you fill in the whole reference? At least also the `UID`. Otherwise we can't distinguish to which object instance the message belongs.

tests/storage_test.go
```diff
@@ -422,6 +428,80 @@ var _ = Describe("Storage", func() {
...
 		})
 
+		Context("With smaller than requested PVCs", func() {
+
+			var pvc string
+
+			BeforeEach(func() {
+
+				// prepare storage with fix small size (20Mi) where fs already has > 10 % some overhead
+				out, err := exec.Command("/bin/sh", "-c", `
```
rmohr: How about creating such a file-system in a privileged pod, then use mount propagation to mount that file-system via a hostDisk into the vmi pod?

rmohr: Here the docu: https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation and here where we use it in a POC: #1565

slintes: I already tried to do this in disk-images-provider, similar to creating the cirros block device there, but it didn't work... so at least for this PR I'd prefer to keep it his way

pkg/host-disk/host-disk.go
slintes: deleted

pkg/host-disk/host-disk.go
slintes: thx, using it now

pkg/host-disk/host-disk.go
slintes: fixed everywhere

pkg/host-disk/host-disk.go
slintes: done

pkg/virt-controller/services/template.go
slintes: deleted

pkg/virt-launcher/notify-client/client.go
slintes: done

slintes:
older CI run looks good I'd say: http://jenkins.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/2576/#showFailuresLink
Let's wait for the current one...

pkg/host-disk/host-disk.go
```diff
@@ -85,19 +91,58 @@ func getPVCDiskImgPath(volumeName string) string {
...
+					// This can be configured with the "pvc-tolerate-less-space-up-to-percent" parameter in the kubevirt-config ConfigMap.
+					// It is provided as argument to virt-launcher.
+					toleratedSize := requestedSize * (100 - int64(hdc.lessPVCSpaceToleration)) / 100
+					if uint64(toleratedSize) > availableSize {
+						return fmt.Errorf("unable to create %s, not enough space, demanded size %d B is bigger than available space %d B, also after taking %v %% toleration into account",
+							hostDisk.Path, uint64(requestedSize), availableSize, hdc.lessPVCSpaceToleration)
+					}
+					diskSize = int64(availableSize)
+
+					msg := fmt.Sprintf("PV size too small: expected %v B, found %v B. Using it anyway, it is within %v %% toleration", requestedSize, availableSize, hdc.lessPVCSpaceToleration)
```
rmohr: Something is not  right here. I see

```
kubevirt-test-default   5s        5s        3         testvmikqcx6zjv8k9kfdhq4p96k4whlt7mbbjlp6g9lpz9xb6mjsdt.15696bb0f4c21ea2   VirtualMachineInstance             Warning   SyncFailed   virt-handler, node01   server error. command Launcher.Sync failed: unable to create /tmp/kubevirt/krzs9chpww/mount/disk.img, not enough space, demanded size 39705950400 B is bigger than available space 33881468928 B, also after taking 10 %!t(MISSING)oleration into account
```
in the event log.

slintes: the same formatting is used for the ToleratedSmallPV event, there it works. I guess in case of an error, the formatted error message ("...taking 10 % toleration...") is used again for formatting somewhere else, and there the single % fails... will try find that place

slintes:
/hold

I have fix ready for the formatting issue, but pushed that to the backport first, in order to let CI process that PR first

PR #1728, opened 2018-11-20
Added code to passthrough custom dhcp options to vm.

gageorsburn:
Adds a struct to the api for configurable dhcp options such as tftp server address and boot file name.

Fixes #

**Release note**:

```release-note
Allow tftpServerName and booFileName to be passed to the internal interface dhcp server.
```


ovirt-infra:

gageorsburn:
@rmohr second attempt at what i spoke to you about like a month ago

pkg/api/v1/schema.go
```diff
@@ -794,6 +794,16 @@ type Interface struct {
 	// If specified, the virtual network interface will be placed on the guests pci address with the specifed PCI address. For example: 0000:81:01.10
 	// +optional
 	PciAddress string `json:"pciAddress,omitempty"`
+	// If specified the network interface will pass additional DHCP options to the VMI
+	// +optional
+	DhcpOptions `json:"dhcpOptions,omitempty"`
```
rmohr: The direction looks good. I would expect this option in `PodNetwork` in `spec.networks`, since we can only guarantee it there. Only there we know for sure that we run a DHCP server. What do you think?

gageorsburn: From what I can see, we run a DHCP server for every interface.

rmohr: Not always (if the interface in the pod has no IP, we just pass through without starting our own dhcp server), but good point. In most cases we do. Especially if they are normal k8s-network extensions.

cluster/examples/vmi-sriov.yaml
```diff
@@ -19,8 +19,10 @@ spec:
         volumeName: cloudinitvolume
       interfaces:
       - bridge: {}
+        dhcpOptions: {}
         name: default
-      - name: sriov-net
+      - dhcpOptions: {}
```
rmohr: Ah that looks suspicios. The struct is not formed correctly. See the comment below.

pkg/api/v1/schema.go
```diff
@@ -796,14 +796,16 @@ type Interface struct {
 	PciAddress string `json:"pciAddress,omitempty"`
 	// If specified the network interface will pass additional DHCP options to the VMI
 	// +optional
-	DhcpOptions `json:"dhcpOptions,omitempty"`
+	DHCPOptions `json:"dhcpOptions,omitempty"`
```
rmohr: That line should look like this:

```go
DHCPOptions *DHCPOptions `json:"dhcpOptions,omitempty"`
```

We don't want to embed the DHCPOptions struct, we want it a s a property.

gageorsburn:
@rmohr I haven't thought much about the automated integration testing of this. I mentioned above I'm using it with virtual routers, which could get heavy and not sure about licensing matters of using them in an automated test so probably best not to go that route.

rmohr:
@gageorsburn maybe we can do a test where we set the options and then manually invoke a dhcp agent from inside the VMI. Let's check if we find a handy tool for that ...

rmohr:
> @gageorsburn maybe we can do a test where we set the options and then invoke a dhcp agent from inside the VMI. Let's check if we find a handy tool for that ...

Ok, turns out that this is not that hard at all. We can simply first release the actual lease inside the VMI:

```bash
#release the old lease
sudo dhclient -1 -r -d -sf eth0
```

And then request a new lease. We point the new lease to `-sf /usr/bin/env`:

```bash
#do a new request and print all new options to stdout
sudo dhclient -1 -d -sf /usr/bin/env --request-options subnet-mask,broadcast-address,time-offset,routers,domain-search,domain-name,domain-name-servers,host-name,nis-domain,nis-servers,ntp-servers,interface-mtu eth0
```

`dhclient` provides to the hook script everything it discovered via environment variables, therefore `/usr/bin/env` prints something like this:

```bash
[...]
requested_interface_mtu=1
requested_ntp_servers=1
requested_nis_servers=1
requested_nis_domain=1
requested_host_name=1
requested_domain_name_servers=1
requested_domain_name=1
requested_domain_search=1
requested_routers=1
requested_time_offset=1
requested_broadcast_address=1
requested_subnet_mask=1
new_expiry=1629130331
new_domain_name=default.svc.cluster.local
new_host_name=vmi-fedora3
new_domain_search=default.svc.cluster.local. svc.cluster.local. cluster.local.
new_interface_mtu=1450
new_dhcp_server_identifier=169.254.75.10
new_domain_name_servers=10.96.0.10
new_dhcp_message_type=5
new_dhcp_lease_time=86313600
new_routers=10.244.0.1
new_subnet_mask=255.255.255.0
new_broadcast_address=10.244.0.255
new_network_number=10.244.0.0
new_next_server=0.0.0.0
new_ip_address=10.244.0.35
[...]
```

@gageorsburn what do you think?

gageorsburn:
@rmohr Having some troubles with test timing out. At first I thought it was the -d flag in dhclient because of attempting the test manually but after taking it out it still hangs. Are there any ways for me to see what's going on?

tests/vmi_networking_test.go
```diff
@@ -568,6 +570,60 @@ var _ = Describe("Networking", func() {
...
+			err = tests.CheckForTextExpecter(dhcpVMI, []expect.Batcher{
+				&expect.BSnd{S: "\n"},
+				&expect.BExp{R: "#"},
+				&expect.BSnd{S: "sudo dhclient -1 -sf /usr/bin/env --request-options subnet-mask,broadcast-address,time-offset,routers,domain-search,domain-name,domain-name-servers,host-name,nis-domain,nis-servers,ntp-servers,interface-mtu,tftp-server-name,bootfile-name eth0\n"},
+				&expect.BExp{R: "new_tftp_server_name=tftp.kubevirt.io"},
+				// &expect.BSnd{S: "echo $?\n"},
+				// &expect.BExp{R: "0"},
+			}, 15)
+
+			// Was planning on executing /usr/bin/env and checking for each option passed in
```
rmohr: Yes, dhclient only passes the environmentvariables to the binary in `-sf`, so it is not persisted in your session. You can try to do two `Bexp` in a row:

```go
&expect.BExp{R: "new_tftp_server_name=tftp.kubevirt.io"},
&expect.BExp{R: "new_bootfile_name=config"}
```

but I don't know if the order is always the same.

Another option is to write the env output somewhere (e.g. `/dhcp-env`):

```go
&expect.BSnd{S: "sudo dhclient -1 -sf /usr/bin/env --request-options subnet-mask,broadcast-address,time-offset,routers,domain-search,domain-name,domain-name-servers,host-name,nis-domain,nis-servers,ntp-servers,interface-mtu,tftp-server-name,bootfile-name eth0 > /dhcp-env\n"},
```

and then cat and check:

```go
&expect.BSnd{S: "cat /dhcp-env\n"},
&expect.BExp{R: "new_tftp_server_name=tftp.kubevirt.io"},
&expect.BSnd{S: "cat /dhcp-env\n"},
&expect.BExp{R: "new_bootfile_name=config"}
```

rmohr:
/cc @SchSeba @yuvalif @booxter 

Can one of you have a look to on the PR too?

rmohr:
> So I got that method sort of working. `/usr/bin/env` does not persist between expecter calls. What do you think?

Looks pretty good already. Added an inline comment on how to make that work.

pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
```diff
@@ -130,6 +133,18 @@ func prepareDHCPOptions(
 	if len(domainName) > 0 {
 		dhcpOptions[dhcp.OptionDomainName] = []byte(domainName)
 	}
+
+	if customDHCPOptions != nil {
```
rmohr: Can you add one or two unit tests to dhcp_test.go?

tests/vmi_networking_test.go
```diff
@@ -568,6 +570,60 @@ var _ = Describe("Networking", func() {
 			checkLearningState(learningDisabledVMI, "0")
 		})
 	})
+
+	Context("VirtualMachineInstance with dhcp options", func() {
+		BeforeEach(func() {
+			tests.BeforeTestCleanup()
+		})
+
+		FIt("should offer extra dhcp options to pod iface", func() {
```
rmohr: `It`

tests/vmi_networking_test.go
gageorsburn: Have to tee it but that works like a charm actually :smile: !

rmohr:
cu test please

rmohr:
ci test please

rmohr:
ci test please

gageorsburn:
@rmohr @booxter @yuvalif @SchSeba anything I need to do for this pr?

rmohr:
@gageorsburn no, looks good from my perspective! :+1:

PR #1729, opened 2018-11-20
Multus webhook support

booxter:
**What this PR does / why we need it**: this adopts kubevirt to support new multus mutating webhook

TODO: add functional test for sriov network crd from different namespace

Fixes #

**Special notes for your reviewer**: this is work in progress, building on top of another PR for sriov support. Posted to start discussion and better visualisation of what's required to make it work in kubevirt context.

**Release note**:

```release-note
Adopt multus mutating webhook for resource requests, which removes the need to explicitly request device plugin resources in VMI spec
```

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -555,8 +555,8 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 
 	volumes = append(volumes, k8sv1.Volume{Name: "infra-ready-mount", VolumeSource: k8sv1.VolumeSource{EmptyDir: &k8sv1.EmptyDirVolumeSource{}}})
 
-	// VirtualMachineInstance target container
-	container := k8sv1.Container{
+	// VirtualMachineInstance target container.
+	compute := k8sv1.Container{
 		Name:            "compute",
 		Image:           t.launcherImage,
 		ImagePullPolicy: imagePullPolicy,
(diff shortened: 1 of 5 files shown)


pkg/virt-controller/services/template.go
```diff
@@ -798,12 +798,22 @@ func getResourceNameForNetwork(network *networkv1.NetworkAttachmentDefinition) s
 	return "" // meaning the network is not served by resources
 }
 
+func getNamespaceAndNetworkName(vmi *v1.VirtualMachineInstance, fullNetworkName string) (namespace string, networkName string) {
```
phoracek: This is still not enough. The full format of the network request is `namespace/network@ifname`. https://github.com/intel/multus-cni/blob/master/k8sclient/k8sclient.go#L151

Could you also handle the last ifname part? We don't need to use it as part of this PR, just not to assume network@ifname is network name.

phoracek: Actually, I'm not sure whether it could ever happen in VMI API. Perhaps we should block networkName that contains request for specific interface name.

booxter: Picking a name of an interface in VMI scope seems like a weird request since the netlink implementation is internal to launcher container. While I agree it may make sense to forbid the format completely on api level it seems to me this is not in scope of the PR.

booxter:
Note: Commit https://github.com/kubevirt/kubevirt/pull/1729/commits/5907d2564cacebb4955e6e06a442579775b1c736 should be dropped before final merge. It's a temporary workaround for a bug in multus webhook.

booxter:
Here is an alternative interim solution for requests / limits filling-in that doesn't depend on multus webhook: https://github.com/kubevirt/kubevirt/pull/1816

booxter:
/remove-lifecycle stale

We'll need this PR once multus gets their hook in.

PR #1746, opened 2018-11-22
Add masquerade binding mechanism.

SchSeba:
Changes:
* added new interface type (proxy)
* added coreos/iptables package into glide
* added unit tests
* added functinal tests
* updated some packages

Add a new binding mechanism call proxy.

This implementation insert iptable rules on the virt-launcher pod.
MASQUERADE all the outgoing traffic from the vm with the ip of the pod.
NAT (SNAT and DNAT) all the ports configured on the vm yaml.
NAT all the localhost traffic from the configured ports to the vm.

**Release note**:

```release-note
proxy binding mechanism
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/1746)


SchSeba:
/cc @rmohr 

Can you please take a look.

SchSeba:
/cc @phoracek @booxter 

Can you please take a look on this PR ?

api/openapi-spec/swagger.json
```diff
@@ -4538,6 +4538,9 @@
        "$ref": "#/definitions/v1.Port"
       }
      },
+     "proxy": {
```
fabiand: Please consider to call is masquerade or nat.

To me at least, proxy is rather a process which does the relaying.
And I'd also like to keep the process bound proxying free, because I hope that we'll have some process bound proxying in future, as it's the only K8s compatible way to do things.

SchSeba:
Rebase

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -97,6 +106,53 @@ func (h *NetworkUtilsHandler) ParseAddr(s string) (*netlink.Addr, error) {
...
+}
+func (h *NetworkUtilsHandler) IptablesAppendRule(table, chain string, rulespec ...string) error {
+	iptablesObject, err := iptables.New()
+	if err != nil {
+		return err
+	}
+
+	return iptablesObject.Append(table, chain, rulespec...)
+}
+func (h *NetworkUtilsHandler) GetAvailableAddrsFromCIDR(s string) ([]string, error) {
```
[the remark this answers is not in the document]
booxter: Perhaps it makes sense in context of what we already have here.

api/openapi-spec/swagger.json
dankenigsberg: Why would the App Operator care if the proxy is implemented by a process, or as suggested here, by the kernel? I would assume that if we find an efficient way to do what you hope, we can just flip the implementation.

Why is kernel-based less k8s compatible?

(I'm not going for a holy war about the name `proxy`, just willing to understand your position)

fabiand: Please reach out to @SchSeba to understand the differences between a kernel and process based proxy approach for kubernetes.

A user would usually not care, but if sees a spec, then I would like to leave the best hints to give a user transparency. That#s why I favor to have a more kernel-ish name for this implementation and `proxy` for a more process-ish implementation.

SchSeba:
There is an error for the masquerade test 
```
{"component":"virt-launcher","kind":"","level":"error","msg":"Defining the VirtualMachineInstance failed.","name":"testvmiswlkx788gh4wjpwz78wlj55nmq8n8zft9rz66snkwq4wgnmf","namespace":"kubevirt-test-default","pos":"libvirt_helper.go:101","reason":"virError(Code=8, Domain=44, Message='invalid argument: could not find capabilities for arch=x86_64 domaintype=kvm ')","timestamp":"2018-12-02T16:40:29.751213Z","uid":"f0262580-f650-11e8-b453-525500d15501"}
``` 

for reference 
https://jenkins.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/2726//artifact/check-patch.k8s-multus-1.11.1-release.el7.x86_64/mock_logs/script/stdout_stderr.log

I will try to run the tests again

SchSeba:
ci test please

SchSeba:
@fabiand Change the interface as requested can you please take a look

SchSeba:
Multiple error on the tests
```
â€¢ Failure [109.562 seconds]
Networking
/root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:48
  VirtualMachineInstance with masquerade binding mechanism
  /root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:616
    should allow regular network connection [It]
    /root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:620

    Expected error:
        <expect.TimeoutError>: 30000000000
        expect: timer expired after 30 seconds
    not to have occurred

    /root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:671
STEP: creating two virtual machines
STEP: waiting for the virtual machines to become ready
{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"Created virtual machine pod virt-launcher-testvmi7tqnq7hzsqmpv45n4jdll8bfkvt5fcvxsq4679bsd9","name":"testvmi7tqnq7hzsqmpv45n4jdll8bfkvt5fcvxsq467g4gdsf6vn48","namespace":"kubevirt-test-default","pos":"utils.go:270","timestamp":"2018-12-02T21:06:03.306216Z","uid":"13adb077-f676-11e8-8b73-525500d15501"}
{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"VirtualMachineInstance defined.","name":"testvmi7tqnq7hzsqmpv45n4jdll8bfkvt5fcvxsq467g4gdsf6vn48","namespace":"kubevirt-test-default","pos":"utils.go:270","timestamp":"2018-12-02T21:06:18.189213Z","uid":"13adb077-f676-11e8-8b73-525500d15501"}
{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"VirtualMachineInstance started.","name":"testvmi7tqnq7hzsqmpv45n4jdll8bfkvt5fcvxsq467g4gdsf6vn48","namespace":"kubevirt-test-default","pos":"utils.go:270","timestamp":"2018-12-02T21:06:18.207509Z","uid":"13adb077-f676-11e8-8b73-525500d15501"}
{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"Created virtual machine pod virt-launcher-testvmi2mlr9ggcv8mrsdz2bt4n2bm5pv4nclwqp9pwcfw9ps","name":"testvmi2mlr9ggcv8mrsdz2bt4n2bm5pv4nclwqp9pwc6cpk8pztr6z","namespace":"kubevirt-test-default","pos":"utils.go:270","timestamp":"2018-12-02T21:07:06.265913Z","uid":"13b2a684-f676-11e8-8b73-525500d15501"}
{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"VirtualMachineInstance defined.","name":"testvmi2mlr9ggcv8mrsdz2bt4n2bm5pv4nclwqp9pwc6cpk8pztr6z","namespace":"kubevirt-test-default","pos":"utils.go:270","timestamp":"2018-12-02T21:07:06.266135Z","uid":"13b2a684-f676-11e8-8b73-525500d15501"}
{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"VirtualMachineInstance started.","name":"testvmi2mlr9ggcv8mrsdz2bt4n2bm5pv4nclwqp9pwc6cpk8pztr6z","namespace":"kubevirt-test-default","pos":"utils.go:270","timestamp":"2018-12-02T21:07:06.266509Z","uid":"13b2a684-f676-11e8-8b73-525500d15501"}
STEP: expose service for the server virtual machine
STEP: check ping to google
STEP: start a tcp server
STEP: Connect from the client
{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"[{1 \r\n\r\n$  [$ ]} {3 echo test | nc -w 1 server-tcp 8080 1\u003e /dev/null\r\nnc: bad address 'server-tcp'\r\n$  [$ ]} {5 echo $?\r\n1\r\n$  []}]","name":"testvmi2mlr9ggcv8mrsdz2bt4n2bm5pv4nclwqp9pwc6cpk8pztr6z","namespace":"kubevirt-test-default","pos":"utils.go:1712","timestamp":"2018-12-02T21:07:48.452142Z","uid":""}
â€¢ [SLOW TEST:40.909 seconds]
```

Run functional tests with focus 

/hold

booxter:
@SchSeba you need to rebase it. Also, is hold label valid still?

tests/vmi_networking_test.go
```diff
@@ -594,6 +594,70 @@ var _ = Describe("Networking", func() {
 		})
 	})
 
+	FContext("VirtualMachineInstance with masquerade binding mechanism", func() {
```
booxter: Wait, you disabled all tests but this one.

SchSeba: Yes I know just want to be sure this test pass on the CI before I run it with all the other tests.

SchSeba:
@booxter I will wait for the CI to finish and then I will remove the hold

SchSeba:
The logs look good now.
on the network provider there is one error related to a know issue we have of the mac starting with "fe"

Now I think this PR is ready to be merge.
/hold cancel

fabiand:
@mfranczy do you mind doing a sanity check?

If @booxter can then give it's blessing the we should be good I guess.

As always: Please keep an eye on the tests.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -317,6 +335,242 @@ func (b *BridgePodInterface) createBridge() error {
...
+func (p *MasqueradePodInterface) discoverPodNetworkInterface() error {
+	link, err := Handler.LinkByName(p.podInterfaceName)
+	if err != nil {
+		log.Log.Reason(err).Errorf("failed to get a link for interface: %s", p.podInterfaceName)
+		return err
+	}
+	p.podNicLink = link
+
+	// Get interface MTU
+	p.vif.Mtu = uint16(p.podNicLink.Attrs().MTU)
```
mfranczy: For me this conversion looks dangerous, the range of `uint16` is 0 to 65535 and MTU is `int` type. I would add additional checks.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -317,6 +335,242 @@ func (b *BridgePodInterface) createBridge() error {
...
+
+	return nil
+}
+
+func (p *MasqueradePodInterface) startDHCPServer() {
+	// Start DHCP Server
+	Handler.StartDHCP(p.vif, p.gatewayAddr, p.bridgeInterfaceName, p.iface.DHCPOptions)
+}
+
+func (p *MasqueradePodInterface) decorateConfig() error {
```
mfranczy: What's the point fo returning an error here when you always return `nil`

tests/vmi_networking_test.go
```diff
@@ -594,6 +594,70 @@ var _ = Describe("Networking", func() {
...
 	})
 
+	Context("VirtualMachineInstance with masquerade binding mechanism", func() {
+		var serverVMI *v1.VirtualMachineInstance
+		var clientVMI *v1.VirtualMachineInstance
+
+		It("should allow regular network connection", func() {
+			By("creating two virtual machines")
+			ports := []v1.Port{{Name: "http", Port: 8080}}
+			serverVMI = tests.NewRandomVMIWithMasqueradeInterfaceEphemeralDiskAndUserdata(tests.ContainerDiskFor(tests.ContainerDiskCirros), "#!/bin/bash\necho 'hello'\n", ports)
```
mfranczy: that's quite a long func name `NewRandomVMIWithMasqueradeInterfaceEphemeralDiskAndUserdata` :), may I suggest to rename it and move it to local `Context` scope? You're using this function only here so I have doubts if we need it in `utils`. My opinion is, if someone will need to use it somewhere else than he/she should moved it then.

pkg/virt-launcher/virtwrap/network/podinterface.go
mfranczy: Okay, but what about `jumbograms`? Do we support it or not?
According to RFC [doc](https://tools.ietf.org/html/rfc2675)

> A "jumbogram" is an IPv6 packet containing a payload longer than
   65,535 octets.  This document describes the IPv6 Jumbo Payload
   option, which provides the means of specifying such large payload
   lengths.  It also describes the changes needed to TCP and UDP to make
   use of jumbograms.

> Jumbograms are relevant only to IPv6 nodes that may be attached to
   **links with a link MTU greater than 65,575 octets**, and need not be
   implemented or understood by IPv6 nodes that do not support
   attachment to links with such large MTUs.

So if some crazy guy will need to use it then this conversion will fail.
Or maybe there is some check somewhere else which prevents using such a high MTU.

mfranczy: Ah forget about `jumbogram` comment, `virtio` does not support it. But I still think there should be some additional check or `VIF.Mtu` field should be the same type. It's just my opinion after all, it's up to you.

SchSeba: I think it will be hard to chnage the variable to int because of the dhcp option
```go
	mtuArray := make([]byte, 2)
	binary.BigEndian.PutUint16(mtuArray, mtu)

	dhcpOptions := dhcp.Options{
		dhcp.OptionSubnetMask:       []byte(clientMask),
		dhcp.OptionRouter:           []byte(routerIP),
		dhcp.OptionDomainNameServer: bytes.Join(dnsIPs, nil),
		dhcp.OptionInterfaceMTU:     mtuArray,
	}
```

I can add if statement to check this < 0 and > 65535

booxter: I think it makes sense to guard the code and not rely on virtio limitations. A if statement is the way to go.

booxter: That being said... Isn't it the same bug that is perhaps present in other places of code? We may want to fix them too. I would be fine if e.g. this part is moved into a separate PR, as long as we are assured this happens.

booxter: I see this bug in at least one other place, for BridgePodInterface. (Not a surprise considering that a lot of this code is a copy-pasted and then adjusted code for bridge implementation.) @SchSeba please tell us what you prefer: fixing the issue here or sending a follow up PR. I am fine with both approaches. (I am not fine with leaving it not handled for long though.)

SchSeba: @booxter I just add this to the code for the bridge and the masquerade bindings. 
Also added a unit test for that please see my last commit

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -193,6 +193,10 @@ func (b *BridgePodInterface) discoverPodNetworkInterface() error {
 		b.vif.MAC = mac
 	}
 
+	if b.podNicLink.Attrs().MTU < 0 || b.podNicLink.Attrs().MTU > 65535 {
+		return fmt.Errorf("MTU value out of range ")
```
booxter: this trailing space... :)

booxter:
@SchSeba please add docs

pkg/virt-launcher/virtwrap/network/podinterface.go
booxter: Also, you could include the failed value too.

PR #1748, opened 2018-11-23
Stabilize test where virt-handler should recover after crash

rmohr:

Stabilize 

```
VMIlifecycle Creating a VirtualMachineInstance when virt-handler crashes should recover and continue management
```
test.

It was necessary to wait longer for the events to come because there are many uncertainties when the pod gets killed. Especially: How fast it gets killed and restarted and what it will then exactly see.

Fixes #

**Release note**:

```release-note
NONE
```


rmohr:
The test passes now in all lanes.

PR #1772, opened 2018-11-28
suggest a design principle: VM should behave like a native workload

dankenigsberg:
Not only do we keep native workloads intact, we aim to have VM-based workloads behave just like native ones.

I understand that this design principle is well know and agreed upon. However it is quite implicit, and its importance was not clear to me. I would like it to be better understood and followed.

@fabiand @rmohr @simon3z 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -92,4 +92,4 @@ to the VMIs managed through KubeVirt.
 Furthormore: if you can run native workloads, and you have KubeVirt installed,
 you should be able to run VM-based workloads, too.
 For example, Application Operators should not require additional permissions
-just because their application runs within a VM.
+to use cluster features for VMs, compared to using that feature with a plain Pod.
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -88,3 +88,8 @@ communicate with Kubernetes and where the additional types are stored:
 KubeVirt is deployed on top of a Kubernetes cluster.
 This means that you can continue to run your Kubernetes-native workloads next
 to the VMIs managed through KubeVirt.
+
+Furthormore: if you can run native workloads, and you have KubeVirt installed,
+you should be able to run VM-based workloads, too.
+For example, Application Operators should not require additional permissions
+just because their application runs within a VM.


dankenigsberg:
My point is that the architecture must ALLOW the cluster owner to say "let anyone run VMs" without compromising their security, without a need to grant further rights to users. The cluster owner can still say that users starting with f* are not allowed to use VMs, but that's besides my point.

For example, if an Operator wants to start a VM with sriov access, he must not need any more rights than a user willing to run sriov in a Pod. This is, currently, not the case, and we need to fix it *because* it contradict a basic design principle.

docs/architecture.md
```diff
@@ -88,3 +88,8 @@ communicate with Kubernetes and where the additional types are stored:
 KubeVirt is deployed on top of a Kubernetes cluster.
 This means that you can continue to run your Kubernetes-native workloads next
 to the VMIs managed through KubeVirt.
+
+Furthormore: if you can run native workloads, and you have KubeVirt installed,
+you should be able to run VM-based workloads, too.
+For example, Application Operators should not require additional permissions
```
rmohr: Maybe we can write it this way:

```
For example, Application Operators should not require additional permissions to use cluster features for VMs, compared to using that feature with a plain Pod.
```
That would mention that this is meant in comparison with the permissions needed on a Pod.

dankenigsberg:
@booxter please take a look. For you this was obvious - but not for me. That's the reason we must make sure sriov in VMs does not require any more user permission relative to sriov in Pods.

fabiand:
Thanks for clarification and adjustment.

fabiand:
Awaiting @booxter feedback, then we can merge it.

booxter:
@dankenigsberg the wording is fine but I think we have a disconnect after all about sriov specifics. As it stands now, a user does NOT need additional privileges allocated to them to run a SR-IOV attached VMI. The issue with privileges with SR-IOV is NOT that users need additional RBAC rules but that launcher pods are running privileged (while there are ways to avoid it now with fresh SR-IOV DP), so anyone who can start a VMI and attach it to a SR-IOV network can effectively get a privileged container available to them. So, again, the issue is NOT that a user can't do something but that the user CAN do more with the SR-IOV enabled VMI than with a regular VMI (perhaps more than they should be able to).

dankenigsberg:
Thanks for approving my text.

I suspect that I don't understand something basic, so please break it down to me.

As an unprivileged user:
* may I receive a Pod with a VF in it? (I hope I do!)
* may I receive a VMI with a VF in it? (I hope I don't, because currently it means that I have gained access to a privileged Pod and I can pwn the host)

fabiand:
IOW:

If a user uses a pod with SRIOV. the container is unprivileged.
If a user uses a VMI with SRIOV, the launcher container is privileged.
If a user uses a VMI without SRIOV, the launcher container is _not_ privileged.

This is probably because we need extra privileges inside the container to permit libvirt to consume the device.

booxter:
What @fabiand said. The only thing to add is that we allocate privileges to the launcher pod so that libvirt can configure the device for vfio consumption. With the latest builds of SR-IOV device plugin*, we should be able to avoid privilege escalation by 1) pre-registering all devices with vfio as pre-deployment step; and 2) removing the privileged=true from launcher pods; and 3) making sure that libvirt doesn't do anything silly like assuming that it should do the registration itself and crashing, or unregistering vfio devices on exit. This is future work.

* the plugin now can register vfio device inodes against container device plugin

booxter:
I've merged the PR since it makes general sense even if it came from a wrong place (specifically, from misunderstanding of the issues we have with SR-IOV privileges - it's not that users may not have enough but that they may have too much). If we would follow up on this particular SR-IOV case here we would probably also state that no feature should escalate privileges for virt-* pods owned by users beyond minimally viable. This would be a separate discussion and a separate PR though.

fabiand:
@booxter can you please file an issue to track the privilege removal for sriov-ed VMs?

PR #1797, opened 2018-12-04
Adding IP into registry disk container as a networking test relies on…

gageorsburn:
**What this PR does / why we need it**:
A networking test is currently broken. The container disk image is the first container in a pod currently. This may change over time and as such we should not rely on the ordering of the containers, rather call out the containers by name. In the meantime This will be a quick fix that adds IP into both 2 containers so that order of the containers does not matter.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:
```release-note
NONE
```

PR #1800, opened 2018-12-04
Fixed build caused by goimports change

booxter:
The following patch landed recently in goimports that introduced named
imports for name/path mismatches:

https://go.googlesource.com/tools/+/ee45598d2ff288037f53f9e13ae0b1a6e2165ad5

Majority of updates to the new requirements landed as part of PR #1797.
Still, these bits were missed.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/crd-generator/crd-generator.go
+++ b/tools/crd-generator/crd-generator.go
@@ -27,7 +27,7 @@ import (
 	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"kubevirt.io/kubevirt/pkg/api/v1"
+	v1 "kubevirt.io/kubevirt/pkg/api/v1"
 )
 
 func generateBlankCrd() *extensionsv1.CustomResourceDefinition {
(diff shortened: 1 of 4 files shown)


rmohr:
@booxter ah ok, so i probably failed on the travis run when merged in master? Maybe we should not use the container cache in travis, which we recently intreoduced #1779. It is only there for stdci and has no advantages on travis (job runtime is the same like uncached) ...

rmohr:
Here what I would do, to ensure that travis never uses the cached builder image:

```diff
diff --git a/.travis.yml b/.travis.yml
index 5bd249567..278ab7b17 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ install:
 
 script:
 - export KUBEVIRT_UPDATE_CACHE_FROM="${TRAVIS_BRANCH}"
-- export KUBEVIRT_CACHE_FROM="${TRAVIS_BRANCH}"
 - if [[ "${TRAVIS_PULL_REQUEST}" != "false" ]]; then make pull-cache; fi # only use the cache in PRs, rebuild otherwise
 - make generate
 - if [[ -n "$(git status --porcelain)" ]] ; then echo "It seems like you need to run
```

booxter:
@rmohr I would argue, perhaps we should not really pull in master of goimports and other linters and linter-like tools, but pin them.

booxter:
@rmohr I think both disabling the cache for travis builds as well as pinning versions should be implemented. (And of course we need this PR anyway.)

rmohr:
@booxter works for me too.

PR #1801, opened 2018-12-04
Pin goimports to release 1.11 branch

booxter:
Running against master means we are prone to breakage if changes to
rules land there.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/docker-builder/Dockerfile
+++ b/hack/docker-builder/Dockerfile
@@ -18,7 +18,10 @@ RUN \
     source /etc/profile.d/gimme.sh && \
     go get github.com/mattn/goveralls && \
     go get -u github.com/Masterminds/glide && \
-    go get -u golang.org/x/tools/cmd/goimports && \
+    go get -d golang.org/x/tools/cmd/goimports && \
+    cd /go/src/golang.org/x/tools/cmd/goimports && \
+    git checkout release-branch.go1.11 && \
+    go install && \
     git clone https://github.com/mvdan/sh.git $GOPATH/src/mvdan.cc/sh && \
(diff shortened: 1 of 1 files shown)


booxter:
In general we should do it for any linters and linter-like tools. I couldn't find any other tools we rely on though, so it's goimports only here. Perhaps there are more that I am not aware.

hack/docker-builder/Dockerfile
```diff
@@ -18,7 +18,10 @@ RUN \
     source /etc/profile.d/gimme.sh && \
     go get github.com/mattn/goveralls && \
     go get -u github.com/Masterminds/glide && \
-    go get -u golang.org/x/tools/cmd/goimports && \
+    go get golang.org/x/tools/cmd/goimports && \
```
rmohr: I think we also have add `-d`, to be most protected against breakage. Otherwise it would try to  compile master too.

booxter: Yeah I even intended to, checked the help message... and forgot. :) Fixed now.

mfranczy:
Can we backport this to branches release-0.[8-10]? I can do it (by cherrypick bot), just asking if anybody see any reason why not. I checked this locally on branch release-0.8, works fine.

rmohr:
Makes sense from my point of view.

mfranczy:
I see backport to release-0.8 has merge conflicts so instead of doing something with it I added goimports changes to my already created PR https://github.com/kubevirt/kubevirt/pull/1817 (which fixes docker build), sorry for the confusion with that.

PR #1802, opened 2018-12-04
Fixed virtio container disk, Ci doesn't support multistage builds

slintes:
Fixed new virtio-conatiner-disk, CI doesn't support multistage docker builds

**Release note**:
```release-note
NONE
```

### DIFF
--- a/images/virtio-container-disk/Dockerfile
+++ b/images/virtio-container-disk/Dockerfile
@@ -16,19 +16,13 @@
 # Copyright 2017 Red Hat, Inc.
 #
 
-FROM fedora:28 as builder
+FROM kubevirt/container-disk-v1alpha
+
+LABEL maintainer="The KubeVirt Project <kubevirt-dev@googlegroups.com>"
 
 RUN curl https://fedorapeople.org/groups/virt/virtio-win/virtio-win.repo -o /etc/yum.repos.d/virtio-win.repo \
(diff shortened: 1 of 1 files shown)


images/virtio-container-disk/Dockerfile
```diff
@@ -16,19 +16,13 @@
...
 #
 
-FROM fedora:28 as builder
+FROM kubevirt/container-disk-v1alpha
+
+LABEL maintainer="The KubeVirt Project <kubevirt-dev@googlegroups.com>"
 
 RUN curl https://fedorapeople.org/groups/virt/virtio-win/virtio-win.repo -o /etc/yum.repos.d/virtio-win.repo \
-&& dnf install -y virtio-win && mkdir -p /disk 
+&& dnf install -y virtio-win && dnf clean all && mkdir -p /disk
```
rmohr: Can you either use a symlink instead of cp, or "mv" the image in the same "RUN" statement? Otherwise we have the disk two times incluided.

slintes: sure, done

slintes: ah wait, according to the comment a symlink doesn't work...? Will change to mv

slintes: mv doesn't follow symlinks, so I copy and delete now

PR #1816, opened 2018-12-05
Automatically fill in requests / limits for resourceNames listed in multus network annotations

booxter:
**What this PR does / why we need it**: this is a temporary workaround that removes the need to explicitly list SR-IOV device requests / limits when a VMI is attached to a SR-IOV network that already lists intel.com/sriov resourceName in its annotations. (The same mechanism can be leveraged by other plugins, for example, bridge-cni or ovs-cni.) This is a temporary solution because in the future Multus team is going to ship a mutating webhook that will serve the same need, at which point we'll drop our implementation.

Fixes #

**Release note**:

```release-note
Automatically fill in resources.limits and resources.requests for SR-IOV network attachments.
```

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -431,6 +431,7 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 	vmiResources := vmi.Spec.Domain.Resources
 
 	resources.Requests = make(k8sv1.ResourceList)
+	resources.Limits = make(k8sv1.ResourceList)
 
 	// Copy vmi resources requests to a container
 	for key, value := range vmiResources.Requests {
@@ -438,20 +439,12 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 	}
 
 	// Copy vmi resources limits to a container
-	if vmiResources.Limits != nil {
-		resources.Limits = make(k8sv1.ResourceList)
-	}
-
 	for key, value := range vmiResources.Limits {
 		resources.Limits[key] = value
 	}
 
 	// Consider hugepages resource for pod scheduling
 	if vmi.Spec.Domain.Memory != nil && vmi.Spec.Domain.Memory.Hugepages != nil {
-		if resources.Limits == nil {
-			resources.Limits = make(k8sv1.ResourceList)
-		}
-
 		hugepageType := k8sv1.ResourceName(k8sv1.ResourceHugePagesPrefix + vmi.Spec.Domain.Memory.Hugepages.PageSize)
 		resources.Requests[hugepageType] = resources.Requests[k8sv1.ResourceMemory]
 		resources.Limits[hugepageType] = resources.Requests[k8sv1.ResourceMemory]
@@ -513,9 +506,6 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		// schedule only on nodes with a running cpu manager
 		nodeSelector[v1.CPUManager] = "true"
 
-		if resources.Limits == nil {
-			resources.Limits = make(k8sv1.ResourceList)
-		}
 		cores := uint32(0)
 		if vmi.Spec.Domain.CPU != nil {
 			cores = vmi.Spec.Domain.CPU.Cores
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -167,6 +167,30 @@ func CPUModelLabelFromCPUModel(vmi *v1.VirtualMachineInstance) (label string, er
 	return
 }
 
+// Request a resource by name. This function bumps the number of resources,
+// both its limits and requests attributes.
+//
+// If we were operating with a regular resource (CPU, memory, network
+// bandwidth), we would need to take care of QoS. For example,
+// https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed
+// explains that when Limits are set but Requests are not then scheduler
+// assumes that Requests are the same as Limits for a particular resource.
+//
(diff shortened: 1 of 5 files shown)


booxter:
@phoracek I think this should also help with your ovs-cni work.

booxter:
ci retest please

booxter:
ci test please

jekader:
CI system maintenance caused some tests for this PR to be left in "Pending" state. To re-run them please add a comment saying "ci test please".

jekader:
CI system maintenance caused some tests for this PR to be left in "Pending" state. To re-run them please add a comment saying "ci test please".

booxter:
ci test please

pkg/virt-controller/services/template.go
```diff
@@ -555,6 +578,21 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
+	networkToResourceMap, err := getNetworkToResourceMap(t.virtClient, vmi)
+	if err != nil {
+		return nil, err
+	}
+
+	// Register resource requests and limits corresponding to attached multus networks.
+	// TODO(ihar) remove when we adopt Multus mutating webhook that handles the job.
+	for _, resourceName := range networkToResourceMap {
+		if resourceName != "" {
+			requestResource(resources, resourceName)
```
SchSeba: Just a small question here.
The resources gets update? you send an object to the function and not a pointer

booxter:
Need to add some unit tests for the new function.

booxter:
The latest upload changes the following:

- pass resources into requestResource via reference;
- fixed a bug in requestResource that made the call for of a resource that is already in the mapping ineffective;
- added unit tests for requestResource.

booxter:
ci test please

pkg/virt-controller/services/template.go
```diff
@@ -144,6 +144,31 @@ func isSRIOVVmi(vmi *v1.VirtualMachineInstance) bool {
...
+		resources.Limits[name] = unitQuantity
+	}
+
+	// Fill in requests
+	val, ok = resources.Requests[name]
+	if ok {
+		val.Add(unitQuantity)
+		resources.Requests[name] = val
+	} else {
+		resources.Requests[name] = unitQuantity
```
vladikr: Usually, it should be fine to avoid setting Requests for a certain resource if a corresponding Limits is set for it.
When only Limits is set for a certain resource k8s will make it be equal to resource's Requests.
If all Limits are equal to Requests (or Requests not set at all) Pod's QOS will be set to `Guaranteed`.
As it is now if the user has set only Limits, we will set Requests to `unitQuantity`, which will affect Pod's QOS... 
I think we only need to set Requests for a certain resource if it's missing Limits.

booxter: @vladikr I will need to think about it, but this logic repeats what was proposed for Multus webhook. If there is an issue in this logic we should bring it up to Multus folks to address too. Let me think about it.

booxter: @vladikr from device plugin design document: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-management/device-plugin.md#end-user-story

```
Devices can be selected using the same process as for OIRs in the pod spec. Devices have no impact on QOS. However, for the alpha, we expect the request to have limits == requests.
```

This wording suggests that we should always set limits == requests for devices managed by a device plugin (for the alpha; the feature is in beta right now). I am not aware of any changes that happened since then in regards to device plugin API that would suggest a change for limits management.

vladikr: @booxter Yeah, I've expected to see limits == requests. Great that it doesn't affect QOS.

But right now, I see that the code above allows setting it differently.
If limits = 2 and no requests set  - requests will be set to 1, no?

booxter: @vladikr yes but that would a problem of those who put incorrect limit that doesn't match requests in the first place.

vladikr: @booxter I'm not sure, I think normally in k8s if a user specifies only limits it implies that requests will match the limits.
This is how it works with cpu and mem, at least.[1]
`Note: If a Container specifies its own memory limit, but does not specify a memory request, Kubernetes automatically assigns a memory request that matches the limit. Similarly, if a Container specifies its own CPU limit, but does not specify a CPU request, Kubernetes automatically assigns a CPU request that matches the limit.`

I don't have a strong opinion about this. If you think that the current behavior will not surprise the users, I'm fine with it. 

[1] https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed

booxter: @vladikr I don't think we should refer to QoS documentation for a feature that explicitly says in its design document that it doesn't affect QoS.

But let's run the idea. How would the implementation look like? I will try something in pseudocode, please tell me if this is what you would have here.

```
func requestResource(resources, resourceName) {
    if resourceName in resources.Limits and resourceName not in resources.Requests {
        resources.Requests[resourceName] = resources.Limits[resourceName]
    }
    resources.Requests[resourceName]++
    resources.Limits[resourceName]++
}
```

vladikr: :) something like that.

But, I think I would just change 
`} else {
    resources.Requests[name] = unitQuantity
`

to something like: 
```
} else {
    val, _ := resources.Limits[name]
    resources.Requests[name] = val
}
```
Since the code already deals with `Limits` we can just set the limits value for the missing `Requests`

vladikr: > I don't think we should refer to QoS documentation for a feature that explicitly says in its design document that it doesn't affect QoS.

@booxter You're right. I've just seen it in other places as well, like `LimitRange` - but it was also realtred to cpu/mem I think.

vladikr:
Let's wait for CI

vladikr:
Test failures don't seem to be related.

PR #1818, opened 2018-12-06
make sure api server validates client certs if provided

mhenriks:

The recent Kubernetes CVE got me thinking about our API servers.  Turns out we are passing the wrong flag to our http server config.  An attacker can send an http request to our api server directly to control vms.  Just have to make sure it has a client cert (we check that one exists but do no validation) and the appropriate headers are set correctly (X-Remote-User, X-Remote-Group).

Fixes #

**Release note**:

```release-note
NONE
```


pkg/virt-api/api.go
```diff
@@ -905,7 +905,7 @@ func (app *virtAPIApp) startTLS() error {
 		// response is given. That status request won't send a peer cert regardless
 		// if the TLS handshake requests it. As a result, the TLS handshake fails
 		// and our aggregated endpoint never becomes available.
-		ClientAuth: tls.RequestClientCert,
+		ClientAuth: tls.VerifyClientCertIfGiven,
```
rmohr: Should we use `RequireAndVerifyClientCert`?

mhenriks: According to this comment, no:  https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-api/api.go#L902-L907

mhenriks: For privileged urls, here is where we check that the request contains a cert: https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-api/rest/authorizer.go#L209

PR #1821, opened 2018-12-09
Fix parsing of custom emulated machines

ahadas:
We now limit kubevirt to Q35 by default but enable overriding it for testing and debugging purposes. This can be done by setting custom `emulated-machines`  within `kubevirt-config` as a comma-separated list. However, users would typically introduce spaces between the values in this list (e.g., "a, b, c") and those space are not treated correctly by the current code. Those spaces will now by trimmed during parsing.

This PR also contains some minor cleanups as separate commits.

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/virt-config/config-map.go
+++ b/pkg/virt-config/config-map.go
@@ -34,10 +34,11 @@ import (
 )
 
 const (
-	configMapName         = "kubevirt-config"
-	featureGateEnvVar     = "FEATURE_GATES"
-	FeatureGatesKey       = "feature-gates"
-	emulatedMachineEnvVar = "VIRT_EMULATED_MACHINES"
+	configMapName          = "kubevirt-config"
+	featureGateEnvVar      = "FEATURE_GATES"
+	FeatureGatesKey        = "feature-gates"
+	emulatedMachinesEnvVar = "VIRT_EMULATED_MACHINES"
+	emulatedMachinesKey    = "emulated-machines"
 )
 
 // We cannot rely on automatic invocation of 'init' method because this initialization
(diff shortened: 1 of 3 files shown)


ahadas:
@cynepco3hahue  apparently, there was another test that used an invalid PCI address on q35 (part of vmi_networking_test) so I pushed a fix for that test as well.

ahadas:
ci test please

rmohr:
@davidvossel @cynepco3hahue should we start using arrays instead of comma/space separated strings?

ahadas:
Please correct me if I'm wrong - a configmap is just a mapping of string to string, right?
So let's say that we save those values as arrays or lists or whatever collection we find best, when parsing the yaml to store it in the config map we'll need one transformation and yet another transformation when parsing the value from the config map into consumable functions in virt-properties.go, is that right? because in this case, I would rather prefer to reduce the number of transformations and stick with the comma/space separated string...

rmohr:
> if that's the case, I would rather prefer to reduce the number of transformations and stick with the comma/space separated string...

For both we have to interpret it. In one case we have to write our own parsing, in the other case it is clear on how it should be interpreted. We don't have to do the switch now.

fabiand:
To be honest - That would be odd (having yalm in strings).

Maybe it's time for a custom resource to have a KubeVirtConfig.

ahadas:
> To be honest - That would be odd (having yalm in strings).
+1

rmohr:
At the end it should probably be it's own CR, but it is completely normal tho have nested yamls in ConfigMaps.

rmohr:
And as said, we don't have to solve it in this PR.

ahadas:
Ah, that looks better than what I had in mind.
But yeah, +1 on doing it separately.

ahadas:
@rmohr Done

cynepco3hahue:
ci test please

PR #1839, opened 2018-12-11
add support for CPU sockets and threads

ksimon1:
Add support for CPU sockets and threads

The performance team identified an issue with how we define CPUs for VMs. They were getting significantly better performance when configuring virtual CPUs as sockets (vs. the current method of specifying cores) - https://bugzilla.redhat.com/show_bug.cgi?id=1653453. This PR allows user to specify socket and threads in vm. 
This patch changes the way how cores are calculated. Previously cpu topology had just 1 socket, 1 thread and n cores. Now cpu topology can have n cores, threads and sockets. If none of cores, threads and sockets are defined in `Spec.Domain.CPU`, virt-launcher looks into `Spec.Domain.Resources.Requests` and `Spec.Domain.Resources.Limits` for `k8sv1.ResourceCPU` and sets this value into sockets (because sockets have better performance than cores). Calculation of `vcpu` is changed to `sockets*cores*threads`.

@MarSik, @fromanirh, @yanirq, @petrkotas, @rmohr please review

**Release note**:
```release-note
Add support for CPU sockets and threads
```

### DIFF
--- a/pkg/util/hardware/hw_utils.go
+++ b/pkg/util/hardware/hw_utils.go
@@ -22,6 +22,8 @@ package hardware
 import (
 	"strconv"
 	"strings"
+
+	v1 "kubevirt.io/kubevirt/pkg/api/v1"
 )
 
 const CPUSET_PATH = "/sys/fs/cgroup/cpuset/cpuset.cpus"
@@ -55,4 +57,25 @@ func ParseCPUSetLine(cpusetLine string) (cpusList []int, err error) {
 		}
 	}
 	return
+}
+
+//GetNumberOfVCPUs returns number of vCPUs
+//It counts sockets*cores*threads
+func GetNumberOfVCPUs(cpuSpec *v1.CPU) int64 {
+	vCPUs := cpuSpec.Cores
+	if cpuSpec.Sockets != 0 {
+		if vCPUs == 0 {
+			vCPUs = cpuSpec.Sockets
+		} else {
+			vCPUs *= cpuSpec.Sockets
+		}
+	}
+	if cpuSpec.Threads != 0 {
+		if vCPUs == 0 {
+			vCPUs = cpuSpec.Threads
+		} else {
+			vCPUs *= cpuSpec.Threads
+		}
+	}
+	return int64(vCPUs)
 }
--- a/pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
+++ b/pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
@@ -41,6 +41,7 @@ import (
 	v1 "kubevirt.io/kubevirt/pkg/api/v1"
 	"kubevirt.io/kubevirt/pkg/log"
 	"kubevirt.io/kubevirt/pkg/util"
+	"kubevirt.io/kubevirt/pkg/util/hardware"
 	"kubevirt.io/kubevirt/pkg/virt-api/webhooks"
 	virtconfig "kubevirt.io/kubevirt/pkg/virt-config"
 )
@@ -677,7 +678,8 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 		limitsMem := spec.Domain.Resources.Limits.Memory().Value()
 		requestsCPU := spec.Domain.Resources.Requests.Cpu().Value()
 		limitsCPU := spec.Domain.Resources.Limits.Cpu().Value()
-		vmCores := int64(spec.Domain.CPU.Cores)
+		vCPUs := hardware.GetNumberOfVCPUs(spec.Domain.CPU)
+
 		// memory should be provided
 		if limitsMem == 0 && requestsMem == 0 {
 			causes = append(causes, metav1.StatusCause{
@@ -721,7 +723,7 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 		}
 
 		// cpu amount should be provided
-		if requestsCPU == 0 && limitsCPU == 0 && vmCores == 0 {
+		if requestsCPU == 0 && limitsCPU == 0 && vCPUs == 0 {
 			causes = append(causes, metav1.StatusCause{
 				Type: metav1.CauseTypeFieldValueInvalid,
 				Message: fmt.Sprintf("either %s or %s or %s must be provided when DedicatedCPUPlacement is true ",
@@ -746,8 +748,8 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 		}
 
 		// cpu resource and cpu cores should not be provided together - unless both are equal
-		if (requestsCPU > 0 || limitsCPU > 0) && vmCores > 0 &&
-			requestsCPU != vmCores && limitsCPU != vmCores {
+		if (requestsCPU > 0 || limitsCPU > 0) && vCPUs > 0 &&
+			requestsCPU != vCPUs && limitsCPU != vCPUs {
 			causes = append(causes, metav1.StatusCause{
 				Type: metav1.CauseTypeFieldValueInvalid,
 				Message: fmt.Sprintf("%s or %s must not be provided at the same time with %s when DedicatedCPUPlacement is true ",
--- a/pkg/virt-launcher/virtwrap/api/converter.go
+++ b/pkg/virt-launcher/virtwrap/api/converter.go
@@ -766,18 +766,17 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 		return err
 	}
 
-	if vmi.Spec.Domain.CPU != nil {
-		// Set VM CPU cores
-		domain.Spec.CPU.Topology = &CPUTopology{
-			Sockets: 1,
-			Cores:   calculateRequestedVCPUs(vmi),
-			Threads: 1,
-		}
-		domain.Spec.VCPU = &VCPU{
-			Placement: "static",
-			CPUs:      calculateRequestedVCPUs(vmi),
-		}
+	// Set VM CPU cores
+	// CPU topology will be created everytime, because user can specify
+	// number of cores in vmi.Spec.Domain.Resources.Requests/Limits, not only
+	// in vmi.Spec.Domain.CPU
+	domain.Spec.CPU.Topology = getCPUTopology(vmi)
+	domain.Spec.VCPU = &VCPU{
+		Placement: "static",
+		CPUs:      calculateRequestedVCPUs(domain.Spec.CPU.Topology),
+	}
 
+	if vmi.Spec.Domain.CPU != nil {
 		// Set VM CPU model and vendor
 		if vmi.Spec.Domain.CPU.Model != "" {
 			if vmi.Spec.Domain.CPU.Model == v1.CPUModeHostModel || vmi.Spec.Domain.CPU.Model == v1.CPUModeHostPassthrough {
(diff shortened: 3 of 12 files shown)


ovirt-infra:

fabiand:
@ksimon1 thanks for the pr.

Please add a little bit more context to the PR description to tell a reviewer on the high-level what you are changing.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -1014,27 +1010,64 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+		return vmi.Spec.Domain.CPU.Threads
+	}
+	return threads
+}
+
+func calculateRequestedVCPUs(cpuTopology *CPUTopology) uint32 {
+	return cpuTopology.Cores * cpuTopology.Sockets * cpuTopology.Threads
+}
+
+func calculateRequestedCPUs(vmi *v1.VirtualMachineInstance) uint32 {
```
MarSik: Shouldn't this be deleted now?

ksimon1: No, because you still need to get cores from the vmi.

tests/vmi_configuration_test.go
```diff
@@ -110,6 +110,64 @@ var _ = Describe("Configurations", func() {
...
+				}
+				vmi.Spec.Domain.Resources = v1.ResourceRequirements{
+					Requests: kubev1.ResourceList{
+						kubev1.ResourceMemory: resource.MustParse("64M"),
+					},
+				}
+
+				By("Starting a VirtualMachineInstance")
+				vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+				Expect(err).ToNot(HaveOccurred())
```
petrkotas: Please add labels to Expects. Without the labels, the error, if any, will not tell much information, eg.
```
Expected err to be nil.
```
With label: `Expect(err).ToNot(HaveOccurred(), "should start vmi")` it will print the label along
```
should start vmi:
Expected err to be nil.
```

tests/vmi_configuration_test.go
```diff
@@ -110,6 +110,64 @@ var _ = Describe("Configurations", func() {
...
+				By("Expecting the VirtualMachineInstance console")
+				expecter, err := tests.LoggedInAlpineExpecter(vmi)
+				Expect(err).ToNot(HaveOccurred())
+				defer expecter.Close()
+
+				By("Checking the number of sockets under guest OS")
+				_, err = expecter.ExpectBatch([]expect.Batcher{
+					&expect.BSnd{S: "grep '^physical id' /proc/cpuinfo | uniq | wc -l\n"},
+					&expect.BExp{R: "3"},
+				}, 15*time.Second)
```
petrkotas: I would increase the time out to more than 15s, our CI tends to be slow sometimes. 60s should work :), as is in the second test.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -1014,27 +1010,64 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+	cpu := vmi.Spec.Domain.CPU
+	if cpu.Sockets != 0 {
+		return cpu.Sockets
+	}
+
+	if cpu.Cores == 0 && cpu.Threads == 0 {
+		//if cores, sockets, threads are not set, take value from domain resources request or limits and
+		//set value into sockets, which have best performance (https://bugzilla.redhat.com/show_bug.cgi?id=1653453)
+		resources := vmi.Spec.Domain.Resources
+		if cpuRequests, ok := resources.Requests[k8sv1.ResourceCPU]; ok {
```
fabiand: CPU Resource can also be in milli format aka `0.100` or alike, thus no int.

But anything else than int is not helpful for number of sockets.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -1014,27 +1010,64 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+			sockets = uint32(cpuRequests.Value())
+		} else if cpuLimit, ok := resources.Limits[k8sv1.ResourceCPU]; ok {
+			sockets = uint32(cpuLimit.Value())
```
fabiand: Same here.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -1014,27 +1010,64 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+}
+
+func calculateRequestedSockets(vmi *v1.VirtualMachineInstance) uint32 {
+	sockets := uint32(1)
+	cpu := vmi.Spec.Domain.CPU
+	if cpu.Sockets != 0 {
+		return cpu.Sockets
+	}
+
+	if cpu.Cores == 0 && cpu.Threads == 0 {
```
fabiand: What happens if one of them is 0?

How does this affect `calculateRequestedVCPUs`?

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -768,14 +768,10 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 
 	if vmi.Spec.Domain.CPU != nil {
 		// Set VM CPU cores
-		domain.Spec.CPU.Topology = &CPUTopology{
-			Sockets: 1,
-			Cores:   calculateRequestedVCPUs(vmi),
-			Threads: 1,
-		}
+		domain.Spec.CPU.Topology = getCPUTopology(vmi)
```
fabiand: The only change that was needed here was to increase `Sockets` instead of `Cores`

pkg/virt-launcher/virtwrap/api/converter.go
fabiand: It would only be an int in the guaranteed case.

pkg/virt-launcher/virtwrap/api/converter.go
ksimon1: if one of them is 0, then functions calculateRequestedThreads or calculateRequestedCPUs will change the value to 1 and the other one will remain the same.

MarSik:
@fabiand We need to be able to specify all sockets, cores and threads for windows cases. The licensing there might limit the number of sockets you are allowed to utilize.

The patch here could just map cores to sockets, but it would be confusing (and breaking the API again) once we change that again to support sockets again in a separate PR. We have unique opportunity to break the API together with everything else we did in 1.4 and not break it again in the future.

pkg/virt-launcher/virtwrap/api/converter.go
MarSik: Making it confusing and preventing the support of full topology in the future..

pkg/virt-launcher/virtwrap/api/converter.go
MarSik: Did you notice this is the original code just moved to different place?

tests/vmi_configuration_test.go
ksimon1: done

fabiand:
> @fabiand We need to be able to specify all sockets, cores and threads for windows cases. The licensing there might limit the number of sockets you are allowed to utilize.

That's what I proposed. Extend the API to let a user specify cores and sockets (threads for completeness as well I guess (this referes to (1) above).

> The patch here could just map cores to sockets, but it would be confusing (and breaking the API again) once we change that again to support full topology in a separate PR. We have unique opportunity to break the API together with everything else we did in 1.4 and not break it again in the future.

I'm not suggesting to map cores to sockets or vice-versa. I'm suggesting to expose a public API where a cuser can specify sockets, cores and threads.

1.4 seems to be a non-KubeVirt version number

> Moreover, adding support for topology means you have to touch all the places you mentioned and there is no way around that. So the additional PR wouĺd look the same as this one again.

I'm not sure how topology would help here.

My point in general is: We know that we should be able to specifc sockets (and cores). Thus let's change the API to specify both of them (and threads becuase it's just one more item).
The other part is that we know that when vcpus are specified, that they should increase sockets instead of cores (as it's done right now).
I'm also recommending to keep it simple, as I heared that there are requests to backport this performance sensitive patch to a stable branch. For this, smaller and well understood patches are better.
I'm proposing to do these ismple changes right now, and any adiditonal work as follow up.

pkg/virt-launcher/virtwrap/api/converter.go
MarSik: But you are right, we need to round the value up to make sure it is a greater than zero  integer.

ksimon1: If i am correct, cpuRequests.Value() should return round int64 number.

MarSik: We were digging into kubernetes source with @vladikr and it seems that there are two possible implementations, [one](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L684) just strips the fractional part (0.5 -> 0) and the [other one](https://github.com/golang/build/blob/master/kubernetes/api/quantity.go#L356) rounds up.

vladikr:
ci test please

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -1014,27 +1010,51 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 	return nil
 }
 
-func calculateRequestedVCPUs(vmi *v1.VirtualMachineInstance) uint32 {
-	cores := uint32(0)
-	if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Cores != 0 {
-		return vmi.Spec.Domain.CPU.Cores
+func getCPUTopology(vmi *v1.VirtualMachineInstance) *CPUTopology {
```
vladikr: would be good if you could add a unit test, to cover all the options, please.

ksimon1:
@petrkotas, @vladikr, @fabiand, @rmohr please can you review this?

pkg/virt-launcher/virtwrap/api/converter_test.go
```diff
@@ -551,36 +553,131 @@ var _ = Describe("Converter", func() {
...
-			Expect(domainSpec.CPU.Topology.Cores).To(Equal(uint32(3)))
-			Expect(domainSpec.CPU.Topology.Sockets).To(Equal(uint32(1)))
-			Expect(domainSpec.CPU.Topology.Threads).To(Equal(uint32(1)))
-			Expect(domainSpec.CPU.Mode).To(Equal("custom"))
-			Expect(domainSpec.CPU.Model).To(Equal("Conroe"))
-			Expect(domainSpec.VCPU.Placement).To(Equal("static"))
-			Expect(domainSpec.VCPU.CPUs).To(Equal(uint32(3)))
-		})
+				Expect(domainSpec.CPU.Topology.Cores).To(Equal(uint32(3)))
+				Expect(domainSpec.CPU.Topology.Sockets).To(Equal(uint32(2)))
```
petrkotas: Please do add the labels here. Otherwise failure, if any will report:
Expected 3 to be 2. Which is not helpful. I would like to see
Expected sockets:
3 to be 2.

Written as `Expect(domainSpec.CPU.Topology.Sockets).To(Equal(uint32(2)), "Expect sockets")`

Please to this for the rest.

pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
```diff
@@ -677,7 +677,22 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 		limitsMem := spec.Domain.Resources.Limits.Memory().Value()
 		requestsCPU := spec.Domain.Resources.Requests.Cpu().Value()
 		limitsCPU := spec.Domain.Resources.Limits.Cpu().Value()
-		vmCores := int64(spec.Domain.CPU.Cores)
+		vCPUs := int64(spec.Domain.CPU.Cores)
```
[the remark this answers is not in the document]
ksimon1: done

petrkotas:
The CI team, had to restart the jenkins. There might a need to trigger the CI again.

ksimon1:
@fabiand, @rmohr, @vladikr Please can you take a look on this PR?

petrkotas:
ci test please

fabiand:
We need Vladik's feedback befoer this PR can be merged.

pkg/util/hardware/hw_utils.go
```diff
@@ -56,3 +58,24 @@ func ParseCPUSetLine(cpusetLine string) (cpusList []int, err error) {
 	}
 	return
 }
+
+//GetNumberOfVCPUs returns number of vCPUs
+//It counts sockets*cores*threads
+func GetNumberOfVCPUs(cpuSpec *v1.CPU) int64 {
```
vladikr: I don't know if hardware utils is the right place for this method, as it calculates something which is related to the vmi spec, not hardware... but I won't block on it.

mfranczy: I ask for that this because it was used as well in webhook, so if you see better place for that then please suggest, thanks.

vladikr:
Seem like the failures aren't related to the change.

PR #1844, opened 2018-12-13
docs: make it clear kubevirt-ansible is recommended for sriov

booxter:
Now that kubevirt-ansible automatically deploys all the relevant
SR-IOV components on cluster, we should recommend using it unless one
has a legit reason not to (for example, to deploy kubevirt from
sources).

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/sriov.md
+++ b/docs/sriov.md
@@ -83,6 +83,13 @@ Now you are ready to set up your cluster.
 You can use your preferred mechanism to deploy your kubernetes cluster as long
 as you deploy on bare metal.
 
+Current recommendation is to use ```kubevirt-ansible``` to deploy the cluster.
+Ansible playbooks will also deploy all the relevant SR-IOV components
+for you. See [here](https://github.com/kubevirt/kubevirt-ansible/).
+
+You may still want to deploy software using `local` provider if you'd like to
+deploy from Kubevirt sources though.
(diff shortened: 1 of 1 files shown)


booxter:
Sebastian I personally don't know if that's possible to use kubevirt-ansible with local provider. Have you tried it?

SchSeba:
Sorry I mean with the external provider.

I just want to said that you can add to the document the ability of deploy only multus and the sriov DP and not the kubevirt.yaml ansible playbook.

booxter:
@SchSeba I'm trying to understand the goal for your suggestion. Is it to document options available in kubevirt-ansible? If so, I don't think kubevirt repository is the right place to deep dive.

My goal with this tiny PR is to make sure that readers are aware of a better tool to deploy discussed components, and that they explicitly opt in to deploy from sources (perhaps because they want to develop some new SR-IOV featuresin kubevirt or one of SR-IOV components). I would think that this tiny PR achieves this goal, and adding more details about all the different options that kubevirt-ansible repository provides would not improve for this particular goal.

SchSeba:
sounds good to me

booxter:
@vladikr could you please help with getting this merged? It's a tiny doc update. Thanks.

PR #1871, opened 2018-12-21
Scale virt-launcher to pick up config changes during tests

rmohr:

* The config changes are only half picked-up because the functests only scale virt-api and virt-controller. virt-handler needs to be scaled too.
* Don't deploy the test-infra via the test-binary. The test-infra is also needed during development. The test-binary deploys it for the usual development-flow too late and also does not keep the test-infra.

Alternative solution to #1869.

**Release note**:

```release-note
NONE
```


rmohr:
@lukas-bednar @slintes had to update it again. See the updated commit message. I think we need to still deploy the test-infra in our scripts, because we also want the test infra to be available during development.

rmohr:
ci test please

cluster/deploy.sh
```diff
@@ -25,19 +25,32 @@ source hack/config.sh
...
         if [[ $manifest =~ .*demo.* ]]; then
             continue
         fi
-        _kubectl create -f $manifest
+        _kubectl apply -f $manifest
     done
 fi
 
 if [[ "$KUBEVIRT_PROVIDER" =~ os-* ]]; then
+    _kubectl create -f ${MANIFESTS_OUT_DIR}/testing/ocp
```
lukas-bednar: you will need to move that scc manifest into this directory.

this change will affect that *-for-k8s.yaml & *-for-ocp.yaml fnmatch which I use to read appropriate manifests for relevant cluster.

rmohr:
Seems like some lanes had temporary connectivity issues:
```
local-volume-provisioner-slrxd   0/1       ImagePullBackOff   0         24m
```
For reference:

http://jenkins.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/2984/artifact/ci_build_summary.html

rmohr:
ci test please

PR #1875, opened 2018-12-26
Test attributes for user access tests

nellyc:
Changed all test levels in api_validation_test to focus (FDescribe, FContext, FIt)
added rfe id, test id, criticality etc

based on issue
https://github.com/kubevirt/containerized-data-importer/issues/359

This is an example on how i plan to change the tests, so we will be able to auto create them in Polarion from the code.
There are 2 steps to it:
first, we create the tests without test ids (if they are new tests)
after running the script which creates the tests in Polarion (kubevirt/qe-tools/polarion-generator), each one gets its own ID,
and we can than add the test ids to the test code (this will allow reporting & changing the test in the future).

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/access_test.go
+++ b/tests/access_test.go
@@ -30,7 +30,7 @@ import (
 	"kubevirt.io/kubevirt/tests"
 )
 
-var _ = Describe("User Access", func() {
+var _ = Describe("[rfe_id:500, crit:high, vendor:cnv-qe@redhat.com, level:component]User Access", func() {
 
 	flag.Parse()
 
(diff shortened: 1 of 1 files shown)


nellyc:
@rmohr, thanks for the review

regarding your question about test refactoring, that would depend on the level of refactoring, but in general, by having the test id on the 'It', we should be safe (while before even a typo fix would affect us)

I was also playing with the thought, let me know what you think, of executing our tool for each PR (running on stdci) which will give feedback to the developer on newly created test IDs, or maybe just run it on stage to make sure nothing is broken

rmohr:
> I was also playing with the thought, let me know what you think, of executing our tool for each PR (running on stdci) which will give feedback to the developer on newly created test IDs, or maybe just run it on stage to make sure nothing is broken

Hm, maybe something which we can add to travis, so that we don't introduce duplicate ids or something?

tests/access_test.go
```diff
@@ -30,7 +30,7 @@ import (
 	"kubevirt.io/kubevirt/tests"
 )
 
-var _ = Describe("User Access", func() {
+var _ = Describe("[rfe_id:500, crit:high, vendor:cnv-qe@redhat.com, level:component]User Access", func() {
```
rmohr: I have one request. Can we align with the way which k8s uses to specify tags? See for instance https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/test/e2e/network/networking_perf.go#L170.

The important difference for me is that they use `[Experimental] [Feature:Networking-IPv6] [Slow]` instead of `[Experimental, Feature:Networking-IPv6, Slow]`

rmohr: While this is not directly related to polarion tags, this  might be an interesting  read too: https://github.com/kubernetes/community/blob/master/contributors/devel/e2e-tests.md#kinds-of-tests

rmohr: And this one: https://github.com/onsi/ginkgo/issues/144

nellyc: done

cynepco3hahue:
@nellyc Can we close it?

nellyc:
> @nellyc Can we close it?

no :)
we need to merge it

nellyc:
ci test please

PR #1877, opened 2018-12-30
[WIP] Add a test which measures the time between state changes

gbenhaim:
I moved the jobs to run on faster servers, let's see how it affects performance, specifically VMI state change.

```release-note
None
```


gbenhaim:
VMI boot time is ~20 seconds.

PR #1883, opened 2019-01-02
propagate vmi uid to libvirt domain

ahadas:
Propagate the UID that a VMI is defined with to libvirt as the UUID of the domain. It is better than letting libvirt generate a different UID for the domain.

**Release note**:

```release-note
None
```


mfranczy:
Converter tests are failing (unit tests), you might want to add output from `make generate` as well.

ahadas:
@mfranczy thanks :) fixed the failing unit test and added the output of `make generate`

ahadas:
oops, do not merge! there's a problem with this that I'm checking

cynepco3hahue:
@ahadas When it will be ready for merge just run `/unhold`

ahadas:
Closing this PR as on the one hand, the UID (in ObjectMetadata) of a VMI is auto-generated by the system and cannot be set by the user. On the other hand, the Name (in ObjectMetadata) of the VMI cannot be updated. So in this situation, I don't see a reason to propagate the ID - when needing to do some operation through libvirt we can use the Name as the identifier.

rmohr:
Yes we explicitly did not use the k8s uuid, so that you can set the firmware uuid.

PR #1890, opened 2019-01-03
Automate vNIC tests

yossisegev:
Automate vNIC tests:
- Verify default NIC model type is VirtIO.
- Verify unsupported NIC model type cannot be applied.

### DIFF
--- a/tests/vmi_networking_test.go
+++ b/tests/vmi_networking_test.go
@@ -362,12 +362,24 @@ var _ = Describe("Networking", func() {
 		})
 
 		Context("VirtualMachineInstance with default interface model", func() {
+			// Unless an explicit interface model is specified, the default interface model is virtio.
 			It("should expose the right device type to the guest", func() {
 				By("checking the device vendor in /sys/class")
+
+				// Taken from https://wiki.osdev.org/Virtio#Technical_Details
+				virtio_vid := "0x1af4"
+
 				for _, networkVMI := range []*v1.VirtualMachineInstance{inboundVMI, outboundVMI} {
 					// as defined in https://vendev.org/pci/ven_1af4/
-					checkNetworkVendor(networkVMI, "0x1af4", "\\$ ")
+					checkNetworkVendor(networkVMI, virtio_vid, "\\$ ")
 				}
+			})
+
+			It("should reject the creation of virtual machine with unsupported interface model", func() {
+				// Create a virtual machine with an unsupported interface model
+				customIfVMI := NewRandomVMIWithInvalidNetworkInterface()
+				_, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(customIfVMI)
+				Expect(err).To(HaveOccurred())
 			})
 		})
 	})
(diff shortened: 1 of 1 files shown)


yossisegev:
@dankenigsberg @myakove 
Please review

tests/utils.go
```diff
@@ -1772,6 +1772,14 @@ func NewRandomVMIWithe1000NetworkInterface() *v1.VirtualMachineInstance {
 	return vmi
 }
 
+func NewRandomVMIWithInvalidNetworkInterface() *v1.VirtualMachineInstance {
```
myakove: I guess this function can be in the test file and not in utils.

SchSeba:
@yossisegev also please run `make generate` and update the PR to make travis happy

tests/utils.go
```diff
@@ -63,12 +63,12 @@ import (
 	"k8s.io/client-go/tools/remotecommand"
 
 	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/datavolumecontroller/v1alpha1"
-	"kubevirt.io/kubevirt/pkg/api/v1"
+	v1 "kubevirt.io/kubevirt/pkg/api/v1"
```
cynepco3hahue: redundant

tests/utils.go
```diff
@@ -39,7 +39,7 @@ import (
 	"time"
 
 	ghodssyaml "github.com/ghodss/yaml"
-	"github.com/google/goexpect"
+	expect "github.com/google/goexpect"
```
cynepco3hahue: use explicit names for packages only when you have a conflict with another package

yossisegev: VSCode entered these explicit package names automatically once I saved each of these files, so I couldn't even remove them (unless I stopped using vscode).

tests/utils.go
yossisegev: VSCode entered these explicit package names automatically once I saved each of these files, so I couldn't even remove them (unless I stopped using vscode).

tests/utils.go
yossisegev: I put this function here as a "sister" function to NewRandomVMIWithe1000NetworkInterface(), which is also implemented here although it is only used in vmi_networking_test.go.

SchSeba: @yossisegev you can do something like this directly in the test.
```golang
vmi := NewRandomVMIWithe1000NetworkInterface()
vmi.Spec.Domain.Devices.Interfaces[0].Model = "gibberish"
_, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
Expect(err).To(HaveOccurred())
```

yossisegev: @SchSeba , I'll go with with moving this function to the test file itself. Using NewRandomVMIWithe1000NetworkInterface() seems out of context of this specific test.

SchSeba: sounds good

tests/utils.go
yossisegev: Resolved this issue in VSCode with the help of Meni.

tests/utils.go
yossisegev: Resolved this issue in VSCode with the help of Meni.

yossisegev:
CI please test

PR #1895, opened 2019-01-04
Add virctl wrapper

petrkotas:
This PR introduces reading config files used to open either VNC or console to the VM, which is designed to be used when opening connection to vm from webui. It allows virtctl to be registered as browser mime extension.
The webui sends file containg protocol and vmname on
single line. File is parsed and parameters are passed
to the virctl, which opens the connection.

This is temporary workaround until the proper
proxy is in place.

@stu-gott currently it relies on vicrtl located in the same folder. Where should I point it to properly?

Fixes #

**Release note**:

```release-note
Adds virctl support for opening config file to start vnc or console. It is designed to be used as MIME extension in web browser.
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/1895)

### DIFF
--- a/cmd/virtctl/virtctl.go
+++ b/cmd/virtctl/virtctl.go
@@ -20,7 +20,9 @@
 package main
 
 import (
+	"fmt"
 	"os"
+	"strings"
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
@@ -28,6 +30,9 @@ import (
 )
 
 func main() {
-	virtctl.Execute()
+	if err := virtctl.Execute(os.Args[1:]); err != nil {
+		fmt.Fprintf(os.Stderr, strings.TrimSpace(err.Error())+"\n")
+		os.Exit(1)
+	}
 	os.Exit(0)
 }
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -20,3 +20,15 @@ func GetNamespace() (string, error) {
 	}
 	return namespaceKubevirt, nil
 }
+
+func FileExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	exists := false
+
+	if err == nil {
+		exists = true
+	} else if os.IsNotExist(err) {
+		err = nil
+	}
+	return exists, err
+}
--- a/pkg/virtctl/root.go
+++ b/pkg/virtctl/root.go
@@ -1,14 +1,18 @@
 package virtctl
 
 import (
+	"bufio"
+	"errors"
 	"fmt"
-	"os"
+	"io/ioutil"
+	"regexp"
 	"strings"
 
 	"github.com/spf13/cobra"
 
 	"kubevirt.io/kubevirt/pkg/kubecli"
 	"kubevirt.io/kubevirt/pkg/log"
+	"kubevirt.io/kubevirt/pkg/util"
 	"kubevirt.io/kubevirt/pkg/virtctl/console"
 	"kubevirt.io/kubevirt/pkg/virtctl/expose"
 	"kubevirt.io/kubevirt/pkg/virtctl/imageupload"
@@ -55,10 +59,110 @@ func NewVirtctlCommand() *cobra.Command {
 	return rootCmd
 }
 
-func Execute() {
+func Execute(args []string) error {
 	log.InitializeLogging("virtctl")
-	if err := NewVirtctlCommand().Execute(); err != nil {
-		fmt.Println(strings.TrimSpace(err.Error()))
-		os.Exit(1)
+	// check whether there is a MIME config file
+	// if yes, then set up the args for commands and execute
+	// otherwise normal operation
+	confArgs, err := parseMime(args)
+	if err != nil {
+		return err
 	}
+
+	cmd := NewVirtctlCommand()
+
+	if confArgs != nil {
+		cmd.SetArgs(confArgs)
+	} else {
+		cmd.SetArgs(args)
+	}
+
+	if err := cmd.Execute(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func parseMime(args []string) ([]string, error) {
+	var parsedArgs []string
+
+	for ind := 0; ind < len(args); ind++ {
+
+		filePath, err := checkForFile(args[ind])
+		if err == nil && filePath != "" {
+			data, err := ioutil.ReadFile(filePath)
+			if err != nil {
+				return nil, fmt.Errorf("Cannot read config file: %s", filePath)
+			}
+
+			mimeArgs, err := parseMimeConfig(string(data))
+			if err != nil {
+				return nil, err
+			}
+			for _, mimeArg := range mimeArgs {
+				parsedArgs = append(parsedArgs, mimeArg)
+			}
+
+			return parsedArgs, nil
+		} else if err != nil {
+			return nil, err
+		}
+
+		if args[ind] == "--kubeconfig" || args[ind] == "--server" {
+			parsedArgs = append(parsedArgs, args[ind], args[ind+1])
+			ind = ind + 1
+		}
+
+	}
+
+	return nil, nil
+}
+
+func checkForFile(arg string) (string, error) {
+	var err error
+	filePath := ""
+
+	if strings.HasSuffix(arg, ".vvv") {
+		exist, _ := util.FileExists(arg)
+		if exist {
+			filePath = arg
+		} else {
+			err = errors.New("File does not exist")
+		}
+	}
+
+	return filePath, err
+}
+
+func parseMimeConfig(data string) ([]string, error) {
+	// read only single line as expected
+	scanner := bufio.NewScanner(strings.NewReader(data))
+	scanner.Scan()
+	line := scanner.Text()
+
+	tokens := strings.Split(line, " ")
+	if len(tokens) != 3 {
+		// invalid param line
+		return nil, fmt.Errorf("Invalid file format, 3 parameters required, %d received", len(tokens))
+	}
+
+	// first token have to be one of vnc|console
+	if tokens[0] != "vnc" && tokens[0] != "console" {
+		return nil, fmt.Errorf("Protocol have to be one of: vnc, console. Got: %s", tokens[0])
+	}
+
+	for _, token := range tokens {
+		matched, err := regexp.MatchString(`^[a-z\d\-_]+$`, token)
+		if err != nil {
+			return nil, fmt.Errorf("Cannot parse token: %s", token)
+		}
+		if !matched {
+			return nil, fmt.Errorf("Token containing illegal character: %s", token)
+		}
+	}
+
+	tokens[1] = "--namespace=" + tokens[1]
+
+	return tokens, nil
 }
--- a/pkg/virtctl/expose/expose_test.go
+++ b/pkg/virtctl/expose/expose_test.go
@@ -80,7 +80,7 @@ var _ = Describe("Expose", func() {
 		Context("With empty set of flags", func() {
 			It("should succeed", func() {
 				cmd := tests.NewVirtctlCommand(expose.COMMAND_EXPOSE)
-				Expect(cmd).ToNot(BeNil())
+				Expect(cmd()).ToNot(BeNil())
 			})
 		})
 	})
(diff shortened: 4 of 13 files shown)


cmd/virtctl/virctl.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash 
+
+# This script is designed to be proxy used to open virtctl from config file
```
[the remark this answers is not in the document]
petrkotas: This script actually is a proxy opener. Not used by a proxy. Althoug I see missing 'a'

cmd/virtctl/virctl.sh
```diff
@@ -0,0 +1,45 @@
...
+
+set -ex
+
+function usage() {
+  echo "Illegal usage."
+  echo "Use ${0} config_file."
+  echo "Config file containes single line: {vnc|console} vm_name<newline>"
```
stu-gott: I still think the literal `<newline>` is confusing. that usually means to the user "replace newline with something appropriate"... :) and in fact, this isn't needed. Try adding trailing whitespace to your VM name. It gets stripped.

petrkotas: ok agree, double checked and it works without the new line :). Thanks for help

cmd/virtctl/virctl.sh
```diff
@@ -0,0 +1,45 @@
...
+fi
+
+if [[ ${VCTL_VM_NAME} == *['!'@#\$%^\&*()+';']* ]]
+then
+  echo "VM_NAME CONTAINS ILEGAL CHARACTER"
+  usage
+  exit 1
+fi
+
+virtctl ${VCTL_PROTOCOL} ${VCTL_VM_NAME}
```
rmohr: It would also require the namespace.

petrkotas: Great catch, thanks.

petrkotas:
Registering directly `virctl` is definitely an option, which we extensively discussed with @jelkosz. Support in browsers for executables as MIME handler requires handling single parameter - path to the file. For this to work with `virctl`, it would require adding support for reading files in the code as first parameter, which in my opinion would create a confusion within the context of current command scheme.

As an example please look at: `virctl /tmp/ff32244ef243/vm.remote`, versus current use: `virctl console --namespace my_namespace my_vm`.

For this reason, I have chosen the indirect path of having wrapper script. However, I am not fixed on this solution and I am open for better solution.

rmohr:
Let's go with `virtctl` directly. You can put that at the right place in the cobra setup. See also #1361 for an example where we read some environment variables first, then set the environment  variables as parameters for cobra and then use our normal execution paths.

petrkotas:
I have done the changes requested. However, as I explained in my previous comment, the usage is now inconsistent. Implementation is also something, I am not going to put my name on. Therefore please consider this PR as prove of concept and not a feature proposal.

To explain the implementation.
The implementation is done in a way that benefits the current code infrastructure. Alternative implementation is possible, but it would require significant code changes to allow initializing vnc and console commands external of rootCmd. 
However, for both cases, the issue with inconsistent behaviour is the same as exaplained in https://github.com/kubevirt/kubevirt/pull/1895#issuecomment-451887519

cmd/virtctl/virtctl.go
```diff
@@ -20,14 +20,60 @@
...
+			// Read line from the file
+			reader, err := os.Open(filePath)
+			if err != nil {
+				fmt.Printf("Cannot read config file: %s\n", filePath)
+				os.Exit(1)
+			}
+
+			scanner := bufio.NewScanner(reader)
+			scanner.Scan()
+			tokens := strings.Split(scanner.Text(), " ")
```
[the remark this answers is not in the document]
petrkotas: Yes, but it is considered wrong file format. I only read single line, which have to contain the whole set.

pkg/util/util.go
```diff
@@ -20,3 +20,15 @@ func GetNamespace() (string, error) {
 	}
 	return namespaceKubevirt, nil
 }
+
+// FileExist return true if file or dir exist
+func FileExist(path string) (bool, error) {
```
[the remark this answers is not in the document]
petrkotas: Godd point. Done.

petrkotas:
@rmohr reworked after suggestions, thanks. Working on the tests, without them this is incomplete. Also need to fix the tests directory to point to new FileExists location.

slintes:
@petrkotas seems you need to run `make generate` and `make` and commit changes

petrkotas:
@slintes Still working on it :) not finished yet.

pkg/virtctl/virctl_internal_test.go
```diff
@@ -0,0 +1,164 @@
...
+		osArgs := []string{"caller_executable_dummy", tempConfig.Name()}
+
+		It("should mark valid file", func() {
+			isFile, filePath := checkForFile(osArgs)
+			Expect(isFile).To(BeTrue(), "should mark existing file")
+			Expect(filePath).To(Equal(tempConfig.Name()), "marked file path should match")
+		})
+
+		It("should parse file properly", func() {
+			fmt.Println("RUN SECOND IT")
```
mfranczy: you might want to remove it

petrkotas: Thanks! Missed it.

petrkotas:
@mfranczy @rmohr @slintes do you have any more remarks?

petrkotas:
@slintes would you review please?

petrkotas:
@slintes done, waiting for the ci

petrkotas:
@slintes ping?

slintes:
lgtm!
Only one last small thing: please update the release note :)

petrkotas:
@slintes done, sorry about that, missed the note when updating description.

slintes:
I'm afraid there is sth going wrong: I just tested virtctl locally, and wasn't able to trigger any normal command anymore. The combination of returning nil/non-nil values and/or errors/no errors in parseMime/checkForFile doesn't seem to work correctly in all cases. When I add this to the beginning of parseMime, normal commands work again, but your unit tests fail:
```
	if len(args) != 2 || !strings.HasSuffix(args[1], ".vvv") {
		return parsedArgs, nil
	}
```
Sorry for catching this so late in the review.

petrkotas:
@slintes fixed that. For the record, this behaviour is the reason I did not want to add it internally. Introducing different behaviour that designed originally makes proper error handling hard.

I had to solve it, in a way that I ignore errors related to the file and go directly to other commands. I try to honour the original design and treat original command superior to file handling.

Would this do?

rmohr:
Petr I really can't follow your concern. I doubt that we will ever introduce a `*.vvv` subcommand.

pkg/virtctl/root.go
```diff
@@ -55,10 +59,98 @@ func NewVirtctlCommand() *cobra.Command {
...
+
+	return parsedArgs, err
+}
+
+func checkForFile(args []string) (string, error) {
+	var err error
+	filePath := ""
+
+	if len(args) == 2 {
+		if strings.HasSuffix(args[1], ".vvv") {
```
rmohr: If it is not a vvv file then treat it as an arbitrary subcommand. It is then not an unsupported file type.

pkg/virtctl/root.go
```diff
@@ -55,10 +59,98 @@ func NewVirtctlCommand() *cobra.Command {
...
+			exist, _ := util.FileExists(args[1])
+			if exist {
+				filePath = args[1]
+			} else {
+				err = errors.New("File does not exist")
+			}
+
+		} else {
```
rmohr: Think for insance about `virtctl version`. It would not work now.

pkg/virtctl/root.go
rmohr: Btw. `vvv` is used by oVirt too, to connect such files with vnc viewers, right? Wouldn't it  be better to use another suffix.

pkg/virtctl/root.go
```diff
@@ -55,10 +59,98 @@ func NewVirtctlCommand() *cobra.Command {
...
-func Execute() {
+func Execute(args []string) error {
 	log.InitializeLogging("virtctl")
-	if err := NewVirtctlCommand().Execute(); err != nil {
-		fmt.Println(strings.TrimSpace(err.Error()))
-		os.Exit(1)
+	// check whether there is a MIME config file
+	// if yes, then set up the args for commands and execute
+	// otherwise normal operation
+	confArgs, _ := parseMime(args)
```
rmohr: Please use `parseMime` also in `NewVirtctlCommand` in the functional tests and add functional tests.

rmohr: There are error cases, like having an invalid file present which are valid errors. Don't ignore them.

rmohr:
@petrkotas any updates on this?

petrkotas:
@rmohr yes finishing up tests. And will be pushing soon.

pkg/virtctl/root.go
petrkotas: Ovirt uses .vv extension. I would keep it here.

pkg/virtctl/root.go
petrkotas: Fixed

pkg/virtctl/root.go
petrkotas: I covered those tests in unit tests https://github.com/kubevirt/kubevirt/pull/1895/files#diff-2a0b79d92584e47990463ad0767f21ba

I designed the code so it can be unit tested and does not bulk up the functional tests.

Would this be enough?

rmohr: @petrkotas using the code in the functional tests should be as simple as replacing these two lines her: https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L2188-L2189 with `virtctl.Execute(append(commandline, args...)`.

In order to tests it then, you can simply run another virtctl command in the vnc test which we already have via a temporary mimetype file.

rmohr: and of course returning an execute function instead of executing it. Something like this:

```go
return func() error {
    return virtctl.Execute(append(commandline, args...)
}
```

petrkotas:
@rmohr fixed. Thanks for catching it up.

pkg/virtctl/root.go
```diff
@@ -55,10 +59,103 @@ func NewVirtctlCommand() *cobra.Command {
...
+	}
+
+	return nil
+}
+
+func parseMime(args []string) ([]string, error) {
+	var parsedArgs []string
+	var err error
+
+	for ind := 0; ind < len(args); ind++ {
```
rmohr: Ok, this function is very weird now. Let me summarize the  expectations again:
 - If a single argument is supplied which ends in `.vvv`, then it either has to be an existing and valid mime-type file, or we have an error.
 - If it is a single argument which does not end with `.vvv`, then we want a normal execution

Here the code:

```go
func parseMime(args []string) ([]string, error) {
  if len(args) == 2 && strings.HasSuffix(args[1], ".vvv") {
    // retuirns (nil, err) in case of no or an invalid file, (args, nil) otherwise
    return readAndParseMime(args[1])
  }
  return args, nil
}
```

The expectation right now is, that people are logged in to their cluster.

petrkotas: The function is weird, because there are optional arguments `--kubeconfig /pat/to/.kubeconfig` and `--server server` which are required in tests.
These two are also required when connecting to the vnc or console.

Thus I have to walk through the parameters and check one by one.

Regarding your expectations, the functions does exactly what you want. It is verified in the tests.

rmohr: Ah I guess you do that because of the functional tests where we pass kubeconfig. You can change them to pass the config parameters via environment variables.

rmohr: Also ok for me, if we assume that the first argument is the mime-file and we simply add any additional flags to the ones in the mime file.

petrkotas: I introduced one more check to verify only --kubeconfig and --server gets copied. Otherwise, I made sure original functionality is unharmed.

There is also one think to keep in mind. This patch will get deprecated once next version is released. The current redirection through api server will be insufficient and the proper connection proxy would have to be introduced.

Would you be ok with this implementation for now?

petrkotas:
I think this PR is no longer needed. If not required. @jelkosz do you agree?

mfranczy:
ping @jelkosz ^

PR #1922, opened 2019-01-09
Update workflow document

irosenzw:
Document major and minor upgrades for kubevirt

Fixes #

**Release note**:

```release-note

```


irosenzw:
@davidvossel @fabiand @rthallisey @ohadlevy @rmohr 
Please review the upgrade flow I tested on upstream

docs/upgrade.md
```diff
@@ -0,0 +1,149 @@
...
+
+This document outlines the step by step process of upgrading kubevirt
+from the perspective of the kubevirt-operator.  These steps can also
+be done by hand.
+
+**Major Upgrade** - Upgrade from 1.0 to 2.0
+**Minor Upgrade** - Upgrade from 1.0 to 1.1
+
+### Major Upgrade
+*kubevirt-1.1 to kubevirt-2.0*
```
rthallisey: Let's change this to upstream versions.  ```kubevirt-0.11.0 to kubevirt-0.12.0```

docs/upgrade.md
```diff
@@ -0,0 +1,149 @@
...
+ - Register kubevirt-2.0 CRD webhooks
+ - Update to kubevirt-2.0 CRDs
+ - Rollout kubevirt-2.0 virt-controllers
+ - Test virt-controller functionality with kubevirt-2.0 data
+ - Rollout kubevirt-2.0 virt-api
+ - Test virt-api functionality with kubevirt-2.0 data
+
+
+### Minor Upgrade
+*kubevirt-1.1 to kubevirt-1.2*
```
rthallisey: I think these instructions are different than below.  Maybe we should just remove?

docs/upgrade.md
```diff
@@ -0,0 +1,149 @@
+# Upgrading KubeVirt
+
+This document outlines the step by step process of upgrading kubevirt
+from the perspective of the kubevirt-operator.  These steps can also
+be done by hand.
+
+**Major Upgrade** - Upgrade from 1.0 to 2.0
+**Minor Upgrade** - Upgrade from 1.0 to 1.1
+
+### Major Upgrade
```
rthallisey: Put the minor upgrade above the major, since major isn't something we're going to handle until a we have a major version to jump to.

docs/upgrade.md
irosenzw: I don't think we should remove it. 
It is an overall plan, we may want to use it in the future.

rthallisey: You have the minor upgrade plan below.  Downstream can replace version numbers in the docs so we don't need to have two sections titled minor upgrade.

docs/upgrade.md
```diff
@@ -0,0 +1,149 @@
...

+
+2. Apply new manifests
+
+    `# cluster/kubectl.sh apply -f ~/kubevirt-12.yaml`
+
+    __Pods status:__
+    ![](upgrade_flow_screenshots/2.png)
+
+    We can see virt-api,virt-controller and virt-handler were created in the ‘kubevirt’ namespace while these the old configuration in kube-system namespace still exist. Thus, they need to be deleted.
```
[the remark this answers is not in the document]
irosenzw: Yes, this is specific for the upgrade between v0.11.0 to v0.12.0  .

dcritch:
So namespaces and API breakage is one thing, but we need to get in front of normal security patches too.

Say a CVE is identified in libvirt. It (probably?) won't be a new major or minor release, but a sysadmin is going to want to roll through any VMs and restart them on the new version as fast as possible, but in a controlled way.

It would be great if the upgrade operator could handle rolling restarts of VMs, perhaps temporarily oc patch'ed so the fresh image is pulled each time, before patching the image pull policy back to normal after all VMs have been restarted. 

Similarly but less complicated, if such a security fix hits the e.g. API or controller, there should be a nice clean way to pull the new image and restart the container(s).

rmohr:
/close

Once we support updates the operator will do it. Work started in #2033.

PR #1932, opened 2019-01-11
Fix virtcl expose

karmab:

Fixes #

**Release note**:

```release-note
fixes nodeport behaviour by getting back the functionality of using a specific port number
```

### DIFF
--- a/pkg/virtctl/expose/expose.go
+++ b/pkg/virtctl/expose/expose.go
@@ -187,7 +187,11 @@ func (o *Command) RunE(cmd *cobra.Command, args []string) error {
 	if port == 0 && len(ports) == 0 {
 		return fmt.Errorf("couldn't find port via --port flag or introspection")
 	} else if port != 0 {
-		ports = []v1.ServicePort{{Name: portName, Protocol: protocol, Port: port, TargetPort: targetPort}}
+		if strServiceType == "NodePort" {
+			ports = []v1.ServicePort{{Name: portName, Protocol: protocol, Port: port, TargetPort: targetPort, NodePort: port}}
+		} else {
+			ports = []v1.ServicePort{{Name: portName, Protocol: protocol, Port: port, TargetPort: targetPort}}
+		}
 	}
 
 	// actually create the service


slintes:
@karmab there is a compile error
```
pkg/virtctl/expose/expose.go:191:88: cannot use port (type int32) as type intstr.IntOrString in field value
pkg/virtctl/expose/expose.go:191:104: cannot use targetPort (type intstr.IntOrString) as type int32 in field value
```

rmohr:
That will not work as expected. Virtctl let's the apiserver assign a random port out of the node port range. We do the same. If the test is flaky then not because of that.

karmab:
> @karmab there is a compile error
> 
> ```
> pkg/virtctl/expose/expose.go:191:88: cannot use port (type int32) as type intstr.IntOrString in field value
> pkg/virtctl/expose/expose.go:191:104: cannot use targetPort (type intstr.IntOrString) as type int32 in field value
> ```

fixed it

slintes:
Hi @karmab, as Roman already noted, virtctl expose works like kubectl expose: you can use it to create a NodePort service, but it doesn't allow you to define the actual NodePort. It will be assigned by the apiserver.
I also don't see how this should fix the flaky test. There we create the service, get it back from the apiserver, and get the NodePort from it for the rest of the test.
So, I would like to close this. Agree?

karmab:
@slintes i m fine closing it since i can mimick the expected behaviour by manually creating a service but note it's actually a regression, as you were able to expose a specific node port some time ago

rmohr:
@karmab, I don't know what you are trying to do but note that users normally neither know the allowed port range nor which ports are already used. The node port can be set as you do by posting a service object, if you really have to, but the port range is shared between different users so you should let the apiserver assign it and then look it up. We always wanted expose work exactly like in k8s, that setting the node port was allowed was actually not intended and users should normally not use it.

Maybe it is the right thing for you to use it but maybe you should change the way you use node port services.

karmab:
@rmohr i think that if you want to allow remote ssh access to your vms, nodeport is a decent option, and from then, from a management side, beeing able to specify a desired port number seems cleaner than having to find out which one got assigned by an api server.
that said, i guess mimicking kubectl expose is a valid reason for not having this feature

rmohr:
@karmab OK, so you should really not do that in a multi-namespace/multi-user environment. As said you share the ports with others. You don't know which port is free to use. Relying on the apiserver and looking the port up is definitely the correct way here.

PR #1936, opened 2019-01-14
Automate service-expose test: Verify service still functions after VM restart

yossisegev:
Automate service-expose test: Verify service still functions after VM restart:
- Expose a cluster IP service on a VM.
- Verify service functionality.
- Restart VM.
- Re-verify service funcitonality.

### DIFF
--- a/tests/expose_test.go
+++ b/tests/expose_test.go
@@ -15,7 +15,6 @@ import (
 	v1 "kubevirt.io/kubevirt/pkg/api/v1"
 	"kubevirt.io/kubevirt/pkg/kubecli"
 	"kubevirt.io/kubevirt/pkg/virtctl/expose"
-	"kubevirt.io/kubevirt/pkg/virtctl/vm"
 	"kubevirt.io/kubevirt/tests"
 )
 
@@ -387,7 +386,7 @@ var _ = Describe("Expose", func() {
 			generateHelloWorldServer(vmi, virtClient, testPort, "tcp")
 		})
 
-		Context("Expose ClusterIP service", func() {
+		Context("Expose ClusterIP service on a VM.", func() {
 			It("[test_id:1538]Connect to ClusterIP services that was set when VM was offline", func() {
 				By("Getting back the cluster IP given for the service")
 				svc, err := virtClient.CoreV1().Services(vm.Namespace).Get(serviceName, k8smetav1.GetOptions{})
@@ -408,55 +407,10 @@ var _ = Describe("Expose", func() {
 
 				Eventually(getStatus, 60, 1).Should(Equal(k8sv1.PodSucceeded))
 			})
-		})
-	})
 
-	Context("Verify that an exposed service on a VM survives after restarting the VM.", func() {
-		const servicePort = "27017"
-		const serviceName = "cluster-ip-vm-restart"
-		var vm_obj *v1.VirtualMachine
+			It("Should verify the exposed service is functional before and after VM restart.", func() {
+				vm_obj := vm
 
-		tests.BeforeAll(func() {
-			By("Creating an VM object")
-			template := tests.NewRandomVMIWithEphemeralDiskAndUserdata(tests.ContainerDiskFor(tests.ContainerDiskCirros), "#!/bin/bash\necho 'hello'\n")
-			template.Labels = map[string]string{"expose": "vm"}
-			vm_obj = NewRandomVirtualMachine(template, false)
-
-			By("Creating the VM")
-			_, err := virtClient.VirtualMachine(tests.NamespaceTestDefault).Create(vm_obj)
-			Expect(err).ToNot(HaveOccurred())
-
-			By("Starting the VMI.")
-			virtctl := tests.NewRepeatableVirtctlCommand("start", "--namespace", vm_obj.Namespace, vm_obj.Name)
-			err = virtctl()
-			Expect(err).ToNot(HaveOccurred())
-
-			By("Verifying the VMI is running.")
-			Eventually(func() bool {
-				vm_obj, err = virtClient.VirtualMachine(vm_obj.Namespace).Get(vm_obj.Name, &k8smetav1.GetOptions{})
-				Expect(err).ToNot(HaveOccurred())
-				return vm_obj.Status.Ready
-			}, 120*time.Second, 1*time.Second).Should(BeTrue())
-
-			By("Exposing a service on the VM using virtctl")
-			virtctl = tests.NewRepeatableVirtctlCommand(expose.COMMAND_EXPOSE, "virtualmachine", "--namespace",
-				vm_obj.Namespace, vm_obj.Name, "--port", servicePort, "--name", serviceName, "--target-port", strconv.Itoa(testPort))
-			err = virtctl()
-			Expect(err).ToNot(HaveOccurred())
-
-			By("Getting the running VMI")
-			var vmi *v1.VirtualMachineInstance
-			Eventually(func() bool {
-				vmi, err = virtClient.VirtualMachineInstance(vm_obj.Namespace).Get(vm_obj.Name, &k8smetav1.GetOptions{})
-				Expect(err).ToNot(HaveOccurred())
-				return vmi.Status.Phase == v1.Running
-			}, 120*time.Second, 1*time.Second).Should(BeTrue())
-
-			generateHelloWorldServer(vmi, virtClient, testPort, "tcp")
-		})
-
-		Context("Verify the exposed service is functional before and after VM restart.", func() {
-			It("Connect to exposed ClusterIP service.", func() {
 				By("Getting back the service's allocated cluster IP.")
 				svc, err := virtClient.CoreV1().Services(vm_obj.Namespace).Get(serviceName, k8smetav1.GetOptions{})
 				Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


yossisegev:
ci please test

yossisegev:
ci please test

tests/expose_test.go
```diff
@@ -409,4 +410,117 @@ var _ = Describe("Expose", func() {
...
+				vmi, err = virtClient.VirtualMachineInstance(vm_obj.Namespace).Get(vm_obj.Name, &k8smetav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				return vmi.Status.Phase == v1.Running
+			}, 120*time.Second, 1*time.Second).Should(BeTrue())
+
+			generateHelloWorldServer(vmi, virtClient, testPort, "tcp")
+		})
+
+		Context("Verify the exposed service is functional before and after VM restart.", func() {
+			It("Connect to exposed ClusterIP service.", func() {
```
myakove: This should be one case, you should have one 'It' and everything should be 'By'.
'It' order is not generated .

tests/expose_test.go
```diff
@@ -407,6 +407,72 @@ var _ = Describe("Expose", func() {
...
+				Eventually(func() bool {
+					vm_obj, err = virtClient.VirtualMachine(vm_obj.Namespace).Get(vm_obj.Name, &k8smetav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred())
+					return vm_obj.Status.Ready
+				}, 120*time.Second, 1*time.Second).Should(BeTrue())
+
+				By("Getting the running VMI.")
+				var vmi *v1.VirtualMachineInstance
+				Eventually(func() bool {
+					vmi, err = virtClient.VirtualMachineInstance(vm_obj.Namespace).Get(vm_obj.Name, &k8smetav1.GetOptions{})
```
davidvossel: There's a race condition here. 

I think it's possible for you to be retrieving the old VMI here before the restart.

You should retrieve the vmi object from the first run and save the vmi.UUID in a variable. Then in this loop you should wait until you get a vmi object with a different UUID from the first vmi.

When you issue the "restart" command, behind the scenes the VMI is just being deleted. The VM controller then spins up a new VMI to replace the old VMI that was shut down.  All this happens asynchronously, so you'll have to wait for the UUID of the VMI to change before you can be sure the restart has occurred on the backend.

yossisegev: Great point, @davidvossel !
I suspect this race condition failed my CI on the previous commits.
I now handle this race, but differently than what you suggested:
 - Stop the VMI.
 - Verify the VMI is stopped.
 - Start the VMI back.
 - Retrieve the new VMI, hence the new UUID.
Thanks!

tests/expose_test.go
```diff
@@ -386,7 +386,7 @@ var _ = Describe("Expose", func() {
 			generateHelloWorldServer(vmi, virtClient, testPort, "tcp")
 		})
 
-		Context("Expose ClusterIP service", func() {
+		Context("Expose ClusterIP service on a VM.", func() {
```
myakove: 'on a VM.' i guess it should be 'to VM'

yossisegev: OK...

yossisegev:
ci please test

tests/expose_test.go
```diff
@@ -407,6 +407,77 @@ var _ = Describe("Expose", func() {
...
+				By("Stopping the running VM.")
+				virtctl := tests.NewRepeatableVirtctlCommand("stop", "--namespace", vm.Namespace, vm.Name)
+				err = virtctl()
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Verifying the VMI is not running.")
+				var vmi *v1.VirtualMachineInstance
+				Eventually(func() bool {
+					vmi, err = virtClient.VirtualMachineInstance(vm_obj.Namespace).Get(vm_obj.Name, &k8smetav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
yossisegev: Accepted.
I now verify both that VMI is in running state and holds new UID.

tests/expose_test.go
```diff
@@ -407,6 +407,77 @@ var _ = Describe("Expose", func() {
...
+
+				By("Starting the VM.")
+				virtctl = tests.NewRepeatableVirtctlCommand("start", "--namespace", vm.Namespace, vm.Name)
+				err = virtctl()
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Verifying the VMI is running.")
+				Eventually(func() bool {
+					vmi, err = virtClient.VirtualMachineInstance(vm_obj.Namespace).Get(vm_obj.Name, &k8smetav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
yossisegev: I now added the verification of new UID.
Ignoring the "NotFound" error (and any other error) is done by waiting in the "Eventually" block, until the grace timeout (120 seconds) elapses.

tests/expose_test.go
```diff
@@ -407,6 +407,77 @@ var _ = Describe("Expose", func() {
...
+					return vmi.Status.Phase == v1.Running
+				}, 120*time.Second, 1*time.Second).Should(BeTrue())
+
+				By("Creating a TCP server on the VM.")
+				generateHelloWorldServer(vmi, virtClient, testPort, "tcp")
+
+				By("Repeating the sequence as prior to restarting the VM: Connect to exposed ClusterIP service.")
+				By("Getting back the service's allocated cluster IP.")
+				svc, err = virtClient.CoreV1().Services(vm_obj.Namespace).Get(serviceName, k8smetav1.GetOptions{})
```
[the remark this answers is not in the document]
yossisegev: Agreed.
I mistakenly assumed the exposed service is also restarted, along with the VMI, and therefore should also be retrieved again. That doesn't make sense, however, because it's actually a cluster service.

tests/expose_test.go
```diff
@@ -408,7 +408,7 @@ var _ = Describe("Expose", func() {
 				Eventually(getStatus, 60, 1).Should(Equal(k8sv1.PodSucceeded))
 			})
 
-			It("Should verify the exposed service is functional before and after VM restart.", func() {
+			FIt("[test_id:345]Should verify the exposed service is functional before and after VM restart.", func() {
```
myakove: I think it should be 'It'. @nellyc ?

tests/expose_test.go
```diff
@@ -342,7 +342,7 @@ var _ = Describe("Expose", func() {
 		})
 	})
 
-	Context("Expose service on an VM", func() {
+	Context("[rfe_id:253][crit:high][vendor:cnv-qe@redhat.com][level:component]Expose service on an VM", func() {
```
myakove: to VM

tests/expose_test.go
```diff
@@ -430,40 +430,30 @@ var _ = Describe("Expose", func() {
 
 				Eventually(getStatus, 60, 1).Should(Equal(k8sv1.PodSucceeded))
 
-				By("Stopping the running VM.")
-				virtctl := tests.NewRepeatableVirtctlCommand("stop", "--namespace", vm.Namespace, vm.Name)
-				err = virtctl()
-				Expect(err).ToNot(HaveOccurred())
-
-				By("Verifying the VMI is not running.")
+				By("Retrieving the current VMI UID, to be compared with the new UID after restart.")
```
myakove: I don't think this should be a 'By'. 'By' is test step. this is just an internal operation that you need to do.

yossisegev:
ci please test

myakove:
@yossisegev please resolve the conflicts.

tests/expose_test.go
```diff
@@ -342,7 +342,7 @@ var _ = Describe("Expose", func() {
 		})
 	})
 
-	Context("Expose service on an VM", func() {
+	Context("[rfe_id:253][crit:high][vendor:cnv-qe@redhat.com][level:component]Expose service to an VM", func() {
```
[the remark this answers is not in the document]
yossisegev: Done.

tests/expose_test.go
```diff
@@ -407,6 +407,67 @@ var _ = Describe("Expose", func() {
...
+					return pod.Status.Phase
+				}
+
+				Eventually(getStatus, 60, 1).Should(Equal(k8sv1.PodSucceeded))
+
+				// Retrieve the current VMI UID, to be compared with the new UID after restart.
+				var vmi *v1.VirtualMachineInstance
+				vmi, err = virtClient.VirtualMachineInstance(vm_obj.Namespace).Get(vm_obj.Name, &k8smetav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				vmi_uid_before_restart := vmi.GetObjectMeta().GetUID()
```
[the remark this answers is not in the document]
yossisegev: Your OCD is my command ;) ....
Done.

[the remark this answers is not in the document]
slintes: you misspelled "convention"... ;)
https://golang.org/doc/effective_go.html#mixed-caps

yossisegev:
@cynepco3hahue As usual - after I resolved my authentic failures in CI, it now still fails because of stuff which is not related to me.
Can you please merge?
Thank you!

tests/expose_test.go
```diff
@@ -407,6 +407,67 @@ var _ = Describe("Expose", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				serviceIP := svc.Spec.ClusterIP
+
+				By("Starting a pod which tries to reach the VMI via ClusterIP.")
+				job := tests.NewHelloWorldJob(serviceIP, servicePort)
+				job, err = virtClient.CoreV1().Pods(vmObj.Namespace).Create(job)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Waiting for the pod to report a successful connection attempt.")
+				getStatus := func() k8sv1.PodPhase {
```
SchSeba: Can you please create a  function for this we call it 3 times in the code

something like 
```golang
func getStatus(job k8sv1.Job) {
pod, err := virtClient.CoreV1().Pods(job.Namespace).Get(job.Name, k8smetav1.GetOptions{})
					return pod.Status.Phase
				}

 				Eventually(getStatus, 60, 1).Should(Equal(k8sv1.PodSucceeded))
}
```

yossisegev: @SchSeba I already merged this PR.
Good point, though. I might implement it on another PR.

PR #1941, opened 2019-01-15
Degrade sriov virt-launcher to unprivileged

### LINKED ISSUE #1784: SR-IOV network attachment grants privileged=true mode to virt-launcher
/kind bug

**What happened**: when a VMI is attached to a sriov multus network, its virt-launcher is started with privileged=true. It was needed before when SR-IOV DP was not allocating necessary /dev/vfio/<NUM> device nodes to containers, but now that it does, we should move away from the practice. This means we will need to register (or instruct admins to) devices with vfio subsystem before SR-IOV DP up; remove privileged=true; perhaps also cleaning up some hostMounts added to launcher pod to facilitate vfio registration by libvirt; and finally, making sure libvirtd correctly handles the allocated vfio device (which means 1. it doesn't crash and 2. it doesn't unregister vfio device on exit).

**What you expected to happen**: launcher pod should not grant privileged=true to user.

**How to reproduce it (as minimally and precisely as possible)**: create a VMI attached to sriov multus network. The virt-launcher pod is started with privileged=true.

**Anything else we need to know?**: SR-IOV is an experimental feature and is hidden behind a feature gate.
**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

booxter:
**What this PR does / why we need it**: it removes the need to run SR-IOV attached VMs as privileged pods.

Fixes #1784

**Release note**:

```release-note
SR-IOV VMIs no longer run inside a privileged pod. Note: it is now required that all SR-IOV VFs that can be allocated by SR-IOV device plugin are pre-registered with vfio subsystem on deployment.
```

### DIFF
--- a/pkg/virt-controller/services/template_test.go
+++ b/pkg/virt-controller/services/template_test.go
@@ -1110,7 +1110,7 @@ var _ = Describe("Template", func() {
 		})
 
 		Context("with sriov interface", func() {
-			It("should run privileged", func() {
+			It("should not run privileged", func() {
 				sriovInterface := v1.InterfaceSRIOV{}
 				domain := v1.DomainSpec{}
 				domain.Devices.Interfaces = []v1.Interface{{Name: "testnet", InterfaceBindingMethod: v1.InterfaceBindingMethod{SRIOV: &sriovInterface}}}
@@ -1125,7 +1125,7 @@ var _ = Describe("Template", func() {
 				Expect(err).ToNot(HaveOccurred())
 
 				Expect(len(pod.Spec.Containers)).To(Equal(1))
-				Expect(*pod.Spec.Containers[0].SecurityContext.Privileged).To(Equal(true))
+				Expect(*pod.Spec.Containers[0].SecurityContext.Privileged).To(Equal(false))
 			})
 			It("should mount pci related host directories", func() {
 				sriovInterface := v1.InterfaceSRIOV{}
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -56,6 +56,7 @@ const GenieNetworksAnnotation = "cni"
 
 const CAP_NET_ADMIN = "NET_ADMIN"
 const CAP_SYS_NICE = "SYS_NICE"
+const CAP_SYS_RESOURCE = "SYS_RESOURCE"
 
 // LibvirtStartupDelay is added to custom liveness and readiness probes initial delay value.
 // Libvirt needs roughly 10 seconds to start.
@@ -315,7 +316,6 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 	var volumes []k8sv1.Volume
 	var volumeDevices []k8sv1.VolumeDevice
 	var userId int64 = 0
-	// Privileged mode is disabled by default.
 	var privileged bool = false
 	var volumeMounts []k8sv1.VolumeMount
 	var imagePullSecrets []k8sv1.LocalObjectReference
@@ -384,14 +384,6 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 				},
 			},
 		})
-
-		// todo: revisit when SR-IOV DP registers /dev/vfio/NN with pod
-		// device group:
-		// https://github.com/intel/sriov-network-device-plugin/pull/26
-		//
-		// Run virt-launcher compute container privileged to allow qemu
-		// to open /dev/vfio/NN for PCI passthrough
-		privileged = true
 	}
 
 	serviceAccountName := ""
@@ -935,6 +927,12 @@ func getRequiredCapabilities(vmi *v1.VirtualMachineInstance) []k8sv1.Capability 
 	}
 	// add a CAP_SYS_NICE capability to allow setting cpu affinity
 	res = append(res, CAP_SYS_NICE)
+
+	if isSRIOVVmi(vmi) {
+		// this capability is needed for libvirt to be able to change ulimits for device passthrough:
+		// "error : cannot limit locked memory to 2098200576: Operation not permitted"
+		res = append(res, CAP_SYS_RESOURCE)
+	}
 	return res
 }
 


booxter:
/hold

This requires a multus piece, as well as a separate 'noop' CNI plugin.

greg-bock:
Should we do something similar to what kata does and handle the unbinding/rebinding in a CNI or other host based process?

greg-bock:
kata + sriov device plugin breaks on VFIO type (it tries to bind the pci device twice). This will break users trying to use kubevirt and kata with SR-IOV on the same cluster. 

You also lose [health checking](https://github.com/intel/sriov-network-device-plugin/blob/70cc5eb1f8b7d65907a96a15fc6632e5d7e69766/pkg/utils/utils.go#L122) when the device plugin is run as vfio.

booxter:
@greg-bock about using with kata, I think the way to go about it right now is having two separate pools, one for ```vfio``` and another for regular ```netlink``` devices.

Health checking is a separate issue. I don't think the way we use it with kubevirt right now is compatible with it anyway because we still unbind the device, just after the ```netlink``` interface is moved inside the pod. But you still loose the ```netlink``` device once ```libvirt``` configures the device for passthrough.

booxter:
As to CNI handling VFIO (un)binding, I think the main problem would be lack of direct communication between device plugins and CNI plugins. DP owns information about the type of the device, so it would need to somehow pass this information into CNI plugin for it to act.

greg-bock:
> @greg-bock about using with kata, I think the way to go about it right now is having two separate pools, one for `vfio` and another for regular `netlink` devices.

We only have one PF to split up, I haven't looked/tried to see if two pools can point to the same real device. 

> Health checking is a separate issue. I don't think the way we use it with kubevirt right now is compatible with it anyway because we still unbind the device, just after the `netlink` interface is moved inside the pod. But you still loose the `netlink` device once `libvirt` configures the device for passthrough.

With the sriov device plugin in netdevice mode, availability of requested resources does influence scheduling in k8s, which in turn influences where the launcher pod will be placed no?

greg-bock:
> As to CNI handling VFIO (un)binding, I think the main problem would be lack of direct communication between device plugins and CNI plugins. DP owns information about the type of the device, so it would need to somehow pass this information into CNI plugin for it to act.

Kata just looks at the parent device's driver and rebinds to it on tear down when in netdevice mode.

booxter:
@greg-bock how does CNI know which mode is it? In SR-IOV DP case, the DP controls / hosts the information about the intended mode. There seems to be no clear path to forward this information from DP through Multus to CNI. Or is this information relayed through network CRD?

As for reusing the same card for two pools with different type, indeed it's probably not possible right now. We should perhaps make DP support this split upstream so that we can keep kubevirt the same. This should be useful for others, not just us. (If nothing else, for testing purposes if one wants to check how multiple SR-IOV networks behave, but on a single card node.)

Thanks for the feedback, I will try to see how we can not break things.

greg-bock:
> how does CNI know which mode is it?

The vfio type adds path where netdevice just passes the pci id if I remember correctly. We may have self-inflicted issues during our initial testing of kata with VFIO as we didn't unbind the device from the normal driver first. In netdevice mode kata does what kubevirt currently does, although I think kubevirt lets libvirt do the heavy lifting, and moves the device from host driver to vfio driver.

Health checking by netdevice for PF and VF are still very important to us.

cluster/k8s-1.13.0-sriov/manifests/multus.yaml
```diff
@@ -119,7 +119,8 @@ spec:
       serviceAccountName: multus
       containers:
         - name: kube-multus
-          image: nfvpe/multus:latest
+          image: nfvpe/multus:snapshot
+          command: ["/entrypoint.sh", "--multus-conf-file=auto"]
```
SchSeba: Nice!

booxter:
@greg-bock AFAIU there can't be a netlink device if you use vfio, so we can't health check using netlink. But we probably can do by other means.

sseetharaman6:
I would be interested in knowing how IPAM is handled by the no-op plugin (or whichever CNI plugin) and how this data is eventually passed on to the SRIOV interfaces on the VM. From conversation with @booxter , I understand that currently cloud-init PR gathers info from the netlink dev in the virt-launcher pod.

greg-bock:
> @greg-bock AFAIU there can't be a netlink device if you use vfio, so we can't health check using netlink. But we probably can do by other means.

Kata unbinds the device from the driver it is on and rebinds it to vfio during the start process and reverses it on teardown: [vfio.go](https://github.com/kata-containers/runtime/blob/780cd5f7b9d3c76522e96f83905b1ef224f8574d/virtcontainers/device/drivers/vfio.go). You get the benefits of health checking and the security of vfio plus other things can use VFs from the pool in whatever manner they choose.

booxter:
@greg-bock in kata, you don't have health checks while a device is in VFIO mode, do you?

booxter:
ci test please

greg-bock:
> @greg-bock in kata, you don't have health checks while a device is in VFIO mode, do you?

No, this is for the SRIOV DP available resource.

booxter:
@greg-bock there is a discussion going in ```sriov-cni``` repo: https://github.com/intel/sriov-cni/issues/37 as to why kata may have different requirements that are not compatible with Intel vision for the plugin. In case of Intel DP, the plugin manages whether to expose /dev/vfio/NN to a pod, while in Kata you need to pass it manually for consumption: https://github.com/kata-containers/runtime/tree/master/virtcontainers#how-to-pass-a-device-using-vfio-passthrough

Could you please chime in there?

booxter:
@greg-bock as to health checks, considering that DP always claims no change in health status for VFIO: https://github.com/intel/sriov-network-device-plugin/blob/db45dc4a722daf895d8c24fc91d99dde86c10c57/pkg/resources/vfioPool.go#L100 I don't think it works now anyway; so this PR shouldn't be a problem per se. My understanding is that health checks belong to device plugin and have little to do with kubevirt. If that's the case, perhaps you can create a new issue for the device plugin to implement health checks for userspace.

greg-bock:
> My understanding is that health checks belong to device plugin and have little to do with kubevirt.

True enough, but scheduling on a VM with VFIO where there is a problem (e.g. link down) isn't very useful.

greg-bock:
> > My understanding is that health checks belong to device plugin and have little to do with kubevirt.
> 
> True enough, but scheduling on a VM with VFIO where there is a problem (e.g. link down) isn't very useful.

Health checking on the parent PF might be enough.

greg-bock:
> Could you please chime in there?

It looks like they are mainly focused on a DPDK use case in that thread, but there is some overlap like adding mac and ipam details to IPAM responses for use further of the plugin chain.

booxter:
@greg-bock discussion in ```sriov-cni``` issue is a bit confusing because it touches on two separate issues. One, original, is how we pass IPAM information in userspace mode; and another is about whether we could move to a model where CNI handles userspace mode binding as needed. I was referring to the latter. It seems design decisions were made that would block us from reusing the same pool for different workloads. As the latest comment suggests, perhaps it can be handled on cluster level by having a top level orchestrator rebalancing pools between types based on utilization, which of course introduces its own complexity.

For health checks, PF check should be enough, we just need to implement it in DP. (Please post an issue there if not already.)

greg-bock:
Yeah, that's why it seemed at first to be simpler to just have a cni plugin from kubevirt move a net device to the vfio driver, and back on tear down, to have interoperability with a functional working kata setup. I'm not as much worried about configuring any net info on the vfio device in the launcher as I would care about being able to glean that data from the vmi object.

We've been thinking about proposing a free form array of strings type on the cni NetConf type to the network sig so plugins could pass things up and down the chain without having to redefine cni types for plugin extension/development.

booxter:
@greg-bock if I follow your idea with free form array of string, it would look as follows:

1. all PFs on a node would be registered with the same allocation pool (assuming they have the same uplink).
2. there will be several SR-IOV network attachment definitions, one per type used. They will all refer to the same resourceName. Each of them will have a tag that would specify the device subtype needed (vfio, netdevice, uio, ...)
3. When pod is attached to one of the networks, multus will extract the subtype from CRD and pass it via the array of strings into DP. (How? The current device plugin API AllocateRequest is an array of device ID strings and nothing else.)
4. DP receives the device requests and corresponding subtypes, allocates the devices; it also handles userspace binding if needed.
5. SR-IOV CNI does nothing if subtype is one of userspace options.

Is my reading correct? If so, how would you fill the gap in step 3?

booxter:
To add some more prior history, it's not the first time when a suggestion to extend device plugin API to serve network devices (and specifically userspace bindings) comes up. For example, here is an old (now closed) issue in kubernetes suggesting DP may need to be aware of namespace of the pod to handle binding itself without relying on CNI at all: https://github.com/kubernetes/kubernetes/issues/63048

greg-bock:
> Is my reading correct? If so, how would you fill the gap in step 3?

I meant only for passing around info within the CNI chain, I haven't dug into DP at all to be honest.

> 4\. DP receives the device requests and corresponding subtypes, allocates the devices; it also handles userspace binding if needed.

Why not have the CNI handle the driver binding/unbinding? For what I initially proposed this doesn't even require changing the SRIOV DP. If it was vfio leave it alone; if not rebind the driver to vfio (and rebind to host on pod teardown). 

You can tell if a device returned from the DP was netdevice or not. [netdevice](https://github.com/intel/sriov-network-device-plugin/blob/acaa87c7261b05a8813f8c8d15332297b3844969/pkg/resources/netDevicePool.go#L76) 
 has an empty DeviceSpecs while [VFIO](https://github.com/intel/sriov-network-device-plugin/blob/acaa87c7261b05a8813f8c8d15332297b3844969/pkg/resources/vfioPool.go#L75) populates information such as HostPath.

I haven't looked back through all versions to see when things were added, but a quick glance at master shows: a map of envs, Mount, and DeviceSpec objects in the [AllocateResponse](https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/pkg/kubelet/apis/deviceplugin/v1alpha/api.proto#L92)

booxter:
@greg-bock the reason why we can't allow CNI to bind is because then /dev/vfio/NN is not registered with container, which won't work if the container is not privileged. (Which is the goal of the PR.)

greg-bock:
Not registered in what way?

booxter:
@greg-bock see here for docker: https://docs.docker.com/engine/reference/commandline/run/#add-host-device-to-container---device

Basically, for docker to allow access to a device for an unprivileged container, runtime should be informed on container started that a particular device should be available to the container, otherwise you get permission denied errors. This is what happens when a DP allocates a device - it passes the paths to each device to be allocated to a container (in our case it's one of /dev/vfio/NN), and kubelet passes this information into CRI.

To add to that, one cannot register a non-existing device (meaning, one cannot postpone / offload creation of /dev/vfio/NN to CNI plugin).

sseetharaman6:
@booxter how would this play out for pure PCI Passthrough of the NIC? binding to vfio ahead of time may not be optimal in this scenario right?

booxter:
@sseetharaman6 what's the exact scenario you describe? I am not sure I follow what you describe.

sseetharaman6:
The use case I am thinking of is  setting sriovMode: false on the Intel SRIOV DP and using the NIC in Passthrough mode (expose the root device to the VM). Would pre-registering the PF to VFIO driver ahead of time render the entire NIC unusable by the host driver if necessary? Would this be a concern at all?

booxter:
@sseetharaman6 I admit I am yet to try pfOnly mode. If you have capacity, perhaps you could try and report how it behaves.

fabiand:
Yup - I think some research would help. The goal is clear and needed.

fabiand:
Ping?

booxter:
@fabiand we need to complete CI (a separate PR; it actually finally passed yesterday, just need to clean it up now); then rebase this PR on top of CI PR to guarantee that this PR doesn't break anything. Code wise, this PR is mostly ready (minus the cleanup that we also need to do here).

booxter:
ci test please

booxter:
Finally got to validate the fix manually. In pod spec:

```
    securityContext:
      capabilities:
        add:
        - NET_ADMIN
        - SYS_NICE
        - SYS_RESOURCE
```

```
[root@zeus08 kubevirt]# ./cluster/kubectl.sh get vmis --all-namespaces                                                         
NAMESPACE   NAME        AGE   PHASE     IP             NODENAME
default     vmi-sriov   81s   Running   10.244.0.103   kubdev
```

I will also check SRIOV test suite later and report back.

booxter:
I've just validated that this works both with older versions of SR-IOV DP (the one that cluster-network-addons-operator pulls) as well as :latest image from upstream. So there should be no issues with merging it for larger ecosystem.

booxter:
It looks like CI hasn't "merged" the PR with latest master and fail with an issue that is already fixed in master. So I just rebased.

booxter:
I've checked whether SRIOV test suite still works with the change, and it does. The only caveat is that because the machine I have for SR-IOV work has Mellanox cards and not Intel, I needed an additional patch for the suite to make it load kernel drivers for VFs inside the guest to pass: https://github.com/kubevirt/kubevirt/pull/2347

While I needed the test change to make the suite pass on the machine I have, I don't think the test change blocks this PR because the suite never worked with Mellanox VFs. (I just didn't test it against non-Intel hosts before.)

booxter:
I've checked test failures, two jobs fail because of MAC address issue (there is a workaround in another PR somewhere here, and it's unrelated to this PR); there is also another job that failed in lots of migration tests which is also unrelated and a fluke. (Wasn't happening in a previous run for the same code.)

pkg/virt-controller/services/template_test.go
```diff
@@ -1125,7 +1125,7 @@ var _ = Describe("Template", func() {
 				Expect(err).ToNot(HaveOccurred())
 
 				Expect(len(pod.Spec.Containers)).To(Equal(1))
-				Expect(*pod.Spec.Containers[0].SecurityContext.Privileged).To(Equal(true))
+				Expect(*pod.Spec.Containers[0].SecurityContext.Privileged).To(Equal(false))
```
phoracek: There is BeFalse() in Gomega, maybe you could use that. Not a must though.

sseetharaman6:
@booxter - I tested SRIOV and PCI passthrough with Kubevirt v0.19.0  :  
1. As expected SRIOV by itself works but assigning IPs via the [cloud-init side car](https://github.com/stackpath/sriov-hook-sidecar) breaks
```
{"component":"cloudinit-hook-sidecar","level":"info","msg":"Hook's Info method has been called","pos":"cloudinit.go:44","timestamp":"2019-07-19T18:52:42.995496Z"}
--
{"component":"cloudinit-hook-sidecar","level":"info","msg":"Hook's PreCloudInitIso callback method has been called","pos":"cloudinit.go:63","timestamp":"2019-07-19T18:52:47.117399Z"}
{"component":"cloudinit-hook-sidecar","level":"info","msg":"starting discovery for: net1","pos":"sriov-discovery.go:156","timestamp":"2019-07-19T18:52:47.118464Z"}
{"component":"cloudinit-hook-sidecar","level":"error","msg":"failed to get a link for interface: net1","pos":"sriov-discovery.go:167","reason":"Link not found","timestamp":"2019-07-19T18:52:47.118652Z"}
{"component":"cloudinit-hook-sidecar","level":"error","msg":"failed to get SR-IOV network details for net1","pos":"sriov-discovery.go:140","reason":"Link not found","timestamp":"2019-07-19T18:52:47.118710Z"}
{"component":"cloudinit-hook-sidecar","level":"error","msg":"cloudInitDiscoverNetworkData Failed","pos":"cloudinit.go:90","reason":"Link not found","timestamp":"2019-07-19T18:52:47.118736Z"}
panic: Link not found
```
2. For PCI passthrough, because we pre-register the PF to use `vfio-pci` driver,  the `discoverHostDevices` routine in Intel SRIOV DP, can't read the device and query the PF's network information (interface name etc) and therefore can't advertise the resource. So PCI passthrough is broken with v0.19.0
```discoverDevices() error adding new device: "open /sys/bus/pci/devices/0000:81:00.1/net: no such file or directory"```
--

booxter:
For the record, I assume that both issues created - in DP and CNI - will track the fixes for the degradation mentioned going forward. The sidecar issue is probably not to fix once https://github.com/intel/sriov-cni/issues/37 is fixed.

leyao-daily:
> @greg-bock see here for docker: https://docs.docker.com/engine/reference/commandline/run/#add-host-device-to-container---device
> 
> Basically, for docker to allow access to a device for an unprivileged container, runtime should be informed on container started that a particular device should be available to the container, otherwise you get permission denied errors. This is what happens when a DP allocates a device - it passes the paths to each device to be allocated to a container (in our case it's one of /dev/vfio/NN), and kubelet passes this information into CRI.
> 
> To add to that, one cannot register a non-existing device (meaning, one cannot postpone / offload creation of /dev/vfio/NN to CNI plugin).

I use kubevirt 0.22.0 and assign vf to a vm, it report permission denied. How to solve it.

PR #1956, opened 2019-01-19
Introduce NPWG as a new NetworkSource and provide cni annotation as JSON list

michaelhenkel:
This PR enables support for the TungstenFabric SDN Controller
Fixes #

**Release note**:

```release-note
This PR enables the usage of TungstenFabric as a SDN Controller.
```


fabiand:
Nice one!

To me this looks "good enough". I could imagine that the biggest issue is the missing functional CI, BUT - just like for i.e. genie - we need to cut the ropes somewhere and can not test any provider integration, thus ensuring that the annotations are set correctly, and connecting to the correct interfaces t me sounds good.

However, @michaelhenkel do you plan to run any integration tests on your side?

pkg/virt-controller/services/template_test.go
```diff
@@ -203,6 +203,32 @@ var _ = Describe("Template", func() {
...
+				vmi := v1.VirtualMachineInstance{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "testvmi",
+						Namespace: "default",
+						UID:       "1234",
+					},
+					Spec: v1.VirtualMachineInstanceSpec{
+						Domain: v1.DomainSpec{},
+						Networks: []v1.Network{
+							{Name: "{\"domain\":\"default-domain\", \"project\": \"k8s-default\",\"name\":\"bla\"}",
```
davidvossel: For this unit test, the 'Name' value should be different than the 'NetworkName' value. That ensures we don't accidentally inject the 'Name' value as the TungstenFabric's annotation value instead of NetworkName.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -974,6 +978,8 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 						prefix = "net"
 					} else if net.Genie != nil {
 						prefix = "eth"
+					} else if net.Tungstenfabric != nil {
+						prefix = "eth"
```
davidvossel: we need a test in converter_test.go to ensure the correct domain xml is generated when a TungstenFabric network is in use.

pkg/virt-controller/services/template_test.go
michaelhenkel: done

pkg/virt-launcher/virtwrap/api/converter.go
michaelhenkel: done:
https://github.com/kubevirt/kubevirt/pull/1956/commits/3865d166e9b73e6be4058db819f5658a4c18d911#diff-406f8a4e9665b5e52c190c5815dbacd2R1020

michaelhenkel:
> userguide

https://github.com/kubevirt/user-guide/pull/196

davidvossel:
> To me this is not very user-friendly, as it's easy to mess up quoting.

I can see two sides to this.

1. That json blob is actually what is used in the cni annotation directly. So by keeping it an opaque string value we're not being opinionated about the tungsten fabric cni api.  This means if the tungsten fabric cni api changes, nothing has to happen on our end to support the change.

2. If the tungsten fabric cni api is both mature and not expected to change much, then the structured/opinionated way of representing the values in our api would offer a better user experience.

michaelhenkel:
Problem is that Annotations are of type map[string]string and this is what we have to use on the POD level.
I want to present the same interface for POD and VMI configuration. Having two separate interfaces, one for POD and one for VMI, will confuse people. And as we cannot change the POD interface, we have to use the current interface for VMI as well.

davidvossel:
> Problem is that Annotations are of type map[string]string and this is what we have to use on the POD level. I want to present the same interface for POD and VMI configuration. Having two separate interfaces, one for POD and one for VMI, will confuse people. And as we cannot change the POD interface, we have to use the current interface for VMI as well.

For a pod, does this annotation always represent adding a single network interface to the pod, or might it be used to represent adding multiple connections to the pod using the tungsten cni plugin?

michaelhenkel:
> For a pod, does this annotation always represent adding a single network interface to the pod, or might it be used to represent adding multiple connections to the pod using the tungsten cni plugin?

For now we only support a single network interface per POD (and as such per VMI). Multi network interface support will follow.

davidvossel:
> For now we only support a single network interface per POD (and as such per VMI). Multi network interface support will follow.

When you all move to multi nic support, this will complicate the situation for kubevirt integration as you have it now.  That json blob you're passing into NetworkName is assumed to represent only a single nic. 

Take our multus cni integration for example. There can be multiple multus network interfaces defined on a VMI. Each one of those interfaces is mapped to a device on the domain. The multus annotation is generated by aggregating all of these defined network interfaces to build the annotation string.

Since the tungsten fabric integration isn't aggregating multiple networks into that annotation string, this current method wouldn't allow for multiple network interface support.

dankenigsberg:
> yes, this will be added to Tungsten Fabric CI

@michaelhenkel can you elaborate how? Stackpath.io have donated a host to kubevirt ci, so we can continuously test kubevirt on their hardware. Do you envisage something similar? Do you intend to watch every kubevirt PR and run your test suite on it?

michaelhenkel:
@dankenigsberg TungstenFabric CI will not test individual kubevirt PRs but it will test TungstenFabric SDN controller commits against the latest release of kubevirt. i.e. whenever a TungstenFabric SDN controller patch is committed, the CI will have one job to install TungstenFabric SDN with the latest k8s and kubevirt and do some basic connectivity tests.

dankenigsberg:
This makes sure that TF does not break compatibility; it does not ensure that kubevirt inadvertently break it. If this happens, we'd know about it only after we release a broken version.

davidvossel:
ok to test

davidvossel:
ci test please

pkg/virt-controller/services/template.go
```diff
@@ -988,10 +990,24 @@ func getCniInterfaceList(vmi *v1.VirtualMachineInstance) (ifaceListString string
...
+			ifaceList = append(ifaceList, network.Tungstenfabric.NetworkName)
+			if cniAnnotation == "" {
+				cniAnnotation = "k8s.v1.cni.cncf.io/networks"
+			}
 		}
 	}
 
-	ifaceListString = strings.Join(ifaceList, ",")
+	if tungstenfabric {
+		ifaceListString = fmt.Sprintf("[\n")
```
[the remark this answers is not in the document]
michaelhenkel: @davidvossel 
My concern is that the annotation can become a very long single line, especially when we enable more attributes for the interface, such as IP and MAC. With the new line the annotation stays readable in the POD definition.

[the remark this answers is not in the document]
michaelhenkel: My concern is that the annotation can become a very long single line, especially when we enable more attributes for the interface, such as IP and MAC. With the new line the annotation stays readable in the POD definition.

dankenigsberg:
@michaelhenkel I have a general question about this PR, it may be worthwhile to repeat it here.

What is the benefit in using TungstenFabric's multiplexing of interfaces over Multus's? Is it possible to set up Multus as k8s's CNI, and have it delegate to TungstenFabric? If it is, we can support TF with no new code in KubeVirt; and can add non-TF CNI to a Pod with TF. We only need to make sure that kubevirt-ansible installs multus properly.

pkg/virt-controller/services/template.go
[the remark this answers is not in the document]
davidvossel: alrighty, that works for me if it helps.

davidvossel:
> it does not ensure that kubevirt inadvertently break it. If this happens, we'd know about it only after we release a broken version.

@dankenigsberg That's true to a certain extent. However, we are validating through the unit tests that both the correct pod annotations for TF are generated and that in the VMI's pod that we're looking for the expected TF network interfaces. 

Those two unit tests at least provide some protection against us inadvertently breaking TF... it's not perfect though. We're in a similar place with sriov support too I believe.

dankenigsberg:
> +1 These changes look good to me.
> 
> @dankenigsberg @fabiand Do either of you have any reservations about introducing TF cni support?

yes, I do. I do not understand why we cannot use the existing `multus` support. I'd love if @michaelhenkel  answers that.

davidvossel:
> yes, I do. I do not understand why we cannot use the existing multus support. I'd love if @michaelhenkel answers that.

Sounds good. I've put this PR on hold until we resolve this discussion.

michaelhenkel:
> > +1 These changes look good to me.
> > @dankenigsberg @fabiand Do either of you have any reservations about introducing TF cni support?
> 
> yes, I do. I do not understand why we cannot use the existing `multus` support. I'd love if @michaelhenkel answers that.



> > Problem is that Annotations are of type map[string]string and this is what we have to use on the POD level.
> > I want to present the same interface for POD and VMI configuration. Having two separate interfaces, one for POD and one for VMI, will confuse people. And as we cannot change the POD interface, we have to use the current interface for VMI as well.
> 
> I'd say we already have a different interface, because we put it into the VMI spec and not in the VMI metadata.
> 
> To some degree this support is comparable to multus integration, there also some pod annotation is used, but on the VMI API we express this with a regular and validated API field.
> 
> I do understand that arbitrary blobs bring freedom, but they also have the drawback that we can not validate them, and thus raising errors early is impossible.
> 
> IOW right now I wonder why we are using a specified API for multus, but don't require it here.
> 
> And again: I'll all for having this in :)

@fabiand The Network Custom Resource Definition standard defines two ways to define two supported annotation formats: 
1. comma separated -> used by multus, limited to network name or domain/network name
2. JSON format -> provides the option to specify more information for the interface such as static IP, MAC address and more.

TF currently only supports the JSON format, comma separated format is added over the next couple of days. 
Having said that the question is if kubevirt is restrictive and only supports comma separated annotation format or if it is flexible and supports both.

michaelhenkel:
> > +1 These changes look good to me.
> > @dankenigsberg @fabiand Do either of you have any reservations about introducing TF cni support?
> 
> yes, I do. I do not understand why we cannot use the existing `multus` support. I'd love if @michaelhenkel answers that.

@dankenigsberg 
We didn't test through multus but I am sure we can make it work.
But it is yet another indirection with yet another layer of configuration which is fine for users planning to use multiple CNIs. However, I think for users using a single CNI it is more appealing to directly interface with the CNI they want to use.
We can certainly work on a multus integration as well but it should be an option and not a mandate.
We wanted to integrate with kubevirt and weren't aware that multus is the way to go.
If possible I'd first like to finish direct kubevirt implementation and then start exploring the multus way.

dankenigsberg:
> > > +1 These changes look good to me.
> > > @dankenigsberg @fabiand Do either of you have any reservations about introducing TF cni support?
> > 
> > 
> > yes, I do. I do not understand why we cannot use the existing `multus` support. I'd love if @michaelhenkel answers that.
> 
> @dankenigsberg
> We didn't test through multus but I am sure we can make it work.
> But it is yet another indirection with yet another layer of configuration which is fine for users planning to use multiple CNIs. However, I think for users using a single CNI it is more appealing to directly interface with the CNI they want to use.
> We can certainly work on a multus integration as well but it should be an option and not a mandate.
> We wanted to integrate with kubevirt and weren't aware that multus is the way to go.
> If possible I'd first like to finish direct kubevirt implementation and then start exploring the multus way.
> Also, there are users having running kubernetes clusters with TF as the CNI and want to add kubevirt without changing the CNI of the running k8s cluster.

I'll try to explain why I worry about this approach - and do correct me if I wrong. It seems TF complies with the DF standard just like Multus; it seems that the code that you are adding here a a dup of the Multus handling code - up to `net`->`eth` change.

May I dare ask what happens if you make the `net`->`eth` change, and run KubeVirt with `multus` on TF NetAttDefs? Wouldn't it just work? If so, we can have the interface prefix as a system-wide configurable, or as dcbw suggested, have KubeVirt choose the interface names and not depend on individual implementation.

I worry that your currently-suggested code is not tested continuously against TF, and thus easily broken by unintending kubevirt devels. If the idea in the former paragraph is viable, we'd have the best of all worlds: support via multus for multiple different CNIs, slim support via TF only, and no untested code path.

BTW, note that currently kubevirt-ansible deploys multus by default. You'd probably need to do something about it if Multus+TF is not supported.

michaelhenkel:
> > > > +1 These changes look good to me.
> > > > @dankenigsberg @fabiand Do either of you have any reservations about introducing TF cni support?
> > > 
> > > 
> > > yes, I do. I do not understand why we cannot use the existing `multus` support. I'd love if @michaelhenkel answers that.
> > 
> > 
> > @dankenigsberg
> > We didn't test through multus but I am sure we can make it work.
> > But it is yet another indirection with yet another layer of configuration which is fine for users planning to use multiple CNIs. However, I think for users using a single CNI it is more appealing to directly interface with the CNI they want to use.
> > We can certainly work on a multus integration as well but it should be an option and not a mandate.
> > We wanted to integrate with kubevirt and weren't aware that multus is the way to go.
> > If possible I'd first like to finish direct kubevirt implementation and then start exploring the multus way.
> > Also, there are users having running kubernetes clusters with TF as the CNI and want to add kubevirt without changing the CNI of the running k8s cluster.
> 
> I'll try to explain why I worry about this approach - and do correct me if I wrong. It seems TF complies with the DF standard just like Multus; it seems that the code that you are adding here a a dup of the Multus handling code - up to `net`->`eth` change.
> 
Yes, there is some code duplication and I reduced it further in the latest commit. But as outlined above, besides the eth vs. net nic name, multus currently only supports the comma separated network name annotation, whereas TF only support the JSON list format. For TF I need to create the JSON list and there is currently no code doing this.
> May I dare ask what happens if you make the `net`->`eth` change, and run KubeVirt with `multus` on TF NetAttDefs? Wouldn't it just work? If so, we can have the interface prefix as a system-wide configurable, or as dcbw suggested, have KubeVirt choose the interface names and not depend on individual implementation.
> 
We cannot use multus in the NetworkAttachmentDefinitions as the TF controller watches for the opencontrail.org prefix:
```
---
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: test4
  namespace: default
  annotations:
    "opencontrail.org/cidr" : "4.0.0.0/24"
    "opencontrail.org/ip_fabric_snat" : "true"
    "opencontrail.org/ip_fabric_forwarding" : "true"
spec:
  config: '{
    “cniVersion”: “0.3.0”,
    "type": "contrail-k8s-cni"
}'
```
We could potentially use multus as the network provider in the VMI network definition but this will fail as well until TF supports comma separated annotations or multus JSON list. Ideally TF and multus support both. And yes, I agree, an interface name agnostic implementation is desirable. 
> I worry that your currently-suggested code is not tested continuously against TF, and thus easily broken by unintending kubevirt devels. If the idea in the former paragraph is viable, we'd have the best of all worlds: support via multus for multiple different CNIs, slim support via TF only, and no untested code path.
> 
With the UTs we should be able to ensure that at least correct POD annotations and virsh definitions are created.
Furthermore I think users should configure the provider they intent to use. If they want multi CNI -> multus, in case they only want TF -> tungstenfabric. It would be awkward to configure multus if they will only want TF.
> BTW, note that currently kubevirt-ansible deploys multus by default. You'd probably need to do something about it if Multus+TF is not supported.
Agreed.

michaelhenkel:
> It seems that you are building a JSON list with only the "name" argument. If you build it with "interface" argument, too, and set it to "netXXX", you can join another part of the multus/TF code paths (assuming that TF respects the "interface" argument, as the de-facto standard requests).

@dankenigsberg can you point me to the interface arg for the pod annotations? I was checking 
https://docs.google.com/document/d/1Ny03h6IDVy_e_vmElOqR7UdTPAG_RNydhVE1Kx54kFQ/edit#
4.1.2.1 JSON List Format Key Definitions and there is no interface parameter. The only required parameter is "name" - which we are using.
Also, the interface list creation in kubevirt for multus interfaces only creates a comma separated list:
https://github.com/kubevirt/kubevirt/blob/c11899605f7818829ab8536847f8ab36fa23adda/pkg/virt-controller/services/template.go#L976
https://github.com/kubevirt/kubevirt/blob/c11899605f7818829ab8536847f8ab36fa23adda/pkg/virt-controller/services/template.go#L984
https://github.com/kubevirt/kubevirt/blob/c11899605f7818829ab8536847f8ab36fa23adda/pkg/virt-controller/services/template.go#L1009
With this code a valid JSON list cannot be passed to the POD annotation. 
But... I like the idea of generalizing the provider and agree that there shouldn't be CNI specific implementation in kubevirt.
In order to achieve that we need to do the following:

1. as you said s/multus/npwgv1/
Do you know if someone is already looking into this?
2. add support for JSON list based POD annotations
I already spent some thoughts on it. Difficulty is to detect if the VMI uses a simple networkName (which results in a comma separated annotation) or a JSON (which results in a JSON list annotation) as both are of type string. So, we have to parse the string to see if it starts with a curly bracket or not. In case it does -> JSON list, otherwise comma separated.

michaelhenkel:
> Please note that kubevirt has to keep supporting `multus` string, so we'd need to canonize to `npwgv1` early on.

what about instead of replacing multus with npwgv1 I replace tungstenfabric with npwgv1. At a later point we can then depreciate the multus string?

dankenigsberg:
> what about instead of replacing multus with npwgv1 I replace tungstenfabric with npwgv1. At a later point we can then depreciate the multus string?

I love removing code, and cringe when it is needlessly added. By replacing `multus`, you immediately earn its CI (both in kubevirt and in kubevirt-ansible's end-to-end tests)

michaelhenkel:
> > what about instead of replacing multus with npwgv1 I replace tungstenfabric with npwgv1. At a later point we can then depreciate the multus string?
> 
> I love removing code, and cringe when it is needlessly added. By replacing `multus`, you immediately earn its CI (both in kubevirt and in kubevirt-ansible's end-to-end tests)

In order to maintain backwards compatibility we cannot replace it right away. We need to transition and have npwgv1 and multus in parallel for a while. So, at least in the schema we need both and also in the UTs.

dankenigsberg:
> So, at least in the schema we need both and also in the UTs.

Yes, that is what I attempted to mean by "canonize". We should behave alike with `multus` or `npwgv1`. Right after (or during?) input validation, input should be changed to `npwgv1`, which the bulk of the code should use exclusively.

davidvossel:
@michaelhenkel just checking in here. Do you have everything you need to move forward now?

michaelhenkel:
Hi David,

yes, besides the time... I hope I can spend some cycles next week.



David Vossel <notifications@github.com> schrieb am Fr. 15. Feb. 2019 um
23:59:

michaelhenkel:
@davidvossel @dankenigsberg 
I started to look into turning the Multus interface into a generic Npwgv1.
The thing I struggle with is the pod interface kernel name. Multus sets it to net* whilst other solutions will expect a different prefix. 
Just configuring the CNI to use net* in all SDN controllers is not a good solution. I'd prefer a solution where the prefix can be defined in the the VMI definition. It can be 

  networks:
  - name: net1
     npwgv1:
       networkName: test3
       interfacePrefix: eth

or more specific

  networks:
  - name: net1
     npwgv1:
       networkName: test3
       interfaceName: eth0

What do you think?

fabiand:
I'd not ask the user to specify it, as it puts the bruden on the user.

Could we i.e. try a list of well known names or somehow identify the
provider and then choose the correct prefiy?

Just my 2ct

dankenigsberg:
> configuring the CNI to use net* in all SDN controllers is not a good solution.

@michaelhenkel  why? why should KubeVirt deal with multiple schema? A reasonable answer can be standard implementation X does not support this feature. My own answer to this answer would be - let us fix X, it should support the de-facto standard properly.

michaelhenkel:
@fabiand 
I agree, interfacePrefix is not very intuitive. We can pimp the POD annotation to always specify the interface which is then correctly picked up by the CNI delegate plugin. Today we create a comma separated list with the network names:
https://github.com/kubevirt/kubevirt/blob/4b2cb00c2a1954d840310c6ae54b4f91511cf9ef/pkg/virt-controller/services/template.go#L984
This will have to change to a JSON list in the format of
[
      { "name": "network1", "interface": "net1"},
      { "name": "network2", "interface": "net2"}
]
This will instruct CNI explicitly to create POD interfaces with the names "net1" and "net2". Prefix "net" is static and index based on the number of networks.
This will be inline with https://docs.google.com/document/d/1Ny03h6IDVy_e_vmElOqR7UdTPAG_RNydhVE1Kx54kFQ/edit#
section 4.1.2.1.5 "interface"
@dankenigsberg 
The above should address your comment. PODs will always have netX prefixed interfaces. As said, we have to change POD network annotation from list to JSON, otherwise we cannot transport the interface name to the CNI.

Are we ok with that?

dankenigsberg:
> Are we ok with that?

I am completely OK with moving to JSON list.

ovirt-infra:

michaelhenkel:
@dankenigsberg @davidvossel @fabiand 
Can you please review?

pkg/virt-api/webhooks/validating-webhook/validating-webhook_test.go
```diff
@@ -1382,7 +1382,7 @@ var _ = Describe("Validating Webhook", func() {
 
 				vmi.Spec.Networks = append(vmi.Spec.Networks,
 					v1.Network{Name: networkName, NetworkSource: v1.NetworkSource{
-						Multus: &v1.CniNetwork{NetworkName: networkName}}})
+						Npwgv1: &v1.CniNetwork{NetworkName: networkName}}})
```
booxter: There should be tests, unit and functional, that still use ```multus``` attribute, to make sure we don't break old manifests.

pkg/virt-controller/services/template.go
```diff
@@ -961,9 +962,15 @@ func getNamespaceAndNetworkName(vmi *v1.VirtualMachineInstance, fullNetworkName
...
 func getNetworkToResourceMap(virtClient kubecli.KubevirtClient, vmi *v1.VirtualMachineInstance) (networkToResourceMap map[string]string, err error) {
 	networkToResourceMap = make(map[string]string)
-	for _, network := range vmi.Spec.Networks {
+	for idx, network := range vmi.Spec.Networks {
 		if network.Multus != nil {
-			namespace, networkName := getNamespaceAndNetworkName(vmi, network.Multus.NetworkName)
+			networkSource := v1.NetworkSource{Npwgv1: network.Multus}
+			npwgv1Network := v1.Network{Name: network.Name, NetworkSource: networkSource}
+			vmi.Spec.Networks[idx] = npwgv1Network
+			network.Npwgv1 = network.Multus
```
booxter: what does this line do? AFAIU you've just overridden the network in vmi.Spec to a network of the new format, so this network is no longer attached to the list. Why do you need to modify it instead of just leaving it to garbage collector since it's, presumably, no longer attached to anywhere?

pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
```diff
@@ -945,10 +945,15 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 				podExists = true
 			}
 
+			// convert Multus to Npwgv1 network for backwards compatibility
```
booxter: Do you do it in-place? So if I post a manifest with ```multus``` used, I still end up with ```npgwv1``` attribute and w/o ```multus``` attribute? Why do we need to do it while we still support both? We'll need to migrate them to the new name when and if we decide to drop support for ```multus``` attribute anyway, so doing it here doesn't seem of help much; and it may raise questions of why previously the same manifest result was different, in a seemingly inconsistent way. I would prefer if we honour user choice of names, and revisit migration of attributes only when we decide to go with incompatible changes.

pkg/virt-controller/services/template.go
```diff
@@ -976,12 +983,23 @@ func getNetworkToResourceMap(virtClient kubecli.KubevirtClient, vmi *v1.VirtualM
...
 	ifaceList := make([]string, 0)
+	ifaceListMap := make([]map[string]string, 0)
 
-	for _, network := range vmi.Spec.Networks {
+	for idx, network := range vmi.Spec.Networks {
 		// set the type for the first network
 		// all other networks must have same type
 		if network.Multus != nil {
-			ifaceList = append(ifaceList, network.Multus.NetworkName)
+			networkSource := v1.NetworkSource{Npwgv1: network.Multus}
```
booxter: It seems to be a pattern in this PR to convert network spec in-place over and over. I don't think we really need to do this. You can introduce a helper function "getNpwgNetwork" that, when called on ```multus``` network, would return a structure of ```npwg``` type, and would return as-is for an ```npwg``` networks. There seems to be no reason to modify the spec in-place, whether on POST or when preparing pod environment. Thoughts?

pkg/virt-controller/services/template.go
```diff
@@ -992,8 +1010,13 @@ func getCniInterfaceList(vmi *v1.VirtualMachineInstance) (ifaceListString string
 			}
 		}
 	}
-
-	ifaceListString = strings.Join(ifaceList, ",")
+	if len(ifaceListMap) > 0 {
+		//ifaceListString, _ = json.Marshal(ifaceListMap)
```
booxter: remove?

pkg/virt-controller/services/template.go
```diff
@@ -992,8 +1010,13 @@ func getCniInterfaceList(vmi *v1.VirtualMachineInstance) (ifaceListString string
...
+		ifaceJsonString, _ := json.Marshal(ifaceListMap)
```
booxter: I don't think you can just ignore the potential error here. What if it failed?

pkg/virt-controller/services/template.go
```diff
@@ -992,8 +1010,13 @@ func getCniInterfaceList(vmi *v1.VirtualMachineInstance) (ifaceListString string
...
+		ifaceListString = fmt.Sprintf("%s", ifaceJsonString)
+	} else {
+		ifaceListString = strings.Join(ifaceList, ",")
```
booxter: What's the reason for us to interchange format of the annotation based on the length of the map-list? Wouldn't it work even if we use json format for a 1-list? If 1-list json still works, this seems to introduce unnecessary complexity.

pkg/virt-launcher/virtwrap/network/network.go
```diff
@@ -70,11 +70,11 @@ func SetupNetworkInterfaces(vmi *v1.VirtualMachineInstance, domain *api.Domain)
...
 		}
 
-		if networks[iface.Name].Multus != nil {
-			// multus pod interfaces named netX
+		if networks[iface.Name].Npwgv1 != nil {
+			// Npwgv1 pod interfaces named netX if interfacePrefix not specified
 			podInterfaceName = fmt.Sprintf("net%d", cniNetworks[iface.Name])
 		} else if networks[iface.Name].Genie != nil {
-			// genie pod interfaces named ethX
+			// genie and tf pod interfaces named ethX
```
booxter: "tf"?

pkg/virt-launcher/virtwrap/network/network.go
```diff
@@ -70,11 +70,11 @@ func SetupNetworkInterfaces(vmi *v1.VirtualMachineInstance, domain *api.Domain)
 			return err
 		}
 
-		if networks[iface.Name].Multus != nil {
-			// multus pod interfaces named netX
+		if networks[iface.Name].Npwgv1 != nil {
+			// Npwgv1 pod interfaces named netX if interfacePrefix not specified
```
booxter: where is ```interfacePrefix``` specified? Is it some remnants of previous work on the PR?

pkg/virt-launcher/virtwrap/network/network_test.go
```diff
@@ -61,19 +61,19 @@ var _ = Describe("Network", func() {
 			err := SetupNetworkInterfaces(vmi, domain)
 			Expect(err).To(BeNil())
 		})
-		It("should configure networking with multus", func() {
+		It("should configure networking with npwgv1", func() {
 			NetworkInterfaceFactory = func(network *v1.Network) (NetworkInterface, error) {
 				return mockNetworkInterface, nil
 			}
-			const multusInterfaceName = "net1"
+			const npwgv1InterfaceName = "net1"
```
booxter: is it even used anywhere?..

tests/vmi_npwgv1_test.go
```diff
@@ -45,7 +45,7 @@ const (
 	sriovConfCRD = `{"apiVersion":"k8s.cni.cncf.io/v1","kind":"NetworkAttachmentDefinition","metadata":{"name":"%s","namespace":"%s","annotations":{"k8s.v1.cni.cncf.io/resourceName":"intel.com/sriov"}},"spec":{"config":"{ \"name\": \"sriov\", \"type\": \"sriov\", \"ipam\": { \"type\": \"host-local\", \"subnet\": \"10.1.1.0/24\" } }"}}`
 )
 
-var _ = Describe("Multus", func() {
+var _ = Describe("Npwgv1", func() {
```
booxter: I think you will need to adjust ```automation/test.sh``` for this suite name change.

api/openapi-spec/swagger.json
```diff
@@ -4907,6 +4907,9 @@
       "description": "Network name.\nMust be a DNS_LABEL and unique within the vm.\nMore info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names",
       "type": "string"
      },
+     "npwgv1": {
```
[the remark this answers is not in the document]
dankenigsberg: yes, `npwgv1` stands for that. It was suggested by @dcbw. Lacking a better name, I accepted it myself. I'm fine with using `npwg` until a `v2` is released.

I don't like `standard` because it is even less descriptive and googleable than `npwg`. `defacto` could be nicer

booxter:
BTW there is probably no hard reason to have two distinct changes in this same PR:

1) switch to json list for networks;
2) introduce the new synonym for ```multu``` called ```npwg```.

It may be easier to land them separately. (I don't insist, just an idea.)

pkg/virt-controller/services/template.go
michaelhenkel: In the next line we check for network.Npwgv1, hence network.Npwgv1 has to be set:
https://github.com/kubevirt/kubevirt/blob/1983659767f9737533edd710e23588bf8e8782a8/pkg/virt-controller/services/template.go#L970-L972

pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
michaelhenkel: @dankenigsberg asked for that:

> So, at least in the schema we need both and also in the UTs.

>> Yes, that is what I attempted to mean by "canonize". We should behave alike with `multus` or >> `npwgv1`. Right after (or during?) input validation, input should be changed to `npwgv1`, >> which the bulk of the code should use exclusively.

dankenigsberg: I requested we *behave* the same way no matter if `multus` or `npwg` is placed by the user.
@booxter suggests that we *modify* the user input, which is not too polite.

dankenigsberg: Sorry about having misled you @michaelhenkel . I now see that my text text was badly written, I never intended to modify the `spec`.

booxter: @dankenigsberg I could even go with modifying input, *if* we would consistently migrate all existing VMIs to the new name. We don't, so with the way PR stands right now, we would have some VMIs use ```multus``` and some ```npwg```, depending on when they were created and whether they got updated after the PR was rolled in. This is the inconsistency I am primarily concerned about. Otherwise, for me personally, the decision could go both ways - either don't ever modify posted spec; or modify all specs, for new and existing VMIs. The former should be easier to pull.

michaelhenkel:
As requested I removed the replacement of the network and instead use a helper function to return the network name, based on the type.
I also renamed npwgv1 to npwg.
I did not touch the UTs, yet. If we are in agreement I will take care for them.

booxter:
@michaelhenkel do you have capacity to update this PR? Perhaps I could take the json list part if you don't mind.

michaelhenkel:
@booxter by updating the PR -> do you mean removing the JSON list from it?

pkg/virt-controller/services/template.go
michaelhenkel: Reason is that I do not know if Genie can deal with JSON list. In case it can -> I agree with your statement and we should use JSON list here as well.

pkg/virt-controller/services/template.go
```diff
@@ -623,8 +624,8 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		return nil, err
 	}
 
-	// Register resource requests and limits corresponding to attached multus networks.
-	// TODO(ihar) remove when we adopt Multus mutating webhook that handles the job.
+	// Register resource requests and limits corresponding to attached npwg/multus networks.
+	// TODO(ihar) remove when we adopt Npwg/Multus mutating webhook that handles the job.
```
booxter: it's really just Multus here :) as in name of the project.

pkg/virt-controller/services/template.go
```diff
@@ -962,8 +966,10 @@ func getNamespaceAndNetworkName(vmi *v1.VirtualMachineInstance, fullNetworkName
 func getNetworkToResourceMap(virtClient kubecli.KubevirtClient, vmi *v1.VirtualMachineInstance) (networkToResourceMap map[string]string, err error) {
 	networkToResourceMap = make(map[string]string)
 	for _, network := range vmi.Spec.Networks {
-		if network.Multus != nil {
-			namespace, networkName := getNamespaceAndNetworkName(vmi, network.Multus.NetworkName)
+		var namespace string
```
booxter: why should these variables be defined in this scope and not under ```if``` from line 971 below?

pkg/virt-controller/services/template.go
booxter: Please add an explanation in the comment on why we keep this logic (about Genie).

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -906,7 +906,7 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 		if network.Pod != nil {
 			numberOfSources++
 		}
-		if network.Multus != nil {
+		if network.Npwg != nil || network.Multus != nil {
```
booxter: (No action required) looks like we are doing this type of check in multiple places. This could be a helper function like ```isMultusNetwork``` that would do that.

pkg/virt-controller/services/template.go
michaelhenkel: done

pkg/virt-controller/services/template.go
michaelhenkel: done

pkg/virt-launcher/virtwrap/api/converter.go
michaelhenkel: Done

michaelhenkel:
@davidvossel @booxter 
can you help with that failure:
https://travis-ci.org/kubevirt/kubevirt/builds/503028384#L966

I don't see that on my local setup:

[root@5b3s32]~/kubevirt# if [[ -n "$(git status --porcelain)" ]] ; then echo "It seems like you need to run `make`. Please run it and commit the changes"; git status --porcelain; false; fi
[root@5b3s32]~/kubevirt# echo $?
0
[root@5b3s32]~/kubevirt#

davidvossel:
> @davidvossel @booxter
can you help with that failure:
https://travis-ci.org/kubevirt/kubevirt/builds/503028384#L966

I think you just need to run "make generate" and add the diff to a commit. That check is just making sure that no changes occur when make generate runs. If changes occur, that means autogenerated files are missing/modified and need to be committed.

michaelhenkel:
hm, that's what I did. When I do 'make generate' and commit/push there is
no change tonbe pushed.
I will try again tomorrow.

David Vossel <notifications@github.com> schrieb am Do. 7. März 2019 um
23:07:

michaelhenkel:
@booxter @davidvossel @dankenigsberg 
the first two gates passed now. Can you review and you ask ci to test please?

slintes:
ci test please

pkg/virt-launcher/virtwrap/api/converter.go
davidvossel: it looks like this is the same as before still?! maybe I'm missing something?

pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
```diff
@@ -967,10 +967,10 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 				podExists = true
 			}
 
-			if network.NetworkSource.Multus != nil {
+			if multusOrNpwgNetworkName := util.GetMultusOrNpwgNetworkName(network); multusOrNpwgNetworkName != "" {
```
davidvossel: this logic isn't exactly the same as before. 

previously we were checking if the multus object existed, now we're checking if either the multus/npwg object exists and whether or not the name is set as well.

I think this means the 'networkNameExistsOrNotNeeded' won't get set the same way it was previously.

pkg/util/util.go
```diff
@@ -20,3 +22,13 @@ func GetNamespace() (string, error) {
 	}
 	return namespaceKubevirt, nil
 }
+
+func GetMultusOrNpwgNetworkName(network v1.Network) string {
```
davidvossel: we probably need a 'IsMultusOrNpwgNetwork(network v1.Network) function as well. 

This might seem a little strange at first, but I'd actually just put these helpers in the pkg/api/v1/schema.go file. They could look something like this.

```
func (network *Network) GetMultusOrNpwgNetworkName() string {
        if network.Npwg != nil {
		return network.Npwg.NetworkName
	}
	if network.Multus != nil {
		return network.Multus.NetworkName
	}
	return ""
}

func (network *Network) IsMultusOrNpwgNetwork() string {
        if network.Npwg != nil || network.Multus != nil {
            return true
        }
        return false 
}

```

pkg/virt-controller/services/template.go
```diff
@@ -1044,8 +1048,8 @@ func getNamespaceAndNetworkName(vmi *v1.VirtualMachineInstance, fullNetworkName
 func getNetworkToResourceMap(virtClient kubecli.KubevirtClient, vmi *v1.VirtualMachineInstance) (networkToResourceMap map[string]string, err error) {
 	networkToResourceMap = make(map[string]string)
 	for _, network := range vmi.Spec.Networks {
-		if network.Multus != nil {
-			namespace, networkName := getNamespaceAndNetworkName(vmi, network.Multus.NetworkName)
+		if multusOrNpwgNetworkName := util.GetMultusOrNpwgNetworkName(network); multusOrNpwgNetworkName != "" {
```
davidvossel: this isn't the same logic as before. We're checking if network name exists rather if the struct is set.  This probably wouldn't matter once the validating webhook logic I pointed out gets fixed.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -1072,7 +1072,7 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 				if value, ok := cniNetworks[iface.Name]; ok {
 					prefix := ""
 					// no error check, we assume that CNI type was set correctly
-					if net.Multus != nil {
+					if net.Npwg != nil || net.Multus != nil {
```
davidvossel: Once those helper funcitons are added to schema.go, we could replace this (and similar if statements) with ```net.IsMultusOrNpwg()```

pkg/virt-launcher/virtwrap/network/network.go
```diff
@@ -51,8 +51,8 @@ func SetupNetworkInterfaces(vmi *v1.VirtualMachineInstance, domain *api.Domain)
 	cniNetworks := map[string]int{}
 	for _, network := range vmi.Spec.Networks {
 		networks[network.Name] = network.DeepCopy()
-		if networks[network.Name].Multus != nil {
-			// multus pod interfaces start from 1
+		if networks[network.Name].Npwg != nil || networks[network.Name].Multus != nil {
```
davidvossel: ```networks[network.Name].IsMultusOrNpwg()``` same with the checks similar checks later on in this function.

dankenigsberg:
> I'm also curious why the 'v1' was dropped off of npwg in the api. Would we just add a version field to the npwg object if v2 of the standard is not backwards compatible?

Precisely.

> are we 100% certain npwg is the right name for this network

Not at all. But that's the name suggested by an author of the de-facto standard (@dcbw), so we went along.

pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
michaelhenkel: done

pkg/util/util.go
michaelhenkel: done

pkg/virt-controller/services/template.go
michaelhenkel: done

pkg/virt-launcher/virtwrap/network/network.go
michaelhenkel: done

pkg/virt-launcher/virtwrap/api/converter.go
michaelhenkel: done

booxter:
@michaelhenkel please adjust release note: it should NOT mentiond tungsten; and it should mention the new npwg attribute as well as the change to how you pass the list of interfaces.

booxter:
I am concerned we are burying json annotation format change into the rest of the PR that introduces a synonym for multus attribute. The more I see where this PR goes the more I am wondering why we have both changes in a single PR. It's clear that the new attribute introduction is currently blocking the annotation format change with no good reason. I prefer them be split into separate pieces. I've posted a new PR that includes just the bare minimal for JSON annotation format: https://github.com/kubevirt/kubevirt/pull/2107 @michaelhenkel please take a look and tell me if you agree with the split and if so whether it includes all the changes needed for Tungsten to proceed (even if using `multus` attribute in VMI specs). I tried to keep authorship of your code in the new PR (set with git commit --amend --author).

booxter:
I would also like to join the unease of @davidvossel about the name of the new attribute. The name is unclear to outsiders, doesn't refer to anything known outside kubevirt (`multus` is at least well known in the community), nor it reflects what the attribute *does*. The *np* prefix is also perhaps redundant since we already have the networking context when dealing with interfaces and networks. So even `wg` would probably be a slightly better option (I saw in other contexts the word "workgroup" being abbreviated like that). But of all options proposed I hate `defacto` the less. At least it's 1) not plugin specific and 2) refers to a real thing known in the community.

michaelhenkel:
> @michaelhenkel please adjust release note: it should NOT mentiond tungsten; and it should mention the new npwg attribute as well as the change to how you pass the list of interfaces.

will do

michaelhenkel:
> I would also like to join the unease of @davidvossel about the name of the new attribute. The name is unclear to outsiders, doesn't refer to anything known outside kubevirt (`multus` is at least well known in the community), nor it reflects what the attribute _does_. The _np_ prefix is also perhaps redundant since we already have the networking context when dealing with interfaces and networks. So even `wg` would probably be a slightly better option (I saw in other contexts the word "workgroup" being abbreviated like that). But of all options proposed I hate `defacto` the less. At least it's 1) not plugin specific and 2) refers to a real thing known in the community.
> 
> Also, `npwg` is a letter soup prone to mixing. Each time I need to write it, I struggle to make the order of letters right. I imagine there will be plenty of cursing when people make mistakes in their specs.

I'm not picky on the name and I am used to change it by now. I just need a final decision. Who besides @davidvossel , @booxter and @dcbw must give their ok?

michaelhenkel:
> I am concerned we are burying json annotation format change into the rest of the PR that introduces a synonym for multus attribute. The more I see where this PR goes the more I am wondering why we have both changes in a single PR. It's clear that the new attribute introduction is currently blocking the annotation format change with no good reason. I prefer them be split into separate pieces. I've posted a new PR that includes just the bare minimal for JSON annotation format: #2107 @michaelhenkel please take a look and tell me if you agree with the split and if so whether it includes all the changes needed for Tungsten to proceed (even if using `multus` attribute in VMI specs). I tried to keep authorship of your code in the new PR (set with git commit --amend --author).

@booxter perfectly fine with that. All TF requires is the JSON list. It will work with multus naming as well as long as the POD annotation is a JSON list.

davidvossel:
> perfectly fine with that. All TF requires is the JSON list. It will work with multus naming as well as long as the POD annotation is a JSON list.

@booxter @michaelhenkel 

Just to get this straight. The JSON list is already compatible with multus and we don't anticipate it breaking anything?

If that's the case, I'm in favor of merging a patch that's reduced to just the JSON list for multus and continuing the discussion about how to support the npwg in the future separate from that pr. I just don't think there's enough agreement or understanding around how to represent npwg in our api yet.

booxter:
@davidvossel yes TF should be able to work with just JSON piece. That's the reason why I split it out - no reason to hold off TF support with our naming discussions.

fabiand:
So, this looks like a ripe fruit - what are the remaining gaps to get this merged?

booxter:
@fabiand we should rebase the PR now that JSON format PR merged. (There are git conflicts.) Release note should be updated to not mention TF.

I guess the only and major question hanging in the air right now is the name for the new synonym attribute. Is everyone ok with `npwg`? Do we believe it's the best and the most clear to understand name we can come up with? (Alternative mentioned so far were: `defacto`, `standard`... Maybe something more generic like `additional` or `advanced` or?..)

If we are ok with the name, it's a matter of rebase and merge.

michaelhenkel:
I will adjust the documentation once we agreed on the name. 
I'm ok with any name but I prefer defacto over the other proposals.
@davidvossel @booxter @fabiand Are you all ok with defacto? If yes, do we need someone else to agree?

booxter:
@michaelhenkel I don't think there is a single person to get an approval for the issue. I've raised the question on the previous weekly call last Wednesday, I haven't seen strong approvals or disapprovals here since then. I tend to lean towards `defacto` but `npwg` is ok too. I will let others decide. I suggest to wait for the final comments another day and then pick one of options, adopt it, and we proceed to merge.

@SchSeba @davidvossel @fabiand @dankenigsberg final thoughts before @michaelhenkel updates the PR with the (new?) name for the `defacto` / `npwg` synonym of `multus` network?

dankenigsberg:
@michaelhenkel can you take the question to sig-network? I can live with `multus` until there is a consensus about a generic term.

I would not like to add the suggested term only to replace it few weeks later.

fabiand:
The term/abbreviation is something that the NetworkPlumbing WG needs to
provide.
I'd not take a generic term like "standard" or "defacto" because it lacks
any context.

If, even if cumbersome to read and remember, I'd take nwpgv1 as it gives at
lesat some context.

ut at best we get a recommendation for an abbreviation fro the Network
Plumibing WG - WHo can speak to them?

davidvossel:
> The term/abbreviation is something that the NetworkPlumbing WG needs to
provide.
I'd not take a generic term like "standard" or "defacto" because it lacks
any context.

I tend to agree with this.

From what I understand we're unblocked (at least for the time being) with the multus json list.

I suggest the next step is to urge the networking WG to provide a reasonable and consistent term to reference the specification they've created. Without that, I'm not going to be able to say with any confidence the term we pick will remain accurate in the future. We're at the point now where once we introduce something into our API it's not trivial to get rid of or replace it later on.

dankenigsberg:
@michaelhenkel would you confirm that https://github.com/kubevirt/kubevirt/pull/2107 is functionally enough to support TF?

michaelhenkel:
@dankenigsberg yes, TF works with 2107 using multus name.

booxter:
@michaelhenkel have you reached to networking plumbing WG requesting a decision on a clear name for the standard they produce?

michaelhenkel:
I reached out on the google group:
https://groups.google.com/forum/#!topic/kubernetes-sig-network/cXTVuCQIqJE

dougbtv:
Saw your post on sig-network, I tend to like npwgv1, or even npwg. I think it's probably right to change it over just simply "multus". No doubt, it's a verbose name for the spec, mostly we were concerned with the names within it, and that the title of the spec be... well, rather specific about what it specifies.

booxter:
@dougbtv do you agree though that `npwg` tells very little to users about what it is, unless they are part of the dev community? The attribute is public API and it would be nice to have a easy to decipher name.

davidvossel:
> Saw your post on sig-network, I tend to like npwgv1, or even npwg. I think it's probably right to change it over just simply "multus". No doubt, it's a verbose name for the spec, mostly we were concerned with the names within it, and that the title of the spec be... well, rather specific about what it specifies.

No one seems to be arguing that npwg is an ideal name for the specification, which doesn't instill much confidence that this acronym will be relevant to anyone 2-3 years from now. The fact that no one was concerned much with the naming of the spec as much as writing the contents of the spec further raises my suspicion that npwg is a temporary place holder for something that will receive a more official title at some point once people need to begin referencing more as we're trying to do here. 

These kinds of changes have large impacts on our api versioning which is why we're hesitant to adopt a name like "npwg" into our api when there's any ambiguity about the long term viability of the name.

PR #1967, opened 2019-01-23
Add test attributes for networkpolicy cases

duyanyan:
Changed all test to have rfe id, test id, criticality etc attributes

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/networkpolicy_test.go
+++ b/tests/networkpolicy_test.go
@@ -39,7 +39,7 @@ func assertPingFail(ip string, vmi *v1.VirtualMachineInstance) {
 	Expect(err).ToNot(HaveOccurred())
 }
 
-var _ = Describe("Networkpolicy", func() {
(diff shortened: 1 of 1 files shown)


tests/networkpolicy_test.go
```diff
@@ -100,13 +100,13 @@ var _ = Describe("Networkpolicy", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
-		It(" should be failed to reach vmia from vmib", func() {
+		It("[test_id:CNV-1511] should be failed to reach vmia from vmib", func() {
```
nellyc: i think some tests should be marked with posneg:negative
no?

duyanyan: I think they are not negative test scenarios, since when we created networkpolicy in the namespace, the traffic will be limited by the rules. The failed connections here are only to validate whether the rule is taking effect.

nellyc: makes sense! thanks

PR #1971, opened 2019-01-24
Add test attributes to L2 network configuration in multus.

yossisegev:
Add test attributes to L2 network configuration in multus.


yossisegev:
@cynepco3hahue , once again - CI failed for errors that are not related to this PR.
Can you please CR and merge?
Thanks.

Issue #1760, opened 2018-11-26
Define supported CPU models and validate VM.spec.cpu.model


yanirq:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind enhancement

Related to https://github.com/kubevirt/kubevirt/pull/1721#issuecomment-441564441 

Now that we are introducing scheduling virtual machines on nodes according to CPU model (https://github.com/kubevirt/kubevirt/pull/1721) by matching the vmi.spec.cpu.model to supported cpu models on nodes (nodes are labelled by [NFD Libvirt](https://github.com/ksimon1/node-feature-discovery-libvirt) [WIP]),  we need to define and validate the cpu models that can be set on a VMI spec according to the CPU map provided by libvirt. 

One reason for that need is if the CPUNodeDiscovery feature-gate will be enabled (disabled by default) 
and a vmi.spec.cpu.model wont match existing labels on nodes (due to a typo for example), the error on vmi / virt-launcher will be a general FailedScheduling message.
(In case the CPUNodeDiscovery feature-gate is disabled,  the error message we would get is a libvirt error for a wrong model)

yanirq:
Adding @rmohr @fabiand  @MarSik  to the discussion : 
- Is this a valid need ? 
- If implemented, what is preferable : open API, validation web-hooks ?

MarSik:
libvirt no loger defines names for all combinations of features ([1]). I think we want to define custom kubevirt CPU model + flags/features combinations, expose them using NFD and use them in virt-launcher.

There are couple of data fields we need:
- Human readable name for the "model"
- Base model + extra features (+ their mode) for libvirt domxml
- Base model + extra features for nodeSelector (compatible with NFD)
- Model for NFD created labels (unique model identifier)

Since this needs to be visible both to NFD and virt-launcher, we should probably create either a CRD (one per model?) or at least a ConfigMap.

A feature gate should be used to switch between pure libvirt, libvirt + custom and custom only models (this would allow extending or limiting per product needs).

[1] For example see SSBD security related flag that must be forced: https://www.redhat.com/archives/libvir-list/2018-May/msg01562.html

MarSik:
As to the validation question:

- libvirt model validation can't be done without libvirt which is only accessible from virt-launcher. But an admission controller could inherit from kubevirt/libvirt base container (without starting the daemon even) to get access to the data
- custom model validation in my proposed direction can be done from anywhere as CRDs and ConfigMaps are easily accessible

[...]

PR #1978, opened 2019-01-28
Support extra CPU flags when starting a VM

yanirq:
1.Add support for CPU flags (features) in a VM
2.Add the VM pod node-selector labels according to the VM's CPU flags so they match nodes supporting these features.
This feature can be enabled/disabled using the feature-gate : "CPUNodeDiscovery" 

Related to #1760 

To complete the picture, cpu features labels should be added to the nodes themselves by the [nfd discovery plugin](https://github.com/ksimon1/cpu-model-nfd-plugin) 
The current form of the labels that should match is:
feature.node.kubernetes.io/cpu-feature-\<feature-X\> = "true" 

**Release note**:

```release-note
To enable scheduling virtual machines on nodes by matching cpu feature on the virtual machine spec
and cpu feature labels on nodes (generated and added to node using node feature discovery plugin) :  
Enable the feature gate CPUNodeDiscovery - add it to the kubevirt-config ConfigMap.
```


yanirq:
[[epic-node-compatibility-1.4] Support extra CPU flags (features) when starting a VM [initiative-vm-high-availability]](https://trello.com/c/FGD3RGsr/91-epic-node-compatibility-14-support-extra-cpu-flags-features-when-starting-a-vm-initiative-vm-high-availability)

yanirq:
@ksimon1 @MarSik @vladikr 
Patch is ready for the first round of reviews.
Please note that currently CPU features can be passed to the VM and all error handling for name/policy are at the pod creation/libvirt level. no current hooks or other validations (there is a talk about validations going on #1760) 
Also, there are no special actions taken for feature policy (other than just passing them on to domain xml creation)
If this needs to be enhanced further we can start a discussion.

pkg/api/v1/schema.go
```diff
@@ -176,12 +176,35 @@ type CPU struct {
 	// Defaults to host-model.
 	// +optional
 	Model string `json:"model,omitempty"`
+	// Features specifies the CPU features list inside the VMI.
+	// In sepcial cases where Model is set to "host-model" or "host-passthrough",
```
ksimon1: sepcial - special?

pkg/virt-controller/services/template.go
```diff
@@ -168,6 +169,18 @@ func CPUModelLabelFromCPUModel(vmi *v1.VirtualMachineInstance) (label string, er
 	return
 }
 
+func CPUFeatureLabelsFromCPUFeatures(vmi *v1.VirtualMachineInstance) []string {
+	var labels []string
+	if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Features != nil {
+		if len(vmi.Spec.Domain.CPU.Features) > 0 {
```
ksimon1: you don't have to have this check here. If vmi.Spec.Domain.CPU.Features has 0 length, then for loop won't start.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -787,6 +787,18 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 			}
 		}
 
+		// Set VM CPU features
+		if vmi.Spec.Domain.CPU.Features != nil {
+			if len(vmi.Spec.Domain.CPU.Features) > 0 {
```
ksimon1: you don't have to have this check here. If vmi.Spec.Domain.CPU.Features has 0 length, then for loop won't start.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -787,6 +787,18 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+				for _, feature := range vmi.Spec.Domain.CPU.Features {
+					domainFeature := Feature{}
```
ksimon1: I would prefer this syntax:
```
domain.Spec.CPU.Features = append(domain.Spec.CPU.Features, Feature{
	Name: feature.Name,
	Policy: feature.Policy,
})
```

pkg/virt-controller/services/template.go
```diff
@@ -168,6 +169,18 @@ func CPUModelLabelFromCPUModel(vmi *v1.VirtualMachineInstance) (label string, er
...
+			for _, feature := range vmi.Spec.Domain.CPU.Features {
+				labels = append(labels, NFD_CPU_FEATURE_PREFIX+feature.Name)
```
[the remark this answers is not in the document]
yanirq: all features will make it to the list, policies will be addressed by libvirt.
The nfd plugin will label the nodes with all supported features regardless of policies.
Having that said, in case the nfd plugin will label nodes according to policies as well then we will adapt, but at the current stage this should be sufficient.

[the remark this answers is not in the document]
vladikr: Does the nfd plugin add __all__ of the cpu flags as labels? all 100+ of them or how does it work?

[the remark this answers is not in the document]
yanirq: @vladikr  the nfd plugin should add all the supported cpu flags (as labels) for each node in the cluster.
(the pluging is still WIP so we might have/expect some fine tuning on that part)

pkg/api/v1/schema.go
```diff
@@ -210,12 +210,33 @@ type CPU struct {
...
 	// +optional
 	DedicatedCPUPlacement bool `json:"dedicatedCpuPlacement,omitempty"`
 }
 
+// CPUFeature allows specifying a CPU feature.
+// ---
+// +k8s:openapi-gen=true
+type CPUFeature struct {
+	// Name of the CPU feature
+	Name string `json:"name,omitempty"`
```
ksimon1: shouldn't be name without omitempty?

pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
```diff
@@ -834,6 +835,27 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 		}
 	}
 
+	// Validate CPU Feature Policies
+	if spec.Domain.CPU != nil && spec.Domain.CPU.Features != nil {
+		isValidePolicy := func(policy string) bool {
```
ksimon1: isValidePolicy -> isValidPolicy?

pkg/virt-launcher/virtwrap/api/schema.go
```diff
@@ -164,9 +164,15 @@ type VCPU struct {
 type CPU struct {
 	Mode     string       `xml:"mode,attr,omitempty"`
 	Model    string       `xml:"model,omitempty"`
+	Features []CPUFeature `xml:"feature"`
 	Topology *CPUTopology `xml:"topology"`
 }
 
+type CPUFeature struct {
+	Name   string `xml:"name,attr,omitempty"`
```
ksimon1: shouldn't be name without omitempty?

pkg/virt-controller/services/template.go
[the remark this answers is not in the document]
vladikr: ah. it just sounds very odd to me that we are submitting code that depends on something that doesn't yet exist.
Why not wait until the plugin is implemented?

I also find it a bit weird that the plugin will expose such a large number of labels! It would be much reasonable (to me, at least) to expose a cpu family + certain flags and we will match against that. Similar to what libvirt does.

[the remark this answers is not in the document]
yanirq: this is a 2 phase submission as we did for cpu families +  this feature is enabled/disabled using a feature gate (as mentioned above). 

for the number of labels exposed, true, this could be large. but then again we will label nodes only with the feature and familys they support and can downsize it further if needed.
(see https://github.com/ksimon1/cpu-model-nfd-plugin)

pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
yanirq: you cant add '?' here, this is bad syntax.

ksimon1: i meant isValidPolicy without question mark :)

pkg/virt-controller/services/template.go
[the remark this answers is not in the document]
vladikr: > for the number of labels exposed, true, this could be large. but then again we will label nodes only with the feature and familys they support and can downsize it further if needed.
> (see https://github.com/ksimon1/cpu-model-nfd-plugin)

What's reasonable to me, is that the plugin would expose cpu family `Nehalem` and a cpu feature `spec-ctrl` and when I'm specifically looking for hosts with `spec-ctrl` that would be clear. But if I'm looking for hosts with a cpu feature `ssse3` I wouldn't expect the node to expose it as a separate cpu feature label but rather we should figure out that `ssse3` is part of the `Nehalem` family.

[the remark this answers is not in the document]
vladikr: > this is a 2 phase submission as we did for cpu families + this feature is enabled/disabled using a feature gate (as mentioned above).

I still find it very hard to undertand why should we keep unfuctioning code that doesn't have it's dependecy merged yet, even if it's behind a feature gate.

pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
yanirq: ah , i see what you mean :)

pkg/virt-controller/services/template.go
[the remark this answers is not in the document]
michalskrivanek: > What's reasonable to me, is that the plugin would expose cpu family `Nehalem` and a cpu feature `spec-ctrl` and when I'm specifically looking for hosts with `spec-ctrl` that would be clear. But if I'm looking for hosts with a cpu feature `ssse3` I wouldn't expect the node to expose it as a separate cpu feature label but rather we should figure out that `ssse3` is part of the `Nehalem` family.

@vladikr right. And I was expecting that we also cut off older models we do not ever plan to support. E.g. Nehalem would be the very first one, and the extra flags would be as reported by libvirt in addition to the base CPU model name. But I guess that's for the plugin discussion rather than here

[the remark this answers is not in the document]
ksimon1: @michalskrivanek cpu plugin already has functionality for cpu model black list, but it works only with kubernetes version of NFD. For openshift NFD some code changes would be needed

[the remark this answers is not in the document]
MarSik: @vladikr But node selectors do not work that way. Kubernetes does not know that ssse3 is included in Nehalem. It just treats all those as simple labels.

We need the feature gate for CI and deployments without NFD. NFD is optional, you can easily write your own method or use manual labels. Kubevirt here should be stupid and just copy whathever the user (or higher level management!) asked for. It should not care about what cpu labels or what cpu families are exposed or supported as labels. CNV level handles that and configures kubevirt accordingly.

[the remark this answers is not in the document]
MarSik: > But I guess that's for the plugin discussion rather than here

@michalskrivanek Exactly. kubevirt should not care.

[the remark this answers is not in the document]
MarSik: Btw just to be clear, we will limit the amount of exposed families / flags on the node label side. The easiest way is to remove all flags that are always supported by the selected baseline. But I think kubevirt just should not care as this will be implemented externally.

pkg/virt-controller/services/template.go
```diff
@@ -168,6 +169,16 @@ func CPUModelLabelFromCPUModel(vmi *v1.VirtualMachineInstance) (label string, er
 	return
 }
 
+func CPUFeatureLabelsFromCPUFeatures(vmi *v1.VirtualMachineInstance) []string {
+	var labels []string
+	if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Features != nil {
+		for _, feature := range vmi.Spec.Domain.CPU.Features {
+			labels = append(labels, NFD_CPU_FEATURE_PREFIX+feature.Name)
```
MarSik: Only labels with policy `require` should be converted to labels (optional, disable and force should not). The same applies to `forbid`, but those must create a negative selector.

yanirq: good point

yanirq:
@vladikr @davidvossel   Can you give another round of reviews ?  (also see the latest comments above by @MarSik )

yanirq:
ci test please

pkg/virt-controller/services/template.go
```diff
@@ -168,6 +169,67 @@ func CPUModelLabelFromCPUModel(vmi *v1.VirtualMachineInstance) (label string, er
 	return
 }
 
+func CPUFeatureLabelsFromCPUFeatures(vmi *v1.VirtualMachineInstance) []string {
+	var labels []string
+	if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Features != nil {
+		for _, feature := range vmi.Spec.Domain.CPU.Features {
+			if feature.Policy == "" || feature.Policy == "require" {
+				labels = append(labels, NFD_CPU_FEATURE_PREFIX+feature.Name)
```
fabiand: Should we sanitize the cpufamily / feature names (i.e. make them lowercase)?

IIRC then Kubernetes is also lowercasing all labels.

yanirq: > IIRC then Kubernetes is also lowercasing all labels.
lowercasing at what stage ?  , in this PR the features/models are translated to nodeselectors (not just labels in their raw form) .
The nodeselector's labels are not lower-cased when added to the virt-launcher pod (creating a vmi with a spec that has Uppercase labels under a node selector will translate to uppercase labels in virt-launcher)

yanirq: the valid [label syntax](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) includes upper-case letters.

tests/vmi_configuration_test.go
```diff
@@ -816,6 +816,32 @@ var _ = Describe("Configurations", func() {
...
+							Name: "apic",
+						},
+						{
+							Name:   "clflush",
+							Policy: "optional",
+						},
+					},
+				}
+
+				By("Starting a VirtualMachineInstance")
```
vladikr: I think this test should be much more expressive.
It needs to show that the VM is scheduled in the right place and has the right flags - or am not understanding something?

pkg/virt-controller/services/template.go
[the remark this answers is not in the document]
vladikr: Sorry for the delay. I was out last week.
I still think kubevirt can be smarter about it, effectively reducing the number of labels that need to be exposed by NFD - that's just weird to expose it all.
After all, kubevirt can easily know which flag belongs to which family and just schedule according to a family (+ additional flag that is an addition - like `spec-ctrl`)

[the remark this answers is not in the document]
yanirq: I think the most common use case here will be to add the cpu model to the vm spec and maybe additional flags (or change some existing flags behaviour).
On kubevirt side i don't see too many cases of adding all flags of a family to the vm spec. 
The overload of labels on the node side (assigned by nfd)  might be reduced also by that assumption (and by what @MarSik  mentioned above). 

Well defining the cpu models and features for NFD/Libvirt/scheduler  as discussed [here](https://github.com/kubevirt/kubevirt/issues/1760#issuecomment-442622633) should probably be a further step for this PR.

yanirq:
ci test please

pkg/virt-controller/services/template.go
fabiand: Right. I just think that I noticed that nfd exposed cpu labels in lower-case

fabiand: But that's not blocking, was just a perception.

pkg/virt-controller/services/template.go
```diff
@@ -168,6 +169,67 @@ func CPUModelLabelFromCPUModel(vmi *v1.VirtualMachineInstance) (label string, er
...
+				RequiredDuringSchedulingIgnoredDuringExecution: &k8sv1.NodeSelector{
+					NodeSelectorTerms: []k8sv1.NodeSelectorTerm{term},
+				},
+			}
+
+			if pod.Spec.Affinity != nil && pod.Spec.Affinity.NodeAffinity != nil {
+				if pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+					terms := pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+					// Since NodeSelectorTerms are ORed , the anti affinity requirement will be added to each term.
+					for i, selectorTerm := range terms {
```
davidvossel: shouldn't we be appending a node selector term here rather than mutating items already in the list?

tests/vmi_configuration_test.go
```diff
@@ -949,6 +957,41 @@ var _ = Describe("Configurations", func() {
 				}, 10*time.Second)
 			})
 		})
+
+		Context("when CPU features defined", func() {
+			It("should start a Virtaul Machine with matching features", func() {
+				vmiModel := "Conroe"
```
davidvossel: is this test going to break if someone attempts to run this test on something that isn't Conroe or Opteron_G1?

tests/vmi_lifecycle_test.go
```diff
@@ -653,32 +662,52 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
+				n.SetLabels(originalLabels)
+				_, err = virtClient.CoreV1().Nodes().Update(n)
+				Expect(err).ToNot(HaveOccurred())
+
 				//FIXME improve the detection if virt-controller already received the config-map change
 				time.Sleep(time.Millisecond * 500)
 			})
 
-			It("[test_id:1639]the vmi with cpu.model matching a nfd label on a node should be scheduled", func() {
+			It("[test_id:1639]the vmi with cpu.model and cpu.features matching nfd labels on a node should be scheduled", func() {
```
davidvossel: This is an overloaded test. We can't distinguish between the cpu model node selector causing the node to be scheduled or the Features.

tests/vmi_lifecycle_test.go
```diff
@@ -687,28 +716,41 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
+				Expect(curVMI.Status.NodeName).To(Equal(node.Name), "Updated VMI name run on the node with matching NFD CPU label")
 
 			})
 
-			It("[test_id:1640]the vmi with cpu.model that cannot match an nfd label on node should not be scheduled", func() {
-
-				nodes, err := virtClient.CoreV1().Nodes().List(metav1.ListOptions{})
-				Expect(err).ToNot(HaveOccurred(), "Should list nodes")
-				Expect(nodes.Items).ToNot(BeEmpty(), "There should be some nodes")
+			It("[test_id:1640]the vmi with cpu.model cpu.features that cannot match nfd labels on a node should not be scheduled", func() {
```
davidvossel: here as well, i don't like that this test is being overloaded. It means we can accidently break one part (feature policy forbid for example) but the test will still pass because the cpumodel prevents it from being scheduled.

pkg/api/v1/schema.go
```diff
@@ -210,12 +210,33 @@ type CPU struct {
 	// Defaults to host-model.
 	// +optional
 	Model string `json:"model,omitempty"`
+	// Features specifies the CPU features list inside the VMI.
+	// For more information see https://libvirt.org/formatdomain.html#elementsCPU.
```
davidvossel: we shouldn't be referencing libvirt documentation here. This is the kubevirt api. How our api maps to a libvirt domain is an implementation detail. 

It's confusing to have people learn about this feature by reading about libvirt. Libvirt doesn't explain how the KubeVirt feature impacts scheduling, or how this is related to the NFD component.  I see that libvirt documentation was referenced for CPU model as well. That shouldn't have made it in. We need to be documenting our API. How our API relates to Libvirt is irrelevant from the KubeVirt user's perspective.

pkg/api/v1/schema.go
```diff
@@ -210,12 +210,33 @@ type CPU struct {
...
+type CPUFeature struct {
+	// Name of the CPU feature
+	Name string `json:"name"`
+	// Policy is the CPU feature attribute which can have the following attributes:
+	// force    - The virtual CPU will claim the feature is supported regardless of it being supported by host CPU.
+	// require  - Guest creation will fail unless the feature is supported by the host CPU or the hypervisor is able to emulate it.
+	// optional - The feature will be supported by virtual CPU if and only if it is supported by host CPU.
+	// disable  - The feature will not be supported by virtual CPU.
+	// forbid   - Guest creation will fail if the feature is supported by host CPU.
+	// Defaults to require
```
davidvossel: It would be nice to indicate how each policy value impacts scheduling here.

pkg/api/v1/schema.go
yanirq: It was indeed taken and placed here according to the link we have in our user guide. 
ill remove the reference here and also in user guide.

tests/vmi_configuration_test.go
yanirq: actually the cpu model here can be removed completely.

pkg/virt-controller/services/template.go
davidvossel: i'm wrong here. we have to add to the match expressions.

"If you specify multiple nodeSelectorTerms associated with nodeAffinity types, then the pod can be scheduled onto a node if one of the nodeSelectorTerms is satisfied."

pkg/api/v1/schema.go
yanirq: Im not sure this is the right place for that.
The explanation for the schema here will be overloaded with specific information because the scheduling decisions made here for forbid/require/non are taken place in case the NFD feature is enabled and the nodes have the matching labels.

davidvossel:
If no one objects, I think this can be merged. I'll leave it for another day for people to raise any concerns.

PR #2003, opened 2019-02-05
Double memory and CPU limits

cynepco3hahue:
- for `container-disk` container
- for `compute` container

Fixes #

I see this PR just like some temporary solution to fix OOM failures under our CI tests.
We still need to provide some better way to calculate overhead of `compue` and `container-disk` containers

**Release note**:

```release-note
NONE
```


cynepco3hahue:
ok the single test that fails among 4 lanes is
```
[Fail] DataVolume Integration Starting a VirtualMachine with a DataVolume using Alpine import [It] should be successfully started and stopped multiple times 
/root/go/src/kubevirt.io/kubevirt/tests/utils.go:3080
```

PR #2008, opened 2019-02-06
enable usb controller

ksimon1:
This PR enables usb controller which was by default turned off. USB controller is needed for this PR: https://github.com/kubevirt/kubevirt/pull/1987 to enable USB bus for tablet. USB controller is enabled only when user specify `spec.domain.features.usbDevice.enabled` to true.

**Release note**:
```release-note
enable usb controller 
```

### DIFF
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -353,6 +353,54 @@ var _ = Describe("Configurations", func() {
 
 		})
 
+		Context("with usb controller", func() {
+			It("should start the VMI with usb controller", func() {
+				vmi := tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskAlpine))
+				_true := true
+				vmi.Spec.Domain.Features = &v1.Features{
+					USBDevice: &v1.FeatureState{
+						Enabled: &_true,
+					},
+				}
+				By("Starting a VirtualMachineInstance")
+				vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+				Expect(err).ToNot(HaveOccurred(), "should start vmi")
+				tests.WaitForSuccessfulVMIStart(vmi)
+
+				By("Expecting the VirtualMachineInstance console")
+				expecter, err := tests.LoggedInAlpineExpecter(vmi)
+				Expect(err).ToNot(HaveOccurred(), "should get console")
+				defer expecter.Close()
+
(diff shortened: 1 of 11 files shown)


tests/vmi_configuration_test.go
```diff
@@ -353,6 +353,54 @@ var _ = Describe("Configurations", func() {
...
+				tests.WaitForSuccessfulVMIStart(vmi)
+
+				By("Expecting the VirtualMachineInstance console")
+				expecter, err := tests.LoggedInAlpineExpecter(vmi)
+				Expect(err).ToNot(HaveOccurred(), "should get console")
+				defer expecter.Close()
+
+				By("Checking the number of usb under guest OS")
+				_, err = expecter.ExpectBatch([]expect.Batcher{
+					&expect.BSnd{S: "ls -l /sys/bus/usb/devices/usb* | wc -l\n"},
```
stu-gott: What we're really trying to check here is that the USB controller is enabled, so it might be safer to check for a value greater than zero instead of checking for exactly 2 devices.

ksimon1:
Right now, usb controller is disabled by default, it is turned on only when user sets usbDevice feature.
Sure thing it can be changed to allow usb controller, when user add usb device.

vladikr:
I agree with Fabian. I don't see a reason to just unconditionally enable the controller without an actual USB device being added.
Also, could you please explain why a `usbDevice` is be a feature? What are we trying to achieve here?

If we are trying to schedule a vm on a node with an arbitrary usb device, shouldn't we be looking for a specific device? Or am I missing something?

ksimon1:
@vladikr I agree the enabling of usb controller when usb device is added is better solution. I put it inside the features, because you can specify if you would like to controller enabled or disabled. 
I had to add this functionality to this PR https://github.com/kubevirt/kubevirt/pull/1987, because it contains attributes which are needed to check if usb should be enabled. Please feel free to review it. I am closing this PR, because it is no longer needed.

michalskrivanek:
@fabiand @vladikr I would rgue that it actually makes more sense to add USB controller by default unless it's really a high perf workload type. After alll that's why we introduce that concept.
The controller itself is not hotpluggable, so you do need it whenever you would want to offer a USB hot pluggability (hotplug of hostdev or SPICE-based USB sharing). I do not think there's actually any overhead to have a single XHCI controller there by default

fabiand:
Right, if it has no overhead on a high-performance VM profile, then it
sounds to me as if we can add it. But adding it does not imply that we need
to expose it in the API.

PR #2015, opened 2019-02-07
Add basic bash completion

rsevilla87:
More working will be required in the future in order to add zsh shell and custom completion functions.

**Release note**:

```release-note
Basic bash completion option to the virtctl CLI
```

### DIFF
--- a/pkg/virtctl/root.go
+++ b/pkg/virtctl/root.go
@@ -36,6 +36,29 @@ func NewVirtctlCommand() *cobra.Command {
 			fmt.Fprint(cmd.OutOrStderr(), cmd.UsageString())
 		},
 	}
+	completionCmd := &cobra.Command{
+		Use:   "completion",
+		Short: "Generates completion scripts for the specified shell",
+		Args:  cobra.MaximumNArgs(1),
+		Long: `To load completion run
+. <(virtctl completion (SHELL TYPE))
+
+To configure bash shell to load completions for each session add to bashrc
+
+# ~/.bashrc or ~/.profile
+. <(virtctl completion (SHELL TYPE))
+`,
+		Run: func(cmd *cobra.Command, args []string) {
+			switch {
+			case len(args) == 0:
+				rootCmd.GenBashCompletion(os.Stdout)
+			case args[0] == "bash":
+				rootCmd.GenBashCompletion(os.Stdout)
+			default:
+				fmt.Fprintf(cmd.OutOrStderr(), "%q is not a supported shell", args[0])
+			}
+		},
+	}
 	optionsCmd.SetUsageTemplate(templates.OptionsUsageTemplate())
 	//TODO: Add a ClientConfigFactory which allows substituting the KubeVirt client with a mock for unit testing
 	clientConfig := kubecli.DefaultClientConfig(rootCmd.PersistentFlags())
@@ -51,6 +74,7 @@ func NewVirtctlCommand() *cobra.Command {
 		version.VersionCommand(clientConfig),
 		imageupload.NewImageUploadCommand(clientConfig),
 		optionsCmd,
+		completionCmd,
 	)
 	return rootCmd
 }


stu-gott:
ok to test

stu-gott:
ci test please

stu-gott:
thanks @rsevilla87!

pkg/virtctl/root.go
```diff
@@ -36,6 +36,29 @@ func NewVirtctlCommand() *cobra.Command {
...
+
+To configure bash shell to load completions for each session add to bashrc
+
+# ~/.bashrc or ~/.profile
+. <(virtctl completion (SHELL TYPE))
+`,
+		Run: func(cmd *cobra.Command, args []string) {
+			switch {
+			case len(args) == 0:
+				rootCmd.GenBashCompletion(os.Stdout)
```
slintes: I would prefer to align argument handling with kubectl, and return an error if no arg is given, instead of defaulting to bash. Also, please check for `len(args) > 1` for giving a better error message. See https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/completion/completion.go#L118

pkg/virtctl/root.go
```diff
@@ -36,6 +36,29 @@ func NewVirtctlCommand() *cobra.Command {
...
+			case args[0] == "bash":
+				rootCmd.GenBashCompletion(os.Stdout)
+			default:
```
slintes: I see that that kubectl does some crazy stuff for zsh (see link above).... but there also is a `GenZshCompletion` function.  Maybe worth a try?

rsevilla87: `GenZshCompletion` generates a similar output than GenBashCompletion
I'll take a look to zsh completion but I think that the "crazy stuff" would be also required here :(

SchSeba:
Hi @slintes I don't have time to work on that yet so please continue with this PR

@rsevilla87 nice job!
I will make a review on this PR  soon

pkg/virtctl/root.go
slintes: ok, understand. Would be awesome if you can add zsh, but if you prefer to skip it for this PR, that's also fine :)

pkg/virtctl/root.go
```diff
@@ -36,6 +36,29 @@ func NewVirtctlCommand() *cobra.Command {
 			fmt.Fprint(cmd.OutOrStderr(), cmd.UsageString())
 		},
 	}
+	completionCmd := &cobra.Command{
+		Use:   "completion",
+		Short: "Generates completion scripts for the specified shell",
+		Args:  cobra.MaximumNArgs(1),
```
rmohr: You can use `cobra.ExactArgs(1)`, then you don't have tho check the length of the arguments.

pkg/virtctl/root.go
```diff
@@ -36,6 +36,29 @@ func NewVirtctlCommand() *cobra.Command {
 			fmt.Fprint(cmd.OutOrStderr(), cmd.UsageString())
 		},
 	}
+	completionCmd := &cobra.Command{
```
rmohr: I think you can copy my exact descriptions from #916. It is itself a copy from `kubectl complete`.

pkg/virtctl/root.go
rmohr: You can use `		ValidArgs: []string{"bash", "zsh"},`.

pkg/virtctl/root.go
```diff
@@ -36,6 +36,29 @@ func NewVirtctlCommand() *cobra.Command {
...
+. <(virtctl completion (SHELL TYPE))
+
+To configure bash shell to load completions for each session add to bashrc
+
+# ~/.bashrc or ~/.profile
+. <(virtctl completion (SHELL TYPE))
+`,
+		Run: func(cmd *cobra.Command, args []string) {
+			switch {
+			case len(args) == 0:
```
rmohr: Validation should not be necessary in the command code itself. You can define it on the command above itself.

pkg/virtctl/root.go
rmohr: I think this has to be `return cmd.Parent().GenBashCompletion(os.Stdout)`.

rmohr: Note the `Parent()`.

PR #2030, opened 2019-02-17
Automate service expose test

yossisegev:
Automate service expose test:
Expose VM as a Cluster IP service, delete the VM and verify the service is no longer functional.

### DIFF
--- a/tests/expose_test.go
+++ b/tests/expose_test.go
@@ -49,6 +49,20 @@ func generateHelloWorldServer(vmi *v1.VirtualMachineInstance, virtClient kubecli
 	Expect(err).ToNot(HaveOccurred())
 }
 
+func waitForJobToCompleteWithStatus(virtClient *kubecli.KubevirtClient, jobPod *k8sv1.Pod, expectedResult string, timeoutSec int) {
+	jobPodPhase := k8sv1.PodFailed
+
+	if expectedResult == "success" {
+		jobPodPhase = k8sv1.PodSucceeded
+	}
+
+	EventuallyWithOffset(1, func() k8sv1.PodPhase {
+		pod, err := (*virtClient).CoreV1().Pods(jobPod.Namespace).Get(jobPod.Name, k8smetav1.GetOptions{})
+		ExpectWithOffset(1, err).ToNot(HaveOccurred())
+		return pod.Status.Phase
+	}, time.Duration(timeoutSec)*time.Second, 1*time.Second).Should(Equal(jobPodPhase))
+}
+
 var _ = Describe("Expose", func() {
 
 	flag.Parse()
@@ -85,12 +99,7 @@ var _ = Describe("Expose", func() {
 				Expect(err).ToNot(HaveOccurred())
 
 				By("Waiting for the pod to report a successful connection attempt")
-				getStatus := func() k8sv1.PodPhase {
-					pod, err := virtClient.CoreV1().Pods(job.Namespace).Get(job.Name, k8smetav1.GetOptions{})
-					Expect(err).ToNot(HaveOccurred())
-					return pod.Status.Phase
-				}
-				Eventually(getStatus, 60, 1).Should(Equal(k8sv1.PodSucceeded))
+				waitForJobToCompleteWithStatus(&virtClient, job, "success", 60)
 			})
 		})
 
@@ -178,12 +187,7 @@ var _ = Describe("Expose", func() {
 					Expect(err).ToNot(HaveOccurred())
 
 					By("Waiting for the pod to report a successful connection attempt")
-					getStatus := func() k8sv1.PodPhase {
-						pod, err := virtClient.CoreV1().Pods(job.Namespace).Get(job.Name, k8smetav1.GetOptions{})
-						Expect(err).ToNot(HaveOccurred())
-						return pod.Status.Phase
-					}
-					Eventually(getStatus, 60, 1).Should(Equal(k8sv1.PodSucceeded))
+					waitForJobToCompleteWithStatus(&virtClient, job, "success", 60)
 				}
 			})
 		})
@@ -219,12 +223,7 @@ var _ = Describe("Expose", func() {
 				Expect(err).ToNot(HaveOccurred())
 
 				By("Waiting for the pod to report a successful connection attempt")
-				getStatus := func() k8sv1.PodPhase {
-					pod, err := virtClient.CoreV1().Pods(job.Namespace).Get(job.Name, k8smetav1.GetOptions{})
-					Expect(err).ToNot(HaveOccurred())
-					return pod.Status.Phase
-				}
-				Eventually(getStatus, 60, 1).Should(Equal(k8sv1.PodSucceeded))
+				waitForJobToCompleteWithStatus(&virtClient, job, "success", 60)
 			})
 		})
 
@@ -266,12 +265,7 @@ var _ = Describe("Expose", func() {
 					Expect(err).ToNot(HaveOccurred())
 
 					By("Waiting for the pod to report a successful connection attempt")
-					getStatus := func() k8sv1.PodPhase {
-						pod, err := virtClient.CoreV1().Pods(job.Namespace).Get(job.Name, k8smetav1.GetOptions{})
-						Expect(err).ToNot(HaveOccurred())
-						return pod.Status.Phase
-					}
-					Eventually(getStatus, 60, 1).Should(Equal(k8sv1.PodSucceeded))
+					waitForJobToCompleteWithStatus(&virtClient, job, "success", 60)
 				}
 			})
 		})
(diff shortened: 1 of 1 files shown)


tests/expose_test.go
```diff
@@ -468,6 +469,51 @@ var _ = Describe("Expose", func() {
...
+					errReason := err.(*errors.StatusError).ErrStatus.Reason
+					return errReason
+				}, 120*time.Second, 2*time.Second).Should(BeEquivalentTo("NotFound"), "The VM could not be deleted.")
+
+				By("Starting a pod which tries to reach the VMI via the ClusterIP service.")
+				job = tests.NewHelloWorldJob(serviceIP, servicePort)
+				job, err = virtClient.CoreV1().Pods(vm.Namespace).Create(job)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Waiting for the pod to report a failed connection attempt.")
```
rmohr: We have a lot of expose tests which already test if we can connect. Are you adding this because of the negative test, to see if we can't connect anymore after the vmi is gone?

yossisegev: Yes, exactly.
It automates a test case we have in Polarion (CNV-343).

tests/expose_test.go
```diff
@@ -468,6 +469,51 @@ var _ = Describe("Expose", func() {
...
+
+				By("Deleting the VM.")
+				Expect(virtClient.VirtualMachine(vm.Namespace).Delete(vm.Name, &k8smetav1.DeleteOptions{})).To(Succeed())
+
+				By("Waiting until VM is fully deleted.")
+				Eventually(func() k8smetav1.StatusReason {
+					_, err := virtClient.VirtualMachine(vm.Namespace).Get(vm.Name, &k8smetav1.GetOptions{})
+					Expect(err).To(HaveOccurred())
+					errReason := err.(*errors.StatusError).ErrStatus.Reason
+					return errReason
```
rmohr: you can do `return errors.IsNotFound(err)`. There is a utility function which does that.

yossisegev: @SchSeba What say you? Can we skip the endpoints check here, assuming it is done in other tests?
If you do think it is necessary here - no problem, I can add it.

SchSeba: Yeah if we check that in other test case it's ok

tests/expose_test.go
```diff
@@ -468,6 +469,51 @@ var _ = Describe("Expose", func() {
...
+				}, 120*time.Second, 2*time.Second).Should(BeEquivalentTo("NotFound"), "The VM could not be deleted.")
+
+				By("Starting a pod which tries to reach the VMI via the ClusterIP service.")
+				job = tests.NewHelloWorldJob(serviceIP, servicePort)
+				job, err = virtClient.CoreV1().Pods(vm.Namespace).Create(job)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Waiting for the pod to report a failed connection attempt.")
+				getStatus = func() k8sv1.PodPhase {
```
myakove: define getStatus outside and use it since you use it twice

tests/expose_test.go
yossisegev: I added the endpoint check anyway.
And replaced the search for the "NotFound" error reason with the utils function.

tests/expose_test.go
yossisegev: Done.
Replaced the getStatus check, which is similar in all the expose tests, with a call to a new service function which performs this check.

tests/expose_test.go
```diff
@@ -460,13 +437,73 @@ var _ = Describe("Expose", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				vmPod := tests.GetRunningPodByVirtualMachineInstance(vmi, vmi.Namespace)
+				Expect(err).ToNot(HaveOccurred())
+				vmPodIpAddress := vmPod.Status.PodIP
+
+				// Get the IP address of the service's endpoint.
+				endpointsName := serviceName
+				svcEndpoints, err := virtClient.CoreV1().Endpoints(vm.Namespace).Get(endpointsName, k8smetav1.GetOptions{})
+				Expect(svcEndpoints).ToNot(BeNil())
+				endptSubsetIpAddress := func() string {
```
SchSeba: No need for this it should be only one address (the virt-launcher pod) if there is more we should fail to test

tests/expose_test.go
```diff
@@ -460,13 +437,73 @@ var _ = Describe("Expose", func() {
...
+						for _, subsetAddr := range epSubset.Addresses {
+							return subsetAddr.IP
+						}
+					}
+					return ""
 				}
+				Eventually(endptSubsetIpAddress).ShouldNot(BeEquivalentTo(vmPodIpAddress))
+
+				svcEndpoints, err = virtClient.CoreV1().Endpoints(vm.Namespace).Get(endpointsName, k8smetav1.GetOptions{})
+				Expect(svcEndpoints).ToNot(BeNil())
```
SchSeba: same comment as below

tests/expose_test.go
```diff
@@ -460,13 +437,73 @@ var _ = Describe("Expose", func() {
...
+					return ""
+				}
+				Eventually(endptSubsetIpAddress).Should(BeEquivalentTo(vmPodIpAddress))
+
+				By("Deleting the VM.")
+				Expect(virtClient.VirtualMachine(vm.Namespace).Delete(vm.Name, &k8smetav1.DeleteOptions{})).To(Succeed())
+				tests.WaitForVirtualMachineToDisappearWithTimeout(vmi, 120)
+
+				By("Verifying the service's endpoints IP address is deleted once the VM was deleted.")
+				svcEndpoints, err = virtClient.CoreV1().Endpoints(vm.Namespace).Get(endpointsName, k8smetav1.GetOptions{})
```
SchSeba: I think you don't need this.
after remove the vmi pod the endpoint needs to have nil as subsets

tests/expose_test.go
yossisegev: @SchSeba The Subsets member of the Endpoints struct is defined as an array; the Addresses member of the EndpointSubset strucy is also an array.
`Ports []EndpointPort`
`Addresses []EndpointAddress`
Therefore, I must go over all instances of each array, in case the IP address I'm looking for is "hiding" in the non-first subset and/or non-first address.

SchSeba: It should be only one because the service point to one pod only.
if there is multiple it should fail (the service point to multiple pods)

tests/expose_test.go
yossisegev: @SchSeba The question is what we want to verify in this step:
If we want to verify that once the VM is deleted (a VM which is exposed as a service, of course), then all subsets are removed from an endpoint, then your comment is correct, and I should fix it accordingly.
But -
if we want to verify that the related IP address is removed once the VM was deleted, then we should keep it like that.
What if the endpoint includes more than one subset (I understand it is possible if the service is exposed via more than one subset, i.e. more than one VM, in our case)? In that case I assume that only the subset, which is related to the VM I have just deleted, is removed, and the rest of the subsets still exist.

tests/expose_test.go
yossisegev: My bad, I mistakenly added this check. It is incorrect and unnecessary.

tests/expose_test.go
yossisegev: OK, in that case I agree. I'll fix this to look for one address in one subset only, and fail the test if there is more than one subset.

tests/expose_test.go
yossisegev: OK, given your comment above - I'll fix this, and verify no subset exist anymore.

tests/expose_test.go
```diff
@@ -460,13 +437,73 @@ var _ = Describe("Expose", func() {
...
+				vmi, err := virtClient.VirtualMachineInstance(vm.Namespace).Get(vm.Name, &k8smetav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				vmPod := tests.GetRunningPodByVirtualMachineInstance(vmi, vmi.Namespace)
+				Expect(err).ToNot(HaveOccurred())
+				vmPodIpAddress := vmPod.Status.PodIP
+
+				// Get the IP address of the service's endpoint.
+				endpointsName := serviceName
+				svcEndpoints, err := virtClient.CoreV1().Endpoints(vm.Namespace).Get(endpointsName, k8smetav1.GetOptions{})
+				Expect(svcEndpoints).ToNot(BeNil())
```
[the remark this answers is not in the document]
yossisegev: You are right, I missed this error check.
Done.

tests/expose_test.go
yossisegev: Done.

tests/expose_test.go
yossisegev: Done.

tests/expose_test.go
yossisegev: Done.

PR #2042, opened 2019-02-20
Automate HTTP expose test

yossisegev:
Automate HTTP expose test:
Verify a VM exposed as service can be used for HTTP connection.

### DIFF
--- a/tests/expose_test.go
+++ b/tests/expose_test.go
@@ -99,7 +99,7 @@ var _ = Describe("Expose", func() {
 				Expect(err).ToNot(HaveOccurred())
 
 				By("Waiting for the pod to report a successful connection attempt")
-				waitForJobToCompleteWithStatus(&virtClient, job, "success", 60)
+				waitForJobToCompleteWithStatus(&virtClient, job, "success", 120)
 			})
 		})
 
@@ -187,7 +187,7 @@ var _ = Describe("Expose", func() {
 					Expect(err).ToNot(HaveOccurred())
 
 					By("Waiting for the pod to report a successful connection attempt")
-					waitForJobToCompleteWithStatus(&virtClient, job, "success", 60)
+					waitForJobToCompleteWithStatus(&virtClient, job, "success", 120)
 				}
 			})
 		})
@@ -223,7 +223,7 @@ var _ = Describe("Expose", func() {
 				Expect(err).ToNot(HaveOccurred())
 
 				By("Waiting for the pod to report a successful connection attempt")
-				waitForJobToCompleteWithStatus(&virtClient, job, "success", 60)
+				waitForJobToCompleteWithStatus(&virtClient, job, "success", 120)
 			})
 		})
 
@@ -265,7 +265,7 @@ var _ = Describe("Expose", func() {
 					Expect(err).ToNot(HaveOccurred())
 
 					By("Waiting for the pod to report a successful connection attempt")
-					waitForJobToCompleteWithStatus(&virtClient, job, "success", 60)
+					waitForJobToCompleteWithStatus(&virtClient, job, "success", 120)
 				}
 			})
 		})
@@ -326,7 +326,7 @@ var _ = Describe("Expose", func() {
 				Expect(err).ToNot(HaveOccurred())
 
 				By("Waiting for the pod to report a successful connection attempt")
-				waitForJobToCompleteWithStatus(&virtClient, job, "success", 60)
+				waitForJobToCompleteWithStatus(&virtClient, job, "success", 120)
 			})
 		})
 	})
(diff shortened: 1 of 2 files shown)


yossisegev:
@rmohr This is a simple validation, to verify that the specific, yet commonly-used, use-case of HTTP traffic is not broken in the flow of the VM exposure as a service.
We don't implement the automation of all possible use-cases of service exposure on HTTP scenarios, as we do with TCP and UDP. We just run this single, simple test, so we can be sure that if a bug occurs in HTTP traffic over exposed service - we'll catch it.

yossisegev:
@SchSeba @cynepco3hahue @rmohr 
All CI failures are not related to my PR. Can you please merge?
Thanks!

tests/utils.go
```diff
@@ -2813,6 +2813,16 @@ func NewHelloWorldJobUDP(host string, port string) *k8sv1.Pod {
 	return job
 }
 
+// NewHelloWorldJobHttp gets an IP address and a port, which it uses to create a pod.
+// This pod tries to contact the host on the provided port, over HTTP.
+// On success - it expects to receive "Hello World!".
+func NewHelloWorldJobHttp(host string, port string) *k8sv1.Pod {
```
myakove: Just out of curiosity why 'New' in the function name?

yossisegev: Same naming convention as the other 2 functions that already exist: NewHelloWorldJob and NewHelloWorldJobUDP.

yossisegev:
@SchSeba Can you please review this PR so it can be merged?
Thanks!

rmohr:
I would prefer if for instance the TCP test would be improved so that it sends more than the MTU, or something like that. This test ere literally just sends another string Over TCP. It can't catch anything in addition to the existing tcp test.

yossisegev:
@rmohr I moved the HTTP test to be part of the basic TCP test, and not a test of its own.
Regarding the improvement of TCP tests (traffic larger than MTU, etc.) - I agree we should improve them and add more complex, non-basic scenarios, but we'll do this on a different PR.

yossisegev:
> The code looks good
> This is a very basic tests I am not sure I understand the added value related to the TCP connection test.

@SchSeba it's just a simple verification that the flow of HTTP traffic is not broken, when we run this traffic using an exposed service.
It's the only HTTP test we are running for exposed services, and we are keeping the more complex scenarios to be tested with TCP and UDP (and as Roman mentioned above - we should add more possible scenarios). From my P.O.V, I prefer adding this small HTTP test for exposed services, instead of discovering someday that this flow is actually broken due to a bug, that should have been revealed a long time ago.

SchSeba:
@yossisegev I am fine with that

rmohr:
ci test please

yossisegev:
@rmohr The failure in the latest CI, which **[you](https://github.com/kubevirt/kubevirt/pull/2042#issuecomment-466958946)** initiated, is because of other errors, not related to my changes.
This PR is getting old and bitter... Will it **ever** be merged?

rmohr:
> This PR is getting old and bitter... Will it **ever** be merged?

@yossisegev so, you had immediately a review once you pushed the test. I pointed out that the test you added had no value. Then there was weekend. Then I finally convinced you that the test has no value. You changed it on Sunday. I run a retest on monday. What do you mean with **ever**?

tests/expose_test.go
```diff
@@ -388,7 +388,20 @@ var _ = Describe("Expose", func() {
...
-				waitForJobToCompleteWithStatus(&virtClient, job, "success", 60)
+				waitForJobToCompleteWithStatus(&virtClient, job, "success", 120)
+
+				By("Starting an HTTP server on the VM, to verify HTTP connection also succeeds using the exposed VM.")
+				vmi, err := virtClient.VirtualMachineInstance(vm.Namespace).Get(vm.Name, &k8smetav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				tests.StartHTTPServer(vmi, testPort)
+
+				By("Starting a pod which tries to reach the VMI via ClusterIP over HTTP.")
+				job = tests.NewHelloWorldJobHttp(serviceIP, servicePort)
```
rmohr: If you really insist on checking `http` explicitly and not arbitrary `tcp`, then please change this test: https://github.com/kubevirt/kubevirt/blob/master/tests/vmi_networking_test.go#L299 which checks cluster IP services with tcp traffic. However I don't see why `http` should be more important here than `tls`  or `ssh`, or why it should work differently. We are talking about cluster IP services and not about openshift routes which only work for `tls` and `http`.

rmohr: Ah ok, fine for me like it is. Still I don't see the value.

PR #2046, opened 2019-02-21
Add mirrors for the fedora image

cynepco3hahue:
Add mirrors for the Fedora image
I chose one with the biggest bandwidth from https://admin.fedoraproject.org/mirrormanager/mirrors/Fedora/29

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,9 +56,8 @@ functest:
 	hack/dockerized "hack/build-func-tests.sh"
 	hack/functests.sh
 
-clean:
+clean: bazel-clean
 	hack/dockerized "./hack/build-go.sh clean ${WHAT} && rm _out/* -rf"
-	hack/dockerized "bazel clean --expunge"
 	rm -f tools/openapispec/openapispec tools/resource-generator/resource-generator tools/manifest-templator/manifests-templator tools/vms-generator/vms-generator
 
 distclean: clean
@@ -126,6 +125,7 @@ builder-publish:
 .PHONY: \
 	bazel-generate \
 	bazel-build \
+	bazel-clean \
 	bazel-push-images \
 	bazel-tests \
 	build \
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,14 @@ bazel-build:
 		--workspace_status_command=./hack/print-workspace-status.sh \
 		//cmd/..."
 
+bazel-build-images:
+	hack/dockerized "bazel build \
+		--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 \
+		--workspace_status_command=./hack/print-workspace-status.sh \
+		--define container_prefix=${CONTAINER_PREFIX} \
+		--define container_tag=${CONTAINER_TAG} \
+		//:build-images"
+
 bazel-push-images:
 	hack/dockerized "bazel run \
 		--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 \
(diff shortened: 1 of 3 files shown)


cynepco3hahue:
@slintes I still can see the problem with Fedora image, so I do not really sure it helps.
I think the problem, that download partially succeeds, but after checking the hash sum fails.

slintes:
Too bad. But still better than only having 1 url only, that way we at least prevent errors when a mirror is down.
Maybe even remove download.fedoraproject.org? Maybe that's the only one causing problems?

cynepco3hahue:
@rmohr Checked bazel cache, and it has the Fedora image, so looks like we have some problems with connectivity to bazel cache

cynepco3hahue:
@rmohr Propose to build artifacts, before running the cluster, so I will try it

rmohr:
@cynepco3hahue removed the cache clean and the error is back. Could it be that stdci recycles the docker volume and your `clean`  really did the trick?

rmohr:
But still not sure if it wouldn't have occurred with the cache clean again.

cynepco3hahue:
@rmohr Clean does not really help, I still can see 3 lanes with the same error. Also, I added `curl` before and it succeeds to download the image without any troubles. I currently do not have any other ideas, any thoughts?

gbenhaim:
- Currently, docker volumes are not being kept between jobs. Can we use a bind mount instead of a docker volume?

- It indeed looks like the download didn't finished successfully, thus the sha256 was wrong. Do we know from where it tried to download the image `greenhouse` or fedora mirrors?

rmohr:
> Currently, docker volumes are not being kept between jobs. Can we use a bind mount instead of a docker volume?

That is perfect. We don't want to keep them. 

> It indeed looks like the download didn't finished successfully, thus the sha256 was wrong. Do we know from where it tried to download the image greenhouse or fedora mirrors?

We still have the Fedora mirror on the first place. So it is still preferred. Maybe we just have to move it to the end of the list. 

Gal could you also verify that all nodes have sufficient space left for docker in docker? It still looks a little bit like it might be related to a single node issue.

gbenhaim:
> Currently, docker volumes are not being kept between jobs. Can we use a bind mount instead of a docker volume?

That is perfect. We don't want to keep them.

> It indeed looks like the download didn't finished successfully, thus the sha256 was wrong. Do we know from where it tried to download the image greenhouse or fedora mirrors?

We still have the Fedora mirror on the first place. So it is still preferred. Maybe we just have to move it to the end of the list.

>> I would like to understand if the connectivity issue is between the ci pod and greenhouse pod, or between the ci pod and the fedora mirror. How can we know if it failed to fetch the image from cache or not?

Gal could you also verify that all nodes have sufficient space left for docker in docker? It still looks a little bit like it might be related to a single node issue.

I've checked it now, all nodes have 200GB+ free space.

cynepco3hahue:
ci test please

rmohr:
ci test please

rmohr:
ci test please

cynepco3hahue:
@rmohr @slintes 
I think I will close this PR, it just does not solve anything, we still have two problems:
- from some reason, build does not take fedora image from greenhouse
- it downloads only part of the image, so checksum does not correct(I think it some bazel bug)

rmohr:
We should really try downgrading to bazel 0.21.

rmohr:
Closing this one works for me.

PR #2047, opened 2019-02-21
Add annotaion to the virt-launcher template for istio

SchSeba:
Related to a work done in istio https://github.com/istio/istio/pull/11321
We introduce a new label for kubevirt to help manage the traffic getting in and out of the envoy process

Fixes #

**Release note**:

```release-note
Istio integration with masquerade
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/2047)

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -73,6 +73,9 @@ const NFD_CPU_FEATURE_PREFIX = "feature.node.kubernetes.io/cpu-feature-"
 
 const MULTUS_RESOURCE_NAME_ANNOTATION = "k8s.v1.cni.cncf.io/resourceName"
 
+// Istio list of virtual interfaces whose inbound traffic (from VM) will be treated as outbound traffic in envoy
+const ISTIO_KUBEVIRT_ANNOTATION = "traffic.sidecar.istio.io/kubevirtInterfaces"
+
 type TemplateService interface {
 	RenderLaunchManifest(*v1.VirtualMachineInstance) (*k8sv1.Pod, error)
 }
@@ -868,6 +871,10 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		annotationsList[k] = v
 	}
 
+	if HaveMasqueradeInterface(vmi.Spec.Domain.Devices.Interfaces) {
+		annotationsList[ISTIO_KUBEVIRT_ANNOTATION] = "k6t-eth0"
+	}
+
 	// TODO use constants for podLabels
 	pod := k8sv1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
@@ -1027,6 +1034,16 @@ func getPortsFromVMI(vmi *v1.VirtualMachineInstance) []k8sv1.ContainerPort {
 	}
 
 	return ports
+}
+
+func HaveMasqueradeInterface(interfaces []v1.Interface) bool {
+	for _, iface := range interfaces {
+		if iface.Masquerade != nil {
+			return true
+		}
+	}
+
+	return false
 }
 
 func getResourceNameForNetwork(network *networkv1.NetworkAttachmentDefinition) string {
--- a/pkg/virt-controller/watch/vmi.go
+++ b/pkg/virt-controller/watch/vmi.go
@@ -424,7 +424,10 @@ func isPodReady(pod *k8sv1.Pod) bool {
 				return false
 			}
 		} else {
-			if containerStatus.Ready == false {
+			// When using istio the istio-proxy container will not be ready
+			// until there is a service pointing to this pod.
+			// We need to start the VM anyway
+			if containerStatus.Name != "istio-proxy" && containerStatus.Ready == false {
 				return false
 			}
 		}
--- a/cluster/k8s-multus-1.13.3/provider.sh
+++ b/cluster/k8s-multus-1.13.3/provider.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-image="k8s-multus-1.13.3@sha256:2ae2dfd20adf163e9cb3923c713932353670692d119f330e547aa1881872de58"
+image="k8s-multus-1.13.3@sha256:23cd2fca6e9d2468256921b4fd7bd5e9aaf0e3b03af742066883e39e5d29d349"
 
 source cluster/ephemeral-provider-common.sh
 
--- a/pkg/virt-controller/services/template_test.go
+++ b/pkg/virt-controller/services/template_test.go
@@ -221,6 +221,33 @@ var _ = Describe("Template", func() {
 				Expect(value).To(Equal("default,test1"))
 			})
 		})
+		Context("with masquerade interface", func() {
+			It("should add the istio annotation", func() {
+				vmi := v1.VirtualMachineInstance{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "testvmi",
+						Namespace: "default",
+						UID:       "1234",
+					},
+					Spec: v1.VirtualMachineInstanceSpec{
+						Domain: v1.DomainSpec{
+							Devices: v1.Devices{
+								Interfaces: []v1.Interface{
+									{Name: "default",
+										InterfaceBindingMethod: v1.InterfaceBindingMethod{Masquerade: &v1.InterfaceMasquerade{}}},
+								},
+							},
+						},
+					},
+				}
+
+				pod, err := svc.RenderLaunchManifest(&vmi)
+				Expect(err).ToNot(HaveOccurred())
+				value, ok := pod.Annotations[ISTIO_KUBEVIRT_ANNOTATION]
+				Expect(ok).To(Equal(true))
+				Expect(value).To(Equal("k6t-eth0"))
+			})
+		})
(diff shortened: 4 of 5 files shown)


booxter:
Very surprised that istio accepted the name of the annotation that includes ```kubevirt``` in it.

dankenigsberg:
> Very surprised that istio accepted the name of the annotation that includes `kubevirt` in it.

I believe they have *requested* this name. I was surprised, too.

davidvossel:
Can we get a user-guide update that explains the relationship between the masquerade interface and isto? I imagine users need to be aware that this only works with a specific version of istio built after your functionality made it in as well.

davidvossel:
once an istio release supports the annotation, it would be nice to add this to our functional test suite as well.

SchSeba:
Hi @davidvossel thanks for the comment.

about the functional test I will create a new provider for istio there already exist a Istio 1.1.0-rc.0 release.

davidvossel:
> about the functional test I will create a new provider for istio there already exist a Istio 1.1.0-rc.0 release

couldn't this be a functional test within our current providers?

SchSeba:
@davidvossel I can create a functional test but it will be the same as the unit test to check if the annotation exist but for a real functional test I need the istio control plane to be installed inside the cluster.

SchSeba:
Hi @davidvossel I also add to the PR the ability to copy annotations from the vmi to the pod I will like to know if this is something we are interesting to allow.

booxter:
@SchSeba this copying of annotations from VMI to pod doesn't seem related to istio, is it? Perhaps we could have it separated into a different PR.

pkg/virt-controller/services/template.go
```diff
@@ -858,6 +858,12 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		v1.DomainAnnotation: domain,
 	}
 
+	if vmi.ObjectMeta.Annotations != nil {
+		for key, value := range vmi.ObjectMeta.Annotations {
```
dankenigsberg: I have to join to @booxter here: are we sure that *all* vmi annotations are meaningful and harmless when applied onto the launcher Pod? I'm not convinced yet, and appreciate a separate PR with a proper explanation.

SchSeba: no problem I will remove it 

@booxter I just use it to tell istio to use the service mesh only for cluster and pod ip's so I am able to install things like nginx from the yum repo

``` yam,
kind: VirtualMachineInstance
metadata:
  labels:
    special: vmi-masquerade
  annotations:
    traffic.sidecar.istio.io/includeOutboundIPRanges: "10.96.0.0/12,10.244.0.0/24"
  name: vmi-masquerade
...
```

SchSeba:
@davidvossel @booxter I remove the annotation copy from vmi and add a new check in the virt-controller can you please review this PR again?

pkg/virt-controller/services/template.go
booxter: @SchSeba if there is a use case where we may want to allow users to specify annotations for their pods in VMI specs (and we should analyze whether it's safe to do it; open ended features are often vectors for attack), then perhaps, instead of overloading the meaning of VMI-specific annotations, we could introduce a separate attribute specifically to carry custom annotations for pods. But let's discuss this in a separate venue.

pkg/virt-controller/watch/vmi.go
```diff
@@ -424,7 +424,7 @@ func isPodReady(pod *k8sv1.Pod) bool {
 				return false
 			}
 		} else {
-			if containerStatus.Ready == false {
+			if containerStatus.Name != "istio-proxy" && containerStatus.Ready == false {
```
booxter: As an uninitiated newbie, two questions:

1. where is this name - ```istio-proxy``` - defined? Is it guaranteed to always have this exact form?
2. why do we think it's safe to proceed even if the Istio proxy container is not ready yet?

Both questions should probably be clarified with corresponding comments in the code itself.

booxter:
@SchSeba I have two silly questions, please take a look. Also, wasn't it originally the plan to add functional tests here? Or does it belong to a separate PR?

SchSeba:
Hi @booxter I think the functional test needs to be part of a separate PR. 
I need to build a new test lane for it that has istio install and configured.

pkg/virt-controller/watch/vmi.go
SchSeba: Done

booxter: @SchSeba thanks. Anything about the name of the container? Is it guaranteed to be this way in all cases?

SchSeba: This is the default if you really want you can change this in the istio-injection-sidecar-configmap

SchSeba:
ci test please

SchSeba:
ci test please

SchSeba:
for reference 
https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/211/artifact/ci_build_summary.html

SchSeba:
ci test please

pkg/virt-controller/watch/vmi.go
booxter: I guess ideally we would accommodate for a different name if configmap is modified, but perhaps it's not critical.

SchSeba:
@davidvossel @rmohr can you please take a look I think this PR is ready to get merge :)

pkg/virt-controller/services/template.go
```diff
@@ -69,6 +69,8 @@ const NFD_CPU_FEATURE_PREFIX = "feature.node.kubernetes.io/cpu-feature-"
 
 const MULTUS_RESOURCE_NAME_ANNOTATION = "k8s.v1.cni.cncf.io/resourceName"
 
+const ISTIO_KUBEVIRT_ANNOTATION = "traffic.sidecar.istio.io/kubevirtInterfaces"
```
davidvossel: can we get a comment in the code that gives us a hint about what this annotation does?

pkg/virt-controller/watch/vmi.go
```diff
@@ -424,7 +424,10 @@ func isPodReady(pod *k8sv1.Pod) bool {
 				return false
 			}
 		} else {
-			if containerStatus.Ready == false {
+			// When using istio the istio-proxy container will not be ready
+			// until there is a service pointing to this pod.
+			// We need to start the VM anyway
```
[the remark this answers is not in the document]
SchSeba: Hi @davidvossel thanks for the comment 

> why do we need to start the VM anyway? is there a reason we shouldn't wait until there is a service point to the pod?

when we have a regular pod the application inside the pod will start even if all the containers are not ready.

If we don't add the check here the VM will not start until there is a service pointing to it but when we create a service it will take time until the application is ready (vm boot time and application initialize) and the is not the case for regular pods. So I wanted to preserve the same behavior like regular pods.

> In general, I don't feel comfortable with hardcoded exceptions like this. There's also no guarantee the sidecar will be called "istio-proxy"

I also don't like to do it but I don't want to implement a module of searching in the istio configmaps to find what is the name of the proxy. I prefer for now to document it (the use of hard coded proxy name)

pkg/virt-controller/services/template.go
SchSeba: Yes for sure I will add

pkg/virt-controller/watch/vmi.go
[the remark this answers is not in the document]
davidvossel: is there anything we can key off of besides name to detect the istio container more generically?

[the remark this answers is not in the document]
SchSeba: I wasn't able to find anything else :(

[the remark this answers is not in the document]
davidvossel: umm. okay. I don't want to block us on that then.

We'll need a unit test for this and make a note about this behavior in the user guide.

pkg/virt-controller/services/template.go
```diff
@@ -861,6 +864,10 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		annotationsList[cniAnnotation] = cniNetworks
 	}
 
+	if HaveMasqueradeInterface(vmi.Spec.Domain.Devices.Interfaces) {
+		annotationsList[ISTIO_KUBEVIRT_ANNOTATION] = "k6t-eth0"
```
booxter: This seems like a hardcoded name (eth0 part). I suspect at one point it may change, breaking this integration. We'll need functional tests to catch this future scenario.

SchSeba: Functional will come in a separate PR because I need to create a provider for that.

pkg/virt-controller/watch/vmi.go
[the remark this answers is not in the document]
SchSeba: done can you please review :)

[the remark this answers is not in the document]
SchSeba: I will open a PR for the user guide

booxter:
@SchSeba I am merging this (since David was already in general agreement with the PR), but please propose user-guide change and functional tests / provider. Link them here. Thanks.

PR #2073, opened 2019-03-04
Reject migration of VMIs with non-migratable condition

### LINKED ISSUE #2072: Migration of VMIs with a non-migratable conditions are not rejected
/kind bug
**What happened**:
In some cases, migrations of VMIs with a non-migratable condition is not being rejected.

vladikr:
This PR fixes an API validation that should reject migration of VMIs with a non-migratable condition
and adds a functional test to verify this.

Fixes #2072

```release-note
NONE
```

### DIFF
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -373,7 +373,57 @@ var _ = Describe("Migrations", func() {
 
 				By("Waiting for VMI to disappear")
 				tests.WaitForVirtualMachineToDisappearWithTimeout(vmi, 240)
+			})
+		})
+		Context("with an Cirros non-shared ISCSI PVC", func() {
+			var pvName string
+			BeforeEach(func() {
+				pvName = "test-iscsi-lun" + rand.String(48)
+				// Start a ISCSI POD and service
+				By("Starting an iSCSI POD")
+				iscsiIP := tests.CreateISCSITargetPOD(tests.ContainerDiskCirros)
+				// create a new PV and PVC (PVs can't be reused)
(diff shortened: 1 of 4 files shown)


tests/migration_test.go
```diff
@@ -265,6 +266,58 @@ var _ = Describe("Migrations", func() {
...
+				expecter.Close()
+
+				for _, c := range vmi.Status.Conditions {
+					if c.Type == v1.VirtualMachineInstanceIsMigratable {
+						Expect(c.Status).To(Equal(k8sv1.ConditionFalse))
+					}
+				}
+
+				// execute a migration, wait for finalized state
+				By("Starting the Migration for iteration")
```
cynepco3hahue: I think you can delete this `By`

vladikr:
> Is there some subtle change in behavior here that fixes something?

Yeah. I didn't understand this at first, took me a while to debug.  `getVMIMigrationCondition` was returning a pointer to an element in the `vmi.Status.Conditions` array that has changed as `range` progressed, returing only the last element. I've simplified all of that and got rid of the helper functions.
```
func getVMIMigrationCondition(vmi *v1.VirtualMachineInstance) (cond *v1.VirtualMachineInstanceCondition) {
for _, c := range vmi.Status.Conditions {	
		if c.Type == v1.VirtualMachineInstanceIsMigratable {	
			cond = &c	
		}	
	}	
	return cond
}
```

vladikr:
Migration tests are passing and this PR doesn't affect anything else. I'll merge this.

PR #2077, opened 2019-03-04
metrics: prometheus: more storage metrics

fromanirh:

We need few more metrics to report data which kubevirt-ui wants to consume, see
https://github.com/kubevirt/web-ui-components/pull/247

Add metrics to report:
- per-drive per-VM I/O traffic aka bytes read/write total
- per-drive per-VM I/O operational times aka total time spent doing I/O
  ops, needed to compute the average latency time (e.g. read latency).
```release-note
NONE
```

### DIFF
--- a/tests/infra_test.go
+++ b/tests/infra_test.go
@@ -23,11 +23,16 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"sort"
+	"strconv"
 	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
+	expect "github.com/google/goexpect"
+
 	k8sv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
@@ -108,32 +113,81 @@ var _ = Describe("Infrastructure", func() {
 			Expect(err).ToNot(HaveOccurred(), "Should create VMI")
 
 			By("Waiting until the VM is ready")
-			tests.WaitForSuccessfulVMIStart(obj)
+			nodeName := tests.WaitForSuccessfulVMIStart(obj)
 
 			By("Finding the prometheus endpoint")
-			l, err := labels.Parse("kubevirt.io=virt-handler")
-			Expect(err).ToNot(HaveOccurred())
-			pods, err := virtClient.CoreV1().Pods(tests.KubeVirtInstallNamespace).List(metav1.ListOptions{LabelSelector: l.String()})
-			Expect(err).ToNot(HaveOccurred())
-			Expect(pods.Items).ToNot(BeEmpty())
+			pod, err := kubecli.NewVirtHandlerClient(virtClient).ForNode(nodeName).Pod()
+			Expect(err).ToNot(HaveOccurred(), "Should find the virt-handler pod")
 
 			By("Scraping the Prometheus endpoint")
-			pod := pods.Items[0] // only one compute node in the test environment
-
 			Eventually(func() string {
-				stdout, _, err := tests.ExecuteCommandOnPodV2(virtClient,
-					&pod, "virt-handler",
-					[]string{
-						"curl",
-						"-L",
-						"-k",
-						fmt.Sprintf("https://%s:%s/metrics", pod.Status.PodIP, "8443"),
-					})
-				Expect(err).ToNot(HaveOccurred())
-				lines := filterMetricsOutput(stdout, "kubevirt")
+				out := getKubevirtVMMetrics(virtClient, pod, "virt-handler")
+				lines := takeMetricsWithPrefix(out, "kubevirt")
 				return strings.Join(lines, "\n")
 			}, 30*time.Second, 2*time.Second).Should(ContainSubstring("kubevirt"))
 		}, 300)
+
+		It("should include the storage metrics for a running VM", func() {
+			By("Creating the VirtualMachineInstance")
+
+			// WARNING: we assume the VM will have a VirtIO disk (vda)
+			// and we add our own vdb on which we do our test.
+			// but if the default disk is not vda, the test will break
+			// TODO: introspect the VMI and get the device name of this
+			// block device?
+			vmi := tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskAlpine))
+			tests.AppendEmptyDisk(vmi, "testdisk", "virtio", "1Gi")
+
+			By("Starting a new VirtualMachineInstance")
+			obj, err := virtClient.RestClient().Post().Resource("virtualmachineinstances").Namespace(tests.NamespaceTestDefault).Body(vmi).Do().Get()
+			Expect(err).ToNot(HaveOccurred(), "Should create VMI")
+
+			By("Waiting until the VM is ready")
+			nodeName := tests.WaitForSuccessfulVMIStart(obj)
+
+			By("Expecting the VirtualMachineInstance console")
+			expecter, err := tests.LoggedInAlpineExpecter(vmi)
+
+			Expect(err).ToNot(HaveOccurred())
+			defer expecter.Close()
+
+			By("Writing some data to the disk")
+			_, err = expecter.ExpectBatch([]expect.Batcher{
+				&expect.BSnd{S: "dd if=/dev/zero of=/dev/vdb bs=1M count=1\n"},
+				&expect.BExp{R: "localhost:~#"},
+				&expect.BSnd{S: "sync\n"},
+				&expect.BExp{R: "localhost:~#"},
+			}, 10*time.Second)
+			Expect(err).ToNot(HaveOccurred())
+			// we wrote data to the disk, so from now on the VM *is* running
+
+			By("Finding the prometheus endpoint")
+			pod, err := kubecli.NewVirtHandlerClient(virtClient).ForNode(nodeName).Pod()
+			Expect(err).ToNot(HaveOccurred(), "Should find the virt-handler pod")
+
+			By("Scraping the Prometheus endpoint")
+			// the VM *is* running, so we must have metrics promptly reported
+			out := getKubevirtVMMetrics(virtClient, pod, "virt-handler")
+			lines := takeMetricsWithPrefix(out, "kubevirt")
+			metrics, err := parseMetricsToMap(lines)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(metrics).To(HaveKey(ContainSubstring("kubevirt_vm_storage_")))
+
+			By("Checking the collected metrics")
+			var keys []string
+			for metric := range metrics {
+				keys = append(keys, metric)
+			}
+			// we sort keys only to make debug of test failures easier
+			sort.Strings(keys)
+			for _, key := range keys {
+				if strings.HasPrefix(key, "kubevirt_vm_storage_") && strings.Contains(key, "vdb") {
+					value := metrics[key]
+					Expect(value).To(BeNumerically(">", float64(0.0)))
+				}
+			}
+		}, 300)
+
 	})
 
 	Describe("Start a VirtualMachineInstance", func() {
@@ -212,7 +266,36 @@ func getNewLeaderPod(virtClient kubecli.KubevirtClient) *k8sv1.Pod {
 	return nil
 }
 
-func filterMetricsOutput(output, prefix string) []string {
+func getKubevirtVMMetrics(virtClient kubecli.KubevirtClient, pod *k8sv1.Pod, containerName string) string {
+	stdout, _, err := tests.ExecuteCommandOnPodV2(virtClient,
+		pod, containerName,
+		[]string{
+			"curl",
+			"-L",
+			"-k",
+			fmt.Sprintf("https://%s:%s/metrics", pod.Status.PodIP, "8443"),
+		})
+	Expect(err).ToNot(HaveOccurred())
+	return stdout
+}
+
+func parseMetricsToMap(lines []string) (map[string]float64, error) {
+	metrics := make(map[string]float64)
+	for _, line := range lines {
+		items := strings.Split(line, " ")
+		if len(items) != 2 {
+			return nil, fmt.Errorf("can't split properly line '%s'", line)
+		}
+		v, err := strconv.ParseFloat(items[1], 64)
+		if err != nil {
+			return nil, err
+		}
+		metrics[items[0]] = v
+	}
+	return metrics, nil
+}
+
+func takeMetricsWithPrefix(output, prefix string) []string {
 	lines := strings.Split(output, "\n")
 	var ret []string
 	for _, line := range lines {
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -3212,3 +3212,22 @@ func GetVmPodName(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineInstance
 
 	return podName
 }
+
+func AppendEmptyDisk(vmi *v1.VirtualMachineInstance, diskName, busName, diskSize string) {
+	vmi.Spec.Domain.Devices.Disks = append(vmi.Spec.Domain.Devices.Disks, v1.Disk{
+		Name: diskName,
+		DiskDevice: v1.DiskDevice{
+			Disk: &v1.DiskTarget{
+				Bus: busName,
+			},
+		},
+	})
+	vmi.Spec.Volumes = append(vmi.Spec.Volumes, v1.Volume{
+		Name: diskName,
+		VolumeSource: v1.VolumeSource{
+			EmptyDisk: &v1.EmptyDiskSource{
+				Capacity: resource.MustParse(diskSize),
+			},
+		},
+	})
+}
--- a/pkg/monitoring/vms/prometheus/prometheus.go
+++ b/pkg/monitoring/vms/prometheus/prometheus.go
@@ -48,7 +48,18 @@ var (
 		[]string{"domain", "drive", "type"},
 		nil,
 	)
-	// from now on: TODO: validate
+	storageTrafficDesc = prometheus.NewDesc(
+		"kubevirt_vm_storage_traffic_bytes_total",
+		"storage traffic.",
+		[]string{"domain", "drive", "type"},
+		nil,
+	)
+	storageTimesDesc = prometheus.NewDesc(
+		"kubevirt_vm_storage_times_ms_total",
+		"storage operation time.",
+		[]string{"domain", "drive", "type"},
+		nil,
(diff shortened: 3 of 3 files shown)


fromanirh:
> Looks good in general. Can we on an existing metrics test add a few more disks to a vmi, write a little bit and check if the metrics appear in the expected fasion?

of course, thanks for the early review. I totally want to add more tests.

fromanirh:
ci test please

fromanirh:
ping?

fromanirh:
Most of tests fails with
ERROR: Analysis of target '//:build-images' failed; build aborted: no such package '@libstdc//file': java.io.IOException: Error downloading [https://dl.fedoraproject.org/pub/fedora/linux/updates/28/Everything/x86_64/Packages/l/libstdc++-8.2.1-6.fc28.x86_64.rpm] to /root/.cache/bazel/_bazel_root/6f347497f91c9a385dcd9294645b76e0/external/libstdc/file/downloaded: GET returned 404 Not Found
INFO: Elapsed time: 90.080s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (26 packages loaded, 47 targets configured)
FAILED: Build did NOT complete successfully (26 packages loaded, 47 targets configured)

slintes:
that was fixed by https://github.com/kubevirt/kubevirt/pull/2117

slintes:
ci test please

fromanirh:
> that was fixed by #2117

Thanks @slintes ! should I rebase this PR on top of git master?

tests/infra_test.go
```diff
@@ -121,19 +126,79 @@ var _ = Describe("Infrastructure", func() {
...
 		}, 300)
+
+		It("should include the storage metrics for a running VM", func() {
+			By("Creating the VirtualMachineInstance")
+
+			// WARNING: we assume the VM will have a VirtIO disk (vda)
+			vmi := tests.NewRandomVMIWithWatchdog()
+			blockPVName := "block-pv-storage-metrics"
+			tests.CreateBlockVolumePvAndPvc(blockPVName, "1Gi")
+			defer tests.DeletePvAndPvc(blockPVName)
```
rmohr: Could you just add an `emptyDisk`? Then you don't have to do any storage management.

tests/infra_test.go
```diff
@@ -121,19 +126,79 @@ var _ = Describe("Infrastructure", func() {
...
+
+			//... and we add our own vdb on which we do our test.
+			// but if the default disk is not vda, the test will break
+			// TODO: introspect the VMI and get the device name of this
+			// block device?
+			tests.AddPVCDisk(vmi, "block-pvc", "virtio", blockPVName)
```
rmohr: Same here? Just another `emptyDisk`?

tests/infra_test.go
rmohr: Here some background to `emptyDisk`: https://kubevirt.io/user-guide/docs/latest/creating-virtual-machines/disks-and-volumes.html

fromanirh: Neat feature, thanks for the suggestion. Changed accordingly.

tests/infra_test.go
fromanirh: Done

tests/infra_test.go
```diff
@@ -212,6 +273,35 @@ func getNewLeaderPod(virtClient kubecli.KubevirtClient) *k8sv1.Pod {
...
+}
+
+func parseMetricsToMap(lines []string) (map[string]float64, error) {
+	metrics := make(map[string]float64)
+	for _, line := range lines {
+		items := strings.Split(line, " ")
+		if len(items) != 2 {
+			return nil, fmt.Errorf("can't split properly line '%s'", line)
+		}
+		v, err := strconv.ParseFloat(items[1], 64)
```
rmohr: Can this fail because of parsing a comment line?

fromanirh: Good point. Let me check this - also I'll try to add a test for it

fromanirh: @rmohr we sidestep this issue thanks to filterMetricsOutput (see below https://github.com/kubevirt/kubevirt/pull/2077/files/262a8dee9be2f08ba27adfcd2d5583513906b73c#diff-29539fc2ce2beadde4ecbf90b4a00816L215), so the input of this function is a slice of lines which start with "kubevirt_" (regex equivalent would be "^kubevirt"). No comments could get there.
I'll push a small patch to rename filterMetricsOutput and to move it in the calling sites, to make this more evident. In a followup PR I'd like to extract this code and move it in a small helper package under `pkg/util`, with unit tests. How this plan sound?

rmohr: @fromanirh making it a little bit more evident that the output is already filtered is enough for me. You don't have to factor it out to pkg/util if you don't need it in the normal code from my point of view.

fromanirh: @rmohr ack, on it.

tests/infra_test.go
```diff
@@ -121,19 +126,77 @@ var _ = Describe("Infrastructure", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			pods, err := virtClient.CoreV1().Pods(tests.KubeVirtInstallNamespace).List(metav1.ListOptions{LabelSelector: l.String()})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(pods.Items).ToNot(BeEmpty())
+
+			By("Scraping the Prometheus endpoint")
+			pod := pods.Items[0] // only one compute node in the test environment
+
+			var lines []string
+			for tryNum := 0; tryNum < 20; tryNum++ {
```
rmohr: Why is that loop needed? Is there a reason to believe that the metrics will not be scraped?

tests/infra_test.go
```diff
@@ -121,19 +126,77 @@ var _ = Describe("Infrastructure", func() {
...
+				out := getKubevirtVMMetrics(virtClient, &pod, "virt-handler")
+				lines = takeMetricsWithPrefix(out, "kubevirt")
+				time.Sleep(2 * time.Second)
+			}
+			Expect(len(lines)).To(BeNumerically(">", 1))
```
rmohr: One test lane failed here with a line number of `1`. Could you think of a reason for that? Here the lane: https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/396//artifact/check-patch.k8s-1.11.0.el7.x86_64/mock_logs/script/stdout_stderr.log

tests/infra_test.go
fromanirh: This loop is needed to handle retries. There are timing issues in the tests, I just don't want to give up after the first failed attempt. 20 tries is overshoot by high margin, I expect success within the first 2-3 tries in the common case, 5-6 worst (from what I've seen so far)

tests/infra_test.go
fromanirh: Looks weird, checking

tests/infra_test.go
rmohr: I see. Maybe using `Eventually` would be more appropriate than this extra loop?

fromanirh:
Uhm, still fails: https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/415//artifact/check-patch.k8s-1.11.0.el7.x86_64/mock_logs/script/stdout_stderr.log

Works just fine on my box. I'll try with longer timeout.

fromanirh:
I removed "Eventually()" as per suggestion. It seems that the k8s-1.10.11 and os-3.11.0-crio lanes fails consistently, but can't reproduce it locally: the first works fine, the second fails for different reasons.
Also tried as suggested with FMeasure, running the tests many times in a row. k8s-1.10.11 works fine, see this:

k8s-1.11.0:
https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/435//artifact/check-patch.k8s-1.11.0.el7.x86_64/mock_logs/script/stdout_stderr.log
• [SLOW TEST:8.367 seconds]
Infrastructure
/root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:48
  Prometheus Endpoints
  /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:58
    should include the metrics for a running VM
    /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:107
• [MEASUREMENT]
Infrastructure
/root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:48
  Prometheus Endpoints
  /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:58
    should include the storage metrics for a running VM
    /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:135

    Ran 42 samples:
    runtime:
      Fastest Time: 17.538s
      Slowest Time: 22.780s
      Average Time: 20.321s ± 1.512s
• [SLOW TEST:27.807 seconds]
Infrastructure
/root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:48
  Start a VirtualMachineInstance
  /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:205
    when the controller pod is not running and an election happens
    /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:206
      should succeed afterwards
      /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:207
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
Waiting for namespace kubevirt-test-default to be removed, this can take a while ...
Waiting for namespace kubevirt-test-alternative to be removed, this can take a while ...

Ran 5 of 290 Specs in 946.018 seconds
SUCCESS! -- 5 Passed | 0 Failed | 0 Pending | 285 Skipped
PASS


k8s-1.10.11:
Infrastructure
/root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:48
  Prometheus Endpoints
  /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:58
    should include the storage metrics for a running VM
    /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:135

    Ran 42 samples:
    runtime:
      Fastest Time: 18.052s
      Slowest Time: 23.631s
      Average Time: 20.458s ± 1.456s
• [SLOW TEST:31.920 seconds]
Infrastructure
/root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:48
  Start a VirtualMachineInstance
  /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:205
    when the controller pod is not running and an election happens
    /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:206
      should succeed afterwards
      /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:207
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
Waiting for namespace kubevirt-test-default to be removed, this can take a while ...
Waiting for namespace kubevirt-test-alternative to be removed, this can take a while ...

Ran 5 of 290 Specs in 973.140 seconds
SUCCESS! -- 5 Passed | 0 Failed | 0 Pending | 285 Skipped
PASS

fromanirh:
Last CI run still failed the test I added. So I pushed PR#2135 to add logs to understand why it fails, because I suspect some weird race on startup (see below). And so I get https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/458/testReport/ where my test never fails. I'm puzzled.

fromanirh:
> I removed "Eventually()" as per suggestion. It seems that the k8s-1.10.11 and os-3.11.0-crio lanes fails consistently, but can't reproduce it locally: the first works fine, the second fails for different reasons.
> Also tried as suggested with FMeasure, running the tests many times in a row. k8s-1.10.11 works fine, see this:
> ## k8s-1.11.0:
> 
> Infrastructure
> /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:48
> Prometheus Endpoints
> /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:58
> should include the storage metrics for a running VM
> /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:135
> 
> ```
> Ran 42 samples:
> runtime:
>   Fastest Time: 17.538s
>   Slowest Time: 22.780s
>   Average Time: 20.321s ± 1.512s
> ```
> 
> Ran 5 of 290 Specs in 946.018 seconds
> SUCCESS! -- 5 Passed | 0 Failed | 0 Pending | 285 Skipped
> PASS
> ## k8s-1.10.11:
> 
> Infrastructure
> /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:48
> Prometheus Endpoints
> /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:58
> should include the storage metrics for a running VM
> /root/go/src/kubevirt.io/kubevirt/tests/infra_test.go:135
> 
> ```
> Ran 42 samples:
> runtime:
>   Fastest Time: 18.052s
>   Slowest Time: 23.631s
>   Average Time: 20.458s ± 1.456s
> ```
> 
> Ran 5 of 290 Specs in 973.140 seconds
> SUCCESS! -- 5 Passed | 0 Failed | 0 Pending | 285 Skipped
> PASS

With KUBEVIRT_PROVIDER=os-3.11.0 the test fail on my laptop after few (6-7) iterations because the VM is not ready (!).
 After staring more at CI logs, I suspect some weird race on startup, because almost every time my test fails, the logs report the metrics collector found no VM sockets in `/var/run/kubevirt`. So I pushed PR#2135 with more verbose logs, but I unexpectedly got the test pass on every lane: https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/458/#showFailuresLink

rmohr:
> With KUBEVIRT_PROVIDER=os-3.11.0 the test fail on my laptop after few (6-7) iterations because the VM is not ready (!).

Ok, so seems like this might reassemble a scenario similar to CI? If you put in the `Eventually`, does it succeed again?

> I suspect some weird race on startup, because almost every time my test fails, the logs report the metrics collector found no VM sockets in /var/run/kubevirt

So do the tests fail after you executed the serial console commands? This really sounds like a bug, the socket must be there, or must have been there at least. Could you try not failing on the error? Instead try to e.g. sleep for an hour in the test if this error occurs, so that no cleanup happens and that you can investigate the VMI and Pod state?
 
> but I unexpectedly got the test pass on every lane

Hm, yes that can happen with flaky tests.

fromanirh:
> > With KUBEVIRT_PROVIDER=os-3.11.0 the test fail on my laptop after few (6-7) iterations because the VM is not ready (!).
> 
> Ok, so seems like this might reassemble a scenario similar to CI? If you put in the `Eventually`, does it succeed again?

Trying next.

> > I suspect some weird race on startup, because almost every time my test fails, the logs report the metrics collector found no VM sockets in /var/run/kubevirt
> 
> So do the tests fail after you executed the serial console commands?

Yes, it seems so. Is it possible that something deletes `/var/run/kubevirt`? I see quite some os.MkdirAll calls, which seems to suggest that when components start, `/var/run/kubevirt` may not be there.

I wonder if this flow is possible:
1. something in the test flow (at least) removes/recreates `/var/run/kubevirt`. of course within a single test, not between tests -the latter is fine
2. `virt-handler` keeps a ref to an old `/var/run/kubevirt` - or just tries to scan it before it is recreated
3. `virt-launcher`, OTOH, recreates the dir and creates the socket in the new one
4. so from each component perspective everything seems fine, yet there is a race and the two component can't talk to each other
5. *eventually* `virt-handler` re-opens the new directory, so it will be able to talk to the socket

This line of thought is the one that made me craft https://github.com/kubevirt/kubevirt/pull/2135/commits/17d46393ce233f16b45a31227149365fb3fc4841

> This really sounds like a bug, the socket must be there, or must have been there at least.
> Could you try not failing on the error? Instead try to e.g. sleep for an hour in the test if this error 
> occurs, so that no cleanup happens and that you can investigate the VMI and Pod state?

Of course, let's try that. I wonder also if other tests do something similar to what I'm doing here - in a nutshell, fire up a VM, connect to it using console, then connect to it using the `virt-handler` -`virt launcher` socket

> > but I unexpectedly got the test pass on every lane
> Hm, yes that can happen with flaky tests.

Yep. This is another hint we have some race on startup. Unexpected successes are the hardest to explain otherwise :)

fromanirh:
ci test please

tests/infra_test.go
```diff
@@ -121,19 +126,78 @@ var _ = Describe("Infrastructure", func() {
...
+
+			By("Finding the prometheus endpoint")
+			l, err := labels.Parse("kubevirt.io=virt-handler")
+			Expect(err).ToNot(HaveOccurred())
+			pods, err := virtClient.CoreV1().Pods(tests.KubeVirtInstallNamespace).List(metav1.ListOptions{LabelSelector: l.String()})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(pods.Items).ToNot(BeEmpty())
+
+			By("Scraping the Prometheus endpoint")
+			pod := pods.Items[0] // only one compute node in the test environment
```
rmohr: Here is the issue.

You have to explicitly look up the pod which belongs to the VMI. The issue you run into, is that sometimes a pod is still there from a previous VMI (for instance not yet completely shut down, we don't wait for that between tests because it can take a very long time depending on the k8s version until it really disappears).

tests/utils.go
```diff
@@ -1109,6 +1109,29 @@ func GetRunningPodByLabel(label string, labelType string, namespace string) *k8s
 	return readyPod
 }
 
+func GetVirtHandlerPodForNode(nodeName, namespace string) *k8sv1.Pod {
```
rmohr: You can do `kubecli.NewVirtHandlerClient(virtClient).ForNode(nodeName).Pod()`

tests/infra_test.go
fromanirh: Oops. This is a leftover of my simplified development environment. Great catch, thanks!

rmohr:
ok, looks like the issue is gone now: https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/491/artifact/ci_build_summary.html. No occurrence of the issue.

Triggering again.

rmohr:
ci test please

rmohr:
@fromanirh can you squash the commits a little bit? Then we can merge.

fromanirh:
@rmohr sure, on it. Going to squash all the test fixes in one commit.

PR #2080, opened 2019-03-05
Fix some inconsistencies introduces with the bazel switch

rmohr:

* Keep DOCKER_PREFIX and DOCKER_TAG also for bazel right now. If we change that, let's change it for everything to CONTAINER_TAG and CONTAINER_PREFIX.
 * Allow builds with bazel on non-amd64 platforms if bazel is installed and bazel is run outside of our build containers. Use the newest rpm rules which don't perform arch or os checks when invoking rpm.
 * Respect hack/config-* settings when invoking `make bazel-push-images`

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster/external/provider.sh
+++ b/cluster/external/provider.sh
@@ -31,6 +31,6 @@ function down() {
 
 function build() {
     # Build code and manifests
-    ${KUBEVIRT_PATH}hack/dockerized "DOCKER_TAG=${DOCKER_TAG} KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER} ./hack/build-manifests.sh"
-    make push
+    ${KUBEVIRT_PATH}hack/dockerized "DOCKER_TAG=${DOCKER_TAG} DOCKER_PREFIX=${docker_prefix} KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER} ./hack/build-manifests.sh"
+    DOCKER_PREFIX=${docker_prefix} DOCKER_TAG=${docker_tag} make bazel-push-images
 }
--- a/.gitignore
+++ b/.gitignore
@@ -32,5 +32,5 @@ tools/vms-generator/vms-generator
 .coverprofile
 coverage.html
 manifests/**/*.tmp
-bazel-*
-.bazelrc
+/bazel-*
+/.bazelrc
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ script:
 
 deploy:
 - provider: script
-  script: docker login -u="$DOCKER_USER" -p="$DOCKER_PASS" && CONTAINER_PREFIX="index.docker.io/kubevirt" CONTAINER_TAG="latest" make bazel-push-images
+  script: docker login -u="$DOCKER_USER" -p="$DOCKER_PASS" && DOCKER_PREFIX="index.docker.io/kubevirt" DOCKER_TAG="latest" make bazel-push-images
   skip_cleanup: true
   on:
     branch: master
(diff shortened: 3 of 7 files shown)


cynepco3hahue:
@rmohr I prefer to change it to CONTAINER_PREFIX and CONTAINER_TAG, it just will make more sense when we will move images to quay. But we can do it after, so we can merge it :smile:

cluster/external/provider.sh
```diff
@@ -31,6 +31,6 @@ function down() {
 
 function build() {
     # Build code and manifests
-    ${KUBEVIRT_PATH}hack/dockerized "DOCKER_TAG=${DOCKER_TAG} KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER} ./hack/build-manifests.sh"
```
j-griffith: Why not just leave everything else and just fix up the external provider here?  So it turns out I can get this to work without the other changes if I fix up the provider script.  It does seem intermittent, but it did complete and I was able to push to my quay.io account.

I'd be happy to s/DOCKER_/CONTAINER_/ it does make logical sense.  I just want to make sure I can get it to work reliably and then update the docs.

j-griffith: Or does it not work with dockerhub?  Maybe that's what you're getting at here.

j-griffith: I took the pieces I was missing and just converted the external-provider to use the new vars correctly, that worked in terms of getting the images built and uploaded to my registry.  The Bazel process appears to hang though, I get here:

```
Successfully built 728500658bb0
Successfully tagged kubevirt:latest
go version go1.11.5 linux/amd64

go version go1.11.5 linux/amd64
Starting local Bazel server and connecting to it...
INFO: Invocation ID: d483f971-eb50-416c-839c-4c87205c01ca
DEBUG: /root/.cache/bazel/_bazel_root/6f347497f91c9a385dcd9294645b76e0/external/bazel_skylib/skylark_library.bzl:23:1: WARNING: skylark_library.bzl is deprecated and will go away in the future, please use bzl_library.bzl instead.
INFO: Analysed target //:push-images (543 packages loaded, 13139 targets configured).
INFO: Found 1 target...
Target //:push-images up-to-date:
  bazel-bin/push-images
INFO: Elapsed time: 80.004s, Critical Path: 35.15s
INFO: 45 processes: 45 processwrapper-sandbox.
INFO: Build completed successfully, 63 total actions
INFO: Build completed successfully, 63 total actions
```
Your full PR here results int he same behavior.  It doesn't appear that anything is attempting to deploy to the K8s cluster in this case.

j-griffith:
Just to clarify, while this PR put the explicit call back in to bazel-push-images, the problem I'm having *is* in fact `push images`.  The name change is fine, I can update the docs to fix that easily; but I can't seem to get the dockerized run of bazel to push images to anything but localhost:5000.

When setting things to quay.io/jgriffith, it does seem like it might be trying to work; it creates all the empty repos in my quay.io registry, but after an hour it's still running and all of the repos are empty (quay.io is painfully slow to respond when running this so who knows, maybe it just takes hours to complete).

I'm not sure what's up with dockerhub; bazel requires the domain but I seem to get various access errors on that, or http vs https response errors.

Finally, the use case I had in the past was my own private registry with self-signed certs.  That no longer works because I can't seem to get the kubevirt container to mount and use my `/etc/docker/certs.d` directory (although I think there's another layer in the kubevirt container I'm currently passing this volume mount to, so that won't work).

I'd like to help out here, if I can figure out more on how `make bazel-push-images` works I'll surely submit a PR.  In the meantime I've isolated things down to make it a bit faster to test out by just calling the make command directly:

```
CONTAINER_PREFIX=192.168.1.109:5000 CONTAINER_TAG=latest make bazel-push-images
```

j-griffith:
This isn't the real problem, so I opened an issue:  https://github.com/kubevirt/kubevirt/issues/2087

rmohr:
Ah yes, now I remember. Quay does not support the newer image formats yet. Neither v2, nor oci. When asking Quay people a month ago they meant that they have it already in staging and that it is close.

cynepco3hahue:
@rmohr `chmod +x` on the bazel script is missing

rmohr:
> @rmohr `chmod +x` on the bazel script is missing

:+1: fixed.

PR #2083, opened 2019-03-05
hyperv: add feature gate and node selector

fromanirh:
Existing, merged support for HyperV feature need up to date host kernel to work properly.
This PR adds a way to make sure we run hyperv-enabled requests only on hosts which can safely
run them.

We do this relying on a host label, which node-feature-discovery will add for us if it detects the host kernel is up to date

For backward compatibility the strict checking is *DISABLED* by default. E.g. nothing changes unless the admin explicitely asks for it.

```release-note
Adds a feature gate to ensure optimal scheduling on hyperv-enabled VMs
```

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -70,6 +70,7 @@ const LibvirtStartupDelay = 10
 //to match a VirtualMachineInstance CPU model(Family) and/or features to nodes that support them.
 const NFD_CPU_MODEL_PREFIX = "feature.node.kubernetes.io/cpu-model-"
 const NFD_CPU_FEATURE_PREFIX = "feature.node.kubernetes.io/cpu-feature-"
+const NFD_KVM_INFO_PREFIX = "feature.node.kubernetes.io/kvm-info-"
 
 const MULTUS_RESOURCE_NAME_ANNOTATION = "k8s.v1.cni.cncf.io/resourceName"
 const MULTUS_DEFAULT_NETWORK_CNI_ANNOTATION = "v1.multus-cni.io/default-network"
@@ -186,6 +187,17 @@ func IsCPUNodeDiscoveryEnabled(store cache.Store) bool {
 	return false
 }
 
+func isHyperVSupportRequired(store cache.Store) bool {
+	value, err := getConfigMapEntry(store, virtconfig.FeatureGatesKey)
+	if err != nil {
+		return false
+	}
+	if !strings.Contains(value, virtconfig.HypervSupportGate) {
+		return false
+	}
+	return true
+}
+
 func CPUModelLabelFromCPUModel(vmi *v1.VirtualMachineInstance) (label string, err error) {
 	if vmi.Spec.Domain.CPU == nil || vmi.Spec.Domain.CPU.Model == "" {
 		err = fmt.Errorf("Cannot create CPU Model label, vmi spec is mising CPU model")
@@ -205,6 +217,10 @@ func CPUFeatureLabelsFromCPUFeatures(vmi *v1.VirtualMachineInstance) []string {
 		}
 	}
 	return labels
+}
+
+func KVMInfoHypervSupportLabel() string {
+	return NFD_KVM_INFO_PREFIX + "hyperv"
 }
 
 func SetNodeAffinityForForbiddenFeaturePolicy(vmi *v1.VirtualMachineInstance, pod *k8sv1.Pod) {
@@ -790,6 +806,11 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		for _, cpuFeatureLable := range CPUFeatureLabelsFromCPUFeatures(vmi) {
 			nodeSelector[cpuFeatureLable] = "true"
 		}
+	}
+
+	if isHyperVSupportRequired(t.configMapStore) {
+		key := KVMInfoHypervSupportLabel()
+		nodeSelector[key] = "true"
 	}
 
 	nodeSelector[v1.NodeSchedulable] = "true"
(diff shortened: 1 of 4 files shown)


fromanirh:
ping?

pkg/virt-controller/services/template.go
```diff
@@ -178,6 +179,46 @@ func IsCPUNodeDiscoveryEnabled(store cache.Store) bool {
 	return false
 }
 
+func isHyperVSupportRequired(store cache.Store) bool {
```
slintes: I see the same pattern in the `IsCPUNodeDiscoveryEnabled()` func above, but I'm  wondering why we don't use `pkg/virt-config/feature-gates.go` for this. You already added a new func there, which is unused by now.

pkg/virt-controller/services/template.go
```diff
@@ -199,6 +240,10 @@ func CPUFeatureLabelsFromCPUFeatures(vmi *v1.VirtualMachineInstance) []string {
 	return labels
 }
 
+func KVMInfoHypervSupportLabel() string {
+	return NFD_KVM_INFO_PREFIX + "hyperv"
```
slintes: just curious: why a func for this, and not just a const?

pkg/virt-controller/services/template.go
fromanirh: I like your approach about moving `IsCPUNodeDiscoveryEnabled` into `pkg/virt-config`, but I wonder if it is OK to add a dep to `go-client` (for `cache.Store`) there. If it is OK I'll happily move the related code here. I'm going to kill the unused function, it is just confusing.

slintes: No need to add go-client there, your function as it is over there already should just work. Have a look at `virt-config/config-map.go`, there the ConfigMap is read, and it's values are put into the env, which are then read in `feature-gates.go`

pkg/virt-controller/services/template.go
fromanirh: > just curious: why a func for this, and not just a const?

No real reason. Let me simplify.

pkg/virt-controller/services/template.go
fromanirh: @slintes ack, on it.

slintes:
something went wrong on the crio lane, let's try again...
https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/414//artifact/check-patch.os-3.11.0-crio.el7.x86_64/mock_logs/script/stdout_stderr.log

slintes:
ci test please

vladikr:
What I'm worried about with this feature, is that virt-laucher doesn't have a way to protect the VM from cashing when the kernel / qemu doesn't support one of these features. We only rely on nfd , but I'm not sure it's always accurate.
For example, how does nfd know that the qemu version in virt-laucher supports all hyperv features?

fromanirh:
> What I'm worried about with this feature, is that virt-laucher doesn't have a way to protect the VM from cashing when the kernel / qemu doesn't support one of these features. We only rely on nfd , but I'm not sure it's always accurate.

TL;DR: I think we are already safe from this issue, I'm trying to protect from a different problem: the host kernel (lack of) support.

Let's recap the issue here.
The chain is: libvirt -> qemu -> kernel. What the user specifies, either explicitely with a baked VM or via a template, comes to libvirt via trivial data translation, so we can rule out everything else and focus on these three items.

Libvirt negotiatets capability with QEMU, so it knows what it can do. In addition, we -kubevirt- know what we put in the virt-launcher image, so we know what libvirt+qemu support.

In other word, up until QEMU we know what we can safely turn on and off wrt hyperv. We have full control over there. If we ship, say, libvirt 4.7 and qemu 2.12, we know which hyperv flag we expose and which we don't, and we don't even expose flags we don't support. SyNIC/SyNICTimer being the exception, see below.

Now, there are hyperv features that require kernel cooperation or even backports/fixes, for example SyNIC on CentOS. *THIS* is the problematic link we are checking/guarding here.

What we could happen if everything *except the kernel* - perhaps because it is too old, and lacks some patches backported in a future version - is OK? Let's keep the SyNIC example.
QEMU checks the availability of the kernel support for SyNIC (https://github.com/qemu/qemu/blob/stable-2.12/target/i386/kvm.c#L701) and exits if the feature is requested but not supported.

So, in this very case, the VM will get scheduled, and won't run unless kernel is up to date.
And kubevirt can't know this without this patch.

Furthermore, it happened in the past oVirt history that we had to required backported kernel with hyperv (or KVM in general) fixes, and this patch could help should this need rise again (quite likely IMO).

> For example, how does nfd know that the qemu version in virt-laucher supports all hyperv features?

It doesn't! But we control that dependency, so (bar bugs ofc) there isn't such a risk. We know what we ship in images, so we know what we can use and what we can't. It is a different issue.

vladikr:
> @vladikr are your worries resolved?

Not really, but I'm not going to hold this work.

fromanirh:
> > @vladikr are your worries resolved?
> 
> Not really, but I'm not going to hold this work.

I don't want to rush bad changes, but OTOH I don't see better ways. We can continue offline/on ML

fromanirh:
hit wrong button when I commented, sorry.

MarSik:
@vladikr Our NFD plugin uses the exact same kubevirt/libvirt image as you do to get the proper qemu and libvirt and we ask for what is supported. We do not have hyperv support implemented in the plugin yet, but that is just coding. 

Also the current code does not check at all either so this should be an improvement even without NFD (static labels), right?

fromanirh:
ci test please

fromanirh:
After discussion, we agreed that the general direction is OK, but we need finer-grained labels to let kubevirt core do better decisions. Will keep this PR open for the time being, but it will be soon replaced by another implementing the aforementioned strategy.

fromanirh:
closing as per previous comment

PR #2094, opened 2019-03-06
tests: replace remaining occurrence of "local" by StorageClassLocal

dollierp:

Allows to change the storage class used during tests by modifying the code in a unique place.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/imageupload_test.go
+++ b/tests/imageupload_test.go
@@ -59,7 +59,7 @@ var _ = Describe("ImageUpload", func() {
 				"--pvc-size", pvcSize,
 				"--uploadproxy-url", fmt.Sprintf("https://127.0.0.1:%d", localUploadProxyPort),
 				"--wait-secs", "30",
-				"--storage-class", "local",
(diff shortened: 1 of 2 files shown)


rmohr:
> Allows to change the storage class used during tests by modifying the code in a unique place.

Sounds like you would actually need a commandline flag to the tests  binary?

dollierp:
@rmohr 

Indeed, I began to implement adding a new command line flag for this but then I figured out that basically we will need one for all those https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L137-L143.

So many flags doesn't seems a viable solution so I stopped here for now.

I guess the right solution would be to use a configuration file but I don't feel confident with my current skills in go to develop such a change.

rmohr:
> I guess the right solution would be to use a configuration file but I don't feel confident with my current skills in go to develop such a change.

Hm, might make sense. The PR as is is ok for me as cleanup PR. Thanks.

PR #2115, opened 2019-03-13
Propagate MAC address to multus via annotations

booxter:
Now that annotations are JSON payloads, we can propagate more of
attributes through them to help CNI configure devices as per VMI spec.

```release-note
Propagate interface MAC address to Multus CNI plugin
```

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -1080,6 +1080,15 @@ func getNetworkToResourceMap(virtClient kubecli.KubevirtClient, vmi *v1.VirtualM
 	return
 }
 
+func getIfaceByName(vmi *v1.VirtualMachineInstance, name string) *v1.Interface {
+	for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
+		if iface.Name == name {
+			return &iface
+		}
+	}
+	return nil
+}
+
 func getCniAnnotations(vmi *v1.VirtualMachineInstance) (cniAnnotations map[string]string, err error) {
 	ifaceList := make([]string, 0)
 	ifaceListMap := make([]map[string]string, 0)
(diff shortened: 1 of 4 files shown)


pkg/virt-controller/services/template.go
```diff
@@ -1056,26 +1063,51 @@ func getNetworkToResourceMap(virtClient kubecli.KubevirtClient, vmi *v1.VirtualM
 	return
 }
 
-func getCniInterfaceList(vmi *v1.VirtualMachineInstance) (ifaceListString string, cniAnnotation string) {
+func getIfaceByName(vmi *v1.VirtualMachineInstance, name string) *v1.Interface {
```
danielBelenky: Just out of curiosity, why not define `getIfaceByName()` on `VirtualMachineInstance`? Wouldn't it be more convenient to use?

booxter:
/hold

gotta write some func tests.

pkg/virt-controller/services/template.go
booxter: @danielBelenky I am a vigorous OOP denier :) so I see no implicit value of tying a helper function to a type. Especially for one-off helper functions like this one.

booxter:
@SchSeba trying to fully understand your comment. Perhaps I can write down the scenario that I believe you describe and you tell me if this is it.

Right now, we pass MAC address in libvirt domain interface definition for virtio to set the corresponding MAC address on each interface. (Assuming it's set on VMI interface in its spec.) If it is passed to libvirt then the VMI will capture the MAC address and use it for L2 interactions.

This PR also passes MAC addresses set for interfaces to CNI plugins. If a CNI plugin is able to handle the passed MAC address, and it configures this address on the netlink interface inside the pod namespace, then we risk having the same MAC address set on the in-pod interface and inside the guest. Duplicate MAC addresses in the network will probably disrupt connectivity to the guest.

You suggest that if, inside virt-launcher, detect that a netlink interface holds MAC address of the corresponding VMI interface definition. If it does, then virt-launcher will remove the MAC address from the netlink interface and replace it with a randomly generated MAC address. With this, no two interfaces will hold the same MAC address, and no disruption will occur.

Is it a fair description of what you are saying?

SchSeba:
@booxter Exactly thanks for the comment

SchSeba:
I will like if you can add this to the PR comment

booxter:
@SchSeba @phoracek could you please check this PR? Sebastian, I think the issues you pointed to are solved.

rmohr:
ci test please

booxter:
```
go version go1.11.5 linux/amd64
INFO: Invocation ID: 26102281-e20f-4530-bbb7-a27776005e27
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

@rmohr are there known issues with e.g. `bazel` that would lock a job?

booxter:
ci test please

rmohr:
@booxter To be honest it is hard to tell what the issue is. I configured bazel to be a little less verbose on travis. If  e.g. a dependency download is super-slow but does not break, it can happen that there is no output printed for a longer time period. I don't think that there is a general issue behind that.

rmohr:
The os-3.11.0-multus lane looks good now.

PR #2116, opened 2019-03-15
tests: read parameters from a config file

dollierp:

This is a naive implementation to add a new `-config` optional command line flag to functests for specifying the path of a JSON formatted file from which the test suite will load its configuration.

This PR aims to facilitate running the functional tests on environments having different configuration.

As a first step, only the names of the StorageClasses used for tests are read from the configuration file. More parameters can be added later when needed.

**Release note**:

```release-note
functests now support a `-config` flag to specify the path of a JSON configuration file to facilitate running the test suite with parameters customized for a particular testing environment.
```


tests/config.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func loadConfig() *KubeVirtTestsConfiguration {
+	config := NewKubeVirtTestsConfiguration()
+
+	if ConfigFile != "" {
+		err := loadConfigFromFile(ConfigFile, config)
+
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Warning: error loading configuration file: %s\n", err)
```
lukas-bednar: If we got up to here, that means user specified config file on command line explicitly. Does it make sense to continue when we are unable to read user's config file ?

slintes:
You need to run `make generate` and commit changes. That will update the bazel configuration with the new go file, and make Travis happy :)
STDCI failed because of another issue which is fixed in the meanwhile, will retrigger tests

slintes:
ci test please

dollierp:
> What do you think about always reading the configuration from file? Maybe rename example-config.json to default-config.json, and use that filename as default value for the config flag? That way we would always test if reading config from file works.

Fine by me. That means you are expecting the tests must be run within the `tests/` folder from the source checkout by default to find the default config file isn't it?

slintes:
> > What do you think about always reading the configuration from file? Maybe rename example-config.json to default-config.json, and use that filename as default value for the config flag? That way we would always test if reading config from file works.
> 
> Fine by me. That means you are expecting the tests must be run within the `tests/` folder from the source checkout by default to find the default config file isn't it?

good point. No, I don't want to change how we run the tests. So the default value for the flag probably needs to be `tests/default-config.json`. Does that make sense?

dollierp:
I pushed an update which should address all your remarks.

tests/config.go
```diff
@@ -0,0 +1,91 @@
...
+	// StorageClass to use to create block-volume PVCs
+	StorageClassBlockVolume string `json:"storageClassBlockVolume"`
+	// StorageClass to use to create rhel PVCs
+	StorageClassRhel string `json:"storageClassRhel"`
+	// StorageClass to use to create windows PVCs
+	StorageClassWindows string `json:"storageClassWindows"`
+}
+
+func NewKubeVirtTestsConfiguration() *KubeVirtTestsConfiguration {
+	return &KubeVirtTestsConfiguration{
```
slintes: I think we can remove this default config now, because we always read from file and panic when it goes wrong

tests/utils.go
```diff
@@ -456,6 +448,8 @@ func BeforeTestSuitSetup() {
 	log.InitializeLogging("tests")
 	log.Log.SetIOWriter(GinkgoWriter)
 
+	Config = loadConfig()
```
slintes: can we move this into the init block of config.go?

dollierp: Unfortunately it does not work since the command line flags are not parsed yet in `init()`.

tests/config.go
dollierp: I hesitated and decided to keep it in case the user provides a partial configuration file overriding just a few values.

tests/config.go
dollierp: PR modified to panic instead of juste writing a message to stderr.

tests/config.go
slintes: Ah ok, I assumed you have to always provide all values. But overriding a few values only makes sense.
However, now we have the default values at 2 places, in code and in file, which is confusing :/
What about this: remove the default config from code, and remove the default value of the config flag, but still always load `tests/default-config.json` first. And additionally load a custom config file afterwards when the config flag is given?

tests/utils.go
slintes: ok, too bad

tests/config.go
dollierp: Sounds smart, I pushed a new commit to implement this.

PR #2121, opened 2019-03-17
Fix and addition to service expose test.

yossisegev:
Fix and addition to service expose test:
1. Remove redundant code from the test of HTTP connection over the exposed ClusterIP service:
 There is actually no need for a dedicated HTTP server; the TCP server that is already running can serve HTTP connection attempts as well as TCP connections.
2. Test multi-reconnecions over the same exposed service.

### DIFF
--- a/tests/expose_test.go
+++ b/tests/expose_test.go
@@ -376,7 +376,9 @@ var _ = Describe("[rfe_id:253][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 		})
 
 		Context("Expose a VM as a ClusterIP service.", func() {
-			It("[test_id:1538]Connect to ClusterIP services that was set when VM was offline", func() {
+			It("[test_id:1538]Connect to ClusterIP service that was set when VM was offline.", func() {
+				// This TC also covers:
+				// [test_id:1795] Exposed VM (as a service) can be reconnected multiple times.
 				By("Getting back the cluster IP given for the service")
 				svc, err := virtClient.CoreV1().Services(vm.Namespace).Get(serviceName, k8smetav1.GetOptions{})
 				Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


cynepco3hahue:
Do not sure that I understand how the context of the PR mirrors PR description.

- How does the lack of the HTTP server on the VM solve the problem with the test?
- I do not see multi-reconnections over the same exposed service test

yossisegev:
> * How does the lack of the HTTP server on the VM solve the problem with the test?
>
Debugging this revealed that the HTTP server wasn't really started; that's probably because I used the same port (testPort, which is the target port for the exposed service), that is already used for the TCP server.
For some reason, starting the HTTP server is declared as successful, so the test doesn't fail.
Considering this scenario a little further, I found that there is actually no need for the HTTP server; the TCP server that is already running can serves HTTP connection attempts as well as TCP connections.
So removing the HTTP server actually cleans the test from redundant code.

> * I do not see multi-reconnections over the same exposed service test
>
First connection
`				job := tests.NewHelloWorldJob(serviceIP, servicePort)`

and second connection
`				job = tests.NewHelloWorldJobHttp(serviceIP, servicePort)`

cynepco3hahue:
> First connection
> `	job := tests.NewHelloWorldJob(serviceIP, servicePort)`
> 
> and second connection
> `	job = tests.NewHelloWorldJobHttp(serviceIP, servicePort)`

But it already existed here, so you do not add something new, so you do not need to specify it under the description.

cynepco3hahue:
> Debugging this revealed that the HTTP server wasn't really started; that's probably because I used the same port (testPort, which is the target port for the exposed service), that is already used for the TCP server.
> For some reason, starting the HTTP server is declared as successful, so the test doesn't fail.
> Considering this scenario a little further, I found that there is actually no need for the HTTP server; the TCP server that is already running can serves HTTP connection attempts as well as TCP connections.
> So removing the HTTP server actually cleans the test from redundant code.

Such description can explain better the PR purpose, than `Fix the test of HTTP connection over the exposed ClusterIP service,
which was found to be in-fact disfunctional.`

yossisegev:
@cynepco3hahue Agreed.
Changed the commit message and the PR description accordingly.

Issue #1348, opened 2018-07-19
Reject VMs with resources requests larger than namespace limits


vladikr:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
This is a follow up to #970, which copied namespace limits to the vm spec, so it will be visible for the users. However, we don't reject the VM spec and are letting the pod be created which later will be rejected by k8s. We could save the additional steps of the pod creation.

**What you expected to happen**:
We should find a way to reject VMs with resources requests higher than the namespace limits, before creating a pod.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- KubeVirt version (use `virtctl version`):
- Kubernetes version (use `kubectl version`):
- VM or VMI specifications:
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

rmohr:
@vladikr @slintes any idea why this is still open? Are we missing something?

slintes:
@rmohr  I understand that the referenced PR is not about this issue, but about a related one, and this is a unresolved follow up? Fabian asked me to have a look a while ago, but tbh I just forgot about it while being busy with other tasks.

PR #2137, opened 2019-03-21
Add default CPU time

ahadas:
Adding a default value for requested CPU time of 100m when a VMI lacks it.

**Release note**:

```release-note
The requested CPU of a VMI defaults to "100m" when not specified.
```

### DIFF
--- a/pkg/virt-api/webhooks/mutating-webhook/mutating-webhook.go
+++ b/pkg/virt-api/webhooks/mutating-webhook/mutating-webhook.go
@@ -25,6 +25,8 @@ import (
 	"net/http"
 
 	"k8s.io/api/admission/v1beta1"
+	k8sv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
@@ -121,6 +123,9 @@ func mutateVMIs(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
 	setDefaultCPUModel(&vmi, config.GetCPUModel())
 	setDefaultMachineType(&vmi, config.GetMachineType())
 	kubev1.SetObjectDefaults_VirtualMachineInstance(&vmi)
+	// Default CPU request is done after the Resources section is initialized, if needed,
+	// in v1.SetObjectDefaults_VirtualMachineInstance. TODO: set default memory here
+	setDefaultCPURequest(&vmi, config.GetCPURequest())
 
 	// Add foreground finalizer
 	vmi.Finalizers = append(vmi.Finalizers, v1.VirtualMachineInstanceFinalizer)
@@ -223,6 +228,15 @@ func setDefaultCPUModel(vmi *kubev1.VirtualMachineInstance, defaultCPUModel stri
 func setDefaultMachineType(vmi *kubev1.VirtualMachineInstance, defaultMachineType string) {
 	if vmi.Spec.Domain.Machine.Type == "" {
 		vmi.Spec.Domain.Machine.Type = defaultMachineType
+	}
+}
+
+func setDefaultCPURequest(vmi *v1.VirtualMachineInstance, defaultCPURequest resource.Quantity) {
+	if _, exists := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceCPU]; !exists {
+		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.DedicatedCPUPlacement {
+			return
+		}
+		vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceCPU] = defaultCPURequest
 	}
 }
 
--- a/pkg/virt-api/webhooks/mutating-webhook/mutating-webhook.go
+++ b/pkg/virt-api/webhooks/mutating-webhook/mutating-webhook.go
@@ -30,7 +30,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
-	kubev1 "kubevirt.io/kubevirt/pkg/api/v1"
 	v1 "kubevirt.io/kubevirt/pkg/api/v1"
 	"kubevirt.io/kubevirt/pkg/log"
 	"kubevirt.io/kubevirt/pkg/util"
@@ -85,7 +84,7 @@ func mutateVMIs(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
 		return webhooks.ToAdmissionResponseError(err)
 	}
 
-	if resp := webhooks.ValidateSchema(kubev1.VirtualMachineInstanceGroupVersionKind, ar.Request.Object.Raw); resp != nil {
+	if resp := webhooks.ValidateSchema(v1.VirtualMachineInstanceGroupVersionKind, ar.Request.Object.Raw); resp != nil {
 		return resp
 	}
 
@@ -122,7 +121,7 @@ func mutateVMIs(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
 	log.Log.Object(&vmi).V(4).Info("Apply defaults")
 	setDefaultCPUModel(&vmi, config.GetCPUModel())
 	setDefaultMachineType(&vmi, config.GetMachineType())
-	kubev1.SetObjectDefaults_VirtualMachineInstance(&vmi)
+	v1.SetObjectDefaults_VirtualMachineInstance(&vmi)
 	// Default CPU request is done after the Resources section is initialized, if needed,
 	// in v1.SetObjectDefaults_VirtualMachineInstance. TODO: set default memory here
 	setDefaultCPURequest(&vmi, config.GetCPURequest())
@@ -211,13 +210,13 @@ func mutateMigrationCreate(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionRespon
 	}
 }
 
-func setDefaultCPUModel(vmi *kubev1.VirtualMachineInstance, defaultCPUModel string) {
+func setDefaultCPUModel(vmi *v1.VirtualMachineInstance, defaultCPUModel string) {
 	//if vmi doesn't have cpu topology or cpu model set
 	if vmi.Spec.Domain.CPU == nil || vmi.Spec.Domain.CPU.Model == "" {
 		if defaultCPUModel != "" {
 			// create cpu topology struct
 			if vmi.Spec.Domain.CPU == nil {
-				vmi.Spec.Domain.CPU = &kubev1.CPU{}
+				vmi.Spec.Domain.CPU = &v1.CPU{}
 			}
 			//set is as vmi cpu model
 			vmi.Spec.Domain.CPU.Model = defaultCPUModel
@@ -225,7 +224,7 @@ func setDefaultCPUModel(vmi *kubev1.VirtualMachineInstance, defaultCPUModel stri
 	}
 }
 
-func setDefaultMachineType(vmi *kubev1.VirtualMachineInstance, defaultMachineType string) {
+func setDefaultMachineType(vmi *v1.VirtualMachineInstance, defaultMachineType string) {
 	if vmi.Spec.Domain.Machine.Type == "" {
 		vmi.Spec.Domain.Machine.Type = defaultMachineType
 	}
--- a/pkg/virt-config/config-map.go
+++ b/pkg/virt-config/config-map.go
@@ -50,6 +50,7 @@ const (
 	imagePullPolicyKey     = "dev.imagePullPolicy"
 	migrationsConfigKey    = "migrations"
 	cpuModelKey            = "default-cpu-model"
+	cpuRequestKey          = "cpu-request"
 
 	ParallelOutboundMigrationsPerNodeDefault uint32 = 2
 	ParallelMigrationsPerClusterDefault      uint32 = 5
@@ -57,6 +58,7 @@ const (
 	MigrationProgressTimeout                 int64  = 150
 	MigrationCompletionTimeoutPerGiB         int64  = 800
 	DefaultMachineType                              = "q35"
+	DefaultCPURequest                               = "100m"
 
 	NodeDrainTaintDefaultKey = "kubevirt.io/drain"
 )
@@ -141,6 +143,7 @@ func defaultClusterConfig() *Config {
 	nodeDrainTaintDefaultKey := NodeDrainTaintDefaultKey
 	progressTimeout := MigrationProgressTimeout
 	completionTimeoutPerGiB := MigrationCompletionTimeoutPerGiB
+	cpuRequestDefault := resource.MustParse(DefaultCPURequest)
 	return &Config{
 		ResourceVersion: "0",
 		ImagePullPolicy: k8sv1.PullIfNotPresent,
@@ -154,6 +157,7 @@ func defaultClusterConfig() *Config {
 			CompletionTimeoutPerGiB:           &completionTimeoutPerGiB,
 		},
 		MachineType: DefaultMachineType,
+		CPURequest:  cpuRequestDefault,
 	}
 }
 
@@ -164,6 +168,7 @@ type Config struct {
 	ImagePullPolicy k8sv1.PullPolicy
 	MachineType     string
 	CPUModel        string
+	CPURequest      resource.Quantity
 }
 
 type MigrationConfig struct {
@@ -202,6 +207,10 @@ func (c *ClusterConfig) GetMachineType() string {
 
 func (c *ClusterConfig) GetCPUModel() string {
 	return c.getConfig().CPUModel
+}
+
+func (c *ClusterConfig) GetCPURequest() resource.Quantity {
+	return c.getConfig().CPURequest
 }
 
 // setConfig parses the provided config map and updates the provided config.
@@ -256,6 +265,10 @@ func setConfig(config *Config, configMap *k8sv1.ConfigMap) error {
 
 	if cpuModel := strings.TrimSpace(configMap.Data[cpuModelKey]); cpuModel != "" {
 		config.CPUModel = cpuModel
+	}
+
+	if cpuRequest := strings.TrimSpace(configMap.Data[cpuRequestKey]); cpuRequest != "" {
+		config.CPURequest = resource.MustParse(cpuRequest)
 	}
 
 	return nil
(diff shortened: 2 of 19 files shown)


rmohr:
@ahadas is there a background story why we have to do this?

pkg/virt-config/virt-properties.go
```diff
@@ -42,3 +44,12 @@ func SupportedEmulatedMachines() []string {
 	}
 	return vals
 }
+
+func DefaultCPUTime() resource.Quantity {
+	config := os.Getenv(defaultCPUTimeEnvVar)
+	if len(config) > 0 {
```
fromanirh: probably better:

  if config != "" {
  }

jeniferh:
Hi @rmohr, this bug has the background: https://bugzilla.redhat.com/show_bug.cgi?id=1679255

Basically large VM cpu performance can be very limited in some cases by the default cpu.shares value of 2 when no cpu requests are defined. 100m cpu requests seems to be a good default value since it overcomes the cpu.shares bottleneck while still allowing for a reasonably large amount of overcommit (10X per cpu).

pkg/virt-config/virt-properties.go
ahadas: well, [both are equivalent](https://groups.google.com/forum/#!topic/golang-nuts/7Ks1iq2s7FA) but changed anyway

pkg/virt-api/webhooks/mutating-webhook/mutating-webhook.go
```diff
@@ -108,6 +110,8 @@ func mutateVMIs(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
 	// Apply default cpu model
 	setDefaultCPUModel(&vmi, informers.ConfigMapInformer.GetStore())
 
+	setDefaultCPUTime(&vmi)
```
rmohr: This needs to be applied after `applyNamespaceLimitRangeValues`, and you will have to adjust `applyNamespaceLimitRangeValues`, to set the default cpu time from the limitrange object. Otherwise you would always override limitranges set in a namespace.

ahadas: @rmohr , Oh, I think I see your point - you mean that if `applyNamespaceLimitRangeValues` is ever changed to set CPU defaultRequest when CPU-request is not already set then setting CPU-request in `setDefaultCPUTime` that is called beforehand would make the aforementioned logic redundant?

That seems to be the case with memory-request at the moment btw (kubevirt sets it to 8192Ki so a similar logic in `applyNamespaceLimitRangeValues` that sets memory-request to memory defaultRequest from limitranges would not work, right?).

I wonder though why to introduce defaults in kubevirt-config then and not to use limitranges - is it because we assume different default requests for VMs and ordinary pods but cannot assume different namespaces for VMs and ordinary pods? If that's the reason, is it possible to we add a LimitType of VMI?

rmohr: > That seems to be the case with memory-request at the moment btw (kubevirt sets it to 8192Ki so a similar logic in `applyNamespaceLimitRangeValues` that sets memory-request to memory defaultRequest from limitranges would not work, right?).

@ahadas good point, you are right that it would not happen for memory requests. I think that this is a bug. I think we should apply the limits and requests from `LimitRange` 1:1 already in virt-api, but without adding any virt specific defaults. The defaults or calculated overheads should probably stay in virt-controller (where the overhead calculation of memory resides). Defaults tend to become dynamic (like memory overhead) and virt-api does not necessarily have all information present which is needed to calculate that (for example, referenced PVCs and PVs, maps and secrets don't have to be present when creating VMIs or pods).

rmohr: Here another related issue: https://github.com/kubevirt/kubevirt/issues/1348

ahadas: @rmohr ,
1. I don't mind fixing it for the memory as well. But the question is how :) currently, the code finds the minimal default limit for the memory although that doesn't seem to be what k8s does when multiple limitranges are set for containers in the namespace (iiuc the results I see then it picks the last one that was set). However, taking the minimal limit does make sense. But for defaultRequest the logic should probably to take the one that is set in the limit we have taken the default limit from, or otherwise (i.e., when no default limit is taken) - take a random defaultRequest from one of the defined limits. What do you think?

2. About not having virt-specific defaults and having defaults dynamic - I don't know, it does make sense to me to provide different defaults for VMs compared to ordinary pods due to their different nature (run for longer time, being more idle) and computing the defaults seems to be more complex than what we planned. @jeniferh , @MarSik  - thoughts?

rmohr: > 2\. About not having virt-specific defaults and having dynamically calculated defaults - I don't know, it does make sense to me to provide different defaults for VMs compared to ordinary pods due to their different nature (run for longer time, being more idle) and computing the defaults seems to be more complex than what we planned. @jeniferh , @MarSik  - thoughts?

Yeah we can have virt defaults. I just meant that they should be calculated and added by virt-controller and not virt-api.

> 1. I don't mind fixing it for the memory as well. But the question is how :) currently, the code finds the minimal default limit for the memory although that doesn't seem to be what k8s does when multiple limitranges are set for containers in the namespace (iiuc the results I see then it picks the last one that was set). But taking the minimal limit does make sense. For defaultRequest, however, the logic should probably to take the one that is set in the limit we have taken the default limit from, or otherwise (i.e., when no default limit is taken) - take a random defaultRequest from one of the defined limits. What do you think?

Anything which is different to what k8s does is wrong here. If we diverge here we will have to change it.

rmohr: So I think it should be like this:

 1) In virt-api apply range limits on the `VMI` for all values the same way like k8s does it on Pods.
 2) In virt-controller add memory-guest-overhead if appropriate, add cpu-requests if appropriate on the `Pod`.

rmohr: I am however not entirely sure here, since we will always have this small time-gap between posting the VMI and creating the actual pods, so there could be a range limit which got active exactly in that small window.

ahadas: > I am however not entirely sure here, since we will always have this small time-gap between posting the VMI and creating the actual pods, so there could be a range limit which got active exactly in that small window.

Right, but I don't think it is that important,
- if the VMI is posted after the range limit was set, it will get the up-to-date limits
- if the VMI is posted before the range limit was set then we may not get the most up-to-date limits but such races would always be there I suppose (even if we had fetched the limits and then set them directly on the pod, a new limit could have been set right after we queried the limits, right?)

So I don't think we can/should promise anything about entities that are received before a range limit is set

ahadas: @rmohr , posted #2150 to change namespace-limits as discussed - could you please review?

ahadas: Moved that code to api/v1/defaults.go so it would be executed after namespace-limits are applied - for that, I had to break the dependency between virt-config and kubecli packages.
The namespace-limits and memory related stuff are handled separately by other PRs.
@rmohr , could you please do a follow-up review?

ahadas:
@rmohr @cynepco3hahue , removed the configuration part - just defaults it to 100m as proposed in: https://bugzilla.redhat.com/show_bug.cgi?id=1679255

ahadas:
@rmohr ping

rmohr:
continued the discussion in bugzilla.

ahadas:
@rmohr ah okok, wasn't aware of that :)

pkg/virt-config/config-map.go
```diff
@@ -248,6 +257,10 @@ func setConfig(config *Config, configMap *k8sv1.ConfigMap) error {
 		config.MachineType = machineType
 	}
 
+	if cpuRequest := strings.TrimSpace(configMap.Data[cpuRequestKey]); cpuRequest != "" {
+		config.CPURequest = cpuRequest
```
rmohr: Can that compile? Shouldn't there be something like `Parse(cpuReuests)`?

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -1162,14 +1160,10 @@ func getCPUTopology(vmi *v1.VirtualMachineInstance) *CPUTopology {
...
+		//if cores, sockets, threads are not set, take value from domain resources request and
 		//set value into sockets, which have best performance (https://bugzilla.redhat.com/show_bug.cgi?id=1653453)
-		resources := vmi.Spec.Domain.Resources
-		if cpuRequests, ok := resources.Requests[k8sv1.ResourceCPU]; ok {
-			sockets = uint32(cpuRequests.Value())
-		} else if cpuLimit, ok := resources.Limits[k8sv1.ResourceCPU]; ok {
-			sockets = uint32(cpuLimit.Value())
-		}
+		cpuRequests := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceCPU]
+		sockets = uint32(cpuRequests.Value())
```
rmohr: I guess the whole setting default here is shaky, but would it make sense to keep the limit check (and give it priority over the requests amount?)

rmohr:
@ahadas I think the direction, is good now. :+1:
What I don't see right now is a way for users and admins to set no requests at all. I guess it could be added later, like we did for memory.

pkg/virt-config/config-map.go
ahadas: No :) the 'verified' flag is really missing here...

pkg/virt-launcher/virtwrap/api/converter.go
ahadas: I think not - from k8s documentation it seems that the memory limit is just a threshold that admins can set to say "if the memory consumed by the container exceeds that value, kill it". so I think admins would, for instance, set a VM with memory-request=2G and memory-limit=8G to say "the VM generally needs 2G to operate, if the memory consumption gets to 8G it is far higher than the expected value so it should be killed".

so if I understand this correctly, setting the guest topology according to the limit value is a bit awkward

rmohr: Hm there can be the following situation if we don't do this default: If we only have one socket, then we can never use more than one CPU at a time. So a request of 1 CPU and a limit of 10 CPUs can only ever use 1 CPU at max. If I am not mistaken.

ahadas: Removed my previous comment - it was stupid, I had memory rather than CPU limit in mind...

I see your point but note that on the other hand, as scheduling is done according to the requested amount, there's no guarantee that in that scenario we'll have more than 1 CPU available for the VM on the node - so if we'll set 10 sockets here and only 1 CPU would actually be allocated to the VM on the node - then the guest should somehow know it needs to avoid using more than 1 socket, right?

Anyway, as we discussed regarding memory once - I can't see a real-world scenario where users define VMs without specifying their CPU topology so I'm ok with any heuristic here.

ahadas: Done

tests/vmi_configuration_test.go
```diff
@@ -269,6 +269,7 @@ var _ = Describe("Configurations", func() {
 				Expect(domXml).To(ContainSubstring("driver name='vhost' queues='3'"))
 			})
 
+			/* Commenting out the following test as CPU request is always set at the moment
```
rmohr: Let's remove it. It is in the git history and will for sure get outdated and ignored by devs.

pkg/virt-api/webhooks/mutating-webhook/mutating-webhook.go
```diff
@@ -123,7 +124,10 @@ func mutateVMIs(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
 	if vmi.Spec.Domain.Machine.Type == "" {
 		vmi.Spec.Domain.Machine.Type = config.GetMachineType()
 	}
-	kubev1.SetObjectDefaults_VirtualMachineInstance(&vmi)
+	v1.SetObjectDefaults_VirtualMachineInstance(&vmi)
+	// Default CPU request is done after the Resources section is initialized, if needed,
+	// in v1.SetObjectDefaults_VirtualMachineInstance. TODO: set default memory here
+	setDefaultCPURequest(&vmi, func() resource.Quantity { return config.GetCPURequest() })
```
rmohr: I don't think that you need the callback here. The config live-updates values. MutateVMI is called for each VMI individually. You should be ablee to just return `config.GetCPURequest` without `()` instead of wrapping it again.

tests/vmi_configuration_test.go
ahadas: Will do

pkg/virt-api/webhooks/mutating-webhook/mutating-webhook.go
ahadas: Right, this is a workaround for the tests - as I couldn't mock the kubevirt-config in the tests of mutating-webhook. That's the issue I referred to in #2142 - if we find a way to set values in kubevirt-config in those tests, I'll drop it

ahadas: Ah, I thought you meant different thing. ok, I'll pass the configuration value as an argument :)

pkg/virt-launcher/virtwrap/api/converter.go
rmohr: @ahadas yep the whole heuristic here is a little bit wacky. Maybe we can find a better one later.

tests/vmi_configuration_test.go
```diff
@@ -269,22 +269,6 @@ var _ = Describe("Configurations", func() {
 				Expect(domXml).To(ContainSubstring("driver name='vhost' queues='3'"))
 			})
 
-			It("[test_id:1666]should reject virtio block queues without cores", func() {
```
rmohr: Now we only need an e2e test which ensures that we really set 100m by default. A simple test which starts a VMI and where we simply check if `100m` got set should be good enough.

ahadas: Done

tests/vmi_configuration_test.go
```diff
@@ -1107,6 +1121,67 @@ var _ = Describe("Configurations", func() {
...
+		})
+
+		It("should set CPU request when it is not provided", func() {
+			vmi := tests.NewRandomVMI()
+			vmi.Spec.Domain.Resources = v1.ResourceRequirements{
+				Requests: kubev1.ResourceList{
+					kubev1.ResourceMemory: resource.MustParse("64M"),
+				},
+			}
+			tests.RunVMIAndExpectLaunch(vmi, false, 30)
```
rmohr: Can you instaed of waiting for `Running` wait for `Scheduled`? That would make the tests significantly more lightweight and  the Pod exists in that state already.

tests/vmi_configuration_test.go
```diff
@@ -1107,6 +1121,67 @@ var _ = Describe("Configurations", func() {
...
+			computeContainer := getComputeContainerOfPod(readyPod)
+			cpuRequest := computeContainer.Resources.Requests[kubev1.ResourceCPU]
+			Expect(cpuRequest.String()).To(Equal("100m"))
+		})
+
+		It("should set CPU request from kubevirt-confg", func() {
+			cfgMap, err := virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Get(kubevirtConfig, metav1.GetOptions{})
+			Expect(err).To(BeNil())
+			cfgMap.Data[defaultCPURequestKey] = "800m"
+			_, err = virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Update(cfgMap)
```
rmohr: In theory there can be a race condition between detecting the config change and starting the VMI. In practice I don't think it matters for as long as the cluster is not in serious trouble, were we don't get reasonable results anyway then.

tests/vmi_configuration_test.go
ahadas: Good idea :thumbsup:

tests/vmi_configuration_test.go
ahadas: Right, note that it is done like that also in "should set machine type from kubevirt-config" and the tests don't seem to be affected by this

ahadas:
> Everything looks like it should. One request to make the tests faster.

Done - I'm checking for both Scheduled and Running as I've noticed we may miss getting the Scheduled phase. To be honest, I didn't see any improvement because the transition to Scheduled was always missed on my setup.

As for the new matcher, I can try to push it to the github.com/onsi/gomega.

rmohr:
> To be honest, I didn't see any improvement because the transition to Scheduled was always missed on my setup.

My mistake you have to look for `Scheduling`. Still it is good that you match for `Scheduled` and `Running` too.

ahadas:
> My mistake you have to look for `Scheduling`. Still it is good that you match for `Scheduled` and `Running` too.

Ok, done

ahadas:
@rmohr assuming the latest change in the new matcher won't break the test, could you please take another look at this? the execution time of the tests is now significantly faster :)

tests/be_in_matcher.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/onsi/gomega/format"
+	"github.com/onsi/gomega/matchers"
+	"github.com/onsi/gomega/types"
+)
+
+//BeIn succeeds if actual is contained in the passed in elements.
+//BeIn() always uses Equal() to perform the match.
+//When the passed in elements are comprised of a single element that is either a Map or Array or Slice, BeIn() behaves
+//as the reverse of ContainElement() that operates with Equal() to perform the match.
+//    Expect(2).Should(BeIn([]int{1, 2}))
```
rmohr: That is a nice matcher, but you can express it already the other way round:

```golang
Expect(map[string]int{"foo": 1, "bar": 2})).To(HaveKey("foo"))
Expect(map[string]int{"foo": 1, "bar": 2})).To(HaveKeyWithValue("foo", 1))
Expect([]string{"a", "b"}).To(ContainElement("a"))
```

Also this is possible: 

```golang
Expect(myfunc()).To(Or(Equal("a"), Equal("b")))
```

ahadas: > That is a nice matcher, but you can express it already the other way round:
> 
> ```go
> Expect(map[string]int{"foo": 1, "bar": 2})).To(HaveKey("foo"))
> Expect(map[string]int{"foo": 1, "bar": 2})).To(HaveKeyWithValue("foo", 1))
> Expect([]string{"a", "b"}).To(ContainElement("a"))
> ```

Right, in those cases, the order is insignificant (besides if we wish to keep the convention of having `ACTUAL` be the computed item and `VALUE` be its desired value consistent across the tests and `ACTUAL` is not a Map/Array/Slice).

But in other cases, the order may be significant. For example, when using `Eventually`:
```golang
Eventually(fn).Should(BeIn(phases))
```

> 
> Also this is possible:
> 
> ```go
> Expect(myfunc()).To(Or(Equal("a"), Equal("b")))
> ```

This alternative also requires me to have the values I compare the returned value of `myfunc` with hard-coded and in my case, I want them to be provided as an argument to `waitForVMIPhase`

rmohr: A nice matcher.

ahadas:
The test failures seem unrelated (also the one in check-patch.k8s-multus-1.13.3.el7.x86_64 - the VM couldn't be started because of some issue with its MAC address)

PR #2141, opened 2019-03-23
change the location of error log.

volume-ji:
change the location of error log
Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/virt-launcher/virtwrap/network/common.go
+++ b/pkg/virt-launcher/virtwrap/network/common.go
@@ -186,9 +186,9 @@ func (h *NetworkUtilsHandler) SetRandomMac(iface string) (net.HardwareAddr, erro
 		if changed {
 			mac, err = Handler.GetMacDetails(iface)
 			if err != nil {
+				log.Log.Reason(err).Errorf("updated MAC for interface: %s - %s", iface, mac)
 				return nil, err
 			}
-			log.Log.Reason(err).Errorf("updated MAC for interface: %s - %s", iface, mac)
 			break
 		}
 	}


ovirt-infra:

volume-ji:
ci test please

rmohr:
ci test please

booxter:
@rmohr @volume-ji what's the point of this PR? It now logs new MAC address ONLY when we failed to get MAC details?..

rmohr:
> @rmohr @volume-ji what's the point of this PR? It now logs new MAC address ONLY when we failed to get MAC details?..

@booxter I see what you mean, but this message reported and Error level log message no matter if an error occured or not. Feel free to add an info-level log message which uses `Infof` in the no-error case.

booxter:
OK so the problem was incorrect log level. It wasn't clear from the PR or commit messages, and the change got in the way of verifying custom MAC implementation for Multus.

I will fix it in a new PR.

rmohr:
> OK so the problem was incorrect log level. It wasn't clear from the PR or commit messages, and the change got in the way of verifying custom MAC implementation for Multus.

Is this something which will be tested in this repo? We use log messages in various of our tests in this repo, so they are strongly connected, but I would not advise to rely on log messages for tests in other repos, since we would, like it seems to be the case here, break tests somewhere else where can't be aware of.

I am not sure what you test, but I would suggest that you either start a VMI and verify it from inside the VMI that the right mac is there, or if you have to check something on the pod, enter the pod namespaces as part of the test and check it there. But just my two cents.

booxter:
@rmohr I merely want to validate that CNI plugin is triggered and had a chance to configure a custom MAC address (because we passed it via annotations). But then kubevirt immediately replaces the address inside the pod with a random address since the address is moved / configured for the VMI. Since most plugins we use don't really require specific actions on CNI plugin side to make a custom MAC address work, just starting a VMI and checking MAC address is set there and works is not enough. One would need a special CNI plugin like OpenShift SDN (with the MAC address actually honored, which is not yet the case) to functionally validate CNI plugin configured the MAC address properly. Or using a OVS CNI plugin that has native support for the custom MAC address.

This message is just one of indicators for QE to verify in a chain of steps for the feature. I noticed this because I told QE that, among other things, they should see the message with old and new MAC address in launcher log, but they couldn't validate that, because of this commit.

I've posted a PR fixing this here: https://github.com/kubevirt/kubevirt/pull/2190

rmohr:
Not sure I can follow. This looks like manual verification steps. Manual verification is pretty much useless. It is good to have log messages, but I would not use them in proper verification steps outside of that repo. At the end you have a measurable effect in the Pod or the vmi, I would recommend trying to verify that. I don't try to stop you, though. Just my opinion.

booxter:
@rmohr Folks doing the verification are definitely doing it manually. I agree with you automation is preferred but this is a separate discussion.

PR #2142, opened 2019-03-24
cleanup in setting default cpu model

ahadas:
This PR changes the way the default CPU model is fetched from kubevirt-config. It is now fetched via the virt-config package. This allows concentrating all configurable virt-properties in a single place and provides a unified way to access them.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-config/config-map.go
+++ b/pkg/virt-config/config-map.go
@@ -49,6 +49,7 @@ const (
 	useEmulationKey        = "debug.useEmulation"
 	imagePullPolicyKey     = "dev.imagePullPolicy"
 	migrationsConfigKey    = "migrations"
+	cpuModelKey            = "default-cpu-model"
 
 	ParallelOutboundMigrationsPerNodeDefault uint32 = 2
 	ParallelMigrationsPerClusterDefault      uint32 = 5
@@ -162,6 +163,7 @@ type Config struct {
 	MigrationConfig *MigrationConfig
 	ImagePullPolicy k8sv1.PullPolicy
 	MachineType     string
+	CPUModel        string
 }
 
 type MigrationConfig struct {
@@ -196,6 +198,10 @@ func (c *ClusterConfig) GetImagePullPolicy() (policy k8sv1.PullPolicy) {
 
 func (c *ClusterConfig) GetMachineType() string {
 	return c.getConfig().MachineType
+}
+
+func (c *ClusterConfig) GetCPUModel() string {
+	return c.getConfig().CPUModel
 }
 
 // setConfig parses the provided config map and updates the provided config.
@@ -246,6 +252,10 @@ func setConfig(config *Config, configMap *k8sv1.ConfigMap) error {
 	// set machine type
 	if machineType := strings.TrimSpace(configMap.Data[machineTypeKey]); machineType != "" {
 		config.MachineType = machineType
+	}
+
+	if cpuModel := strings.TrimSpace(configMap.Data[cpuModelKey]); cpuModel != "" {
+		config.CPUModel = cpuModel
 	}
 
 	return nil
--- a/pkg/virt-api/webhooks/mutating-webhook/mutating-webhook.go
+++ b/pkg/virt-api/webhooks/mutating-webhook/mutating-webhook.go
@@ -112,14 +112,13 @@ func mutateVMIs(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
 			},
 		}
 	}
-	// Apply default cpu model
-	setDefaultCPUModel(&vmi, informers.ConfigMapInformer.GetStore())
 
 	// Apply namespace limits
 	applyNamespaceLimitRangeValues(&vmi, informers.NamespaceLimitsInformer)
 
 	// Set VMI defaults
 	log.Log.Object(&vmi).V(4).Info("Apply defaults")
+	setDefaultCPUModel(&vmi, config.GetCPUModel())
 	if vmi.Spec.Domain.Machine.Type == "" {
 		vmi.Spec.Domain.Machine.Type = config.GetMachineType()
 	}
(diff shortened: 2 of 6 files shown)


rmohr:
Hm, I don't think this is the right way to go. Before your change you could simply change the config map and virt-api picked the change up. Now one would have to restart virt-api. I think the better way is to over-time let all components detect config changes, immediately pick them up, reconfigure itself, or simply exits so that it gets restarted with the new config.

I think that values which can be updated dynmaically should use the informer in the config package and we should add more dynamic config mechanisms (like callbacks) over time, so that kubevirt reconfigures itself completely. Here a simple example in a PR which did not get merged, regarding to using the informers in the config: https://github.com/kubevirt/kubevirt/pull/1404/commits/037f153050690b7f3c97d1ad55ce916fa2d82dc4#diff-091ace02b49a46f55f2586cfec89a160.

ahadas:
@rmohr , Fair enough, oVirt also supports getting the most up-to-date configuration without restart for some configuration values - that can be useful. I would still claim that having all the configurable virt-properties placed together makes the code more organized though - how about having pkg/virt-config/config-map listen to changes in the config-map and reloading the updated configuration?

rmohr:
> how about having pkg/virt-config/config-map listen to changes in the config-map and reloading the updated configuration?

Works for me. You can have a look on the referenced commit. It does that. You can copy from there anything you want if it helps.

rmohr:
@ahadas current master contains now a reworked `config-map.go` where you could tie in your changes.

ahadas:
@rmohr , done.
could you please take a look at mutating-webhook_test.go - I tried to mock kubevirt-config with some value for 'cpu-model' but it doesn't seems to work. what am I missing?

ahadas:
> @rmohr , done.
> could you please take a look at mutating-webhook_test.go - I tried to mock kubevirt-config with some value for 'cpu-model' but it doesn't seems to work. what am I missing?

and it seems to work just fine when deployed - that's why the mocking seems problematic

rmohr:
> and it seems to work just fine when deployed - that's why the mocking seems problematic

@ahadas I made `MakeFakeClusterConfig` in testutils over-complex. I guess that is the issue. I would suggest something like this in testutils:

```golang
func MakeFakeClusterConfig(configMap v1.ConfigMap) (*virtconfig.ClusterConfig, cache.Store) {
	store := cache.NewStore(cache.DeletionHandlingMetaNamespaceKeyFunc)
	copy := configMap.DeepCopy()
	copy.Name = "kubevirt-config"
	copy.Namespace = "kubevirt"
	copy.ResourceVersion = rand.String(10)
	store.Add(copy)
	return virtconfig.NewClusterConfig(store, "kubevirt"), store
}

func UpdateFakeClusterConfig(config v1.ConfigMap, store cache.Store) {
	copy := config.DeepCopy()
	copy.Name = "kubevirt-config"
	copy.Namespace = "kubevirt"
	// Change the resource version, or the config will not be updated
	copy.ResourceVersion = rand.String(10)
	store.Update(config)
}
```

Changes should then always be immediately picked up when passing a changed config

ahadas:
> > and it seems to work just fine when deployed - that's why the mocking seems problematic
> 
> @ahadas I made `MakeFakeClusterConfig` in testutils over-complex. I guess that is the issue. I would suggest something like this in testutils:
> 
> ```go
> func NewFakeClusterConfig(configMap v1.ConfigMap) (*virtconfig.ClusterConfig, cache.Store) {
> 	store := cache.NewStore(cache.DeletionHandlingMetaNamespaceKeyFunc)
> 	copy := configMap.DeepCopy()
> 	copy.Name = "kubevirt-config"
> 	copy.Namespace = "kubevirt"
> 	copy.ResourceVersion = rand.String(10)
> 	store.Add(copy)
> 	return virtconfig.NewClusterConfig(store, "kubevirt"), store
> }
> 
> func UpdateFakeClusterConfig(config v1.ConfigMap, store cache.Store) {
> 	copy := config.DeepCopy()
> 	copy.Name = "kubevirt-config"
> 	copy.Namespace = "kubevirt"
> 	// Change the resource version, or the config will not be updated
> 	copy.ResourceVersion = rand.String(10)
> 	store.Update(config)
> }
> ```
> 
> Changes should then always be immediately picked up when passing a changed config

Ack, thanks.
As it is still not clear to me how to use that as part of the tests of the mutating-webhook (I don't see how to create an informer that can be set in webhooks#informers out of the returned store), how about handling it separately? the current state of the PR is IMO improves the current state of the code, and the tests worked in a similar way before...

pkg/virt-config/config-map.go
```diff
@@ -248,6 +254,8 @@ func setConfig(config *Config, configMap *k8sv1.ConfigMap) error {
 		config.MachineType = machineType
 	}
 
+	config.CPUModel = strings.TrimSpace(configMap.Data[cpuModelKey])
```
rmohr: I guess that `CPUModel` has a default of `""`, still could you follow the pattern from above, but it is not clear now how you want to interpret it when it it is missing. Could you clarify that in the code?

See all examples above, they explicitly treat `""` as "keep the default". So a missing field or an empty string have the same meaning. It is also possible to differentiate between `""` and a not existing key if you need to.

rmohr:
> As it is still not clear to me how to use that as part of the tests of the mutating-webhook (I don't see how to create an informer that can be set in webhooks#informers out of the returned store), how about handling it separately? the current state of the PR is IMO improves the current state of the code, and the tests worked in a similar way before...

Yep needs a refactoring. Now with the cluster config in place it is now possible when wiring the webhook to pass in the final configuration of the cluster config, and not the informers.
I would prefer to see it in this PR but do not insist on it. In general the webhooks should be moved to a struct instead of being standalone functions. I think the time for that change has come.

pkg/virt-config/config-map.go
ahadas: > I guess that `CPUModel` has a default of `""`, still could you follow the pattern from above

No problem, it is equivalent - changed.

> , but it is not clear now how you want to interpret it when it it is missing. Could you clarify that in the code?
> See all examples above, they explicitly treat `""` as "keep the default". So a missing field or an empty string have the same meaning. It is also possible to differentiate between `""` and a not existing key if you need to.

I want both cases to be the same: no default CPU model should be set in those cases. Therefore the empty string is returned and is ignored in mutating-webhook. Could you please be more concrete about what needs to be clarified in the code and where?

rmohr: You did the change I wanted. :+1: Basically keeping the default if the string is `""` instead of setting it to `""`. The result is the same, because the default is `""` but it is explicit now.

ahadas:
@rmohr I see your point, but I think it would make this PR too broad. Let's keep its original scope and introduce those changes separately, please.

pkg/virt-config/config-map.go
```diff
@@ -49,6 +49,7 @@ const (
 	useEmulationKey        = "debug.useEmulation"
 	imagePullPolicyKey     = "dev.imagePullPolicy"
 	migrationsConfigKey    = "migrations"
+	cpuModelKey            = "default-cpu-model"
```
rmohr: ugh, I never notices that we have a mixture of `-` and camel case keys.

ahadas: :)
That's one of the benefits of having all those placed together - I'll change

rmohr: your style is right now the common one. I don't think that you have to change anything. Just noticed it.

rmohr:
Test failures are unrelated.

PR #2153, opened 2019-03-27
tests: skip migration tests if LiveMigration is not enabled

dollierp:

Skip Migrations tests (i.e those from `tests/migration_test.go`) if LiveMigration feature gate is not enabled instead of running them and seeing them fail.

This is the same behaviour as DataVolume tests (`tests/datavolume_test.go`) which are skipped when DataVolumes feature gate is not enabled.

**Release note**:

```release-note
NONE
```


slintes:
good catch, looks good to me, let's wait for CI

PR #2157, opened 2019-03-28
virtctl imageupload auto detects ingress/route

igoihman:
Instead of passing Upload Proxy URL as argument to `virtctl imageupload`, the tool will lookup for CDIConfig object and find the  Upload Proxy URL (if Ingress/Route exists).
If it doesn't exist, the program will exit with an error message.

Fixes #

**Release note**:

```release-note
virtctl image upload now auto detects Ingress/Route details
```

### DIFF
--- a/pkg/virtctl/imageupload/imageupload.go
+++ b/pkg/virtctl/imageupload/imageupload.go
@@ -55,6 +55,8 @@ const (
 
 	//UploadProxyURI is a URI of the upoad proxy
 	UploadProxyURI = "/v1alpha1/upload"
+
+	configName = "config"
 )
 
 var (
@@ -104,7 +106,6 @@ func NewImageUploadCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
 	}
 	cmd.Flags().BoolVar(&insecure, "insecure", insecure, "Allow insecure server connections when using HTTPS.")
 	cmd.Flags().StringVar(&uploadProxyURL, "uploadproxy-url", "", "The URL of the cdi-upload proxy service.")
-	cmd.MarkFlagRequired("uploadproxy-url")
 	cmd.Flags().StringVar(&pvcName, "pvc-name", "", "The destination PVC.")
 	cmd.MarkFlagRequired("pvc-name")
 	cmd.Flags().StringVar(&pvcSize, "pvc-size", "", "The size of the PVC to create (ex. 10Gi, 500Mi).")
@@ -175,11 +176,20 @@ func (c *command) run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	if uploadProxyURL == "" {
+		uploadProxyURL, err = getUploadProxyURL(virtClient.CdiClient())
+		if err != nil {
+			return err
+		}
+		if uploadProxyURL == "" {
+			return fmt.Errorf("Upload Proxy URL not found")
+		}
+	}
 	u, err := url.Parse(uploadProxyURL)
 	if err != nil {
 		return err
 	} else if u.Scheme == "" {
-		uploadProxyURL = fmt.Sprintf("http://%s", uploadProxyURL)
+		uploadProxyURL = fmt.Sprintf("https://%s", uploadProxyURL)
 	}
 	fmt.Printf("Uploading data to %s\n", uploadProxyURL)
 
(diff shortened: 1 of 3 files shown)


igoihman:
@rmohr can you review?

pkg/virtctl/imageupload/imageupload.go
```diff
@@ -411,3 +419,17 @@ func getAndValidateUploadPVC(client kubernetes.Interface, namespace, name string
 
 	return pvc, nil
 }
+
+func getUploadProxyURL(client cdiClientset.Interface) (string, error) {
+	cdiConfig, err := client.CdiV1alpha1().CDIConfigs().Get("config", metav1.GetOptions{})
```
rmohr: Does the CDI installation take care of RBAC? Is this accessible via RBAC aggregation? Does it only contain non-sensitive information by definion or is it just an "admin" helper?

pkg/virtctl/imageupload/imageupload_test.go
```diff
@@ -224,6 +250,15 @@ var _ = Describe("ImageUpload", func() {
 			validatePVC()
 		})
 
+		It("Use CDI Config UploadProxyURL", func() {
```
rmohr: This test and ...

pkg/virtctl/imageupload/imageupload_test.go
```diff
@@ -249,6 +284,13 @@ var _ = Describe("ImageUpload", func() {
 			Expect(cmd()).NotTo(BeNil())
 		})
 
+		It("uploadProxyURL not configured", func() {
```
rmohr: ... this test look pretty similar. What is the difference?

igoihman: oops, forgot to update this one.

pkg/virtctl/imageupload/imageupload.go
igoihman: This is just an admin helper, non-sensitive information. CDI takes care of RBAC.

pkg/virtctl/imageupload/imageupload.go
```diff
@@ -411,3 +421,17 @@ func getAndValidateUploadPVC(client kubernetes.Interface, namespace, name string
 
 	return pvc, nil
 }
+
+func getUploadProxyURL(client cdiClientset.Interface) (string, error) {
+	cdiConfig, err := client.CdiV1alpha1().CDIConfigs().Get(configName, metav1.GetOptions{})
```
rmohr: Do you need special handling for errors.IsNotFound()? Something like, "no default upload proxy configured, use --uploadproxy-url to pass it manually"? Is it supposed to always exist, or should the client gracefully work with older cdi versions (like e.g. a nice error message like I mentioned?

pkg/virtctl/imageupload/imageupload.go
```diff
@@ -175,11 +176,20 @@ func (c *command) run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	if uploadProxyURL == "" {
+		uploadProxyURL, err = getUploadProxyURL(virtClient.CdiClient())
+		if err != nil {
+			return err
+		}
+		if uploadProxyURL == "" {
```
rmohr: I have see  the pattern to return a tripple too, something like ` func() (url string, exists bool, err error)`

pkg/virtctl/imageupload/imageupload_test.go
```diff
@@ -249,6 +285,17 @@ var _ = Describe("ImageUpload", func() {
...
+			testInit(http.StatusOK)
+			cmd := tests.NewRepeatableVirtctlCommand(commandName, "--pvc-name", pvcName, "--pvc-size", pvcSize,
+				"--insecure", "--image-path", imagePath)
+			config, err := cdiClient.CdiV1alpha1().CDIConfigs().Get(configName, metav1.GetOptions{})
+			Expect(err).To(BeNil())
+			config.Status.UploadProxyURL = nil
+			updateCDIConfig(config)
+			Expect(cmd()).NotTo(BeNil())
```
rmohr: Maybe also good to check that the error is the one you expect. Btw. there exists also a nice meaningful `BeNil()` check for errors, namely `NotTo(Succeed())`

pkg/virtctl/imageupload/imageupload_test.go
```diff
@@ -249,6 +285,17 @@ var _ = Describe("ImageUpload", func() {
 			Expect(cmd()).NotTo(BeNil())
 		})
 
+		It("uploadProxyURL not configured", func() {
```
rmohr: Maybe rename the test to something like "should fail to upload if no uploadProxyURL is configured"

pkg/virtctl/imageupload/imageupload.go
igoihman: It is supposed to always exist. CDIConfig is created when CDI is deployed since 1.7 release.
I can change the error message to a nicer message, please note that `virtctl` won't work with older CDI versions after this change.

pkg/virtctl/imageupload/imageupload_test.go
igoihman: it's part of ""Upload fails" context, we will have a duplicate title if I change the test name.

rmohr:
Let's wait for CI.

PR #2159, opened 2019-03-28
possibility to set machine type from kubevirt-config

mfranczy:
The feature allows to read machine type from kubevirt-config configMap

Fixes #

todo:
- [x] integration tests

**Release note**:

```release-note
Read machine type from kubevirt-config
```

### DIFF
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -20,7 +20,6 @@
 package tests_test
 
 import (
-	"encoding/xml"
 	"flag"
 	"fmt"
 	"regexp"
@@ -43,7 +42,6 @@ import (
 	"kubevirt.io/kubevirt/pkg/kubecli"
 	"kubevirt.io/kubevirt/pkg/log"
 	hw_utils "kubevirt.io/kubevirt/pkg/util/hardware"
-	launcherApi "kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
 	"kubevirt.io/kubevirt/tests"
 )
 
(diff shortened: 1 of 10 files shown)


michalskrivanek:
nice and simple!

tests/vmi_configuration_test.go
```diff
@@ -989,6 +987,56 @@ var _ = Describe("Configurations", func() {
...
+			cfgMap, err := virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Get(kubevirtConfig, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			cfgMap.Data[defaultMachineTypeKey] = ""
+
+			_, err = virtClient.CoreV1().ConfigMaps(namespaceKubevirt).Update(cfgMap)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should set machine type from to VMI spec", func() {
```
stu-gott: minor typo. "from to"

mfranczy: fixed, thanks

cynepco3hahue:
@mfranczy Hm it looks like the same functionallity of `emulatedMachinesEnvVar = "VIRT_EMULATED_MACHINES"`
variable.

mfranczy:
@cynepco3hahue huh, I was not aware of that, I will check it.

/hold

ahadas:
The `VIRT_EMULATED_MACHINES` config sets the group that the **specified** machine type is validated against but it does not specify the default. The default [is set to q35 in api/v1/defaults.go](https://github.com/kubevirt/kubevirt/blob/master/pkg/api/v1/defaults.go#L127). I'm currently working on [a PR that allows settings a configurable default in defaults.go](https://github.com/kubevirt/kubevirt/pull/2137) that could help with this one.

mfranczy:
@cynepco3hahue `emulatedMachinesEnvVar` is for validation https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-config/virt-properties.go#L34 I understand you can extend allowed machine types. This PR is more about setting a default machine type for a cluster, however, thanks for pointing this because virt-config pkg might be a better place for this code, let me check how it fits there.

mfranczy:
> I'm currently working on a PR that allows settings a configurable default in defaults.go that could help with this one

@ahadas If you're working on something where you can customize all default values then I think I should drop that PR and we should have one common way of overwriting defaults (in that case your way). What do you think?

ahadas:
> @ahadas If you're working on something where you can customize all default values then I think I should drop that PR and we should have one common way of overwriting defaults (in that case your way). What do you think?

Sure, fine by me.

mfranczy:
agreed with @ahadas that I will re-open this PR because it might take more time to deliver customizable defaults

mfranczy:
@vladikr currently a default machine type is set [here](https://github.com/kubevirt/kubevirt/blob/master/pkg/api/v1/defaults.go#L112), I think you're right and it makes sense to move it to virt-config, will do.

mfranczy:
@vladikr I changed the PR according to your comment, is that what you meant?

vladikr:
> @vladikr I changed the PR according to your comment, is that what you meant?

:+1: Yep, thanks. Maybe as a follow up, please move this `const DefaultMachineType = "q35"` to  `pkg/virt-config/config-map.go
` where other defaults are.

mfranczy:
@vladikr done.

PR #2160, opened 2019-03-28
Collect runtime statistics

danielBelenky:
We want to collect some data so we'll be able to analyze it
later and get a better picture of bottlenecks and performance.

My only concern is that the _profile function will slow things down a bit.
On the other hand, we'll have good data to analyze within a few days.

**Release note**:
```release-note
NONE
```


danielBelenky:
ci test please

PR #2162, opened 2019-03-28
hyperv: add feature gate and node selector

fromanirh:
Add support to check HyperV support on the host side, depending on the labels.
Label will be added by https://github.com/fromanirh/kvm-info-nfd-plugin

Fixes possible, although rare, VM launch failures if hyperv features are requested on not up-to-date hosts, see:
https://lists.ovirt.org/archives/list/users@ovirt.org/thread/GHUBLKYNM4G5QOIEFY5H7QKPLBDU5H7Z/#GHUBLKYNM4G5QOIEFY5H7QKPLBDU5H7Z

**Release note**:

```release-note
Add feature gate to enable strict checking of the hyperv support from the host
```

### DIFF
--- a/pkg/virt-config/config-map.go
+++ b/pkg/virt-config/config-map.go
@@ -44,13 +44,21 @@ const (
 // We cannot rely on automatic invocation of 'init' method because this initialization
 // code assumes a cluster is available to pull the configmap from
 func Init() {
-	cfgMap := getConfigMap()
+	InitFromConfigMap(getConfigMap())
+}
+
+func InitFromConfigMap(cfgMap *k8sv1.ConfigMap) {
 	if val, ok := cfgMap.Data[FeatureGatesKey]; ok {
 		os.Setenv(featureGateEnvVar, val)
 	}
 	if val, ok := cfgMap.Data[emulatedMachinesKey]; ok {
 		os.Setenv(emulatedMachinesEnvVar, val)
 	}
+}
+
+func Clear() {
+	os.Unsetenv(featureGateEnvVar)
+	os.Unsetenv(emulatedMachinesEnvVar)
 }
 
 func getConfigMap() *k8sv1.ConfigMap {
--- a/pkg/virt-config/feature-gates.go
+++ b/pkg/virt-config/feature-gates.go
@@ -30,12 +30,13 @@ import (
 )
 
 const (
-	dataVolumesGate      = "DataVolumes"
-	cpuManager           = "CPUManager"
-	ignitionGate         = "ExperimentalIgnitionSupport"
-	liveMigrationGate    = "LiveMigration"
-	SRIOVGate            = "SRIOV"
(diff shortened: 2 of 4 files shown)


fromanirh:
CC @fabiand @rmohr @vladikr @slintes - this is the updated PR after our mtg.
We need a minor update to user-guide because the feature gate has now a more expressive name.

slintes:
The code looks good to me, but i would like to have 2nd opinion from @vladikr, since he is more familiar with the topic.

And I have a question @fromanirh: I understand that this prevents starting VMs which use hyperv kernel features, when the kernel doesn't provide those features, which is a rare situation according to your PR description. But this assumes that NFD with your KVM plugin is running, right?
So, correct me if I'm wrong, but I am concerned that we prevent more VM starts because NDF/your plugin is not running, than because the kernel is missing a feature. Or do I miss something?
Can we check somehow if your plugin is running, and only apply the node selectors if it does?

fromanirh:
> The code looks good to me, but i would like to have 2nd opinion from @vladikr, since he is more familiar with the topic.

Sure, thanls for your review
 
> And I have a question @fromanirh: I understand that this prevents starting VMs which use hyperv kernel features, when the kernel doesn't provide those features, which is a rare situation according to your PR description. But this assumes that NFD with your KVM plugin is running, right?

Yes, this is the case we need to enforce if the feature gate is turned on (which is not by default) and that the kubevirt-ssp-operator is going to guarantee.

IOW, the SSP-operator (https://github.com/MarSik/kubevirt-ssp-operator) is going to guarantee that the NFD + kvm info plugin will run, so we will have good labelling.

At that point the user can choose to enable the feature gate to make sure we have optimal/guaranteed scheduling instead of best effort scheduling.

This is also the reason why I think it is better to have the feature gate disabled by default.

> So, correct me if I'm wrong, but I am concerned that we prevent more VM starts because NDF/your plugin is not running, than because the kernel is missing a feature. Or do I miss something?
> Can we check somehow if your plugin is running, and only apply the node selectors if it does?

You got it right. I'm not aware of any clean way to check for the NFD to be running.
I think that the feature gate + ssp operator is a good enough UX, but I'm open for discussion here.

I just came to realize that we may want to do REVERSE labelling, and have my kvm NFD plugin label nodes which LACK key hyperv features. This may be even better because HyperV support is expected to work fine in the 99% of the cases.
I'm open to this approach, but we need to reach a final agreement pretty soon.

fabiand:
> But this assumes that NFD with your KVM plugin is running, right?

Just picky here: We assum the labels are there, and NFD is a convenient solution to get them.

slintes:
>  think that the feature gate + ssp operator is a good enough UX

ah yes, for a moment I forgot that this is feature gated, sorry. Works for me!

> I just came to realize that we may want to do REVERSE labelling

nice idea! But I have no clear opinion if it would be worth the additional work. LGTM as it is

PR #2164, opened 2019-03-29
Reject floppy disks

### LINKED ISSUE #2016: Drop floppy support
/kind enhancement

Today our API supports floppy drives: https://kubevirt.io/api-reference/master/definitions.html#_v1_floppytarget

However, newer qemu machine types don#t support floppies anymore.
Because none of our core functionality is depending on floppies, and because most floppy uses can be safely replaced by CD-ROM based flows (CD-ROM can still be used), I'm proposing to drop the floppy support from our API.

slintes:
Reject floppy disks

Fixes #2016 

**Release note**:
```release-note
Deprecated floppy disks: they are rejected now and will be removed from the API soon 
```

### DIFF
--- a/pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
+++ b/pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
@@ -125,6 +125,16 @@ func validateDisks(field *k8sfield.Path, disks []v1.Disk) []metav1.StatusCause {
 				Field:   field.Index(idx).Child("name").String(),
 			})
 		}
+
+		// Reject Floppy disks
+		if disk.Floppy != nil {
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeFieldValueNotSupported,
+				Message: fmt.Sprintf("Floppy disks are deprecated and will be removed from the API soon."),
+				Field:   field.Index(idx).Child("name").String(),
+			})
+		}
+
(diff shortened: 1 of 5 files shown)


pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
```diff
@@ -125,6 +125,16 @@ func validateDisks(field *k8sfield.Path, disks []v1.Disk) []metav1.StatusCause {
 				Field:   field.Index(idx).Child("name").String(),
 			})
 		}
+
+		// Reject Floppy disks
+		if disk.Floppy != nil {
```
stu-gott: The discussion in https://bugzilla.redhat.com/show_bug.cgi?id=1674443 is somewhat ambiguous to me. Was the final decision to outright reject floppies altogether (as is done here), or to only reject them for q35+ machine types?

Since KubeVirt only supports q35 and future, it's not really clear to me that there's a distinction here, but just wanted to double check.

vladikr:
Looks good to me. Just need to adjust this test `[test_id:1596]should reject POST if validation webhoook deems the spec is invalid`

pkg/virt-api/webhooks/validating-webhook/validating-webhook.go
slintes: As clarified offline, agreement is to always reject floppy disks

PR #2168, opened 2019-04-03
Improve start and stop event messages for vm

legionus:

PR clarifies the essence of what is happening when creating/deleting an virtual machine instance. The previous event message is confusing to users because it does not mention an instance.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1694614

@fabiand please review

### DIFF
--- a/pkg/virt-controller/watch/replicaset.go
+++ b/pkg/virt-controller/watch/replicaset.go
@@ -293,11 +293,11 @@ func (c *VMIReplicaSet) scale(rs *virtv1.VirtualMachineInstanceReplicaSet, vmis 
 				if err != nil {
 					// We can't observe a delete if it was not accepted by the server
 					c.expectations.DeletionObserved(rsKey, controller.VirtualMachineKey(deleteCandidate))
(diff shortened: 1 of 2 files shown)


pkg/virt-controller/watch/replicaset.go
```diff
@@ -320,11 +320,11 @@ func (c *VMIReplicaSet) scale(rs *virtv1.VirtualMachineInstanceReplicaSet, vmis
 				vmi, err := c.clientset.VirtualMachineInstance(rs.ObjectMeta.Namespace).Create(vmi)
 				if err != nil {
 					c.expectations.CreationObserved(rsKey)
-					c.recorder.Eventf(rs, k8score.EventTypeWarning, FailedCreateVirtualMachineReason, "Error creating virtual machine: %v", err)
+					c.recorder.Eventf(rs, k8score.EventTypeWarning, FailedCreateVirtualMachineReason, "Error creating the instance of virtual machine: %v", err)
```
slintes: Can you change this to `Error creating the virtual machine instance`, in order to have it consistent with all other places, please?

legionus: @slintes Fixed

PR #2174, opened 2019-04-04
Move the live migration config to virt-config

vladikr:
So far an additional live migration configuration was part of the migration object.
This PR moves these parameters to be part of the cluster-wide kubevirt config.

```release-note
NONE
```


vladikr:
Thank! I think the failures aren't related. The cancelation test is just the migration didn't start, I think this is due to slowness - it did pass on other tests. 
The failure on CRI-O is because of the containerDisk doesn't terminate after a migration. I've recently noticed it, but it's not related to this change.

PR #2181, opened 2019-04-10
update libvirt image to v5.1.0

vladikr:
Update libvirt version to v5.1.0 mainly to support live migration with cephfs.

```release-note
NONE
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -47,17 +47,17 @@ http_archive(
 # Libvirt dependencies
 http_file(
     name = "libvirt_libs",
-    sha256 = "95d317fd645fb52745d642578263a9bcb0d796beadf00aeadebc0d692f5529ba",
+    sha256 = "45448bd85014d575fbd79f6f302749fa85dd24fcbed1cf05ac9b19931e49810f",
     urls = [
-        "https://libvirt.org/sources/libvirt-libs-5.0.0-1.fc28.x86_64.rpm",
+        "https://libvirt.org/sources/libvirt-libs-5.1.0-1.fc28.x86_64.rpm",
     ],
 )
 
 http_file(
     name = "libvirt_devel",
-    sha256 = "6573a047d777ed00f6858c2e75c780053b1f898ae1c3f7415e991c94c5ccdd70",
+    sha256 = "885edf8828ab4e22537da5721deae1d852edb94bbfba061a14ae61631270476d",
     urls = [
-        "https://libvirt.org/sources/libvirt-devel-5.0.0-1.fc28.x86_64.rpm",
+        "https://libvirt.org/sources/libvirt-devel-5.1.0-1.fc28.x86_64.rpm",
     ],
 )
 
@@ -154,7 +154,7 @@ container_pull(
 # Pull base image libvirt
 container_pull(
     name = "libvirt",
-    digest = "sha256:081f113a73748775e5f37d8fb877a574f595df1551e39e48ebbe8e8afd501d3b",
+    digest = "sha256:64d075a7d5976ce6d69f7483ae4d2d9997e6a18a04a1a435995dbc152c8eb0db",
     registry = "index.docker.io",
     repository = "kubevirt/libvirt",
     #tag = "5.0.0",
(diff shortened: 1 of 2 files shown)


vladikr:
something has changed in libvirt which now fails to lauch qemu in virt-launcher, with 
```Failed to start QEMU binary /usr/bin/qemu-system-x86_64 for probing: qemu-system-x86_64: cannot create PID file: Cannot open pid file: Permission denied","pos":"qemuProcessQMPLaunch:8530```

Potentially because 
```
commit 5724dca13560ff1f5ae012e8ad401602439da2a9
Author: Chris Venteicher <cventeic@redhat.com>
Date:   Wed Feb 13 17:22:31 2019 +0100

    qemu_process: Use unique directories for QMP processes
    
    Multiple QEMU processes for QMP commands can operate concurrently.
    
    Use a unique directory under libDir for each QEMU process to avoid
    pidfile and unix socket collision between processes.
    
    The pid file name is changed from "capabilities.pidfile" to "qmp.pid"
    because we no longer need to avoid a possible clash with a qemu domain
    called "capabilities" now that the processes artifacts are stored in
    their own unique temporary directories.
    
    "Capabilities" was changed to "qmp" in the pid file name because these
    processes are no longer specific to the capabilities usecase and are
    more generic in terms of being used for any general purpose QMP message
    exchanges with a QEMU process that is not associated with a domain.
```

**Comment by fabiand** [comment]
Ouch. Does this require a larger change then?

**Comment by vladikr** [comment]
> Ouch. Does this require a larger change then?

Yes. At least need to figure out what's going on.

**Comment by vladikr** [comment]
I see that the qmp-XXXXX directory is being created in `/var/lib/libvirt/qemu/` but it belongs to `root` and not `qemu` 

```
/var/lib/libvirt/qemu/qmp-Epmjsk:
0 drwx------. 2 root root   6 Apr 11 00:54 qmp-Epmjsk
```

**Comment by vladikr** [comment]
@nertpinx perhaps you have an insight?

File: WORKSPACE [hunk 1]
```diff
@@ -154,7 +154,7 @@ container_pull(
 # Pull base image libvirt
 container_pull(
     name = "libvirt",
-    digest = "sha256:081f113a73748775e5f37d8fb877a574f595df1551e39e48ebbe8e8afd501d3b",
+    digest = "sha256:64d075a7d5976ce6d69f7483ae4d2d9997e6a18a04a1a435995dbc152c8eb0db",
     registry = "index.docker.io",
     repository = "kubevirt/libvirt",
     #tag = "5.0.0",
```
cynepco3hahue: Can you change the tag under the comment just for visibility?

cynepco3hahue:
/hold
Breaking change under the libvirt

cynepco3hahue:
@vladikr Maybe we can set correct permissions for the directory via file https://github.com/kubevirt/libvirt/blob/master/augconf

rmohr:
@vladikr we rebuilt the libvirt image with the latest rpm rebuild. Let's see.

vladikr:
@cynepco3hahue @rmohr I think the problem is that we are running libvirtd as root and it seems that libvirt doesn't set permissions for this temp. directory.
https://github.com/libvirt/libvirt/blob/master/src/qemu/qemu_process.c#L8461

vladikr:
> Maybe we can set correct permissions for the directory via file https://github.com/kubevirt/libvirt/blob/master/augconf

@cynepco3hahue I'm not sure how can we do it. It's a temp directory that libvirt creates in `/var/lib/libvirt/qemu/` which qemu should use.
I thought as a workaround to set g+s and change umask on /var/lib/libvirt/qemu/

nertpinx:
@vladikr Running libvirtd as root is fine, that should not be the problem. The problem is a combination of the aforementioned commit libvirt/libvirt@5724dca13560ff1f5ae012e8ad401602439da2a9 and libvirt/libvirt@a2d3dea9d41dba313d9566120a8ec9d358567bd0, which makes it work outside of a container. I have a fix for labelling the paths correctly, will post it after some testing (to see whether it helped or not).

nertpinx:
Posted patch (nertpinx/libvirt@0e68f22354d6d4d6ce607485da4ee370d10021e0) upstream:

https://www.redhat.com/archives/libvir-list/2019-April/msg00876.html

fabiand:
@vladikr should this be closed?

vladikr:
yes. Let's close this

PR #2186, opened 2019-04-15
allow cluster administrators to disable libvirt's migration verification

### LINKED ISSUE #2194: VMI with DataVolume shared PVC can't be successfully migrated.
> /kind bug

**What happened**:
 An incorrect migration method is set. BlockMigration instead of LiveMigration.
```
Status:
  Conditions:
    Last Probe Time:       <nil>
    Last Transition Time:  <nil>
    Status:                True
    Type:                  LiveMigratable
    Last Probe Time:       <nil>
    Last Transition Time:  2019-02-28T20:52:22Z
    Status:                True
    Type:                  Ready
  Interfaces:
    Ip Address:      10.129.0.100
    Mac:             0a:58:0a:81:00:64
    Name:            default
  Migration Method:  BlockMigration
  Node Name:         cnv-executor-dshchedr-node1.example.com
  Phase:             Running
```

A Migration with a shared DataVolume should be allowed

vladikr:

In some rare cases, it may be required to disable libvirt's live migration verifications.
For example, when users are using a shared filesystem of a new type that libvirt doesn't have complete support for and mistakenly considers the migration unsafe. 
More specifically, CEPHFS support has been introduced, however, due to various limitations, we won't be able to consume it, at least until libvirt v5.3.0.

This PR allows the administrators to disable libvirt's live migration verifications by updating the kubevirt-config config map with an unsafeMigrationOverridesetting parameter.

This PR also considers DataVolumes when performing per-live migration validations.
So far the code has ignored DataVolume and didn't reject a migration of a VMI with a non-shared DataVolume nor set a correct migration method.

Fixes #2194 

```release-note
None
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/manager.go
+++ b/pkg/virt-launcher/virtwrap/manager.go
@@ -279,7 +279,7 @@ func classifyVolumesForMigration(vmi *v1.VirtualMachineInstance) *migrationDisks
 	}
 	for _, volume := range vmi.Spec.Volumes {
 		volSrc := volume.VolumeSource
-		if volSrc.PersistentVolumeClaim != nil ||
+		if volSrc.PersistentVolumeClaim != nil || volSrc.DataVolume != nil ||
 			(volSrc.HostDisk != nil && *volSrc.HostDisk.Shared) {
 			disks.shared[volume.Name] = true
 		}
--- a/pkg/virt-launcher/virtwrap/manager.go
+++ b/pkg/virt-launcher/virtwrap/manager.go
@@ -243,11 +243,14 @@ func (l *LibvirtDomainManager) setMigrationResultHelper(vmi *v1.VirtualMachineIn
 
 }
 
-func prepareMigrationFlags(isBlockMigration bool) libvirt.DomainMigrateFlags {
+func prepareMigrationFlags(isBlockMigration bool, isUnsafeMigration bool) libvirt.DomainMigrateFlags {
 	migrateFlags := libvirt.MIGRATE_LIVE | libvirt.MIGRATE_PEER2PEER
 
 	if isBlockMigration {
 		migrateFlags |= libvirt.MIGRATE_NON_SHARED_INC
+	}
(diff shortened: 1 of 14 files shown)


fromanirh:
Code per se looks fine, but from oVirt experience, overruling libvirt safety checks is almost always suspcious, so I'd suggest to
1. add a concrete example in the commit message for our future reference
2. add a loud warning in the logs when unsafe migration is used

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -249,6 +250,12 @@ func prepareMigrationFlags(isBlockMigration bool) libvirt.DomainMigrateFlags {
 	if isBlockMigration {
 		migrateFlags |= libvirt.MIGRATE_NON_SHARED_INC
 	}
+	val, isSet := os.LookupEnv("UNSAFE_MIGRATION_OVERRIDE")
```
rmohr: How do you set that env variable? It has to be on virt-launcher, right? So it would have to be set in virt-controller? Or do you plan to use a pod preset for that?

rmohr:
I don't feel comfortable with the fact that normal users can influence that via presets. :/

pkg/virt-launcher/virtwrap/manager.go
vladikr: I was thinking that it should be set using a pod preset, same as LIBVIRT_DEBUG is being set.
This way, it won't be part of the api and will stay hidden, to be used with only specific rare cases.

vladikr:
> 2\. add a loud warning in the logs when unsafe migration is used

Good point, thanks.

vladikr:
> I don't feel comfortable with the fact that normal users can influence that via presets. :/

What do you suggest?

vladikr:
Ci test please

vladikr:
the cancelation failure in 'check-patch.k8s-1.10.11.el7.x86_64' is not related to this change

PR #2188, opened 2019-04-16
Make OLM package name configurable

slintes:
Make OLM package name configurable, because it has to be unique per Marketplace instance, but might be pulled from multiple sources (e.g. operatorhub.io vs KubeVirt's Quay repo vs. developer's Quay repo)

- The package name for KubeVirt's Quay repo changed to "kubevirt-operatorhub"  (feel free to propose a better name)
- That will make it possible to use "kubevirt" for operatorhub.io
- Default value is "kubevirt-dev" for development

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,8 @@ deploy:
     condition: $TRAVIS_BRANCH != revert-*
 - provider: script
   script: docker login -u="$DOCKER_USER" -p="$DOCKER_PASS" && DOCKER_PREFIX="index.docker.io/kubevirt"
-    DOCKER_TAG="$TRAVIS_TAG" QUAY_REPOSITORY="kubevirt" sh -c 'make bazel-push-images && make manifests && make olm-push'
+    DOCKER_TAG="$TRAVIS_TAG" QUAY_REPOSITORY="kubevirt" PACKAGE_NAME="kubevirt-operatorhub"
+    sh -c 'make bazel-push-images && make manifests && make olm-push'
   skip_cleanup: true
   file:
   on:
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -41,7 +41,8 @@ deploy:
     condition: $TRAVIS_BRANCH != revert-*
 - provider: script
   script: docker login -u="$DOCKER_USER" -p="$DOCKER_PASS" && DOCKER_PREFIX="index.docker.io/kubevirt"
-    DOCKER_TAG="$TRAVIS_TAG" QUAY_REPOSITORY="kubevirt" sh -c 'make bazel-push-images && make manifests && make olm-push'
+    DOCKER_TAG="$TRAVIS_TAG" QUAY_REPOSITORY="kubevirt" PACKAGE_NAME="kubevirt-staging"
```
rmohr: shouldn't this one here be `kubevirt`? This one should only be executed when we are on tags, see the whole config. I think `kubevirt-staging` would have to be on the script step above.

rmohr:
> * The package name for KubeVirt's Quay repo changed to "kubevirt-staging"  (feel free to propose a better name. I came up with this because we push every release automatically to that repo, before adding it manually to operatorhub.io, so this is a kind of a staging environment)

Maybe `kubevirt-latest`?

.travis.yml
slintes: at least for the moment we only push (`make olm-push`) on tags/releases to Quay, to our own repository. In order to be able to use plain "kubevirt" for operatorhub.io (which is coming from another Quay repository, and a manual process of adding our manifests to another git repository [1]) , I want to use something different than "kubevirt" here.

[1] https://github.com/operator-framework/community-operators/tree/master/upstream-community-operators

slintes:
> > * The package name for KubeVirt's Quay repo changed to "kubevirt-staging"  (feel free to propose a better name. I came up with this because we push every release automatically to that repo, before adding it manually to operatorhub.io, so this is a kind of a staging environment)
> 
> Maybe `kubevirt-latest`?

good idea, done

.travis.yml
rmohr: ok thx.

slintes:
after some discussion, it's kubevirt-operatorhub now

PR #2190, opened 2019-04-16
Log new / old MAC address on success

booxter:
A previous commit (aaade4bcdc2f9561f9ae92b86bd6a206e33ee922) incorrectly
hid the log message under error condition branch. The commit did it
because it was assumed by the author and reviewers that if Errorf is
used then it means that the message is indicating an error in program
flow.

In reality, the message belongs to no-error branch of the program flow.
So the issue is not the condition under which the message is logged but
the log level used for it.

This commit returns the message where it was before the incorrect
commit, only switching the log level to INFO.

```release-note
NONE
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/network/common.go
+++ b/pkg/virt-launcher/virtwrap/network/common.go
@@ -200,7 +200,7 @@ func (h *NetworkUtilsHandler) SetRandomMac(iface string) (net.HardwareAddr, erro
 			if err != nil {
 				return nil, err
 			}
-			log.Log.Infof("updated MAC for interface: %s - %s -> %s", iface, currentMac, mac)
+			log.Log.Infof("updated MAC for %s interface: old: %s -> new: %s", iface, currentMac, mac)
 			break
 		}
 	}
--- a/pkg/virt-launcher/virtwrap/network/common.go
+++ b/pkg/virt-launcher/virtwrap/network/common.go
@@ -198,9 +198,9 @@ func (h *NetworkUtilsHandler) SetRandomMac(iface string) (net.HardwareAddr, erro
 		if changed {
 			mac, err = Handler.GetMacDetails(iface)
 			if err != nil {
-				log.Log.Reason(err).Errorf("updated MAC for interface: %s - %s -> %s", iface, currentMac, mac)
 				return nil, err
 			}
+			log.Log.Infof("updated MAC for interface: %s - %s -> %s", iface, currentMac, mac)
 			break
 		}
 	}


booxter:
@rmohr please take a look

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -198,9 +198,9 @@ func (h *NetworkUtilsHandler) SetRandomMac(iface string) (net.HardwareAddr, erro
 		if changed {
 			mac, err = Handler.GetMacDetails(iface)
 			if err != nil {
-				log.Log.Reason(err).Errorf("updated MAC for interface: %s - %s -> %s", iface, currentMac, mac)
 				return nil, err
 			}
+			log.Log.Reason(err).Infof("updated MAC for interface: %s - %s -> %s", iface, currentMac, mac)
```
rmohr: 'Reason(err)' is wrong. That is one of the reasons why we moved it to the error branch. Can you make the message more descriptive? Was there before but without looking at the source code it would be impossible for me to understand its meaning.

booxter: It shouldn't be an error at all, missed that. Sec..

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -198,9 +198,9 @@ func (h *NetworkUtilsHandler) SetRandomMac(iface string) (net.HardwareAddr, erro
 		if changed {
 			mac, err = Handler.GetMacDetails(iface)
 			if err != nil {
-				log.Log.Reason(err).Errorf("updated MAC for interface: %s - %s -> %s", iface, currentMac, mac)
 				return nil, err
 			}
+			log.Log.V(4).Infof("updated MAC for interface: %s - %s -> %s", iface, currentMac, mac)
```
rmohr: V(4) is above the default log level. Is that intended?

booxter: No just a copy paste from below. I've now checked `docs/logging.md` and looks like I should avoid `V(*)`?

rmohr: You can set it to something likte 4 or above it is mostly meant for debugging and would pollute the logs otherwise. Just don't set it if you want that people see it by default. Every kubevirt component has a `-v` flag which defaults to `3`. The same like in kubernetes.

booxter:
@rmohr you mean you don't like the arrow? Changing the message is not the goal of this PR.

rmohr:
> @rmohr you mean you don't like the arrow? Changing the message is not the goal of this PR.

Yes it is very implicit and in combination with the variable names it is hard to decrypt. At least for me.

booxter:
OK I've modified the message a bit to hopefully be more explicit about what's going on.

rmohr:
> OK I've modified the message a bit to hopefully be more explicit about what's going on.

That is all it needed. Thanks.

PR #2198, opened 2019-04-18
Protect migration pods during node drain

### LINKED ISSUE #2196: Evictions are not always rejected on VMIs when we want to migrate on drains

>
/kind bug
> /kind enhancement

**What happened**:

When the target pod gets ready, the source pod may not yet be done with the migration, but evictions would be granted. As a result the migration does fail and the source pod gets deleted.

Both pods should be protected during a migration by the PodDisruptionBudget.

Create a VMI and stress it a little. This will create an overlap on the pods where both are seen as available by the BDB. The source pod can now be evicted.

**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

rmohr:

Update the PodDisruptionBudget to at least require two pods online. This avoids that evictions on a source pod may succeed because there is an overlap where the source- and the target pod are both seen as healthy.

Fixes #2196 

**Release note**:

```release-note
Ensure long-running migrations are not disturbed by eviction calls, if the VMI has the Migrate evictionStrategy set.
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -231,3 +231,11 @@ http_file(
         "https://dl.fedoraproject.org/pub/fedora/linux/updates/28/Everything/x86_64/Packages/q/qemu-guest-agent-2.11.2-4.fc28.x86_64.rpm",
     ],
 )
+
+http_file(
+    name = "stress",
+    sha256 = "bd93021d826c98cbec15b4bf7e0800f723f986e7ed89357c56284a7efa6394b5",
+    urls = [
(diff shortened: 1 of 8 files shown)


rmohr:
for reference: https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_kubevirt_standard-check-pr/756/artifact/ci_build_summary.html

rmohr:
ci test please

PR #2255, opened 2019-05-07
Tests: add the relevant test_ids to live migration tests

vladikr:
This simply adds relevant test_ids to live migration functional tests.

```release-note
NONE
```


tests/migration_test.go
```diff
@@ -748,7 +748,7 @@ var _ = Describe("Migrations", func() {
 				// create a new PV and PVC (PVs can't be reused)
 				tests.DeletePvAndPvc(pvName)
 			}, 60)
-			It("should reject migrations for a non-migratable vmi", func() {
+			It("[test_id:1862]should reject migrations for a non-migratable vmi", func() {
```
dshchedr: for negative tests we also need to add: [posneg:negative]
so, it should be:
It("[test_id:1862][posneg:negative]should reject migrations for a non-migratable vmi", func() {

vladikr: added

PR #2257, opened 2019-05-07
[Testing] add hostdisk test_id

vladikr:
```release-note
None
```


vladikr:
@ILpinto What ref_id should be used here?

mfranczy:
@vladikr could we merge that in the current state? I am adding to these tests one parent `Context` which will contain a ref_id

PR #2259, opened 2019-05-07
[Testing] Add more tests for ioThreadPolicies

stu-gott:
```release-note
NONE
```

### DIFF
--- a/tests/vmi_iothreads_test.go
+++ b/tests/vmi_iothreads_test.go
@@ -22,6 +22,7 @@ package tests_test
 import (
 	"encoding/xml"
 	"flag"
+	"fmt"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


tests/vmi_iothreads_test.go
```diff
@@ -75,5 +79,148 @@ var _ = Describe("IOThreads", func() {
 			Expect(len(newVMI.Spec.Domain.Devices.Disks)).To(Equal(1))
 		})
 
+		It("[test_id:864][ref_id:2065] Should honor a mix of shared and dedicated ioThreadsPolicy", func() {
+			policy := v1.IOThreadsPolicyShared
+			vmi.Spec.Domain.IOThreadsPolicy = &policy
+
+			// The disk that came with the VMI
+			dedicated := true
+			vmi.Spec.Domain.Devices.Disks[0].DedicatedIOThread = &dedicated
```
sHaggYcaT: What's a point declare same variable here and in the 137 lines in the different contexts? Maybe better to declare it globally? And set it in the test cases how it requires?

tests/vmi_iothreads_test.go
```diff
@@ -75,5 +79,148 @@ var _ = Describe("IOThreads", func() {
...
+			shr3, err := getDiskByName(domSpec, "shr2")
+			Expect(err).ToNot(HaveOccurred())
+			shr4, err := getDiskByName(domSpec, "shr2")
+			Expect(err).ToNot(HaveOccurred())
+
+			// the ioThreads for disks sh1 through shr4 will vary based on how many CPUs there are
+			// but we already verified the total number of threads, so we know they're spread out
+			// across the proper threadId pool.
+
+			By("Ensuring disk0 has a unique threadId")
```
sHaggYcaT: How do you think, maybe better to dump xml, unmarshall it, and parse it instead of use this function?

stu-gott: @sHaggYcaT It's honestly unclear what you're suggesting. lines 164 and 167 of this test are dumping and unmarshalling the domain XML:

```
By("Fetching the domain XML from the running pod")
domain, err := tests.GetRunningVirtualMachineInstanceDomainXML(virtClient, vmi)
domSpec := &api.DomainSpec{}
Expect(xml.Unmarshal([]byte(domain), domSpec)).To(Succeed())
```

So I guess the part that's unclear to me is when you suggest "and parse it". What exactly are you picturing?

tests/vmi_iothreads_test.go
stu-gott: fixed thanks!

tests/vmi_iothreads_test.go
sHaggYcaT: Yeah, you're right @stu-gott

sHaggYcaT:
lgtm

PR #2263, opened 2019-05-09
[Testing] Add test for multiple vm creation with oc/kubectl

petrkotas:
Enhances the test coverage of kubevirt by introducing the negative test rejecting multiple VM creation using `oc/kubectl create -f vm.yaml`

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/2263)

### DIFF
--- a/tests/vm_test.go
+++ b/tests/vm_test.go
@@ -23,6 +23,7 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


tests/vm_test.go
```diff
@@ -978,6 +979,49 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+
+		AfterEach(func() {
+			if vmJson != "" {
+				err = os.Remove(vmJson)
+				Expect(err).ToNot(HaveOccurred())
+				vmJson = ""
+			}
+		})
+
+		It("[rfe_id:273][crit:medium][vendor:cnv-qe@redhat.com][level:component]should create VM only once", func() {
```
dshchedr: rfe_id should be used in "Describe" section (it can combine all tests that have one rfe_id)

 In "It" we need to add test_id:243. Also in negative scenarios we need to add: [posneg:negative], so this string should be:
		It("[test_id:243][posneg:negative]should create VM only once", func() {
￼

tests/vm_test.go
```diff
@@ -978,6 +979,49 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+				Expect(err).ToNot(HaveOccurred())
+				return newVM.Name
+			}, 15*time.Second, 1*time.Second).Should(Equal(vm.Name),
+				"New VM was not created")
+
+			By("Creating the VM again")
+			_, stdErr, err := tests.RunCommand(k8sClient, "create", "-f", vmJson)
+			Expect(err).To(HaveOccurred())
+
+			Expect(strings.HasPrefix(stdErr, "Error from server (AlreadyExists): error when creating")).To(BeTrue())
```
stu-gott: If this fails the error is going to be "false != true". Please add an explanation.

tests/vm_test.go
```diff
@@ -978,6 +979,49 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+			Expect(err).ToNot(HaveOccurred(), "Cannot generate vmJson")
+
+			By("Creating VM with DataVolumeTemplate entry with k8s client binary")
+			_, _, err = tests.RunCommand(k8sClient, "create", "-f", vmJson)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Verifying VM is created")
+			Eventually(func() string {
+				newVM, err := virtClient.VirtualMachine(vm.Namespace).Get(vm.Name, &v12.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
```
stu-gott: This block appears wrong. The way this is written, the `Expect(err).ToNot(HaveOccurred())` will halt the test if it ever fails (e.g. NotFoundError). I don't think you need the `Eventually` here. We're not waiting for some event in the VM's lifecycle--we're just trying to fetch the record.

tests/vm_test.go
petrkotas: Thanks for copy paste solution.

tests/vm_test.go
petrkotas: You are right. The record is there. No need for waiting.

PR #2265, opened 2019-05-09
[Testing] Add test to fail when deleting nonexist VM

petrkotas:
Expand test coverage for deleting non-existent VM with oc/kubectl.
It will fail when the VM does not exist

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/2265)

### DIFF
--- a/tests/vm_test.go
+++ b/tests/vm_test.go
@@ -1131,6 +1131,15 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
 			waitForVMIStart(virtClient, vmi)
 		})
 
+		It("[test_id:233][posneg:negative]should fail when deleting nonexistent VM", func() {
+			vmi := tests.NewRandomVMWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskAlpine))
+
+			By("Creating VM with DataVolumeTemplate entry with k8s client binary")
+			_, stdErr, err := tests.RunCommand(k8sClient, "delete", "vm", vmi.Name)
+			Expect(err).To(HaveOccurred())
+			Expect(strings.HasPrefix(stdErr, "Error from server (NotFound): virtualmachines.kubevirt.io")).To(BeTrue(), "should fail when deleting non existent VM")
+		})
+
 	})
 })
 


tests/vm_test.go
```diff
@@ -978,6 +979,33 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+
+		AfterEach(func() {
+			if vmJson != "" {
+				err = os.Remove(vmJson)
+				Expect(err).ToNot(HaveOccurred())
+				vmJson = ""
+			}
+		})
+
+		It("[rfe_id:273][crit:medium][vendor:cnv-qe@redhat.com][level:component]should fail when deleting non existing VM", func() {
```
dshchedr: Could you please change it to:
		It("[test_id:233][posneg:negative]should fail when deleting non existing VM", func() {
￼

tests/vm_test.go
```diff
@@ -978,6 +979,33 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+			vm = tests.NewRandomVMWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskAlpine))
+
+			By("Creating VM with DataVolumeTemplate entry with k8s client binary")
+			_, stdErr, err := tests.RunCommand(k8sClient, "delete", "vm", vm.Name)
+			Expect(err).To(HaveOccurred())
+			Expect(strings.HasPrefix(stdErr, "Error from server (NotFound): virtualmachines.kubevirt.io")).To(BeTrue())
```
stu-gott: If this fails the error is going to be "false != true". Would you mind providing an explanation? e.g.

```
Expect(strings.HasPrefix(stdErr, "Error from server (NotFound): virtualmachines.kubevirt.io")).To(BeTrue(), "Expected a NotFound error")
```

fromanirh: Can we depend on this string being stable? In general setting expectations (in testing or in any code) looks fragile and always makes me wary. But I guess it's ok if k8s makes promises about the content of the errors.

stu-gott: @fromanirh I actually had the same thought when I read this code. I agree that depending on the (exact) console output of a command line tool is somewhat error prone as that might not be regarded by the authors as an "api".

I wonder if doing a case insensitive regex match such as "not\s*found" would be more robust.

petrkotas: In my opinion testing case insensitive is still fragile. Unfortunately the error code is still 1. So testing only test for pattern is doable.
Luckily, it can be easily fixed if it breaks due to error string change, if it happens.

Moreover, I have not seen the change in strings for a few version. K8s promises stable api.

fromanirh: Not happy about this state, but not your fault Petr. It really seems it's the best we can do.
I also agree that given the constraints we face, better *not* be clever and check the exact string, so breakages are easy to spot. So works for me.

stu-gott:
@petrkotas Can you rebase this PR?

ILpinto:
The test LGTM

PR #2276, opened 2019-05-13
Use go modules instead of glide

cynepco3hahue:
- replace glide with go modules
- update k8s repositories to 1.13.4 under the vendor directory

Fixes #

**Release note**:

```release-note
Use go modules instead of glide
```


cynepco3hahue:
Eh CI looks pretty well, but I have troubles with unit tests, we use at least 3 fake clients under unit tests, from repositories that generated it with old client-go code. I will try to solve it somehow.

cynepco3hahue:
Ok the single problematic repo is CDI one, I will create PR to move it to 1.13.4 too.

cynepco3hahue:
@davidvossel 
> This PR is huge. It's kind of funny, I can barely opening it in my browser.
> 
> Here are some thoughts.
> 
> * This is a big change. If possible I think it might be worthwhile to switch to go modules and then update our 1.13.4 dependency.. or do the 1.13.4 change and this switch to go modules.

That the reason why I divided it into relevant commits. And also the most part of changes relevant to update to 1.13.4, so go module change is a small one, also looks like we really have a lot of unused dependencies and this PR will reduce our repo size significantly.

> * Can we update to golang 1.12.x for our build container now as well since it has a more recent version of go modules implementation?

I prefer to wait until @mfranczy will have d/s patch for 1.12 and do not want to delay this PR.

> * golang 1.11 was the first version that supports modules. Are we locked into building using golang 1.11 or greater if we make this change? It's unclear to me how integrated all this into the build system beyond just doing dependency management.

Go modules purpose to do dependency management only and so far I saw it works much faster and much better than glide, it has some other advantages like:
- multi-version support
- caching outside of vendor directory(not really relevant for us because we use bazel, but can be useful for d/s when we will have our artifactory)
- a single tool to manage dependencies, currently k8s uses `go mod`, we are using `glide` and OpenShift `dep`

What the reason to build on go version lesser than 1.11?

cynepco3hahue:
ci test please

cynepco3hahue:
ci test please

PR #2278, opened 2019-05-13
[Testing] add test id's for hostDisk and PV initialization tests

mfranczy:
Test id's for hostDisk and PV initialization tests

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/storage_test.go
+++ b/tests/storage_test.go
@@ -363,7 +363,7 @@ var _ = Describe("Storage", func() {
 						Eventually(getStatus, 30, 1).Should(Equal(k8sv1.PodSucceeded))
 					})
 
-					It("[test_id:847]Should use existing disk image and start", func() {
+					It("[test_id:2306]Should use existing disk image and start", func() {
 						By("Starting VirtualMachineInstance")
 						vmi := tests.NewRandomVMIWithHostDisk(diskPath, v1.HostDiskExists, nodeName)
 						tests.RunVMIAndExpectLaunch(vmi, 30)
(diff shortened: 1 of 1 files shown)


tests/storage_test.go
```diff
@@ -300,218 +300,230 @@ var _ = Describe("Storage", func() {
...
+
+						By("Checking if disk.img exists")
+						vmiPod := tests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
+						output, _ := tests.ExecuteCommandOnPod(
+							virtClient,
+							vmiPod,
+							vmiPod.Spec.Containers[0].Name,
+							[]string{"find", "/var/run/kubevirt-private/vmi-disks/disk0/", "-name", "disk.img", "-size", "1G"},
+						)
+						Expect(strings.Contains(output, "disk.img")).To(BeTrue())
```
stu-gott: Failure here will result in "false != true". Could you please add an explanation as to what failed?

tests/storage_test.go
```diff
@@ -300,218 +300,230 @@ var _ = Describe("Storage", func() {
...
-						virtClient,
-						vmiPod,
-						vmiPod.Spec.Containers[0].Name,
-						[]string{"find", hostDiskDir, "-name", diskName},
-					)
-					Expect(strings.Contains(output, diskPath)).To(BeTrue())
+						Eventually(getStatus, 30, 1).Should(Equal(k8sv1.PodSucceeded))
+					})
+
+					It("[test_id:847]Should use existing disk image and start", func() {
```
stu-gott: Test ID 847 seems to lay out a different test than what's here. Please double check. I'm not saying this test is wrong, perhaps the test_id is typo-ed?

mfranczy: Good catch! I should add one more test for our kubevirtci and add one more in polarion to reflect this one.

tests/storage_test.go
mfranczy: Done

tests/storage_test.go
mfranczy: Test added, I asked @ILpinto to add one additional scenario for positive tests.

mfranczy:
all ids are in place, pr finished from my perspective

stu-gott:
ci test please

PR #2293, opened 2019-05-17
[Testing] add test 'check guest memory different from requested memory'

mfranczy:

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -347,6 +347,49 @@ var _ = Describe("Configurations", func() {
 				Expect(err).ToNot(HaveOccurred())
 
 			})
+		})
+
+		Context("[rfe_id:140][crit:medium][vendor:cnv-qe@redhat.com][level:component]with support memory over commitment", func() {
+			It("[test_id:755]should show the requested memory different than guest memory", func() {
+				vmi := tests.NewRandomVMIWithEphemeralDiskAndUserdata(tests.ContainerDiskFor(tests.ContainerDiskCirros), "#!/bin/bash\necho 'hello'\n")
+				guestMemory := resource.MustParse("256Mi")
+				vmi.Spec.Domain.Resources.Requests[kubev1.ResourceMemory] = resource.MustParse("64Mi")
+				vmi.Spec.Domain.Resources.OvercommitGuestOverhead = true
+				vmi.Spec.Domain.Memory = &v1.Memory{
+					Guest: &guestMemory,
+				}
+
+				vmi, err := virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+				Expect(err).ToNot(HaveOccurred())
+				tests.WaitForSuccessfulVMIStart(vmi)
+
+				expecter, err := tests.LoggedInCirrosExpecter(vmi)
+				Expect(err).ToNot(HaveOccurred())
+				defer expecter.Close()
+
+				res, err := expecter.ExpectBatch([]expect.Batcher{
+					&expect.BSnd{S: "free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2\n"},
+					&expect.BExp{R: "236"},
+					&expect.BSnd{S: "swapoff -a && dd if=/dev/zero of=/dev/shm/test bs=1k count=118k; echo $?\n"},
+					&expect.BExp{R: "0"},
+				}, 20*time.Second)
+				log.DefaultLogger().Object(vmi).Infof("%v", res)
+				Expect(err).ToNot(HaveOccurred())
+
+				pod := tests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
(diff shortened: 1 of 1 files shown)


mfranczy:
I am wondering if I should increase the guest and requested memories, the original scenario has 2Gi and 1Gi, however, is it not too much for our CI?

The test has a check if host memory usage is higher than 746Mi (after memory stress on VMI), I think host usage is usually higher, so it seems to be wrong.

Do you think I should increase memory values for VMI?

ILpinto:
LGTM

mfranczy:
I am not going to request such a big VMI, I resized from 1Gi to 256Mi.

ILpinto:
@mfranczy 
Sure the test can be with 256M also, NP

tests/vmi_configuration_test.go
```diff
@@ -347,6 +347,49 @@ var _ = Describe("Configurations", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				tests.WaitForSuccessfulVMIStart(vmi)
+
+				expecter, err := tests.LoggedInCirrosExpecter(vmi)
+				Expect(err).ToNot(HaveOccurred())
+				defer expecter.Close()
+
+				res, err := expecter.ExpectBatch([]expect.Batcher{
+					&expect.BSnd{S: "free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2\n"},
+					&expect.BExp{R: "236"},
```
stu-gott: Testing that exactly 236M is free makes me a bit nervous. Would you be open to something along the lines of:

```
[ "$(free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2)" -gt 200 ] && echo "pass" || echo "fail\n"
```

Where we test for something more general like "more than 200M free"?

tests/vmi_configuration_test.go
```diff
@@ -347,6 +347,49 @@ var _ = Describe("Configurations", func() {
...
+					pod,
+					"compute",
+					[]string{"/usr/bin/bash", "-c", "cat /sys/fs/cgroup/memory/memory.usage_in_bytes"},
+				)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Checking if pod memory usage is > 64Mi")
+				m, err := strconv.Atoi(strings.Trim(podMemoryUsage, "\n"))
+				Expect(err).ToNot(HaveOccurred())
+				Expect(m > 67108864).To(BeTrue())
```
stu-gott: So 64 * 1024 * 1024 == 67108864, but it's not immediately clear that's where the number comes from when reading the code. This would be more readable if that were a constant.

gszasz: 256 Mi & 64Mi (67108864) are test inputs and maybe we would like to put them all together, at least as:
```go
const (
    requested_mem = "64Mi"
    requested_mem_int = 67108864  // 64Mi = 64 * 1024^2
    guest_mem = "256Mi"
)
```

tests/vmi_configuration_test.go
gszasz: This minimal expected free memory of the guest directly depends on one of the test inputs. Either calculate it as, e.g., 78% of total guest memory, or specify it in a constant block together with the other test inputs.

```go
const (
    requested_mem = "64Mi"
    guest_mem = "256Mi"
    min_expected_free_guest_mem = 200
)
```

mfranczy:
Oh, I didn't receive any email with comments, that's why I didn't notice your suggestions, looking into them, thanks.

tests/vmi_configuration_test.go
mfranczy: @stu-gott @gszasz good points, thanks.

mfranczy:
@gszasz I prefer not to use constants for this, I think it's better to describe it by using `By("<message>")` with an appropriate message, adding constants for each single test will make code grow quite fast.

mfranczy:
@stu-gott @gszasz please take a look again.

tests/vmi_configuration_test.go
mfranczy: done, but not with constants approach. I used `By` and description field from `Expect`

gszasz:
> @gszasz I prefer not to use constants for this, I think it's better to describe it by using `By("<message>")` with an appropriate message, adding constants for every single test will make code grow quite fast.

1. Separation of the test inputs from the test procedure allows a more straightforward adaptation to the test requirement changes.
2. Replacing literals with constants or variables is a self-documenting approach, and it can prevent errors, especially when the same literal is used multiple times.
3. If one test input can be generated from another test input (e.g., "64Mi" => 64*1024*1024 = 67108864), using a constant/variable can reduce redundancy in the test inputs.
4. In the end, it is not a problem to include a constant/variable into the By("<message>").

Now we need to close the automation gap quickly, so I will not insist on that; however, it should be taken into consideration.

mfranczy:
> Now we need to close the automation gap quickly, so I will not insist on that; however, it should be taken into consideration.

It was taken into consideration, I generally agree with this what you said, but not for short and simple tests like this where you don't reuse values multiple times :P

PR #2304, opened 2019-05-21
Replace default requested memory with cluster overcommit ratio

ahadas:
This PR removes the debatable default configuration of memory-request of a VMI altogether.

First, memory-request is not set to 8Mi when not specified. When no memory is specified (either guest-memory, hugepages or memory-request), the appropriate (existing) validation error is returned.

Second, when either hugepages or guest-request is specified but memory-request is not set - the memory-request is set according to a new configurable ratio that sets the "cluster memory overcommit". For instance, if guest-memory is set to 3072M and this ration is set to 150% then memory-request would be set to 2048M.

If both guest-memory and hugepages are set, we keep returning an error.

**Release note**:

```release-note
Memory-overcommit can now be set at the cluster-level via `kubevirt-config` using the `memory-overcommit` key (in percentage). This renders default memory-request redundant.
```

### DIFF
--- a/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
+++ b/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
@@ -132,16 +132,12 @@ func (mutator *VMIsMutator) setDefaultMachineType(vmi *v1.VirtualMachineInstance
 }
 
 func (mutator *VMIsMutator) setDefaultResourceRequests(vmi *v1.VirtualMachineInstance) {
-	if _, exists := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory]; !exists {
-		if vmi.Spec.Domain.Resources.Requests == nil {
-			vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{}
-		}
-		vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = mutator.ClusterConfig.GetMemoryRequest()
-	}
-
 	if _, exists := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceCPU]; !exists {
 		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.DedicatedCPUPlacement {
 			return
+		}
+		if vmi.Spec.Domain.Resources.Requests == nil {
+			vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{}
 		}
 		vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceCPU] = mutator.ClusterConfig.GetCPURequest()
 	}
--- a/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
+++ b/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
@@ -26,6 +26,7 @@ import (
 
 	"k8s.io/api/admission/v1beta1"
 	k8sv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	v1 "kubevirt.io/kubevirt/pkg/api/v1"
@@ -132,6 +133,30 @@ func (mutator *VMIsMutator) setDefaultMachineType(vmi *v1.VirtualMachineInstance
 }
 
 func (mutator *VMIsMutator) setDefaultResourceRequests(vmi *v1.VirtualMachineInstance) {
+	if _, exists := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory]; !exists {
+		var memory *resource.Quantity
+		if vmi.Spec.Domain.Memory != nil && vmi.Spec.Domain.Memory.Guest != nil {
+			memory = vmi.Spec.Domain.Memory.Guest
+		}
+		if memory == nil && vmi.Spec.Domain.Memory != nil && vmi.Spec.Domain.Memory.Hugepages != nil {
+			if hugepagesSize, err := resource.ParseQuantity(vmi.Spec.Domain.Memory.Hugepages.PageSize); err == nil {
+				memory = &hugepagesSize
+			}
+		}
+		if memory != nil && memory.Value() > 0 {
+			if vmi.Spec.Domain.Resources.Requests == nil {
+				vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{}
+			}
+			if mutator.ClusterConfig.GetMemoryOvercommit() == 100 {
+				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = *memory
+			} else {
+				value := memory.Value() * int64(100)
+				value = value / int64(mutator.ClusterConfig.GetMemoryOvercommit())
+				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = *resource.NewQuantity(value, memory.Format)
+			}
+		}
+	}
+
 	if _, exists := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceCPU]; !exists {
 		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.DedicatedCPUPlacement {
 			return
--- a/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
+++ b/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
@@ -144,6 +144,7 @@ func (mutator *VMIsMutator) setDefaultResourceRequests(vmi *v1.VirtualMachineIns
 			}
 		}
 		if memory != nil && memory.Value() > 0 {
+			log.Log.Object(vmi).V(4).Info("Setting memory-request in light of memory-overcommit")
 			if vmi.Spec.Domain.Resources.Requests == nil {
 				vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{}
 			}
--- a/pkg/virt-config/config-map.go
+++ b/pkg/virt-config/config-map.go
@@ -49,7 +49,7 @@ const (
 	MigrationsConfigKey       = "migrations"
 	CpuModelKey               = "default-cpu-model"
 	CpuRequestKey             = "cpu-request"
-	MemoryRequestKey          = "memory-request"
+	MemoryOvercommitKey       = "memory-overcommit"
 	LessPVCSpaceTolerationKey = "pvc-tolerate-less-space-up-to-percent"
 	NodeSelectorsKey          = "node-selectors"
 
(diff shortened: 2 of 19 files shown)


ahadas:
@rmohr Hi, we talked about this (dropping/changing the default setting of requested memory) - I think this is a nice solution as it defaults the requested memory to a more reasonable value + enables to configure the default overcommit at the cluster level (as in oVirt). What do you think?

ahadas:
@rmohr ping?

pkg/virt-config/config-map.go
```diff
@@ -60,7 +60,7 @@ const (
 	MigrationCompletionTimeoutPerGiB         int64  = 800
 	DefaultMachineType                              = "q35"
 	DefaultCPURequest                               = "100m"
-	DefaultMemoryRequest                            = "8Mi"
+	DefaultMemoryOvercommit                         = 100
```
rmohr: In general I like this setting. Is this something which oVirt has too?

pkg/virt-config/config-map.go
```diff
@@ -60,7 +60,7 @@ const (
 	MigrationCompletionTimeoutPerGiB         int64  = 800
 	DefaultMachineType                              = "q35"
 	DefaultCPURequest                               = "100m"
-	DefaultMemoryRequest                            = "8Mi"
```
rmohr: We were talking about a default value of 1Gi too. Is that off the table?

pkg/virt-config/config-map.go
ahadas: Yes, there is such setting of [memory-overcommit at the cluster level](https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/common/src/main/java/org/ovirt/engine/core/common/businessentities/Cluster.java#L55) in oVirt that the default minimum allocated memory of the VM (that is used for scheduling) is derived from.

pkg/virt-config/config-map.go
ahadas: I think so, at least at the level of kubevirt-core.

When one of guest-memory or hugepages is provided, the 'cluster memory-overcommit' ratio would be applied to default memory-request so there will be no use of such default setting.

When neither guest-memory nor hugepages nor memory-request is provided, I think it makes more sense to fail the request rather than default it to a pre-define value as it really depends on what is inside the guest, assuming clients would most probably default it to a reasonable value appropriate for their users or use common-templates that should default it according to the operating system being used.

ahadas:
/retest 
Seems that the failures are unrelated to those changes, but I really want to see the CI pass :)

ahadas:
> What happens if requests and limit is set? For getting the QoS class `Guaranteed` assigned, they need to be the same on the pod, but I guess we can't do an overcommit in this case?

Yeah, so it should work ok with memory request and limit in the positive flow.
For instance, if request is set to 1024M, limit to 2048M and overcommit=150, then the VM should start with
request=1024M
guest-memory=1536M
limit=2048M

In the negative flow, when the product of a computation that rely on the overcommit ratio conflicts with the memory-limit, a validation error would be returned (i.e., "spec.domain.memory.guest: spec.domain.memory.guest '1536M' must be equal to or less than the memory limit spec.domain.resources.limits.memory '1024M'").

I think it should be ok for two reasons:
1. The memory overcommit is disabled by default (set to 100%)
2. It acts the same as when the conflict is specified explicitly (e.g., request=limit=1G and guest-memory=2G)

rmohr:
> Yeah, so it should work ok with memory request and limit in the positive flow.
> For instance, if request is set to 1024M, limit to 2048M and overcommit=150, then the VM should start with
> request=1024M
> guest-memory=1536M
> limit=2048M

Oh, you actually really request what was set. Only the guest has more. That sounds actually good. In other words: If `request` or `limit` is set, it is left completely alone?

ahadas:
> > Yeah, so it should work ok with memory request and limit in the positive flow.
> > For instance, if request is set to 1024M, limit to 2048M and overcommit=150, then the VM should start with
> > request=1024M
> > guest-memory=1536M
> > limit=2048M
> 
> Oh, you actually really request what was set. Only the guest has more. That sounds actually good. In other words: If `request` or `limit` is set, it is left completely alone?

Exactly

rmohr:
> Exactly

Very good. So one last thing I want to discuss: I tried to play this through in my head:

 1) I request 1GB of memory and I find out that my VM gets 1.5GB inside
 2) I request 1GB of memory and a limit of 1.2 GB of memory and I get an error that this is not in the right range, right?
 3) I set guest memory only. If I don't set a namespace default resource I would actually expect no memory request to be set, but I would see the calculated cluster-overcommit?

If these scenarios apply, I would have huge issues to understand that as a user (who is probably not even aware of such an over-commit config for admins). @ahadas what do you think?

ahadas:
> Very good. So one last thing I want to discuss: I tried to play this through in my head:
> 
>     1. I request 1GB of memory and I find out that my VM gets 1.5GB inside
> 
>     2. I request 1GB of memory and a limit of 1.2 GB of memory and I get an error that this is not in the right range, right?
> 
>     3. I set guest memory only. If I don't set a namespace default resource I would actually expect no memory request to be set, but I would see the calculated cluster-overcommit?
> 
> 
> If these scenarios apply, I would have huge issues to understand that as a user (who is probably not even aware of such an over-commit config for admins). @ahadas what do you think?

These scenarios apply, but let me try to convince you that it's ok nevertheless :)

Regarding the first two scenarios, I agree that when some properties are derived from other properties that are set at a different level this can be confusing. But I think that:
(1) Users would eventually be aware of the implication of setting cluster memory-overcommit as k8s users are now aware of the implication of setting namespace-limits on their pods. I think admins that have prior experience with virtualization would be less surprised by that
(2) This will not happen with the default memory overcommit but only upon changing it to, e.g.,150%

That being said, I'll add a log as we have when applying namespace limits to ease debugging.

As for the third scenario, I think it is similar in a way to the current state where the memory-request defaults to 8Mi, right? It's just that it's unlikely that setting it to 8Mi will lead to any validation/scheduling issue whereas higher value may results in such issues. But the drawback of setting it to an extremely low value like 8Mi is that it may lead to issues at runtime.

In the virtualization context I think that treating memory-request as minimum memory allocation that matches the guest-memory by default (when no overcommit is set) or computed according to the memory-overcommit ratio otherwise, would make more sense than setting it to arbitrary higher value like 1G.

rmohr:
> These scenarios apply, but let me try to convince you that it's ok nevertheless :)
> 
> Regarding the first two scenarios, I agree that when some properties are derived from other properties that are set at a different level this can be confusing. But I think that:
> (1) Users would eventually be aware of the implication of setting cluster memory-overcommit as k8s users are now aware of the implication of setting namespace-limits on their pods. I think admins that have prior experience with virtualization would be less surprised by that

That is in my opinion not the same thing. In the scenario you describe, one either gets what is asked for (a scheduled pod), or it is denied (including warning events and such, which even describe to the user that and why it can't get scheduled).

In this implementation here, I would request 1GB of RAM, and find out that I get 1.5GB, kind of. That is very very strange. It is not even the desired overcommit then, because users say 1GB, and the admin-desired overcommit would only be if I would see 1GB inside the VM but bless than 1GB actually requested from the cluster.

> (2) This will not happen with the default memory overcommit but only upon changing it to, e.g.,150%
> 

In this case I said exactly what I need for my VM and what I want guaranteed. It is strange that I would by default then get guest memory calculated which  is outside of the specified range. Also again, it would not really be the desired overcommit effect.

> That being said, I'll add logs as we have when applying namespace limits to ease debugging.
> 
> As for the third scenario, I think it is similar in a way to the current state where the memory-request defaults to 8Mi, right? It's just that it's unlikely that setting it to 8Mi will lead to any validation/scheduling issue whereas higher value may results in such issues. But the drawback of setting it to an extremely low value like 8Mi is that it may lead to issues at runtime.
> 

8Mi is a remnant which should never have been there for so long. I think we agreed on that. 

> In the virtualization context I think that treating memory-request as minimum memory allocation that matches the guest-memory by default (when no overcommit is set) or computed according to the memory-overcommit ratio otherwise, would make more sense than setting it to an arbitrary higher value like 1G.

Yes, it is perfectly fine to fail and not set it to e.g. 1G, and the validation already expected that one memory setting is at least required. Just the 8Gi remnant prevented this from happening.

rmohr:
What might work instead:

 * If only guest memory is set, then calculate a memory request based on the overcommit ratio. It is much more expected that some defaults are applied if you leave concrete requests empty. This should also ties seamless into k8s QOS classes. Normally I would even say that we should not even try to set a request if none is set, but since VMIs can't give memory dynamically back without ballooning we should help k8s here a little bit.
 * If limit or request is set, the VMI should behave like a Pod (if only limits are set, requests should be set to the same value). We can then still copy the request to the guest memory. This would be mostly to be backward compatible. Guest memory should really have been a required value.

rmohr:
@ahadas created #2381, to finally have the QOS expectations for VMIs in tests. That should make the boundaries also for this PR clearer. Would be great if you have a look.

ahadas:
@rmohr Thanks for the feedback. I can't say that I agree with everything (as I commented on #2381) but I can reduce the scope of this change to what you suggested and it would still be better (in my opinion) than the current state.

ahadas:
> What might work instead:
> 
> If only guest memory is set, then calculate a memory request based on the overcommit ratio. It is much more expected that some defaults are applied if you leave concrete requests empty. This should also ties seamless into k8s QOS classes. Normally I would even say that we should not even try to set a request if none is set, but since VMIs can't give memory dynamically back without ballooning we should help k8s here a little bit.
> If limit or request is set, the VMI should behave like a Pod (if only limits are set, requests should be set to the same value). We can then still copy the request to the guest memory. This would be mostly to be backward compatible. Guest memory should really have been a required value.

Done

ahadas:
> Done

@rmohr , Hi, any suggestion for how to push this forward?

rmohr:
@ahadas looking now again. Thanks for the ping.

rmohr:
> Third, if memory-request is set and neither guest-memory nor hugepages is set then guest-memory is set according to the "cluster overcommit".

Is this still the case (in the PR description)?

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
```diff
@@ -133,16 +134,37 @@ func (mutator *VMIsMutator) setDefaultMachineType(vmi *v1.VirtualMachineInstance
...
+		if vmi.Spec.Domain.Memory != nil && vmi.Spec.Domain.Memory.Guest != nil {
+			memory = vmi.Spec.Domain.Memory.Guest
+		}
+		if memory == nil && vmi.Spec.Domain.Memory != nil && vmi.Spec.Domain.Memory.Hugepages != nil {
+			if hugepagesSize, err := resource.ParseQuantity(vmi.Spec.Domain.Memory.Hugepages.PageSize); err == nil {
+				memory = &hugepagesSize
+			}
+		}
+		if memory != nil && memory.Value() > 0 {
+			log.Log.Object(vmi).V(4).Info("Setting memory-request in light of memory-overcommit")
```
rmohr: I guess it makes more sense to have the info message print only when it does something, and what it does.

ahadas:
> > Third, if memory-request is set and neither guest-memory nor hugepages is set then guest-memory is set according to the "cluster overcommit".
> 
> Is this still the case (in the PR description)?

No, thanks, updated the PR description

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
ahadas: Sure, will do separately

rmohr:
@ahadas can you please open a documentation PR in the user-guide?

PR #2306, opened 2019-05-21
tests: skip ignition tests if ExperimentalIgnitionSupport is not enabled

dollierp:

Skip Ignition tests (i.e those from `tests/vmi_ignition_test.go`) if ExperimentalIgnitionSupport feature gate is not enabled instead of running them and seeing them fail.

This is the same behaviour as:
- DataVolume tests (`tests/datavolume_test.go`) which are skipped when DataVolumes feature gate is not enabled 
- Migrations tests (`tests/migration_test.go`) which are skipped when LiveMigration feature gate is not enabled. 

See also #2153.

Please also cherry-pick this PR on 0.17 release series.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -3401,6 +3401,10 @@ func HasCDI() bool {
 	return HasFeature("DataVolumes")
 }
 
+func HasExperimentalIgnitionSupport() bool {
+	return HasFeature("ExperimentalIgnitionSupport")
+}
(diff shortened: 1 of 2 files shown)


fabiand:
+1

/cc @stu-gott

rmohr:
ci test please

PR #2325, opened 2019-05-28
Create the ovs br1 on worker nodes for ocp 3.11

SchSeba:
This PR fixes an issue on the ocp provider.

When we request more the one node for the cluster the additional workers
are reinstalled using the openshift-ansible scale-up playbook.

This playbook removes all the ovs configuration done previously,
so this PR just run on every node using the ovs daemonset installed in the
cluster and recreate the br1 bridge if not exist.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/2325)

### DIFF
--- a/cluster/os-3.11.0-multus/provider.sh
+++ b/cluster/os-3.11.0-multus/provider.sh
@@ -4,4 +4,4 @@ set -e
 
 source cluster/os-3.11.0/provider.sh
 
-image="os-3.11.0-multus@sha256:278b5eb240fcf968add7bb2ae32b629ac056b56b8dd1d2d33446edb44f7e27b3"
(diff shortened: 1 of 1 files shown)


cluster/os-3.11.0/provider.sh
```diff
@@ -23,4 +23,8 @@ function up() {
 
     # Make sure that local config is correct
     prepare_config
+
+    if [[ $provider_prefix == "os-3.11.0-multus" ]]; then
+        ./cluster/kubectl.sh get po -n openshift-sdn | grep ovs | awk '{print "./cluster/kubectl.sh exec -n openshift-sdn",$1,"-- ovs-vsctl --may-exist add-br br1"}' | sh
```
rmohr: This looks like the run script in kubevirtci should do that for every node, can that be?

SchSeba: I will try to move it into the node01.sh in kubevirtci

PR #2330, opened 2019-05-29
WIP use linux bridge instead of ovs

phoracek:


phoracek:
/test pull-kubevirt-e2e-k8s-multus-1.13.3

Failed on networkaddonsconfig wait timeout. We should add a code that shows the state of networkaddonsconfig in that case (and probably increase the timeout).

https://deck-kubevirt-prow.apps.ovirt.org/view/gcs/kubevirt/pr-logs/pull/kubevirt_kubevirt/2330/pull-kubevirt-e2e-k8s-multus-1.13.3/1133777434357796864

PR #2331, opened 2019-05-29
Forward all network traffic in masquerade

SchSeba:
This PR will forward all the incoming traffic to the vm if not specific port was requested

```release-note
forward all ports using masquerade
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/2331)

### DIFF
--- a/pkg/virt-launcher/virtwrap/network/podinterface.go
+++ b/pkg/virt-launcher/virtwrap/network/podinterface.go
@@ -542,6 +542,15 @@ func (p *MasqueradePodInterface) createNatRules() error {
 		return err
 	}
 
+	if len(p.iface.Ports) == 0 {
+		err = Handler.IptablesAppendRule("nat", "KUBEVIRT_PREINBOUND",
+			"-j",
+			"DNAT",
+			"--to-destination", p.vif.IP.IP.String())
+
+		return err
+	}
+
 	for _, port := range p.iface.Ports {
 		if port.Protocol == "" {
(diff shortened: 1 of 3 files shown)


pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -542,6 +542,15 @@ func (p *MasqueradePodInterface) createNatRules() error {
 		return err
 	}
 
+	if len(p.iface.Ports) == 0 {
```
rmohr: Why can't we just always do that? Pods also always forward every traffic, no matter if pods are specified or not.

SchSeba: we can't do it because it will break istio intregration we can't forward everything into the vm if istio is running and the istio-proxy container is injected into the virt-launcher pod.

This is why I am going to add a PR into kubeivrt user-guide to explain that if the user use istio it **must** configure the ports

rmohr: Ok, can't say  that I like behaving different  by default to fit into the case where istio might be there or not. A unit test is missing. Apart from that good.

SchSeba: Thanks I will work on the unit test for the case

SchSeba:
@rmohr added some unit tests as required can you please make another review on this PR?

PR #2342, opened 2019-06-04
Fixed Openshift check for tests

slintes:
Fixed Openshift check for tests, the /version/openshift endpoint doesn't exist anymore

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1714991

**Release note**:
```release-note
NONE
```


slintes:
logs of running tests look good, the only test using this is skipped on os-3.11, but not on k8s-1.11, will merge

PR #2347, opened 2019-06-05
tests: make sriov suite work with mellanox cards

booxter:
Mellanox VFs require additional kernel module loaded inside the guest for
netlink device to pop up. This patch makes the fedora vmi load the kernel
module. Qemu agent is used to report back to the suite about cloud-init phase
completion.

This patch also deduplicates code of the sriov test cases as well as
introduces a single test function to wait for AgentConnected status that
is then reused in all other test cases in the tree that need this wait.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/vmi_multus_test.go
+++ b/tests/vmi_multus_test.go
@@ -664,17 +664,8 @@ var _ = Describe("Multus", func() {
 				getOptions := &metav1.GetOptions{}
 				var updatedVmi *v1.VirtualMachineInstance
 
-				// Need to wait for cloud init to finnish and start the agent inside the vmi.
-				Eventually(func() bool {
-					updatedVmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(agentVMI.Name, getOptions)
-					Expect(err).ToNot(HaveOccurred())
-					for _, condition := range updatedVmi.Status.Conditions {
-						if condition.Type == "AgentConnected" && condition.Status == "True" {
-							return true
-						}
-					}
-					return false
-				}, 420*time.Second, 2).Should(BeTrue(), "Should have agent connected condition")
+				// Need to wait for cloud init to finish and start the agent inside the vmi.
+				tests.WaitAgentConnected(virtClient, agentVMI)
 
 				Eventually(func() bool {
 					updatedVmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(agentVMI.Name, getOptions)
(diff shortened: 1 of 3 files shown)


booxter:
@phoracek @SchSeba could you please review this test-only fix? thanks.

booxter:
Some jobs fail with a migration test failure so I was suspicious about those since I touch migration test code in this PR. But I've checked other independent PRs and they have the same failure, so it's unrelated to this PR. Otherwise the runs seem rather clean.

tests/vmi_multus_test.go
```diff
@@ -662,27 +658,16 @@ var _ = Describe("Multus", func() {
...
-					return false
-				}, 420*time.Second, 2).Should(BeTrue(), "Should have agent connected condition")
+				// Need to wait for cloud init to finish and start the agent inside the vmi.
+				tests.WaitAgentConnected(virtClient, agentVMI)
 
+				getOptions := &metav1.GetOptions{}
 				Eventually(func() bool {
-					updatedVmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(agentVMI.Name, getOptions)
+					updatedVmi, _ := virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(agentVMI.Name, getOptions)
 					return len(updatedVmi.Status.Interfaces) == 4
```
phoracek: I know this is not a regression, but it looks like this would cause null pointer exception in case VirtualMachineInstance failed. Maybe we can just `continue` in case err != nil?

booxter:
@phoracek I'll see what I can do. But the fix is adding cloud-init that installs / loads kernel module + wait for AgentConnected (because we need cloud-init to complete before checking guest interfaces).

phoracek:
@booxter if you will feel that it is a waste of time and i'm just whining (I feel like that a little bit), just say a word

booxter:
@phoracek tell me if this split is more clear

booxter:
@phoracek ouch. I can't do anything right the first attempt can I. Now should be better.

tests/vmi_multus_test.go
```diff
@@ -662,27 +658,19 @@ var _ = Describe("Multus", func() {
...
 				tests.WaitForSuccessfulVMIStart(agentVMI)
 
-				getOptions := &metav1.GetOptions{}
-				var updatedVmi *v1.VirtualMachineInstance
+				// Need to wait for cloud init to finish and start the agent inside the vmi.
+				tests.WaitAgentConnected(virtClient, agentVMI)
 
-				// Need to wait for cloud init to finnish and start the agent inside the vmi.
+				getOptions := &metav1.GetOptions{}
 				Eventually(func() bool {
```
SchSeba: just a small nit, if we add the `tests.WaitAgentConnected` should we remove the eventually and just get the object?

tests/vmi_multus_test.go
SchSeba: +1

vladikr: This has been addressed on line 667, right?

tests/vmi_multus_test.go
vladikr: I thought that it may time for the agent to report the interface, even if connected. Am I wrong?

SchSeba: Oh you are right!

tests/vmi_multus_test.go
SchSeba: I don't think. in the 667 we just get the vmi object but we need to check if the interfaces array inside status is not nil.

vladikr: I'm not sure how can this happen as at this stage the vm should be running (we've passed tests.WaitForSuccessfulVMIStart(agentVMI)) so there should be at least one interface... but I guess it doesn't hurt checking... @booxter what do you think?

booxter: I should have pushed against fixing the original issue with not checking the error at all; if I would do it, I wouldn't get more requests to fix unrelated code. :)

booxter: @vladikr @SchSeba I have no plan to update the patch unless someone explicitly says they will block it without me fixing this unrelated issue. Does anyone of you plan to block this PR for the issue being discussed?

booxter:
@vladikr @davidvossel @rmohr can we nudge this in and merge? This slows down progress by @fedepaol on SR-IOV CI somewhat. There are three approvals here; the only comment that could be interpreted as not having some closure is a suggestion to fix a hypothetical bug in a test case that is not even SR-IOV related and touched only because I tried to deduplicate the code in the tree. I would hope we don't block this PR just for that. But if so, I'd love to hear that's the case.

Thanks in advance.

PR #2348, opened 2019-06-05
Limit bazel to a maximum of 4 parallel jobs

rmohr:

Bazel tries to come up with a reasonable number of concurrent jobs, but it's capabilities to detect what's available are limited inside docker containers. Limit it for CI to a reasonable number. Otherwise on concurrent builds we can easily see average CPU utilization values above 150 with 12 parallel builds.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -160,6 +160,7 @@ cat >.bazelrc <<EOF
 startup --host_jvm_args=-Dbazel.DigestFunction=sha256
 build --remote_local_fallback
 build --remote_http_cache=${BAZEL_CACHE}
+build --jobs=4
 EOF
 
 # build all images with the basic repeat logic


akalenyu:
@rmohr Hi!
Bumped into this while investigating some failures in downstream pipelines.
Should we revisit now that https://github.com/bazelbuild/bazel/pull/20644 is in?

rmohr:
I guess in principle yes, but kubevirt is still on an old bazel version. Seems like at least 7.1 is needed: https://github.com/kubevirt/kubevirt/blob/main/.bazelversion

akalenyu:
> What's the exact failure? Too slow build?

Looked like an OOM (container running bazel was limited at 6Gi):
`Server terminated abruptly (error code: 14, error message: 'Socket closed'`

PR #2354, opened 2019-06-06
Add utility flags for downstream repo selection

petrkotas:
Added flags enable to select different tags for upstream and downstream,
since they are not the same. Even when the code is the same. 
Moreover testing containers are only provided from upstream repository,
and are not mirrored. Therefore different docker prefix have to be used.

Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=1716432

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/2354)

### DIFF
--- a/tests/subresource_api_test.go
+++ b/tests/subresource_api_test.go
@@ -247,7 +247,7 @@ func testClientJob(virtCli kubecli.KubevirtClient, withServiceAccount bool, reso
 			Containers: []k8sv1.Container{
 				{
 					Name:    name,
-					Image:   fmt.Sprintf("%s/subresource-access-test:%s", tests.KubeVirtRepoPrefix, tests.KubeVirtVersionTag),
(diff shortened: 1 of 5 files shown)


dshchedr:
> @dshchedr Can you confirm whether or not this fixes the issue you're having?

probably, this question should be addressed to Denis @dollierp ?

slintes:
mh, 2 lanes failed for these 2 tests, coincidence? I don't see why they could fail because of this PR though...

```
CloudInitHookSidecars VMI definition with CloudInit hook sidecar should call Collect and PreCloudInitIso on the hook sidecar expand_more	3m7s
CloudInitHookSidecars VMI definition with CloudInit hook sidecar should have cloud-init user-data from sidecar expand_more
```

/retest

PR #2355, opened 2019-06-06
Removed unused SCC modification for tests

slintes:
Removed unused SCC modification for tests

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1714991

It changes SCC for the test user in kube-system ns, but the user is in
another ns... so this file can't have any effect at all...?

**Release note**:
```release-note
NONE
```


manifests/testing/ocp/readme.txt
```diff
@@ -0,0 +1 @@
+place manifests for ocp here, ending in *for-ocp-yaml.in
```
phoracek: it is already in ocp/ directory, why bother with the suffix?

cluster/deploy.sh
```diff
@@ -50,7 +50,10 @@ EOF
 _kubectl apply -f ${MANIFESTS_OUT_DIR}/release/kubevirt-operator.yaml
 
 if [[ "$KUBEVIRT_PROVIDER" =~ os-* ]] || [[ "$KUBEVIRT_PROVIDER" =~ okd-* ]]; then
-    _kubectl create -f ${MANIFESTS_OUT_DIR}/testing/ocp
+
+    if [ -d "${MANIFESTS_OUT_DIR}/testing/ocp" ]; then
+        _kubectl create -f ${MANIFESTS_OUT_DIR}/testing/ocp
```
phoracek: There is also os/okd section in hack/cluster-clean.sh removing SCC, any change it is related to the change?

manifests/testing/ocp/readme.txt
slintes: it is used in testing code, which iterates all manifests in testing and subdirectories

cluster/deploy.sh
slintes: mh, good catch, but it at least doesn't hurt to leave it there, in case we add SCC later on again.

slintes:
rebased

slintes:
@rmohr what do you think of this? Do you remember why the test scc was added? I does not have any effect (anymore), os lanes are green.

phoracek:
And it was failing OKD

rmohr:
@slintes the purpose is that one can use a `kubevirt-testing` service acount to run all tests from a pod service account. Not sure if QE uses it right now. But I don't think so because it would fail on OKD for them too. In theory we could drop this SCC and just add the `kubevirt-testing` service account to the `privileged` scc. But we can also drop it right now.

rmohr:
Oh, and the disk image provider uses the service account, but only the creator of the pod needs privileges to create it. Not the used service account. So we should be good.

hack/cluster-deploy.sh
```diff
@@ -52,7 +52,10 @@ EOF
 _kubectl apply -f ${MANIFESTS_OUT_DIR}/release/kubevirt-operator.yaml
 
 if [[ "$KUBEVIRT_PROVIDER" =~ os-* ]] || [[ "$KUBEVIRT_PROVIDER" =~ okd-* ]]; then
-    _kubectl create -f ${MANIFESTS_OUT_DIR}/testing/ocp
+
+    if [ -d "${MANIFESTS_OUT_DIR}/testing/ocp" ]; then
```
rmohr: Was this the only file in ocp? If yes, then we could just remove the whole if statement.

slintes: @rmohr yes it was, I wanted to keep this in case we need it later on. But I'm fine with removing it. Should I remove all logic for ocp specific manifests then? There is also something in tests/utils.go, which checks for manifests ending in for-ocp.yaml. Wdyt?

slintes: cleaned up everything. Also removed another statement about adding the operator to the prvileged SCC, which is not needed because there are RBAC rules for it.

PR #2361, opened 2019-06-09
use Linux bridge CNI in vmi_multus_test.go

phoracek:
The preferred CNI providing L2 connectivity has been changed from
OVS to Linux bridge. Because of that, and since we don't configure
OVS bridge on OKD local setup, this patch switches to Linux bridge
CNI.

**Release note**:

```release-note
NONE
```


phoracek:
pull-kubevirt-e2e-k8s-multus-1.13.3

cynepco3hahue:
I can see this 3 failures
```
Multus [rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:component]VirtualMachineInstance using different types of interfaces. VirtualMachineInstance with Linux bridge CNI plugin interface [test_id:1577]should create two virtual machines with one interface expand_more	1m30s
Multus [rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:component]VirtualMachineInstance using different types of interfaces. VirtualMachineInstance with Linux bridge CNI plugin interface [test_id:1578]should create two virtual machines with two interfaces expand_more	1m30s
Multus [rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:component]VirtualMachineInstance using different types of interfaces. VirtualMachineInstance with Linux bridge CNI plugin interface and custom MAC address. [test_id:676]should configure valid custom MAC address on Linux bridge CNI interface. expand_more	1m51s
```

phoracek:
/hold

@cynepco3hahue yeah, persistent retests cured timeouts on all the other suites, I guess this one really needs a fix.

petrkotas:
@phoracek how new is the change? I have the feeling our multus tests are still using the previous ovs.

phoracek:
@petrkotas they are using OVS, that's why I posted this PR. The local kubernetes and openshift clusters contain pre-installed OVS CNI pluggins as well as pre-created OVS bridge br1. Recently we switched to using cluster-network-addons-operator in order to install network components on kubevirtci images. While on kubernetes and openshift cluster we still have original OVS deployment in place, with okd that's not the case. We want to test whatever CNI L2 plugin is given by network-addons-operator, hence this change.

phoracek:
As @SchSeba found out, we have to configure iptables on the node. We can do it either as a part of kubevirtci, or using https://github.com/containernetworking/plugins/tree/master/plugins/meta/firewall (after patching it for L2).

```
https://github.com/containernetworking/plugins/tree/master/plugins/meta/firewall
```

CREATE_BRIDGE.yml
```diff
@@ -0,0 +1,45 @@
+apiVersion: apps/v1
```
cynepco3hahue: I thought it should be the network operator the one that creates it?

phoracek: Don't look! It is just testing that this approach would work. It will become a part of kubevirtci.

We don't have node network configuration implemented (yet). Once, it's done. This workaround will be dropped from kubevirtci containers and replaced with proper components, as we did with CNI and multus before.

phoracek:
@cynepco3hahue could you please take another look at this approach? Thanks.

phoracek:
/assign @vladikr

All affected suites have passed, could you please take a look?

tests/vmi_multus_test.go
```diff
@@ -766,5 +769,109 @@ func pingVirtualMachine(vmi *v1.VirtualMachineInstance, ipAddr, prompt string) {
...
+					Containers: []k8sv1.Container{
+						{
+							Name: "network-config",
+							// Reuse image which is already installed in the cluster. All we need is chroot.
+							// Local OKD cluster doesn't allow use to pull from the outside.
+							Image: "registry:5000/kubevirt/virt-operator:devel",
+							Command: []string{
+								"sh",
+								"-c",
+								"set -x; chroot /host ip link add br10 type bridge; chroot /host iptables -I FORWARD 1 -i br10 -j ACCEPT; touch /tmp/ready; sleep INF",
```
qinqon: We have to use && instead of ; if not command is going to be always creating ready even with failures

tests/vmi_multus_test.go
```diff
@@ -766,5 +769,109 @@ func pingVirtualMachine(vmi *v1.VirtualMachineInstance, ipAddr, prompt string) {
...
+			Template: k8sv1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{"name": "network-config"},
+				},
+				Spec: k8sv1.PodSpec{
+					Containers: []k8sv1.Container{
+						{
+							Name: "network-config",
+							// Reuse image which is already installed in the cluster. All we need is chroot.
+							// Local OKD cluster doesn't allow use to pull from the outside.
```
fedepaol: I _think_ there is a typo in "allow use"

tests/vmi_multus_test.go
```diff
@@ -766,5 +769,109 @@ func pingVirtualMachine(vmi *v1.VirtualMachineInstance, ipAddr, prompt string) {
...
+						},
+					},
+					HostNetwork: true,
+				},
+			},
+		},
+	}
+
+	// Helper function returning existing network-config DaemonSet if exists
+	getNetworkConfigDaemonSet := func() *appsv1.DaemonSet {
```
fedepaol: Does it really need to be a closure? Can't it be an helper function taking the daemonset name?

tests/vmi_multus_test.go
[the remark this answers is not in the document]
phoracek: What MTU issue? I'm not sure, I just wasn't able to pull any new images from the internet (and why should we if we can reuse what we have)

tests/vmi_multus_test.go
```diff
@@ -766,5 +769,109 @@ func pingVirtualMachine(vmi *v1.VirtualMachineInstance, ipAddr, prompt string) {
...
+							// Reuse image which is already installed in the cluster. All we need is chroot.
+							// Local OKD cluster doesn't allow use to pull from the outside.
+							Image: "registry:5000/kubevirt/virt-operator:devel",
+							Command: []string{
+								"sh",
+								"-c",
+								"set -x; chroot /host ip link add br10 type bridge; chroot /host iptables -I FORWARD 1 -i br10 -j ACCEPT; touch /tmp/ready; sleep INF",
+							},
+							SecurityContext: &k8sv1.SecurityContext{
+								Privileged: newBool(true),
```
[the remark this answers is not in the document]
phoracek: Looks like test/utils.go already has these functions, lemme reuse those

tests/vmi_multus_test.go
phoracek: that's intended, if the bridge already exists (it fails), we just continue.

tests/vmi_multus_test.go
phoracek: Done

tests/vmi_multus_test.go
phoracek: We are using closures a lot in tests, so I (ab)used that to keep the whole logic in a single function. It can be done outside of it, the question is whether one of those options is better. I personally prefer this keeping this (kind of ugly hack) in a tiny package, waiting to be replaced with nmstate. But if you think keeping it outside the function would be better, I'm ok with changing it.

fedepaol: Yes makes sense, also because it is going to be replaced. Let's keep it like this

tests/utils.go
```diff
@@ -2231,6 +2231,10 @@ func NewInt32(x int32) *int32 {
 	return &x
 }
 
+func NewInt64(x int64) *int64 {
```
cynepco3hahue: k8s has a nice library for all these pointers, `k8s.io/utils/pointer`

tests/vmi_multus_test.go
```diff
@@ -457,14 +460,14 @@ var _ = Describe("Multus", func() {
 					Expect(is_present).To(BeTrue())
 					Expect(ifc.MAC).To(Not(BeZero()))
 				}
-				Expect(interfacesByName["default"].MAC).To(Not(Equal(interfacesByName["ovs"].MAC)))
+				Expect(interfacesByName["default"].MAC).To(Not(Equal(interfacesByName["linux-bridge"].MAC)))
```
cynepco3hahue: can we create a constant for `linux-bridge`, something like `defaultInterfaceName`

tests/vmi_multus_test.go
```diff
@@ -766,5 +769,101 @@ func pingVirtualMachine(vmi *v1.VirtualMachineInstance, ipAddr, prompt string) {
...
+func configureNodeNetwork(virtClient kubecli.KubevirtClient) {
+	// Privileged DaemonSet configuring host networking as needed
+	networkConfigDaemonSet := appsv1.DaemonSet{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "apps/v1",
+			Kind:       "DaemonSet",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "network-config",
+			Namespace: "kube-system",
```
cynepco3hahue: metav1 has constant for system namespace, please replace all `kube-system` with it

phoracek:
@qinqon @booxter and @fedepaol are from the network team @cynepco3hahue

tests/vmi_multus_test.go
phoracek: defaultInterfaceName is "default" though (on the left).

i don't want to do any refactoring in this PR, i want to keep it to the point (switch from OVS to linux bridge).

tests/utils.go
phoracek: we should replace NewBool and NewInt32 in this module then, ack

tests/vmi_multus_test.go
phoracek: Done

cynepco3hahue:
@phoracek Please run `make generate` and re-commit

phoracek:
@cynepco3hahue done

PR #2364, opened 2019-06-10
Port name validation

SchSeba:
This PR adds a port name validation for the kubevirt validating webhook.

Before this PR a port name with upper case was allowed but the pod creation was failing.

Related log:
```
Warning  FailedCreate  1s (x5 over 39s)  virtualmachine-controller  (combined from similar events): Error creating pod: Pod "virt-launcher-testvmilgdmt5p8tf7x4j7mt8wzpt2v924z8j4qt74mlc8dcp" is invalid: [spec.containers[1].ports[1].name: Invalid value: "testPort-tcp": must contain only alpha-numeric characters (a-z, 0-9), and hyphens (-), spec.containers[1].ports[3].name: Invalid value: "testPort-udp": must contain only alpha-numeric characters (a-z, 0-9), and hyphens (-)]
```

**Release note**:

```release-note
Fix port name validation
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/2364)

### DIFF
--- a/pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter_test.go
+++ b/pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter_test.go
@@ -1344,6 +1344,25 @@ var _ = Describe("Validating VMICreate Admitter", func() {
 			Expect(len(causes)).To(Equal(1))
 			Expect(causes[0].Field).To(Equal("fake.domain.devices.interfaces[0].name"))
 		})
+		It("should reject a bad port name", func() {
+			vm := v1.NewMinimalVMI("testvm")
+			vm.Spec.Domain.Devices.Interfaces = []v1.Interface{{
+				Name: "default",
+				InterfaceBindingMethod: v1.InterfaceBindingMethod{
+					Masquerade: &v1.InterfaceMasquerade{},
+				},
+				Ports: []v1.Port{{Name: "Test", Port: 80}}}}
+
+			vm.Spec.Networks = []v1.Network{{
+				Name:          "default",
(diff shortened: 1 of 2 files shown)


pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter_test.go
```diff
@@ -1344,6 +1344,25 @@ var _ = Describe("Validating VMICreate Admitter", func() {
...
+				Ports: []v1.Port{{Name: "Test", Port: 80}}}}
+
+			vm.Spec.Networks = []v1.Network{{
+				Name:          "default",
+				NetworkSource: v1.NetworkSource{Pod: &v1.PodNetwork{}},
+			},
+			}
+
+			causes := ValidateVirtualMachineInstanceSpec(k8sfield.NewPath("fake"), &vm.Spec, config)
+			Expect(len(causes)).To(Equal(1))
```
petrkotas: I believe, that in the case of a failure, this exception will say: Expected N to equal 1. In my opinion this does not tell much about the nature of the exception. Would you please add a brief description to it? Thanks you.

SchSeba: done can you please take a look if it is good now?

petrkotas:
@SchSeba thanks for the quick update. The PR is good now. Please also run make generate and push the update. Travis is complaining.

SchSeba:
@petrkotas ops thanks!

rmohr:
Test failures are not related to the change.

PR #2375, opened 2019-06-12
Add regression test for wrong pci address in VM

petrkotas:
This tests introduces regression test, where VM creation is blocked due to wrong PCI address.

Explanation:
For PCI address "0000:YY:XX.0" where XX > 0 and YY > 0.
if XX is (0,1,2) - kubevirt will block it (Bugzilla 1645069 ).
Ff XX > 0 then libvirt will block it ( as seen in the error msg in Bug 1645069)
and then allow only one slot , only 0 when YY is > 0.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/2375)

### DIFF
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -1310,6 +1310,35 @@ var _ = Describe("Configurations", func() {
 
 			checkPciAddress(vmi, vmi.Spec.Domain.Devices.Disks[0].Disk.PciAddress, "\\$")
 		})
+
+		It("[test_id:1020]should not create the VM with wrong PCI adress", func() {
+			By("setting disk1 Pci address")
+
+			wrongPciAddress := "0000:04:10.0"
+
+			vmi.Spec.Domain.Devices.Disks[0].Disk.PciAddress = wrongPciAddress
+			vmi.Spec.Domain.Devices.Disks[0].Disk.Bus = "virtio"
+			vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+			Expect(err).ToNot(HaveOccurred())
+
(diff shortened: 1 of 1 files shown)


tests/vmi_configuration_test.go
```diff
@@ -1310,6 +1310,32 @@ var _ = Describe("Configurations", func() {
...
+					for _, cond := range vmi.Status.Conditions {
+						if cond.Type == v1.VirtualMachineInstanceConditionType(v1.VirtualMachineInstanceSynchronized) && cond.Status == kubev1.ConditionFalse {
+							vmiCondition = cond
+							return true
+						}
+					}
+				}
+				return false
+			}, 60*time.Second, time.Second).Should(BeTrue())
+			Expect(vmiCondition.Message).To(ContainSubstring("Invalid PCI address 0000:04:10.0"))
```
gszasz:
```suggestion
Expect(vmiCondition.Message).To(ContainSubstring(fmt.Sprintf("Invalid PCI address %s", wrongPciAddress))
```

petrkotas:
Hi @gszasz thanks for pointing this out. I have fixed it.

tests/vmi_configuration_test.go
```diff
@@ -1310,6 +1310,35 @@ var _ = Describe("Configurations", func() {
 
 			checkPciAddress(vmi, vmi.Spec.Domain.Devices.Disks[0].Disk.PciAddress, "\\$")
 		})
+
+		FIt("[test_id:1020]should not create the VM with wrong PCI adress", func() {
```
mfranczy: Probably you want to replace `Fit` with `It` :)

PR #2381, opened 2019-06-13
Ensure that QOS  class guaranteed is indeed supported like documented

rmohr:

 * Ensure that QOS class *guaranteed* can be achieved by setting limits and requests to identical values, also when no dedicated cpus are asked for.
 * Ensure that WOS class *guaranteed* can be achieved by only setting limits. Limits are copied to requests in this case, like it is done for pods.
 * Add tests to ensure that this behaviour will not break.
 * Add a `status.qosClass` field like it exists on pods, to make it easier to discover the qos class.

Fixes #

**Release note**:

```release-note
Fix QOS guaranteed support if VMIs without dedicated CPUs want this QOS class.
```

### DIFF
--- a/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
+++ b/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
@@ -147,7 +147,24 @@ func (mutator *VMIsMutator) setDefaultPullPoliciesOnContainerDisks(vmi *v1.Virtu
 }
 
 func (mutator *VMIsMutator) setDefaultResourceRequests(vmi *v1.VirtualMachineInstance) {
-	if _, exists := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory]; !exists {
+
+	resources := &vmi.Spec.Domain.Resources
+
+	if !resources.Limits.Cpu().IsZero() && resources.Requests.Cpu().IsZero() {
+		if resources.Requests == nil {
+			resources.Requests = k8sv1.ResourceList{}
+		}
+		resources.Requests[k8sv1.ResourceCPU] = resources.Limits[k8sv1.ResourceCPU]
+	}
+
+	if !resources.Limits.Memory().IsZero() && resources.Requests.Memory().IsZero() {
+		if resources.Requests == nil {
+			resources.Requests = k8sv1.ResourceList{}
+		}
+		resources.Requests[k8sv1.ResourceMemory] = resources.Limits[k8sv1.ResourceMemory]
+	}
+
+	if _, exists := resources.Requests[k8sv1.ResourceMemory]; !exists {
 		var memory *resource.Quantity
 		if vmi.Spec.Domain.Memory != nil && vmi.Spec.Domain.Memory.Guest != nil {
 			memory = vmi.Spec.Domain.Memory.Guest
@@ -158,28 +175,28 @@ func (mutator *VMIsMutator) setDefaultResourceRequests(vmi *v1.VirtualMachineIns
 			}
 		}
 		if memory != nil && memory.Value() > 0 {
-			if vmi.Spec.Domain.Resources.Requests == nil {
-				vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{}
+			if resources.Requests == nil {
+				resources.Requests = k8sv1.ResourceList{}
 			}
 			overcommit := mutator.ClusterConfig.GetMemoryOvercommit()
 			if overcommit == 100 {
-				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = *memory
+				resources.Requests[k8sv1.ResourceMemory] = *memory
 			} else {
 				value := (memory.Value() * int64(100)) / int64(overcommit)
-				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = *resource.NewQuantity(value, memory.Format)
+				resources.Requests[k8sv1.ResourceMemory] = *resource.NewQuantity(value, memory.Format)
 			}
-			memoryRequest := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory]
+			memoryRequest := resources.Requests[k8sv1.ResourceMemory]
 			log.Log.Object(vmi).V(4).Infof("Set memory-request to %s as a result of memory-overcommit = %v%%", memoryRequest.String(), overcommit)
 		}
 	}
 
-	if _, exists := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceCPU]; !exists {
+	if _, exists := resources.Requests[k8sv1.ResourceCPU]; !exists {
 		if vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.DedicatedCPUPlacement {
 			return
 		}
-		if vmi.Spec.Domain.Resources.Requests == nil {
-			vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{}
+		if resources.Requests == nil {
+			resources.Requests = k8sv1.ResourceList{}
 		}
-		vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceCPU] = mutator.ClusterConfig.GetCPURequest()
+		resources.Requests[k8sv1.ResourceCPU] = mutator.ClusterConfig.GetCPURequest()
 	}
 }
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -41,6 +41,7 @@ import (
 	v1 "kubevirt.io/client-go/api/v1"
 	"kubevirt.io/client-go/kubecli"
 	"kubevirt.io/client-go/log"
+	kubevirt_hooks_v1alpha2 "kubevirt.io/kubevirt/pkg/hooks/v1alpha2"
 	hw_utils "kubevirt.io/kubevirt/pkg/util/hardware"
 	"kubevirt.io/kubevirt/tests"
 )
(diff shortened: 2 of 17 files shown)


rmohr:
@cynepco3hahue I see that my pull requests seems to remove glog from the vendor folder. How could that be included? Is that correct what I see?

tests/vmi_configuration_test.go
```diff
@@ -468,7 +548,7 @@ var _ = Describe("Configurations", func() {
 			})
 		})
 
-		Context("[rfe_id:140][crit:medium][vendor:cnv-qe@redhat.com][level:component]with namespace memory limits lower than VMI required memory", func() {
+		Context("[rfe_id:140][crit:medium][vendor:cnv-qe@redhat.com][level:component]with namespace memory imits lower than VMI required memory", func() {
```
ahadas: unintended change?

pkg/api/v1/types.go
```diff
@@ -271,6 +276,17 @@ func (v *VirtualMachineInstance) IsCPUDedicated() bool {
...
 
+// WantsToHaveQOSGuaranteed checks if cpu and memoyr limits and requests are identical on the VMI.
+// This is the indicator that people want a VMI with QOS of guaranteed
+func (v *VirtualMachineInstance) WantsToHaveQOSGuaranteed() bool {
+	resources := v.Spec.Domain.Resources
+	if !resources.Requests.Memory().IsZero() && resources.Requests.Memory().Cmp(*resources.Limits.Memory()) == 0 &&
+		!resources.Requests.Cpu().IsZero() && resources.Requests.Cpu().Cmp(*resources.Limits.Cpu()) == 0 {
+		return true
+	}
+	return false
```
ahadas: can it be replaced with
return !resources.Requests.Memory().IsZero() && resources.Requests.Memory().Cmp(*resources.Limits.Memory()) == 0 && !resources.Requests.Cpu().IsZero() && resources.Requests.Cpu().Cmp(*resources.Limits.Cpu()) == 0 
?

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
```diff
@@ -132,17 +132,33 @@ func (mutator *VMIsMutator) setDefaultMachineType(vmi *v1.VirtualMachineInstance
...
-	if _, exists := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory]; !exists {
-		if vmi.Spec.Domain.Resources.Requests == nil {
-			vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{}
+	resources := &vmi.Spec.Domain.Resources
+
+	if !resources.Limits.Cpu().IsZero() && resources.Requests.Cpu().IsZero() {
+		if resources.Requests == nil {
+			resources.Requests = k8sv1.ResourceList{}
+		}
+		resources.Requests[k8sv1.ResourceCPU] = resources.Limits[k8sv1.ResourceCPU]
```
ahadas: I wonder what is the expected result of:
cpu-request = 0
guest-cpu = 4 cpus
cpu-limit = 8000m
I read this as "I want my VM to have 4 vcpus but should never consume more than 8 cpus", right?
This code would set cpu-request=8000m which means that for scheduling purposes we require 8 cpus for this VM although this is more than what it really needs. I think that unlike ordinary pods, we should take into account the guest cpus here

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
```diff
@@ -132,17 +132,33 @@ func (mutator *VMIsMutator) setDefaultMachineType(vmi *v1.VirtualMachineInstance
...
+	}
+
+	if !resources.Limits.Memory().IsZero() && resources.Requests.Memory().IsZero() {
+		if resources.Requests == nil {
+			resources.Requests = k8sv1.ResourceList{}
+		}
+		resources.Requests[k8sv1.ResourceMemory] = resources.Limits[k8sv1.ResourceMemory]
```
ahadas: Same here -
I understand that it's a matter of perception, I take it to the terms in ovirt and vsphere kind of systems where if you define memory-request=1G and set memory-overcommit to 200% then you'd expect the VM to be started with 2G virtual memory, whereas you take it to the world of openshift that doesn't take into account the guest properties.

But more generally, why would we ever want to set memory limit for a VM?
I understand why k8s want to do this - you deploy your application within a pod and you may have memory leakage or unexpected pressure that may affect other pods on the node so you expect it to be killed if it exceeds the threshold.  But in case of VMs:
(1) We most probably don't want the VM to be killed
(2) We set the VM with memory=X so if there's a problem in my application, it would anyway be bounded to that X while running within the guest - worst case, new memory allocations would fail. The only reason I see for setting limits is for bugs in qemu that would result in allocating too much memory which is very unlikely, and again - I'm not sure if we necessarily want the VM to be killed in that case

rmohr: > But more generally, why would we ever want to set memory limit for a VM?

It is a requirement to reach specific QOS classes. And QOS classes give you guarantees regarding to Pod/VM eviction, how much exclusive resources you get and they influence the eviction order in case nodes are under pressure.

rmohr: > I understand that it's a matter of perception, I take it to the terms in ovirt and vsphere kind of systems where if you define memory-request=1G and set memory-overcommit to 200% then you'd expect the VM to be started with 2G virtual memory

SO I say in ovirt vsphere that I want a VM with 1G of memory, and there is a hidden setting which makes 2G out of it?

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
rmohr: I want to be *Pod* compatible. In order to get req: 4, limit: 8 and guest:4 I would expect that you set all three values.

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
ahadas: > > But more generally, why would we ever want to set memory limit for a VM?
> 
> It is a requirement to reach specific QOS classes. And QOS classes give you guarantees regarding to Pod/VM eviction, how much exclusive resources you get and they influence the eviction order in case nodes are under pressure.

I see, so we need this in order to leverage such k8s features that require it. so maybe we should adjust those values in a way they will not conflict with the settings of the virtual resources

ahadas: > > I understand that it's a matter of perception, I take it to the terms in ovirt and vsphere kind of systems where if you define memory-request=1G and set memory-overcommit to 200% then you'd expect the VM to be started with 2G virtual memory
> 
> SO I say in ovirt vsphere that I want a VM with 1G of memory, and there is a hidden setting which makes 2G out of it?

No, if you say "I want a VM with 8G of memory" that's what you'll get.
I think it is equivalent to setting Memory.Guest=8G in kubevirt.
But if you say "I want to guarantee 4G for my VM" and you set overcommit to 150%, then you'd expect to see virtual memory = 6G.
I see your point in saying "I set memory-request=4G" so I want my guest to be set with 4G and not with 6G, but I think that if we consider memory-request to be minimum guaranteed memory rather than guest-memory, then it actually makes sense

rmohr: > But if you say "I want to guarantee 4G for my VM" and you set overcommit to 150%, then you'd expect to see virtual memory = 6G.

Do I set both values in the same dialogue, or is this overcommit value something which is somehow made known to all users? I can understand overcommit ratio when I only set the guest memory. Then the system can request less for me. But the other way round, especially if the two settings are not discoverable together, sounds a little bit strange to me.

ahadas: The memory-overcommit value is set at the cluster level.
In the add/edit VM, when setting the memory or minimum guaranteed memory property, the other one is automatically computed according to the cluster memory overcommit ratio.

I agree that the more common scenario is for someone to provide the virtual memory setting rather than the minimum memory guaranteed so it is not that significant. But as I expect frontend clients to operate like oVirt webadmin in that context, I think it would be better for the backend component to do the same

rmohr: So it looks like we agree now on everything except the request to guest-memory propagation. Can we break this one thing out into an issue and look for a tie breaker? I think we can proceed with the rest. How does that sound?

ahadas: Sure, I was about to drop this part from my PR for the same reason :)

I think that with one more consideration it can make more sense though:
(1) if we are provided with memory/cpu-limits and guest-memory/cpu is not provided then copy the limit to the request as k8s does for POD and set the virtual resources (guest-memory/cpu) accordingly, in case of memory, without taking into account the memory-overcommit.
So if:
memory-request=0
memory-limit=4G
guest-memory=0
then in order to be POD compatible do:
memory-request=memory-limit=guest-memory=4G

(2) if we are provided with memory/cpu-limits and guest-memory/cpu is also provided, set the request according to the virtual resource.
So if:
memory-request=0
memory-limit=4G
guest-memory=3G
then do:
memory-request=3G
memory-limit=4G
guest-memory=3G

I understand (2) might conflict with the behavior expected for ordinary PODs but I strongly believe that requiring more memory than what the guest needs is not the right way to go

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
```diff
@@ -132,17 +132,33 @@ func (mutator *VMIsMutator) setDefaultMachineType(vmi *v1.VirtualMachineInstance
 }
 
 func (mutator *VMIsMutator) setDefaultResourceRequests(vmi *v1.VirtualMachineInstance) {
-	if _, exists := vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory]; !exists {
-		if vmi.Spec.Domain.Resources.Requests == nil {
-			vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{}
+	resources := &vmi.Spec.Domain.Resources
+
+	if !resources.Limits.Cpu().IsZero() && resources.Requests.Cpu().IsZero() {
```
vladikr: I was sure that this is happening automatically in k8s. When you only specify the limits, requests will be set to equal the limits.

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
ahadas: Btw, it may be too late to reconsider this now - but can't we unify memory-limit and guest-memory, why do we need both?

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
rmohr: It happens automatically on the Pod. Yes. Since the VMI is our atomic entity I wanted to make it visible on the VMI too. Does that make sense?

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
vladikr: but we still can have a VM with a lower amount of memory than what we request for the pod, right?

ahadas: Is it useful? I think of it as if minimum guaranteed memory in oVirt would be set to a higher value than the guest memory and I find no use for this - is there in kubevirt?

rmohr: > Btw, it may be too late to reconsider this now - but can't we unify memory-limit and guest-memory, why do we need both?

Your question makes a lot of sense.

We can't remove any fields at this stage anymore. But setting guest memory if limit is specified to the limit might still be a possibility.

rmohr: > Is it useful? I think of it as if minimum guaranteed memory in oVirt would be set to a higher value than the guest memory and I find no use for this - is there in kubevirt?

I think vladik refers to the libvirt/qemu overhead.

rmohr: Oh and actually setting requests and a higher guest memory than that without setting a memory limit makes sense. It allows overcommiting of memory because if a limit is set then it is reserved on the scheduling level already.

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
vladikr: Yeah, sure, it's better to express it explicitly.

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go
ahadas: > Oh and actually setting requests and a higher guest memory than that without setting a memory limit makes sense. It allows overcommiting of memory because if a limit is set then it is reserved on the scheduling level already.

Oh I didn't know this part - can you please point me to the documentation where this part is explained? I thought that the scheduler only looks at the request

vladikr: > I think vladik refers to the libvirt/qemu overhead.

Yes, sorry, for such a brief response. I also think that in some cases we might underestimate this overhead and it's important for us to be able to manually increase it. Does it make sense?

ahadas: > > I think vladik refers to the libvirt/qemu overhead.
> 
> Yes, sorry, for such a brief response. I also think that in some cases we might underestimate this overhead and it's important for us to be able to manually increase it. Does it make sense?

It is, but then I would expect a cluster-wide setting of this overhead rather than expecting users to provide is as part of the VM spec...

rmohr: My last comment regarding to limits was nonsense. Your question regarding to limits and guest memory points out the right thin, I think.

tests/vmi_configuration_test.go
rmohr: Yep. :+1:

pkg/api/v1/types.go
rmohr: Done.

PR #2391, opened 2019-06-14
Fixed RBAC for non admin users

slintes:
Non admin users weren't able to start, restart or stop VMs, and to access Migrations

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1720433 and https://bugzilla.redhat.com/show_bug.cgi?id=1721016

There are several fixes in here:
- added start and stop RBAC rules for admin and edit roles
- added migration RBAC rules for admin and edit roles
- fixed the check if accessing a subresource is allowed (http verb to RBAC verb conversion)
- added tests

**Release note**:
```release-note
Fixed RBAC rules for non admin users
```


rmohr:
Looks good. Now we need a test to avoid another regression.

slintes:
yep, will see if I can get at a functional test ready on monday, or create an issue for it

cynepco3hahue:
the code looks good, but unittest needs fix and as @rmohr said we need functional tests

slintes:
also fixed missing RBAC rule for migartion, and added tests

pkg/virt-api/rest/authorizer_test.go
```diff
@@ -36,144 +36,162 @@ import (
...
-			close(done)
-		}, 5)
-
-		It("should reject unauthorized user", func(done Done) {
-
-			req.Request.TLS = &tls.ConnectionState{}
-			req.Request.TLS.PeerCertificates = append(req.Request.TLS.PeerCertificates, fakecert)
-
-			result, err := app.generateAccessReview(req)
+var _ = Describe("Authorizer", func() {
```
slintes: note to reviewers: this looks worse (at least on github) than it is: I only put the complete old test in a new decsribe block, and added a new block at the end

PR #2395, opened 2019-06-18
Make containerdisks secure and supportable

rmohr:

 * Allow having `scratch` as `containerDisk` base images
 * Mount image in `containerDisk` instad of copying it
 * Since the `containerDisk` is now a mount, only use it as backing store to avoid layer-copying because of file manipulations.

Implementation:

`virt-handler` is responsible for mounting and unmounting the images. No privileges are required on `containerDisk` images. `virt-handler`  ships a static binary in an init directory which will be started in the containerDisk. The only purpose of that binary is to create a socket and then keep the container open. `virt-handler` can detect these sockets and based on it find the disk image. Once it has the disk image, it can mount it into the `compute` container. Then `virt-handler` performs sanity checks on the images (with `qemu-img` shipped in `virt-launcher`). Finally the `compute` container with libvirt now sees the image and can use it as a backing store.

Fixes #

**Release note**:

```release-note
Containerdisks are now secure and they are not copied anymore on every start.
Old containerdisks can still be used in the same secure way, but new containerdisks can't be used on older kubevirt releases.
```


kfox1111:
FYI, the Kubernetes 1.15 release supports ephemeral csi pod inline and I have tested https://github.com/kubernetes-csi/csi-driver-image-populator with it. So you can use an image directly as a volume now. We're continuing to polish it to get the driver solid and further improvements to 1.16.

fabiand:
@rmohr what's your plan with this pr?

rmohr:
@fabiand still not ripe. Some minor issues, but close. Want to finish it soon.

pkg/virt-handler/vm.go
```diff
@@ -986,6 +988,35 @@ func (d *VirtualMachineController) processVmCleanup(vmi *v1.VirtualMachineInstan
...
+		// let containerdisks gracefully terminate if they still managed to run somehow
+		for _, file := range files {
+			path := filepath.Join(mountDir, file.Name())
+			if strings.HasSuffix(path, ".sock") {
+				continue
+			}
+			if mounted, err := isolation.NodeIsolationResult().IsMounted(path); err != nil {
+				return fmt.Errorf("failed to check mount point for containerDisk %v: %v", path, err)
+			} else if mounted {
+				out, err := exec.Command("/usr/bin/chroot", "--mount", "/proc/1/ns/mnt", "umount", path).CombinedOutput()
```
davidvossel: what happens if the umount never occurs here? will the pod terminate still?

pkg/virt-handler/vm.go
```diff
@@ -1352,6 +1383,56 @@ func (d *VirtualMachineController) processVmUpdate(origVMI *v1.VirtualMachineIns
 	}
 
 	if d.isPreMigrationTarget(vmi) {
+		for i, volume := range vmi.Spec.Volumes {
```
davidvossel: can we break this out into a helper function?

pkg/virt-handler/vm.go
```diff
@@ -1386,6 +1467,60 @@ func (d *VirtualMachineController) processVmUpdate(origVMI *v1.VirtualMachineIns
 			d.recorder.Event(vmi, k8sv1.EventTypeNormal, v1.Migrating.String(), "VirtualMachineInstance is migrating.")
 		}
 	} else {
+
+		if !vmi.IsRunning() && !vmi.IsFinal() {
```
davidvossel: helper function here?

pkg/virt-handler/vm.go
rmohr: The pod will terminate and disappears. But you will see that the kubelet periodically tries to unmount the containerDisk layer and it fails because another mount is still there. I can actually improve this by tracking in an extra file vms for which we did not unmount and periodically do the cleanup for these. This will allow to always clean up successfully as soon as virt-handler is started, no matter if VMIs disappeared or not in the meantime.

rmohr:
> so, the only reason the containerDisk's entrypoint creates the socket is so we can detect the namespace?

Yes. If we would find an alternative we could even shrink it down to the `pause` binary from k8s for just keeping the pod open.

> 
> Is there anything being executed in the containerDisk's container namespace or are all the execs in the compute container?

There are no executables in the containerDisks. It is completely empty.

davidvossel:
> Yes. If we would find an alternative we could even shrink it down to the pause binary from k8s for just keeping the pod open.

pause binary with a fake cli arg that has the uid of the vmi. then scan /proc/ for to discover the pid, ns, etc... keying off the uid arg. I kind of hate that idea though and prefer what you're doing.

pkg/virt-handler/vm.go
davidvossel: > I can actually improve this by tracking in an extra file vms for which we did not unmount and periodically do the cleanup for these

doesn't watchdog file account for this though? we dont' delete the watchdog until the vmi is 100% cleaned up locally.

davidvossel:
> There are no executables in the containerDisks. It is completely empty.

alrighty, that's what I thought based on previous discussions, just wanted to clarify one more time.

The overall direction of the PR looks good to me. From a code standpoint, it would be nice to have some of the complexity hid from major code paths (things like the places I requested helper function for), but that's about it.

I'm slightly concerned about introducing Privileged=true and the SELinux capabilities to the containerDisk images though. Is that our only option here?

pkg/virt-handler/vm.go
rmohr: yes you are right.

rmohr:
> I'm slightly concerned about introducing Privileged=true and the SELinux capabilities to the containerDisk images though. Is that our only option here?

I am still investigating a few config options for selinux on our pods.

rmohr:
> I'm slightly concerned about introducing Privileged=true and the SELinux capabilities to the containerDisk images though. Is that our only option here?

fixed now. virt-handler sets the correct selinux labels and vthe container disk has now 0 permissions.

pkg/virt-handler/vm.go
rmohr: Done

pkg/virt-handler/vm.go
rmohr: Done

rmohr:
> FYI, the Kubernetes 1.15 release supports ephemeral csi pod inline and I have tested https://github.com/kubernetes-csi/csi-driver-image-populator with it. So you can use an image directly as a volume now. We're continuing to polish it to get the driver solid and further improvements to 1.16.

@kfox1111 that sounds nice. As soon as it does not download the images via buildah and can ask the CRI, I am **very** happy to remove this logic here. :+1:

rmohr:
ok, @davidvossel can you do another review round? Test results look good now.

rmohr:
/cc @slintes @cynepco3hahue @davidvossel @vladikr 

would be great to have multiple eyes on this one.

cmd/container-disk-v2alpha/BUILD.bazel
```diff
@@ -0,0 +1,62 @@
...
+    name = "fedora-cloud-container-disk-image",
+    directory = "/disk",
+    files = ["@fedora_image//file"],
+    visibility = ["//visibility:public"],
+)
+
+container_image(
+    name = "fedora30-cloud-container-disk-image",
+    directory = "/disk",
+    entrypoint = ["/entry-point.sh"],
```
davidvossel: does this need the entry-point.sh still?

cmd/container-disk-v2alpha/README.md
```diff
@@ -6,8 +6,7 @@ KubeVirt runtime.
 
 This Base Container is compatible with disk type ContainerDisk:v1alpha
 
-# ContainerDisk:v1alpha
-## Storing Disks in Container Registry
+# Storing Disks in Container Registry
```
davidvossel: we're going to need some updates here in the user-guide.

https://kubevirt.io/user-guide/docs/latest/creating-virtual-machines/disks-and-volumes.html#containerdisk

cmd/container-disk-v2alpha/main.go
```diff
@@ -0,0 +1,86 @@
...
+	var healthCheck bool
+
+	logger := log.DefaultLogger()
+
+	flag.StringVar(&copyPath, "copy-path", "", "Image target path")
+	flag.BoolVar(&healthCheck, "health-check", false, "Do a health check")
+	flag.Parse()
+
+	if !healthCheck && copyPath == "" {
+		logger.Error("No copy-path provides.")
```
davidvossel: provides/provided

cmd/container-disk-v2alpha/main.go
```diff
@@ -0,0 +1,86 @@
...
+					os.Exit(0)
+				}
+				os.Exit(1)
+			}
+		}
+	}()
+
+	for {
+		time.Sleep(1 * time.Second)
+		if _, err := os.Stat(socket); os.IsNotExist(err) {
```
davidvossel: is there any reason to trap signals to avoid shutting down before the socket goes away?

cmd/virt-launcher/virt-launcher.go
```diff
@@ -312,28 +312,19 @@ func writeProtectPrivateDir(uid string) {
...
+	if err != nil {
+		log.Log.Reason(err).Errorf("could not clean up ephemeral disk directory: %s", ephemeralDiskDir)
 	}
 }
 
 func main() {
 	qemuTimeout := pflag.Duration("qemu-timeout", defaultStartTimeout, "Amount of time to wait for qemu")
 	virtShareDir := pflag.String("kubevirt-share-dir", "/var/run/kubevirt", "Shared directory between virt-handler and virt-launcher")
 	ephemeralDiskDir := pflag.String("ephemeral-disk-dir", "/var/run/kubevirt-ephemeral-disks", "Base directory for ephemeral disk data")
+	containerDiskDir := pflag.String("container-disk-dir", "/var/run/kubevirt/container-disks", "Base directory for container disk data")
```
davidvossel: does this mean that every vmi's pod can access every other vmi's containerDisks now?

cmd/container-disk-v2alpha/main.go
rmohr: Since I don't trap it, do you mean that we may have to trap it? I don't think so. The mounted disk is available until virt-handler is done with the cleanup and unmounts the disk.

cmd/virt-launcher/virt-launcher.go
rmohr: No, the containerdisks are directly mounted into the correct pod at these paths. If you go to /var/run/kubevirt/container-disks on the node, you will just see an empty directory.

cmd/container-disk-v2alpha/main.go
rmohr: Fixed

cmd/container-disk-v2alpha/BUILD.bazel
rmohr: Fixed

rmohr:
test failures are unrelated, but still:

/retest

cmd/virt-launcher/virt-launcher.go
davidvossel: yay, alrighty. that's what I thought, i just second guessed myself. volumes involve so much indirection at times

cmd/container-disk-v2alpha/main.go
davidvossel: i guess a better way of saying this is... will there be any issues if the containerDisk containers are terminated before virt-launcher?

rmohr: Worst thing is that the kubelet already tries to unmount the containerDisk root, and that you see an error that this failed. As soon as the vmi is down, virt-handler will unmount all containerDisk mounts and the kubelet can proceed with the mount point cleanup.

davidvossel: The umount looks like it depends on the isolation pid detection though. It's unclear to me how the umount works after the fact if the pid inside the container exits.

rmohr: Luckily not. It works like this:

 * Every `compute` pod has a container-disk directory mounted which is on the host : `/var/run/kubevirt/container-disks/<uid>` and inside the pod: `/var/run/kubevirt/container-disks/`
 * This means that the pods can only see their disks (if there are any) or an empty dir in `/var/run/kubevirt/container-disks`
 * Further this mount is `HostToContainer` mount propagation
 * Further we have `/var/run/kubevirt` mounted with `Bidirectional` mount propagation to virt-launcher
 * `virt-handler` does a mount from the host perspective (to see the container disk mounts) to `/var/run/kubevirt/container-disks/<uid>/<disk>`.
 * Now `virt-handler` and `virt-launcher` both see the disks in their namespaces, because of the mount propagation
 * `virt-handler` can now anytime unmount based on the watchdog file, by just checking  for mounts in the container-disks folder of the VMI. No pid is needed anymore at this stage.

pkg/virt-handler/container-disk/mount.go
```diff
@@ -0,0 +1,108 @@
...
+					return fmt.Errorf("failed to create mount point target %v: %v", targetFile, err)
+				}
+				f.Close()
+
+				out, err := exec.Command("/usr/bin/chroot", "--mount", "/proc/1/ns/mnt", "mount", "-o", "ro,bind", strings.TrimPrefix(sourceFile, nodeRes.MountRoot()), targetFile).CombinedOutput()
+				if err != nil {
+					return fmt.Errorf("failed to bindmount containerDisk %v: %v : %v", volume.Name, string(out), err)
+				}
+			}
+			if verify {
```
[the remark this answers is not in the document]
rmohr: Hm, I wanted to do that, but the issue is that `qemu-img` is in the virt-launcher pod. It can't see the disks until the mount occured.

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -629,12 +631,37 @@ func (l *LibvirtDomainManager) PrepareMigrationTarget(vmi *v1.VirtualMachineInst
 		logger.Reason(err).Error("failed to read pod cpuset.")
 		return fmt.Errorf("failed to read pod cpuset: %v", err)
 	}
-
+	// Check if PVC volumes are block volumes
+	isBlockPVCMap := make(map[string]bool)
```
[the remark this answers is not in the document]
rmohr: It had no impact because the context information was by accident not needed on the migration target and did not make any difference (until now).

cmd/container-disk-v2alpha/main.go
davidvossel: oh man. this makes my brain hurt. bottom line is it looks like it works correctly, which is all I care about.

pkg/virt-handler/container-disk/mount.go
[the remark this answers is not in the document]
davidvossel: oh yeah... man this stuff is confusing. makes sense though

pkg/virt-handler/selinux/labels.go
```diff
@@ -10,13 +10,26 @@ import (
...
-	_, err = os.Stat("/proc/1/root/usr/bin/chcon")
-	return
+func isPresent() (present bool, err error) {
+	_, err = os.Stat("/proc/1/root/usr/sbin/getenforce")
+	if os.IsNotExist(err) {
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+	out, err := exec.Command("/usr/bin/chroot", "--mount", "/proc/1/ns/mnt", "exec", "--", "/usr/sbin/getenforce").CombinedOutput()
```
davidvossel: Do you even need to check to see if SELinux is enabled or not? if the setenforce binary exists, doesn't that mean SELinux is at least installed. Shouldn't the same logic be applied regardlesss if selinux is set to enforcing or disabled at the moment this check occurs?

pkg/virt-handler/selinux/labels.go
```diff
@@ -10,13 +10,26 @@ import (
...
+func NewSELinux() (SELinux, bool, error) {
+	present, err := isPresent()
+	return &SELinuxImpl{}, present, err
 }
 
-func isPresent() (err error) {
-	_, err = os.Stat("/proc/1/root/usr/bin/chcon")
-	return
+func isPresent() (present bool, err error) {
+	_, err = os.Stat("/proc/1/root/usr/sbin/getenforce")
```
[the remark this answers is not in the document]
rmohr: I can check `/sbin/`, `/usr/sbin/`, `/bin/` and `/usr/bin`. I think everything else is a little bit unrealistic.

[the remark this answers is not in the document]
rmohr: There are some go bindings for selinux. I did not fully understand yet how they work, but that might be another solution: https://github.com/opencontainers/selinux

pkg/virt-handler/selinux/labels.go
rmohr: Yes, because I can't do the changes if it is Disabled, only when it is Envorced or Permissive. It would be hard to see if a command should work and fails on an arbitrary error, or libvirt is disabled.

pkg/virt-handler/selinux/labels.go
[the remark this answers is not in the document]
rmohr: Fixed.

cmd/container-disk-v2alpha/README.md
rmohr: Done: https://github.com/kubevirt/user-guide/pull/273

davidvossel:
merge conflict :(

sorry. I think i introduced it with my csv-generator pr. I think you need to re-run make dep-update

PR #2412, opened 2019-06-21
[Test] Adding OverCommit VM memory tests

vatsalparekh:

Fixes #

**Release note**:

```release-note
Added OverCommit VM memory tests
```

### DIFF
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -407,6 +407,44 @@ var _ = Describe("Configurations", func() {
 
 		})
 
+		Context("[rfe_id:609][crit:medium][vendor:cnv-qe@redhat.com][level:component]Support memory over commitment test", func() {
+			vmi := tests.NewRandomVMIWithEphemeralDiskAndUserdata(tests.ContainerDiskFor(tests.ContainerDiskCirros), "#!/bin/bash\necho 'hello'\n")
+			vmi.Spec.Domain.Resources.Requests[kubev1.ResourceMemory] = resource.MustParse("128M")
+			vmi.Spec.Domain.Resources.OvercommitGuestOverhead = true
+
+			// Create VMI inside BeforeEach because it gets cleanedup, see BeforeEach at the top
+			BeforeEach(func() {
+				vmi, err := virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+				Expect(err).ToNot(HaveOccurred())
+				tests.WaitForSuccessfulVMIStart(vmi)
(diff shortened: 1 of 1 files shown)


vatsalparekh:
This test includes new test id:731, and porting of the context from here https://github.com/kubevirt/kubevirt-ansible/blob/master/tests/high_performance_vm_test.go#L62

tests/vmi_configuration_test.go
```diff
@@ -407,6 +407,44 @@ var _ = Describe("Configurations", func() {
...
+				expecter, err := tests.LoggedInCirrosExpecter(vmi)
+				Expect(err).ToNot(HaveOccurred())
+				defer expecter.Close()
+
+				// Check on the VM, if the Free memory is roughly what we expected
+				res, err := expecter.ExpectBatch([]expect.Batcher{
+					&expect.BSnd{S: "[ $(free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2) -gt 920 ] && echo 'pass' || echo 'fail'\n"},
+					&expect.BExp{R: "pass"},
+				}, 15*time.Second)
+				log.DefaultLogger().Object(vmi).Infof("%v", res)
```
petrkotas: Is the log really needed here? It will spam the log output of the tests.

tests/vmi_configuration_test.go
```diff
@@ -407,6 +407,44 @@ var _ = Describe("Configurations", func() {
 
 		})
 
+		Context("[rfe_id:609][crit:medium][vendor:cnv-qe@redhat.com][level:component]Support memory over commitment test", func() {
+			vmi := tests.NewRandomVMIWithEphemeralDiskAndUserdata(tests.ContainerDiskFor(tests.ContainerDiskCirros), "#!/bin/bash\necho 'hello'\n")
+			vmi.Spec.Domain.Resources.Requests[kubev1.ResourceMemory] = resource.MustParse("1G")
```
petrkotas: Is the 1G hard requirement? The tests would work for much less memory. Generally our tests should require only resources that are necessary. Making the memory smaller will improves the overall CI run.

tests/vmi_configuration_test.go
rmohr: It is good practice to have it. All logs are going to the log collector and are only visible if the test fails.

tests/vmi_configuration_test.go
vatsalparekh: Not a hard requirement, it was mentioned in the Polarion test case but shouldn't matter if it's something like 128M, changed now

PR #2415, opened 2019-06-23
Run make target under the travis

cynepco3hahue:
Before we run `make build` but this target does not exist anymore.
In additional it should version of packages when it compiled with pure `go`.

Fixes #

**Release note**:

```release-note
NONE
```


cynepco3hahue:
> Why did we remove `build`?

@rmohr I believe we remove once we created `bazel-build` and `go-build`, so we built all artifacts only on unittest run, and it built with go and not with the bazel it was the difference between travis and my environment.

I checked that version has the correct value, both for go and bazel builds so it should be ok now.

PR #2429, opened 2019-06-26
Add RNG to all Fedora examples

booxter:
This is needed on newer guest kernels that may lock booting process waiting for entropy that never arrives. (With this fix, RNG used is external to guest state.)

```release-note
NONE
```

### DIFF
--- a/examples/vm-template-fedora.yaml
+++ b/examples/vm-template-fedora.yaml
@@ -39,6 +39,7 @@ objects:
             - disk:
                 bus: virtio
               name: cloudinitdisk
+            rng: {}
           machine:
             type: ""
           resources:
--- a/examples/vmi-fedora.yaml
+++ b/examples/vmi-fedora.yaml
@@ -15,6 +15,7 @@ spec:
       - disk:
           bus: virtio
         name: cloudinitdisk
+      rng: {}
     machine:
       type: ""
     resources:
--- a/examples/vmi-genie-multiple-net.yaml
+++ b/examples/vmi-genie-multiple-net.yaml
@@ -20,6 +20,7 @@ spec:
         name: default
       - bridge: {}
         name: ptp
+      rng: {}
     machine:
       type: ""
     resources:
--- a/examples/vmi-genie-ptp.yaml
+++ b/examples/vmi-genie-ptp.yaml
@@ -18,6 +18,7 @@ spec:
       interfaces:
       - bridge: {}
         name: ptp
+      rng: {}
     machine:
       type: ""
     resources:
--- a/examples/vmi-masquerade.yaml
+++ b/examples/vmi-masquerade.yaml
@@ -22,6 +22,7 @@ spec:
         - name: http
           port: 80
           protocol: TCP
+      rng: {}
     machine:
       type: ""
     resources:
--- a/examples/vmi-multus-multiple-net.yaml
+++ b/examples/vmi-multus-multiple-net.yaml
@@ -20,6 +20,7 @@ spec:
         name: default
       - bridge: {}
         name: ptp
+      rng: {}
     machine:
       type: ""
     resources:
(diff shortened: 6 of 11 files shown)


SchSeba:
can you please create a PR in the user-guide?

there is an error in the travis related to python-client, I don't think this error is related we need to run the travis build again.

cynepco3hahue:
hm maybe we just need to add rng device by default?

rmohr:
Do we want RNG enabled by default?

cynepco3hahue:
I think we should enable it by default, but maybe I am missing some negative sides
@fromanirh @MarSik Does it have some negative impact to enable `rng` device for every VM?

fromanirh:
> I think we should enable it by default, but maybe I am missing some negative sides
> @fromanirh @MarSik Does it have some negative impact to enable `rng` device for every VM?

In general from guest perspective I don't recall any issue.
Highperf VM disable any unneeded device, which could be overzelous for RNG devices, but I need to recheck.
From Host perspective I can't recall any issue in handling the backend for many rng-enabled VMs, but worth a search in the bugzilla archives.

petrkotas:
We had the discussion about enabling the Rng by default when it was implemented. There were two arguments agains having it enabled by default"

1) Windows does not understand the rng device until there is a driver. It will show yellow warning.
2) User does not expect device he did not added. 

To work around this, we settled for a UI solution where kubevirt web UI will set the Rng by default, which is the same solution as is in oVirt.

As for us, we have to communicate in the documentation, the rng should be enabled for better performance.

slintes:
> there is an error in the travis related to python-client, I don't think this error is related we need to run the travis build again.

retriggered and green now

booxter:
@petrkotas where did this discussion happen? I don't recollect being part of a discussion that decided on UI solution.

booxter:
@petrkotas considering that UI will assume that RNG should always be present, should we then maybe also make ALL examples (not just Fedora based) to have the device added to them?

booxter:
For the record, if we are looking for prior art, then in OpenStack, they do the following:

In libvirt driver for nova, when generating the domain XML for the virtual machine, if the Glance image with the guest OS is tagged in its metadata with hardware RNG model == virtio, and if the virtual machine flavor (~template) allows RNG, then Libvirt domain gets a new device for virtio-rng added. In Nova context, there is no configuration to do for the virtual machine itself; instead, Glance image owner defines that the image supports virtio RNG, and admin allows the user to use hardware RNG; if both conditions are satisfied then the device is automatically added.

In KubeVirt, we don't have images carrying metadata about which devices they support, so it's impossible to distinguish between those images that would benefit from RNG and those that would not.

petrkotas:
Hey @booxter most of the discussion happend in the design document and in the pull request. 
PR: https://github.com/kubevirt/kubevirt/pull/1385
Design: https://docs.google.com/document/d/1G75lPm0dCf3ZcFnG6NJpUDZgTaQnFKKM7_glpXzH9uc/edit#heading=h.nbihdrjh9f07

And you are right with openstack, it was one of the reason to not enable it by default via API. Higher level logic, can do that, eg templates, UI.

Some of the discussion happend directly with UI team lead by @jelkosz . I am sorry if you were not part of that discussion. Will make sure to include you next time.

petrkotas:
> @petrkotas considering that UI will assume that RNG should always be present, should we then maybe also make ALL examples (not just Fedora based) to have the device added to them?

I am not sure how alpine and cirros handle virtio drivers. They are bare minimal OSes, and I guess they do not include unnecesary drivers. In my opinion Fedora will be enough.

booxter:
@rmohr I am not sure how to decide whether we want to make RNG added by default to all VMIs, or leave it up to manifest deployer. It doesn't seem like we know of any issues adding the device; on the other hand, it may be useless for some lightweight images like cirros. We have other "core" devices of similar kind, for example, `clock`, that are not added by default. Perhaps we should consider them the consistently. Thoughts?

PR #2436, opened 2019-06-27
Add remaining tests for, secrets and configmap as volumes

kbidarkar:
1) This is port from kubevirt-ansible repo

2) Add test, configmap and secret as volumes on
   the same VM.

3) Add test, Secret with SSH Keys - Single Volume.

Signed-off-by: Kedar Bidarkar <kbidarka@redhat.com>

**What this PR does / why we need it**:  Ports tests related to `secrets and config-maps` from kubevirt-ansible to this repo. 
Ports the below mentioned PR:
https://github.com/kubevirt/kubevirt-ansible/pull/478

Fixes #

**Release note**:

```
NONE
```

### DIFF
--- a/tests/config_test.go
+++ b/tests/config_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/pborman/uuid"
 
 	"kubevirt.io/client-go/kubecli"
+	"kubevirt.io/client-go/log"
 	"kubevirt.io/kubevirt/pkg/config"
 	"kubevirt.io/kubevirt/tests"
 )
@@ -34,6 +35,8 @@ import (
 var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:component]Config", func() {
 
 	tests.FlagParse()
+	virtClient, err := kubecli.GetKubevirtClient()
+	tests.PanicOnError(err)
 
 	BeforeEach(func() {
 		tests.BeforeTestCleanup()
@@ -46,9 +49,6 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 				configMapName string
 				configMapPath string
 			)
-
-			virtClient, err := kubecli.GetKubevirtClient()
-			tests.PanicOnError(err)
 
 			BeforeEach(func() {
 				configMapName = "configmap-" + uuid.NewRandom().String()
@@ -143,9 +143,6 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 				secretName string
 				secretPath string
 			)
-
-			virtClient, err := kubecli.GetKubevirtClient()
-			tests.PanicOnError(err)
 
 			BeforeEach(func() {
 				secretName = "secret-" + uuid.NewRandom().String()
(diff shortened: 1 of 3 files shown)


tests/config_test.go
```diff
@@ -286,4 +282,201 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+					vmiPod,
+					vmiPod.Spec.Containers[1].Name,
+					[]string{"cat",
+						configMapPath + "/config1",
+						configMapPath + "/config2",
+						configMapPath + "/config3",
+					},
+				)
+				Expect(err).To(BeNil())
+				Expect(podOutput_cfgMap).To(Equal(expectedOutput_cfgMap))
```
petrkotas: Expect like this would read: Expected valuevalue to equal value1value2value3. Without any context this will not be helping to understand what happend on the first sight.

Would you please add descriptors to expect similar to this in your PR to make it more readable? Thanks.

tests/config_test.go
```diff
@@ -286,4 +282,201 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+					// mount ConfigMap image
+					&expect.BSnd{S: "sudo su -\n"},
+					&expect.BExp{R: "#"},
+					&expect.BSnd{S: "mount /dev/sda /mnt\n"},
+					&expect.BSnd{S: "echo $?\n"},
+					&expect.BExp{R: "0"},
+					&expect.BSnd{S: "cat /mnt/config1 /mnt/config2 /mnt/config3\n"},
+					&expect.BExp{R: expectedOutput_cfgMap},
+				}, 200*time.Second)
+				Expect(err).ToNot(HaveOccurred())
```
petrkotas: You can add logging for better debug messages. As noted here https://github.com/kubevirt/kubevirt/pull/2412/files#diff-d47bf036797965d742615e64adc3b777R443

The same for other expecters.

kbidarkar:
> I don't understand why we copy two tests ([1](https://github.com/kubevirt/kubevirt/blob/master/tests/config_test.go#L42),[2](https://github.com/kubevirt/kubevirt/blob/master/tests/config_test.go#L139)) scenarios which already exist and we merge them under one context. What's the point of doing this?

Short answer: `The idea was to test whether both secrets and config-map work with 1 single VM.`

Firstly, I was referring to a polarion test_case id `786` to automate that test case, which was part of the test_plan.

Secondly,  at times certain things may seem redundant at first but would help catch issues/bugs hence the addition of that test I believe. 

While working on the CPU manager feature, 

scenario1) VM without cpu-pinning test PASSED
scenario2) VM with cpu-pinning test PASSED
scenario3) VM without cpu-pinning test FAILED after a VM with cpu-pinning was started.

Now, scenario3 seems redundant but it helped find a bug. https://bugzilla.redhat.com/show_bug.cgi?id=1646372

mfranczy:
> The idea was to test whether both secrets and config-map work with 1 single VM.
following that, what about service account and other volumes types? We would have to have a combination of each volume type.

> While working on the CPU manager feature,
Not sure if I would compare these tests to that, but I don't have good arguments here (didn't try to understand the CPU test scenario yet).

In my opinion, we shouldn't do that, but I might be wrong. We need to have a different opinion about that. 

@slintes since you were working on a service account, what do you think about that?

petrkotas:
When checking the codebase, I agree with @mfranczy. This PR should be refactor of existing code, rather than adding copy of existing code with a slight variation.

Regarding the variations. Testing for every possible test case, will put a lot of stress on our CI. I can see the added value of testing the variations, however the cost can be more flaky tests because of the stress on our CI. Would it make sense to do these types of tests separately?

kbidarkar:
@slintes, @mfranczy, @petrkotas, Will just port these tests written for kubevirt-ansible to over here, which making use of golang stuff. Plan is to move to the python framework moving forward.

I don't intend to add many of such tests further, there could be 1 or 2 more such tests and the reason they were added is to avoid rework as they were already automated in golang but in kubevirt-ansible.

Earlier I was answering about the variations as to why they were part of the test plan. 

Shall we merge these? Any new effort would move to the new python framework hereafter.

tests/config_test.go
```diff
@@ -286,4 +283,206 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+				secretPath    string
+			)
+
+			BeforeEach(func() {
+				configMapName = "configmap-" + uuid.NewRandom().String()
+				configMapPath = config.GetConfigMapSourcePath(configMapName + "-disk")
+				secretName = "secret-" + uuid.NewRandom().String()
+				secretPath = config.GetSecretSourcePath(secretName + "-disk")
+
+				config_data := map[string]string{
```
slintes: please use camelCase names

tests/config_test.go
```diff
@@ -286,4 +283,206 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+					"config1": "value1",
+					"config2": "value2",
+					"config3": "value3",
+				}
+
+				secret_data := map[string]string{
```
slintes: same

tests/config_test.go
```diff
@@ -286,4 +283,206 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+			})
+
+			AfterEach(func() {
+				tests.DeleteConfigMap(configMapName)
+				tests.DeleteSecret(secretName)
+			})
+
+			It("[test_id:786]Should be that cfgMap and secret fs layout same for the pod and vmi", func() {
+				expectedOutput_cfgMap := "value1value2value3"
+				expectedOutput_Secret := "adminredhat"
```
slintes: same on both lines above

tests/config_test.go
```diff
@@ -286,4 +283,206 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+
+		Context("With a single volume", func() {
+			var (
+				secretName string
+				secretPath string
+			)
+
+			var bitSize int = 2048
+			privateKey, _ := tests.GeneratePrivateKey(bitSize)
+			publicKeyBytes, _ := tests.GeneratePublicKey(&privateKey.PublicKey)
```
slintes: no need to check for the error in both lines above?

mfranczy:
@kbidarkar you need to add output from `make generate` as well.

tests/config_test.go
kbidarkar: Will update

kbidarkar:
These are some of the serviceAccount tests that will be ported from here, https://github.com/kubevirt/kubevirt-ansible/pull/493

kbidarkar:
But probably `serviceAccount` stuff I was planning to do in a separate PR.

tests/utils.go
```diff
@@ -3642,3 +3647,50 @@ func WaitAgentConnected(virtClient kubecli.KubevirtClient, vmi *v1.VirtualMachin
...
+
+// generatePrivateKey creates a RSA Private Key of specified byte size
+func GeneratePrivateKey(bitSize int) (*rsa.PrivateKey, error) {
+	// Private Key generation
+	privateKey, err := rsa.GenerateKey(cryptorand.Reader, bitSize)
+	if err != nil {
+		return nil, err
+	}
+
+	// Validate Private Key
```
sHaggYcaT: As far as I know, code styles recommends avoiding obvious comments, like this:

// Adding a to b
c = a + b

tests/utils.go
```diff
@@ -3642,3 +3647,50 @@ func WaitAgentConnected(virtClient kubecli.KubevirtClient, vmi *v1.VirtualMachin
...
+	return publicKeyBytes, nil
+}
+
+// encodePrivateKeyToPEM encodes Private Key from RSA to PEM format
+func EncodePrivateKeyToPEM(privateKey *rsa.PrivateKey) []byte {
+	// Get ASN.1 DER format
+	privDER := x509.MarshalPKCS1PrivateKey(privateKey)
+
+	// pem.Block
+	privateBlock := pem.Block{
```
sHaggYcaT: same as previous

tests/utils.go
```diff
@@ -3642,3 +3647,50 @@ func WaitAgentConnected(virtClient kubecli.KubevirtClient, vmi *v1.VirtualMachin
 		return false
 	}, 420*time.Second, 2).Should(BeTrue(), "Should have agent connected condition")
 }
+
+// generatePrivateKey creates a RSA Private Key of specified byte size
+func GeneratePrivateKey(bitSize int) (*rsa.PrivateKey, error) {
+	// Private Key generation
```
sHaggYcaT: same as next

kbidarkar:
@mfranczy Added the output from `make generate` as well

kbidarkar:
Updated for all the review comments here.

tests/utils.go
kbidarkar: Removed the unnecessary comment, mentioned above.

mfranczy:
@kbidarkar tests fail on `panic: You are trying to make an assertion, but Gomega's fail handler is nil. `

sHaggYcaT:
LGTM

mfranczy:
I am not going to block this since people have a different opinion than me.

Issue #1847, opened 2018-12-15
Fedora vmi console got stuck at "[  OK  ] Started Login Service."


duyanyan:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
> /kind bug

**What happened**:

```
# oc get pod
virt-launcher-vmi-fedora-27znw         2/2       Running   0          5m        10.128.0.111
# oc get vmi
NAME         AGE
vmi-fedora   5m
# oc describe vmi vmi-fedora
Name:         vmi-fedora
Namespace:    kubevirt
Labels:       kubevirt.io/nodeName=dev-test
              special=vmi-fedora
Annotations:  <none>
API Version:  kubevirt.io/v1alpha2
Kind:         VirtualMachineInstance
Metadata:
  Creation Timestamp:  2018-12-15T05:03:25Z
  Generation:          1
  Resource Version:    178612
  Self Link:           /apis/kubevirt.io/v1alpha2/namespaces/kubevirt/virtualmachineinstances/vmi-fedora
  UID:                 c0917da1-0026-11e9-9a46-80c16e6c5154
Spec:
  Domain:
    Devices:
      Disks:
        Disk:
          Bus:        virtio
        Name:         containerdisk
        Volume Name:  registryvolume
        Disk:
          Bus:        virtio
        Name:         cloudinitdisk
        Volume Name:  cloudinitvolume
    Machine:
      Type:  
    Resources:
      Requests:
        Memory:                      1024M
  Termination Grace Period Seconds:  0
  Volumes:
    Container Disk:
      Image:  kubevirt/fedora-cloud-container-disk-demo:latest
    Name:     registryvolume
    Cloud Init No Cloud:
      User Data:  #cloud-config
password: fedora
chpasswd: { expire: False }
    Name:  cloudinitvolume
Status:
  Node Name:  dev-17.rhev.lab.eng.brq.redhat.com
  Phase:      Running
Events:
  Type    Reason            Age              From                                              Message
  ----    ------            ----             ----                                              -------
  Normal  SuccessfulCreate  6m               virtualmachine-controller                         Created virtual machine pod virt-launcher-vmi-fedora-27znw

  Normal  Created           5m (x2 over 5m)  virt-handler, dev-test  VirtualMachineInstance defined.

  Normal  Started           5m               virt-handler, dev-test  VirtualMachineInstanced started.
# virtctl console vmi-fedora
[...]

PR #2441, opened 2019-06-28
Always add rng device to fedora VMs in functests

booxter:
Newer kernels may block booting the guest waiting for entropy for some
of its services.

Related to #1847

```release-note
NONE
```


mfranczy:
@booxter shouldn't you update [console test](https://github.com/kubevirt/kubevirt/blob/master/tests/console_test.go#L93) as well?

PR #2444, opened 2019-07-01
Vfiosh

fedepaol:

This PR fixes a couple of problems I found in the setup script while working on CI for sr-iov devices:

- Some devices do not allow us to write on `sriov_numvfs` in order to activate the virtual functions. In that case we want to skip them and enable the ones that allow us to do so
- When no driver was bound to the device, the script skipped it and did not bind it to the vfio-pci driver

**Release note:**

```release-note
NONE
```

### DIFF
--- a/tools/util/vfio.sh
+++ b/tools/util/vfio.sh
@@ -12,7 +12,12 @@ for file in $(find /sys/devices/ -name *sriov_totalvfs*); do
     pfroot=$(dirname $file)
 
     # enable all available VFs. If this fails we skip the device continuing with the others.
-    cat $file > $pfroot/sriov_numvfs || continue
+    cat $file > $pfroot/sriov_numvfs
+    if [ $? -ne 0 ]; then
+        echo "Skipping $pfroot, not able to activate virtual functions for it"
+        continue
+    fi
+
 
     # bind all VFs with vfio
     for virtfn in $(ls -d $pfroot/virtfn*); do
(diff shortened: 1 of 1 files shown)


tools/util/vfio.sh
```diff
@@ -11,15 +11,15 @@ modprobe vfio-pci
 for file in $(find /sys/devices/ -name *sriov_totalvfs*); do
     pfroot=$(dirname $file)
 
-    # enable all available VFs
-    cat $file > $pfroot/sriov_numvfs
+    # enable all available VFs. If this fails we skip the device continuing with the others.
```
SchSeba: Can you please add a warning if the enable VFs fails so it will be better for the user to know there was an issue when he run the script.

lets not silent errors :)

rmohr:
@fedepaol is that still something we want to merge?

fedepaol:
@rmohr now the recommended way to perform the sr-iov related configuration is by using the openshift sr-iov operator, so I think we could even remove completely the file (in a new PR).

In any case this pr is an improvement over the current one, so merging it won't bite.

@booxter what do you think?

booxter:
@fedepaol let's just kill the file. Operator works, and ideally we in kubevirt don't care about intricacies of how to configure the node for SR-IOV resources.

fedepaol:
> @fedepaol let's just kill the file. Operator works, and ideally we in kubevirt don't care about intricacies of how to configure the node for SR-IOV resources. Note that this implies cleaning up docs in `docs/sriov.md` (perhaps it's a good moment to discuss if we want to carry that much information there in the first place, not just about how to bind drivers.)

Ok, will do, thanks

PR #2447, opened 2019-07-02
Changing qos class of virt-api and virt-controller to Guaranteed

ahadas:
Set the memory and CPU requests+limits of virt-api and virt-controller in a way that ensures those pods are set with the guaranteed QoS.
By default memory is set to 512M and CPU is set to 500m, but that can be overridden using KubeVirt CR.

**Special notes for your reviewer:**
Test(s) for the configuration part is missing, will be added separately.

**Release note**:

```release-note
The virt-api and virt-contoller pods are set with the guaranteed quality of service. By default they are set with 512M of memory and 500m of CPU. This can be overridden via KubeVirt CR.
```


slintes:
Code looks good to me :+1: 
How did come up with the default values? I do not doubt that they make sense, just curious :slightly_smiling_face: 

/lgtm

slintes:
Mh, wait. I am unsure if a change of cpu/memory values on the KubeVirt CR will actually trigger a redeployment. Can you add a functional test for that?

/lgtm cancel

slintes:
I expect that the problem will be, that the installStrategy uses the imageRegistry and imageTag as "key". Since that won't change, it will not create a new one.

ahadas:
@slintes as for the default values - I saw k8s requests 250m for both pods so I assume that 500m will be enough. And I doubt they will need more than 512M of memory - nothing more than feeling.
As for re-deploymeny, I didn't check thus. Will setting those as keys be something easy that I can add separately?

slintes:
> Will it be something easy that I can add separately or will it require changing the modeling significantly?

good question, not sure yet. Actually I think I'm running into the same problem on a PR I'm working on. We need to make sure that we create a new installStrategy when something relevant for deployments changes.

ahadas:
> > Will it be something easy that I can add separately or will it require changing the modeling significantly?
> 
> good question, not sure yet. Actually I think I'm running into the same problem on a PR I'm working on. We need to make sure that we create a new installStrategy when something relevant for deployments changes.

Ack.
The configuration part as a whole is of lower priority in this  PR so I can either drop the third commit or we can merge it as is and adapt it later on to enable the redeployment when it changes, if that part would block this

slintes:
I would be fine with keeping it as is and adapt later, but don't want to decide that alone.
@rmohr wdyt?

rmohr:
Didn't we agree that some scale testing should be done? I think that @ipinto already ran tests with more than 2500 VMIs. The VMIs don't need to be schedulable and will still show some usage.

slintes:
regarding the issue with redeployments:  I'm working on sth in the context of #2483, which will take the complete spec of the KubeVirt CR (beside other things like important env vars) into account for finding and potentially recreating the correct installStrategy. No need to do it here, or remove the config from this PR

ahadas:
> Didn't we agree that some scale testing should be done? I think that @ipinto already ran tests with more than 2500 VMIs. The VMIs don't need to be schedulable and will still show some usage.

You probably meant @ILpinto , right?

@ILpinto, I looked at the results you've shared with me on the scaling tests (thanks!):
1. I see that you wait 2s between each call to `virtctl start`
2. The graphs I see reflect the Node Utilization

In this context, we look for reasonable defaults for memory & CPU consumption (requests and limits) for virt-api and virt-controller.
The trade-off: setting low values may result in those pods being killed (due to memory limit) or starve (due to CPU limit). Setting high values may prevent those pods from being scheduled or for too much resources being reserved at the expense of other workloads.

So in order to assess this consumption, we are interested in the extreme cases of memory and CPU consumption by those pods. For example, it may be that opening a console leads to high CPU consumption by virt-api and then we'll want to open as many consoles as we theoretically can.

Specifically, for virt-api:
1. We need to find out what takes the highest CPU consumption - either it be to start a VMI/opening a console/delete VMIs/etc. Then, we need to trigger these actions considering the max number of VMIs and pods that can execute on a node.
2. Same for memory.
3. We need to see whether/how much it is affected by the number of VMIs running in the cluster in terms of memory and CPU consumption. (for example, it may that during eviction, we trigger migration commands that go through the API in addition to requests that come "from the outside world").

For virt-controller:
1. It is involved in starting VMIs so we need to find the memory and CPU consumption of it during this operation.
2. We need to understand whether/how much it is affected by the number of VMIs running in the cluster in terms of memory and CPU consumption.

In particular, we need to have graphs that reflect the consumption over time at a stable-reasonable state and when the load incrementally increase until we reach such peaks.

With that, we could analyze the results and determine the defaults. For instance, if we'll see that virt-api consumes 64M of memory that can go up to 256M for load X that is likely to be used by 80% of the clients and can reach 2G for the other extreme cases, we may set it to 256M - assuming that clients that reach those peaks can adjust their configuration. In other example, we may see that virt-api consumes up to 512M at peeks and we consider 512M a reasonable portion of memory to reserve so we'll use it as default even if in a regular load it reaches up to 64M.

So I guess we need the following information:
1. Graphs that show the memory and CPU consumption of those pods over time, specifically, at reasonable load and at extreme (supported) loads. You may need to consult with @rmohr  regarding what is likely to significantly affect the memory and CPU consumption of these pods and how it is affected by VMIs that are running in the cluster.

2. Whether you want to test it with the current, unlimited, settings of these pods or with the new settings. With the new settings it may be easier to detect reaching the limits (e.g., the pods may be killed when exceeding the memory limit or operations may get stuck when those pods don't get enough CPU time).

3. Whether you think it worth your time :)  You may just say "ok guys, please merge this beautiful PR - I'll test it with reasonable loads and see how it goes, worst case - we'll change the defaults later " (the defaults are configurable). ".. I think this is not the right approach to set limits on these nodes - you should try to make sure they run on master nodes with CPU requests as the corresponding pods in k8s and address the problem of preventing those pods from being evicted differently, not by setting limits to these pods. But as a temporary workaround it is fine - just make it configurable".

rmohr:
@ahadas is this still on your list? Would be important to switch to the right qos class.

PR #2450, opened 2019-07-03
Fix loopback not found error on kind clusters

### LINKED ISSUE #2449: Unable to run disks-images-provider with docker in docker scenario

/kind bug

**What happened**:
In order to make sriov-ci work, we are running a [kind](https://github.com/kubernetes-sigs/kind) cluster inside a pod spawn by Prow.

The final picture is:
- a container with docker in docker enabled
- a running container that resembles the cluster's node
- pods inside this "node"

In this particular scenario, the `disks-image-provider` pod fails to start with the following error:
```
root@7dbdc157d542:/workspace# kubectl logs -n kubevirt disks-images-provider-zcpbw
converting cirros image from qcow2 to raw, and copying it to local-storage directory, and creating a loopback device from it
losetup: /local-storage/cirros.img.raw: failed to set up loop device: No such file or directory
```

The `disks-image-provider` pod to run.

from a kubevirt checkout:

git clone https://github.com/fedepaol/diskimageprovidertest.git && ./diskimageprovidertest/run.sh

This will run a container using the same image and the same options used on prow, run a kind cluster and install kubevirt on it.

Wait until it checks for the pods to be ready:

```
+ echo 'Waiting for all containers to become ready ...'
+ kubectl get pods --all-namespaces '-ocustom-columns=status:status.containerStatuses[*].ready,metadata:metadata.name' --no-headers
true      cdi-apiserver-566f6bc5bc-p8czh
true      cdi-deployment-bc58c8f79-9zhpf
true      cdi-operator-8f98cf6f6-j967r
true      cdi-uploadproxy-79d7b47d9-gwwdd
true      coredns-86c58d9df4-87n87
true      coredns-86c58d9df4-knw86
true      etcd-sriov-ci-control-plane
true      kube-apiserver-sriov-ci-control-plane
true      kube-controller-manager-sriov-ci-control-plane
true      kube-flannel-ds-amd64-nmhvf
true      kube-multus-ds-amd64-ffllp
true      kube-proxy-t66x2
true      kube-scheduler-sriov-ci-control-plane
true      cdi-http-import-server-6b89dcf5c7-28nxb
false     disks-images-provider-l4brl
true      virt-api-b9d84f757-wflk5
true      virt-api-b9d84f757-xsgqs
true      virt-controller-7bc896dd9f-btfkk
true      virt-controller-7bc896dd9f-ctnm5
true      virt-handler-bjq5k
true      virt-operator-886d5bb7d-762pp
true      virt-operator-886d5bb7d-tjp9n
+ sleep 10
```

The final effect is that the disks-images-provider pod will crash:

```
docker exec imageprovidertest bash -c "KUBECONFIG=/root/.kube/kind-config-sriov-ci kubectl get pods --all-namespaces | grep disks-images-provider"
kubevirt      disks-images-provider-l4brl                      0/1     CrashLoopBackOff   4          4m30s
```

Also note that randomly I saw it work, but more often than not it fails.

The error is happening in the entrypoint of the container: https://github.com/kubevirt/kubevirt/blob/master/images/disks-images-provider/entrypoint.sh#L30

fedepaol:
When running a tests on a kind cluster, the `disk-images-provider` pod crashes on start with the following error:
```
losetup: /local-storage/cirros.img.raw: failed to set up loop device: No such file or directory
```
See #2449 for details on how to reproduce it. 

This PR creates a brand new loopback device, attempts to detach it and mounts the image on it instead of relying on the fact that we are going to find an available one.

Fixes #2449

**Release note**:

```release-note
NONE
```

### DIFF
--- a/images/disks-images-provider/entrypoint.sh
+++ b/images/disks-images-provider/entrypoint.sh
@@ -27,9 +27,13 @@ mkdir -p /images/datavolume1 /images/datavolume2 /images/datavolume3
 echo "converting cirros image from qcow2 to raw, and copying it to local-storage directory, and creating a loopback device from it"
 # /local-storage will be mapped to the host dir, which will also be used by the local storage provider
 qemu-img convert -f qcow2 -O raw /images/cirros/disk.img /local-storage/cirros.img.raw
-LOOP_DEVICE=$(losetup --find --show /local-storage/cirros.img.raw)
+
+mknod /dev/loop99 b 7 0
+# attempt a detach since with docker in docker scenarios the created loopback device is already busy
+losetup -d /dev/loop99 || true 
+losetup /dev/loop99 /local-storage/cirros.img.raw
 rm -f /local-storage/cirros-block-device
-ln -s $LOOP_DEVICE /local-storage/cirros-block-device
+ln -s dev/loop99 /local-storage/cirros-block-device
 
 echo "converting fedora image from qcow2 to raw"
 qemu-img convert -f qcow2 -O raw /images/fedora-cloud/disk.qcow2 /images/fedora-cloud/disk.img
--- a/images/disks-images-provider/entrypoint.sh
+++ b/images/disks-images-provider/entrypoint.sh
@@ -30,7 +30,7 @@ qemu-img convert -f qcow2 -O raw /images/cirros/disk.img /local-storage/cirros.i
 
 mknod /dev/loop99 b 7 0
 # attempt a detach since with docker in docker scenarios the created loopback device is already busy
-losetup -d /dev/loop99 || true 
+losetup -d /dev/loop99 || true
 losetup /dev/loop99 /local-storage/cirros.img.raw
 rm -f /local-storage/cirros-block-device
 ln -s dev/loop99 /local-storage/cirros-block-device


fedepaol:
cc @slintes @cynepco3hahue

fedepaol:
Just to add that I know this is probably the best way to solve the issue, but it's what made me go on with the setup, so if there are any alternative approaches suggested, are more than welcome.

images/disks-images-provider/entrypoint.sh
```diff
@@ -27,9 +27,13 @@ mkdir -p /images/datavolume1 /images/datavolume2 /images/datavolume3
 echo "converting cirros image from qcow2 to raw, and copying it to local-storage directory, and creating a loopback device from it"
 # /local-storage will be mapped to the host dir, which will also be used by the local storage provider
 qemu-img convert -f qcow2 -O raw /images/cirros/disk.img /local-storage/cirros.img.raw
-LOOP_DEVICE=$(losetup --find --show /local-storage/cirros.img.raw)
+
+mknod /dev/loop99 b 7 0
```
rmohr: Would `mknod` fail if the container restarts?

rmohr: Ah that is probably not an issue, since that would mean that the container is set up again and `/dev/loop99` does not exist then.

fedepaol: To be honest I did not think about that possibility but yes, it should not. I also made a quick try that confirms that. If I stop & restart a container `/dev/loop99` is gone

fedepaol:
@rmohr sure, thanks for the quick review!

fedepaol:
/hold

since speaking with @SchSeba we are not sure /dev is shared with host for priviledged pods / containers. Let me check if mknod could actually fail and will get back here (soon).

rmohr:
> since speaking with @SchSeba we are not sure /dev is shared with host for priviledged pods / containers. Let me check if mknod could actually fail and will get back here (soon).

If you don't mount /dev explicitly into the container it is separate, even if you are privileged.

fedepaol:
/hold cancel
Verified on a local priviledged container, it won't be there after restarting it.

PR #2451, opened 2019-07-03
Add/remove PodDisruptionBudget on creation vs. deletion of control plane deployments

dhiller:

During deployment of `virt-api` and `virt-controller` create [PodDisruptionBudgets with minAvailable of 1](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#how-disruption-budgets-work) to ensure that at least one pod for the control plane components is always available.

Fixes #

**Release note**:

```release-note
Protect control plane components against voluntary evictions with a PodDisruptionBudget of MinAvailable=1
```


pkg/virt-operator/creation/components/deployments.go
```diff
@@ -531,3 +533,20 @@ func criticalAddonsToleration() ([]byte, error) {
...
+	minAvailable := intstr.FromInt(int(1))
+	podDisruptionBudget := &v1beta1.PodDisruptionBudget{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: deployment.Namespace,
+			Name:      pdbNameForDeployment,
+			Labels:    deployment.Labels,
+		},
+		Spec: v1beta1.PodDisruptionBudgetSpec{
+			MinAvailable: &minAvailable,
+			Selector:     deployment.Spec.Selector,
```
slintes: same as above, selector is a pointer too.

pkg/virt-operator/install-strategy/create.go
```diff
@@ -25,10 +25,13 @@ import (
 	"reflect"
 	"strings"
 
+	"kubevirt.io/kubevirt/pkg/virt-operator/creation/components"
```
slintes: please move down to the other kubevirt imports

pkg/virt-operator/install-strategy/create.go
```diff
@@ -1415,6 +1418,66 @@ func addOrRemoveSSC(targetStrategy *InstallStrategy,
 	return nil
 }
 
+func syncPodDisruptionBudgetForDeployment(deployment *appsv1.Deployment, clientset kubecli.KubevirtClient, kv *v1.KubeVirt, expectations *util.Expectations, stores util.Stores) error {
+	podDisruptionBudget := components.NewPodDisruptionBudgetForDeployment(deployment)
+
+	imageTag := kv.Status.TargetKubeVirtVersion
+	imageRegistry := kv.Status.TargetKubeVirtRegistry
+
+	injectOperatorMetadata(kv, &podDisruptionBudget.ObjectMeta, imageTag, imageRegistry)
```
slintes: related to an earlier comment: here you would also potentially modify the deployment labels, if you don't use a copy. In this case not an issue, because it will just set an already existing label, but you never know...

pkg/virt-operator/install-strategy/create.go
```diff
@@ -1415,6 +1418,66 @@ func addOrRemoveSSC(targetStrategy *InstallStrategy,
...
+
+		// Add Labels and Annotations Patches
+		labelAnnotationPatch, err := createLabelsAndAnnotationsPatch(&deployment.ObjectMeta)
+		if err != nil {
+			return err
+		}
+		ops = append(ops, labelAnnotationPatch...)
+
+		// Add Spec Patch
+		newSpec, err := json.Marshal(deployment.Spec)
```
slintes: you are going to patch the deployment spec into the pdb?

pkg/virt-operator/kubevirt_test.go
```diff
@@ -638,6 +716,19 @@ var _ = Describe("KubeVirt Operator", func() {
 
 	}
 
+	makePodDisruptionBudgetsReady := func() {
```
slintes: nit: this is a `wait...`, no `make...`

pkg/virt-operator/install-strategy/create.go
dhiller: Eagle eyes, thanks!

pkg/virt-operator/creation/components/deployments.go
```diff
@@ -531,3 +533,20 @@ func criticalAddonsToleration() ([]byte, error) {
...
 }
+
+func NewPodDisruptionBudgetForDeployment(deployment *appsv1.Deployment) *v1beta1.PodDisruptionBudget {
+	pdbNameForDeployment := deployment.Name + "-pdb"
+	minAvailable := intstr.FromInt(int(1))
+	podDisruptionBudget := &v1beta1.PodDisruptionBudget{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: deployment.Namespace,
+			Name:      pdbNameForDeployment,
+			Labels:    deployment.Labels,
```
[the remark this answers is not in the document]
dhiller: Good points, thanks!

pkg/virt-operator/kubevirt_test.go
dhiller: Yes, I just didn't want to use a different style, as all the others also follow the pattern `makeXXXReady`...

dhiller:
Need to have a look at the failing os-3.11 tests
/hold

dhiller:
Waiting for feedback from QE on test plan and test id integration
/hold

tests/virt_control_plane_test.go
```diff
@@ -0,0 +1,111 @@
+/*
```
rmohr: You can add it to `infra_test.go`.

tests/virt_control_plane_test.go
```diff
@@ -0,0 +1,111 @@
...
+		return
+	}
+
+	uncordonNode := func(node string) (err error) {
+		err = runCommandOnNode("uncordon", node)
+		return
+	}
+
+	AfterEach(func() {
+		err = uncordonNode("node01")
```
rmohr: You are not supposed to make assumptions about the nodes. The tests will then only work on kubevirtci and not in general.

A few suggestions:

 1) If a test requires more than one node, you have to skip the test if only one node is there. We do that normally like this:

```
		nodes := tests.GetAllSchedulableNodes(virtClient)
		Expect(nodes.Items).ToNot(BeEmpty(), "There should be some compute node")

		if len(nodes.Items) < 2 {
			Skip("Migration tests require at least 2 nodes")
		}
```

Hape if you make something like this out of this:

```
SkipIfSchedulableNodesBelow(2)
```

2) in order to not depend on the count of available nodes, you will have to be a little creative:
 * Get a schedulable node
 * Add a nodeSelector for that node to our deployment
 * Wait until the pod is running on that node
 * drain it

dhiller: @rmohr thanks for your input, great points, will integrate a more flexible setup concerning the number of nodes. IIUC I should restructure the test as to expect a drain on the _last node_ to fail (because of the PDB) _iif_ all other nodes are cordoned. I saw something concerning iterating on present nodes in `tests.BeforeTestCleanup()`, will adapt on this.

Regarding the number of nodes: the PDB is about safeguarding a pod for voluntary evictions, so a test with one node wouldn't make sense, right? I mean in that case there would be manually intervention required by the admin anyway.

rmohr: > IIUC I should restructure the test as to expect a drain on the _last node_ to fail (because of the PDB) _iif_ all other nodes are cordoned.

You don't have to cordon anything. Set a nodeSelector on the Deployment which selects the node where the pods are currently running on. Then the pods can't bescheduled to any other node. Therefore any `evict` should fail.

rmohr: If you would cordon all nodes that would be fun if you have more nodes :D

dhiller: @rmohr if you have time it'd be great if you could have another look

pkg/virt-operator/kubevirt_test.go
```diff
@@ -1404,13 +1524,17 @@ var _ = Describe("KubeVirt Operator", func() {
 			Expect(kv.Status.Phase).To(Equal(v1.KubeVirtPhaseDeploying))
 			Expect(len(kv.Status.Conditions)).To(Equal(0))
 
-			// 2 because waiting on controller and virt-handler daemonset until API server deploys successfully
-			expectedUncreatedResources := 2
+			// 3, 2 because waiting on controller and virt-handler daemonset until API server deploys successfully
```
slintes: Maybe it's just me, but I had to read this 3 times before understanding that this means 3 = 2 + 1 :laughing:

pkg/virt-operator/kubevirt_test.go
```diff
@@ -1404,13 +1524,17 @@ var _ = Describe("KubeVirt Operator", func() {
...
+			// 	and
+			// 1 because uncreated PDB for virt-controller
+			expectedUncreatedResources := 3
 
 			// 1 because a temporary validation webhook is created to block new CRDs until api server is deployed
 			expectedTemporaryResources := 1
 
 			Expect(totalAdds).To(Equal(resourceCount - expectedUncreatedResources + expectedTemporaryResources))
+			//+ expectedPDBsCreated))
```
slintes: please delete

pkg/virt-operator/kubevirt_test.go
```diff
@@ -1487,8 +1616,14 @@ var _ = Describe("KubeVirt Operator", func() {
...
 			// while controllers are available to process them.
-			Expect(totalPatches).To(Equal(patchCount - 2))
+
+			// 3 because 2 for virt-controller service and deployment
+			// and
+			// 1 because of the pdb of virt-controller
+			Expect(totalPatches).To(Equal(patchCount - 3))
 			Expect(totalUpdates).To(Equal(updateCount))
+
+			Expect(resourceChanges["poddisruptionbudgets"][Patched]).To(Equal(1))
```
slintes: nice to track changes in more detail than before :+1:

tests/virt_control_plane_test.go
```diff
@@ -0,0 +1,266 @@
...
+			DefaultStabilizationTimeoutInSeconds, DefaultPollIntervalInSeconds,
+		).Should(Equal(0))
+		Eventually(func() int { return countReadyPodsOnNodes(podNames, nodeNames...) },
+			DefaultStabilizationTimeoutInSeconds, DefaultPollIntervalInSeconds,
+		).Should(BeNumerically(">=", expectedNumberOfReadyPods))
+	}
+
+	BeforeEach(func() {
+		tests.SkipIfNoCmd("kubectl")
+		tests.SkipIfSchedulableNodesLessThan(1)
```
slintes: Just curious: under which circumstances can this happen?
Or should this be 2?

pkg/virt-operator/install-strategy/create.go
```diff
@@ -1384,6 +1386,66 @@ func addOrRemoveSSC(targetStrategy *InstallStrategy,
...
+
+	injectOperatorMetadata(kv, &podDisruptionBudget.ObjectMeta, imageTag, imageRegistry)
+
+	var cachedPodDisruptionBudget *policyv1beta1.PodDisruptionBudget
+
+	pdbClient := clientset.PolicyV1beta1().PodDisruptionBudgets(deployment.Namespace)
+
+	obj, exists, _ := stores.PodDisruptionBudgetCache.Get(podDisruptionBudget)
+
+	if exists {
```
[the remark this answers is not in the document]
dhiller: You're referring to [here](https://github.com/dhiller/kubevirt/blob/865b2420cbc6134cfa45ebf21f1e0b1a8c87acee/pkg/virt-operator/install-strategy/create.go#L221) and [here](https://github.com/dhiller/kubevirt/blob/865b2420cbc6134cfa45ebf21f1e0b1a8c87acee/pkg/virt-operator/install-strategy/create.go#L158), right?

I found this less confusing to read because the whole `if exists` block applies only if the item exists in cache. I found the other blocks more confusing to read because on a closer look the pattern is like

```
obj, exists := getfromcache
if exists {
  ... cast object to use it later in other blocks
}
if !exists {
  ... do things
} else if !objectMatchesVersion {
  ... do other things
} else {
  ... do even other things
}
```

This confused me insofar as I could not really see where the casted object was used at first. So I changed it into this. 

If you think this is more confusing, I'll change it to match to the other patterns. But I think the scope of usage for `cachedPodDisruptionBudget` is mroe clear now. I could even remove the declaration from line 1397 and convert 1404 into a `:=`. What do you think?

pkg/virt-operator/kubevirt_test.go
```diff
@@ -615,9 +689,13 @@ var _ = Describe("KubeVirt Operator", func() {
 		all = append(all, components.NewPrometheusService(NAMESPACE))
 		all = append(all, components.NewApiServerService(NAMESPACE))
 		apiDeployment, _ := components.NewApiServerDeployment(NAMESPACE, registry, version, imagePullPolicy, verbosity)
+		apiDeploymentPdb := components.NewPodDisruptionBudgetForDeployment(apiDeployment)
+		injectVersionAnnotation(apiDeploymentPdb, version, registry)
```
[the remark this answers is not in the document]
dhiller: Mh, good point, will have a look.

pkg/virt-operator/kubevirt_test.go
dhiller: Yeah, right...

dhiller: Better?
```
// 3 in total are yet missing at this point
// because waiting on controller, controller's PDB and virt-handler daemonset until API server deploys successfully
```

tests/virt_control_plane_test.go
dhiller: Actually this is just a leftover. Will remove.

pkg/virt-operator/install-strategy/create.go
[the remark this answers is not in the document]
dhiller: You're referring to [here](https://github.com/dhiller/kubevirt/blob/865b2420cbc6134cfa45ebf21f1e0b1a8c87acee/pkg/virt-operator/install-strategy/create.go#L221) and [here](https://github.com/dhiller/kubevirt/blob/865b2420cbc6134cfa45ebf21f1e0b1a8c87acee/pkg/virt-operator/install-strategy/create.go#L158), right?

I found this less confusing to read because the whole `if exists` block applies only if the item exists in cache. I found the other blocks more confusing to read because on a closer look the pattern is like

```
obj, exists := getfromcache
if exists {
  ... cast object to use it later in other blocks
}
if !exists {
  ... do things
} else if !objectMatchesVersion {
  ... do other things
} else {
  ... do even other things
}
```

This confused me insofar as I could not really see where the casted object was used at first. So I changed it into this. 

If you think this is more confusing, I'll change it to match to the other patterns. But I think the scope of usage for `cachedPodDisruptionBudget` is mroe clear now. I could even remove the declaration from line 1397 and convert 1404 into a `:=`. What do you think?

pkg/virt-operator/kubevirt_test.go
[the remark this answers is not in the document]
dhiller: Don't remember where that came from honestly. Looked like a leftover, removed it, still worked. :man_shrugging:

pkg/virt-operator/kubevirt_test.go
dhiller: Yeah actually I needed this myself to understand more about what was going on. In a further step we might use this to increase the expectation granularity.

pkg/virt-operator/install-strategy/create.go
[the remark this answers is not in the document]
slintes: > You're referring to here and here, right?

Not only, the createOrUpdate* funcs use the same pattern as well.
Mh, not sure which pattern is easier to read, I have no preference on that. But IMHO we should use the same everywhere.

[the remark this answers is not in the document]
dhiller: OK, I'll change it back to the other pattern.

pkg/virt-operator/kubevirt_test.go
slintes: yes :)

pkg/virt-operator/install-strategy/create.go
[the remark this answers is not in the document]
dhiller: Done.

tests/virt_control_plane_test.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+		return countPods
+	}
+
+	getPodList := func() (*v1.PodList, error) {
+		virtCli, err := kubecli.GetKubevirtClient()
+		if err != nil {
+			return nil, err
+		}
+		podList, err := virtCli.CoreV1().Pods("kubevirt").List(metav1.ListOptions{})
```
slintes: the install namespace can change, please use the `KubeVirtInstallNamespace` var

tests/virt_control_plane_test.go
```diff
@@ -0,0 +1,265 @@
...
+		// Add label to node that was selected for test
+		if selectedNode.Labels == nil {
+			selectedNode.Labels = make(map[string]string)
+		}
+		selectedNode.Labels[labelKey] = labelValue
+		_, err = kubecli.KubevirtClient.CoreV1(virtCli).Nodes().Update(selectedNode)
+		tests.PanicOnError(err)
+
+		// Add nodeSelector to deployments so that they migrate to selectedNode
+		deploymentsClient := kubecli.KubevirtClient.AppsV1(virtCli).Deployments("kubevirt")
```
slintes: dito

tests/virt_control_plane_test.go
```diff
@@ -0,0 +1,265 @@
...
+	})
+
+	AfterEach(func() {
+		uncordonNodes(nodeNames...)
+
+		virtCli, err := kubecli.GetKubevirtClient()
+		tests.PanicOnError(err)
+
+		// Remove nodeSelector from deployments
+		deploymentsClient := kubecli.KubevirtClient.AppsV1(virtCli).Deployments("kubevirt")
```
slintes: dito

slintes:
test results look good now, nice work Daniel :+1: 
/lgtm
/approve

PR #2452, opened 2019-07-03
fix 'composite literal uses unkeyed fields' error on vmi-mutator_test.go

ahadas:
When running the tests in `pkg/virt-api/webhooks/mutating-webhook/mutators` with default settings of golang, the tests fail with the aforementioned error.

**Release note**:

```release-note
NONE
```


rmohr:
Let's wait for travis.

PR #2453, opened 2019-07-03
Move the put verb of the subresources rules from cluster roles to operator rules

slintes:

In an earlier commit the put verb was removed, because it is not needed
and even isn't a valid verb. Unfortunately that broke updates, because
the operator needs all rules of previous versions, so the put verb was
re-addeed. With this commit the put verb is moved from the cluster roles
to the operator's rules, in order to be able to remove it completely as
soon as an upgrade path from the current version v0.18 is not supported
anymore.

PR where put was removed: https://github.com/kubevirt/kubevirt/pull/2391
PR where put was re-added: https://github.com/kubevirt/kubevirt/pull/2440

/cc @davidvossel 
/cc @rmohr 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/manifests/generated/rbac-cluster.authorization.k8s.yaml.in
+++ b/manifests/generated/rbac-cluster.authorization.k8s.yaml.in
@@ -60,7 +60,6 @@ rules:
   - virtualmachines/restart
   verbs:
   - update
-  - put
 - apiGroups:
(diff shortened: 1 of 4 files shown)


slintes:
come on k8s-1.13.3, even crio is green!

/retest

PR #2464, opened 2019-07-08
Add error message in case of abnormal disconnect

dhiller:
Bugzilla # 1723878: in case of abnormal disconnect provide an error
message explaining the possible reasons (i.e. another user connected
or network issues).

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1723878

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virtctl/console/console.go
+++ b/pkg/virtctl/console/console.go
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
- * Copyright 2017, 2018 Red Hat, Inc.
+ * Copyright 2017 - 2019 Red Hat, Inc.
  *
  */
 
@@ -24,6 +24,7 @@ import (
 	"io"
 	"os"
 	"os/signal"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


slintes:
> lgtm, but Travis is unhappy, you need to run `make generate` and commit changes

ah, I think you just did :sunglasses:

pkg/virtctl/console/console.go
```diff
@@ -177,6 +178,11 @@ func (c *Console) Run(cmd *cobra.Command, args []string) error {
 	terminal.Restore(int(os.Stdin.Fd()), state)
 
 	if err != nil {
+		if strings.Contains(err.Error(), "abnormal closure") {
```
fedepaol: A more idiomatic way to check for errors is to treat them as types.
See for example what we do here https://github.com/kubevirt/kubevirt/blob/0acbdb417a0f7721fca9b73bd4be97391bd0ed54/staging/src/kubevirt.io/client-go/kubecli/vmi.go#L121

For more info, see for example https://medium.com/@sebdah/go-best-practices-error-handling-2d15e1f0c5ee 

So the if would be something like
```
	if e, ok := err.(*websocket.CloseError); ok && e.Code == websocket.CloseAbnormalClosure {

dhiller: Oh, interesting! Thanks for that!

slintes: Ah yes, nice! I always forget about this, because we don't use custom error types ourselves that much.

PR #2467, opened 2019-07-08
Increase test timeout

slintes:
Several test lanes are running in timeouts sometimes, so let's try an additional hour.
Prow config is updated to the same value already.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/functests.sh
+++ b/hack/functests.sh
@@ -43,4 +43,4 @@ if [[ ${TARGET} == openshift* ]]; then
     oc=${kubectl}
 fi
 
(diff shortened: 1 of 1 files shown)


rmohr:
/approve
/lgtm

I don't understnad though why we need to increase it so drastically lately. We only added a few tests. This looks a little bit suspicious.

slintes:
> I don't understnad though why we need to increase it so drastically lately. We only added a few tests. This looks a little bit suspicious.

Agree, this might need some investigation. However, it should not prevent merges until we find out more.
I will trigger a full test run on my server, let's see how it compares to CI servers.

slintes:
@rmohr  just compared the number of running tests ("SLOW TEST" string in logs) in a timed out and a succesful test run, the timed out only had 2 tests less. So we seem to be at least only a bit over the old 5 hours limit sometimes...

PR #2472, opened 2019-07-10
Add ability to specify SMBIOS details in kubevirt-config

vatsalparekh:
This PR adds functionality to mention cluster wide SMBios details in `kubevirt-config`, these details will further be taken into domain of the VMs
Details that can be mentioned are:- Family, Product, Manufacturer, Sku, Version

Example configMap
```yaml
apiVersion: v1
data:
  feature-gates: DataVolumes, CPUManager, LiveMigration, ExperimentalIgnitionSupport,
    Sidecar
  smbios: '{"manufacturer":"None","product":"test","version":"1.0","sku":"1.0","family":"test"}'
kind: ConfigMap
metadata:
  creationTimestamp: 2019-08-21T06:54:55Z
  labels:
    kubevirt.io: ""

```
or `smbios` can also be mentioned like this
```yaml
  smbios: |-
    manufacturer=None
    product=test
    version=2.0
    sku=1.0
    family=test
```

**Release note**:

```release-note
Added ability to specify a cluster-wide SMBios details in kubevirt-config configMap, which will be taken into all VM domains
Details that can be mentioned are:- Family, Product, Manufacturer, Sku, Version
```


pkg/virt-handler/vm.go
```diff
@@ -1386,7 +1388,7 @@ func (d *VirtualMachineController) processVmUpdate(origVMI *v1.VirtualMachineIns
 			d.recorder.Event(vmi, k8sv1.EventTypeNormal, v1.Migrating.String(), "VirtualMachineInstance is migrating.")
 		}
 	} else {
-		err = client.SyncVirtualMachine(vmi)
+		err = client.SyncVirtualMachine(vmi, smbios)
```
vladikr: Perhaps it's better to pass a struct here, like `options` which will have smbios as an attribute.
Similar to `MigrateVirtualMachine` options, on line 1377 above.

vatsalparekh: Done

slintes:
> we need a backward compatible way to pass an additional context. Can you have a look?

If I understand grpc/protobuf correct, just adding a new field to the VMIRequest message is easy to handle:

- new client (handler), old server (launcher): server will just not read the new field, and works as before adding the new feature
- old client, new server: the server needs to deal with the fact that the new field might be empty / have its default value. This needs to be done here [0] 

See [1], 2nd bullet

[0] https://github.com/kubevirt/kubevirt/pull/2472/files#diff-5cb08b427cc4fd814e525c608650acaeR173
[1] https://developers.google.com/protocol-buffers/docs/proto3#updating

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -65,7 +65,7 @@ import (
 const LibvirtLocalConnectionPort = 22222
 
 type DomainManager interface {
-	SyncVMI(*v1.VirtualMachineInstance, bool) (*api.DomainSpec, error)
+	SyncVMI(*v1.VirtualMachineInstance, bool, cmdclient.VirtualMachineOptions) (*api.DomainSpec, error)
```
vladikr: I think it'll be better to have options as a pointer

rmohr:
> Can someone explain why these SMBIOS details don't belong on the VMI's spec? I'm okay with the cluster wide kubevirt-config options, it's just unclear why we're avoiding adding this information to the VMI directly.

We can add them to the VMI too if we want. Right now it is not needed there. The background is, that OS vendors (like microsoft) check the SMBIOS details to find out which manager is managing them (vmware, openstack, ovirt, ...). This allows them to install e.g. virtio drivers automatically on boot. I guess on the long run we will have applications which will require custom values there too. I would have no objections if it is configurable on the VMI too.

pkg/virt-config/config-map.go
```diff
@@ -202,6 +212,16 @@ func setConfig(config *Config, configMap *k8sv1.ConfigMap) error {
 		}
 	}
 
+	// set smbios values if they exist
+	smbiosConfig := strings.TrimSpace(configMap.Data[SmbiosConfigKey])
+	if smbiosConfig != "" {
+		// only set values if they were specified, default  values stay intact
+		err := yaml.NewYAMLOrJSONDecoder(strings.NewReader(smbiosConfig), 1024).Decode(config.SmbiosConfig)
```
vladikr: This will be an issue as `config.SmbiosConfig` is currently `nil`. It has been initialized in the default config on line 127.

From my pov, it's better to have a default set. However, if you prefer not to have a default, you then you'll need to check `config.SmbiosConfig` is not nil and if it is you need to parse the `smbiosConfig` and set it, something like:

                        var newSMBconfig *SmbiosConfig
			err := json.Unmarshal([]byte(smbiosConfig), &newSMBconfig)
1024).Decode(newSMBconfig)
			if err != nil {
				return fmt.Errorf("somehting")
			}
			config.SmbiosConfig = newSMBconfig

pkg/virt-config/config_test.go
```diff
@@ -256,4 +256,15 @@ var _ = Describe("ConfigMap", func() {
...
+
+	table.DescribeTable("SMBIOS values from kubevirt-config", func(value string, result virtconfig.SmbiosConfig) {
+		clusterConfig, _ := testutils.NewFakeClusterConfig(&kubev1.ConfigMap{
+			Data: map[string]string{virtconfig.SmbiosConfigKey: value},
+		})
+		smbios := clusterConfig.GetSMBIOS()
+		Expect(smbios).To(Equal(result))
+	},
+		table.Entry("When no values are set, should be nil", `{"smbios":{}`, virtconfig.SmbiosConfig{}),
+		table.Entry("when values set, should equal to result", `{"smbios":{"Family":"KubeVirt","Product":"KubeVirt", "Manufacturer":"None"}}`, virtconfig.SmbiosConfig{Family: "KubeVirt", Product: "KubeVirt", Manufacturer: "None"}),
```
vladikr: the `smbios` key in your json is not necessary as you already have a key when you are setting the value `Data: map[string]string{virtconfig.SmbiosConfigKey: value},`

davidvossel:
> We can add them to the VMI too if we want. Right now it is not needed there. 

If there's any chance it ever makes sense to add these options to the VMI, we need to do it now.

If we don't do it now, then we're going to be stuck with an awkward handler->launcher communication someday where SMBIOs info is passed in both on the VMI and potentially using extra args that we'll never be able to get rid of because of backwards compatibility.

pkg/virt-launcher/virtwrap/manager.go
vatsalparekh: Added

pkg/virt-config/config-map.go
vatsalparekh: Actually, @fabiand had suggested on the card that config should Default to ("KubeVirt", "KubeVirt", None), so adding it as part of `defaultClusterConfig`

pkg/virt-config/config_test.go
vatsalparekh: corrected

pkg/virt-config/config_test.go
```diff
@@ -256,4 +257,22 @@ var _ = Describe("ConfigMap", func() {
...
+		smbiosJson, err := json.Marshal(smbios)
+		Expect(err).ToNot(HaveOccurred())
+
+		resultJson, err := json.Marshal(result)
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(string(smbiosJson)).To(BeEquivalentTo(string(resultJson)))
+	},
+		table.Entry("when values set, should equal to result", `{"Family":"test","Product":"test", "Manufacturer":"None"}`, virtconfig.SmbiosConfig{Family: "test", Product: "test", Manufacturer: "None"}),
+		table.Entry("When an invalid smbios value is set, should return default values", `{"invalid":"invalid"}`, virtconfig.SmbiosConfig{Family: "KubeVirt", Product: "None", Manufacturer: "KubeVirt"}),
```
mfranczy: When invalid smbios values are set, shouldn't we return an error instead of overwriting by default values?

tests/vmi_configuration_test.go
```diff
@@ -1857,4 +1858,47 @@ var _ = Describe("Configurations", func() {
 			})
 		})
 	})
+
+	FContext("Check SMBios with default and custom values", func() {
```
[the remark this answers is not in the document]
vatsalparekh: Was for local testing purpose, removed

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -656,6 +657,18 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
+				Value: c.SMBios.Manufacturer,
+			},
+			{
+				Name:  "family",
+				Value: c.SMBios.Family,
+			},
+			{
+				Name:  "product",
+				Value: c.SMBios.Product,
+			},
```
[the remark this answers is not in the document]
vatsalparekh: @mfranczy Corrected

tests/vmi_configuration_test.go
```diff
@@ -1857,4 +1858,47 @@ var _ = Describe("Configurations", func() {
...
+			Expect(domXml).To(ContainSubstring("<entry name='product'>None</entry>"))
+			Expect(domXml).To(ContainSubstring("<entry name='manufacturer'>KubeVirt</entry>"))
+
+		})
+
+		It("[test_id:2752]test custom SMBios values", func() {
+			// Set a custom test SMBios
+			test_smbios := &v1.VirtualMachineInstanceSMBios{Family: "test", Product: "test", Manufacturer: "None"}
+			smbiosJson, err := json.Marshal(test_smbios)
+			fmt.Println(string(smbiosJson))
```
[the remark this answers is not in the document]
vatsalparekh: removed

staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -1287,3 +1287,10 @@ const (
 const (
 	EvictionStrategyLiveMigrate EvictionStrategy = "LiveMigrate"
 )
+
+// VirtualMachineInstanceSMBios defines the SMBios configuration for VMs
+type VirtualMachineInstanceSMBios struct {
+	Family       string `json:"Family,omitempty"`
+	Product      string `json:"Product,omitempty"`
+	Manufacturer string `json:"Manufacturer,omitempty"`
+}
```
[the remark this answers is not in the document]
vatsalparekh: Added

pkg/virt-handler/cmd-client/client.go
```diff
@@ -233,9 +238,33 @@ func IsDisconnected(err error) bool {
...
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	response, err := c.v1client.SyncVirtualMachine(ctx, request)
+
+	err = handleError(err, "SyncVMI", response)
+	return err
 
+	// return c.genericSendVMICmd("SyncVMI", c.v1client.SyncVirtualMachine, vmi)
```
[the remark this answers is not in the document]
vatsalparekh: I feel it would be a overhead for `genericSendVMICmd` to expect `Options` because they're not not needed for any other operations except sync and maybe create in future.
Do you think it's a hard requirement to move it to `genericSendVMICmd`?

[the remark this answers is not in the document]
vatsalparekh: your thoughts here @vladikr @rmohr ?

pkg/virt-launcher/virtwrap/api/converter.go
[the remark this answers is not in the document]
mfranczy: Did you push changes? I cannot see the fix.

pkg/virt-config/config_test.go
vatsalparekh: We do throw error [here](https://github.com/kubevirt/kubevirt/pull/2472/files#diff-ff0b8fb5d8669fdd2f69f9351f27d341R293)
But the defaultConfig would set the value in that case
I feel that is okay behavior, but what are your suggestion here? what should we change?

pkg/virt-launcher/virtwrap/api/converter.go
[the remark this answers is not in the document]
vatsalparekh: pushed now

pkg/virt-handler/cmd-client/client.go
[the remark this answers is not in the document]
vladikr: I agree that it would be nice to extend `genericSendVMICmd`, but from my pov it's not a hard requirement at the moment. At some point, we could generalize `genericSendVMICmd` as `MigrateVirtualMachine` copies it too.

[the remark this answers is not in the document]
vladikr: Please remove the remarked function.

pkg/virt-handler/cmd-client/client.go
```diff
@@ -238,9 +238,33 @@ func IsDisconnected(err error) bool {
...
+	}
+
+	request := &cmdv1.VMIRequest{
+		Vmi: &cmdv1.VMI{
+			VmiJson: vmiJson,
+		},
+		Options: optionsJson,
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
vladikr: Please note that `genericSendVMICmd` is using a different timeout if you are coping and not extending, please use the same timeout.

vladikr: That's the reason why many tests are failing with `DeadlineExceeded`

pkg/handler-launcher-com/cmd/v1/cmd.proto
```diff
@@ -22,6 +22,7 @@ message VMI {
 
 message VMIRequest {
   VMI vmi = 1;
+  bytes options = 2;
```
slintes: I see you just did the same as for MigrationRequest, but I am wondering why we use `bytes` here?
Wouldn't it be easier to introduce a `SyncOptions` message, with a `SMBIOS` message, which contains the SMBIOS fields? I think that makes maintenance easier, as gRPC handles (de)-serialization, and new and deleted messages and fields. At the end that was the reason to introduce gRPC for handler/launcher communication.
@vladikr do you remember why we use bytes for the migration options?

vatsalparekh: I also feel having a separate `VMIOptions` struct is better here, I was just trying to follow Migrations so `bytes` was followed without any reason.

vatsalparekh: Actually, I think now I get why `byte` was used
if `options` [here](https://github.com/kubevirt/kubevirt/pull/2472/files#diff-db6785c5e48bfb0b8649e70a67e2d047R257) is not `bytes`, `optionsJson` would not be able to assign it directly
We'll either need a dumb wrapper struct on the proto side, or will need to extract fields of options like `SMBios` and have a separate `json` for it (which I think we should not do, to keep the options generic here)
So `bytes` is more easy way to pass the whole `optionsJson` from the client

pkg/virt-handler/cmd-client/client.go
vatsalparekh: Corrected

fabiand:
Maybe worth checking what is wrong with the tests?

pkg/handler-launcher-com/cmd/v1/cmd.proto
slintes: Mh, tbh I do not see why we need bytes and json. We could just use the struct generated for gRPC (will be in in cmd.pb.go as soon as you declare the messages and run make generate) in vm.go instead of the struct you are defining yourself atm in types.go.

fabiand:
Please adjust the title and release note to mention which details can be provided.

vatsalparekh:
@fabiand updated description now

pkg/virt-config/config-map.go
```diff
@@ -252,6 +260,16 @@ func (c *ClusterConfig) SetConfigModifiedCallback(cb ConfigModifiedFn) {
 	go c.configModifiedCallback()
 }
 
+// SmbiosConfig struct to take the mentioned values from Kubevirt-Config
+// and same values are synced with VMs
+type SmbiosConfig struct {
```
slintes: I wondering if we can/should use the generated struct here as well? It also has the json tags, so it shoiuld work oyt of the box.

pkg/virt-handler/cmd-client/client.go
```diff
@@ -238,9 +238,28 @@ func IsDisconnected(err error) bool {
...
+	vmiJson, err := json.Marshal(vmi)
+	if err != nil {
+		return err
+	}
+
+	request := &cmdv1.VMIRequest{
+		Vmi: &cmdv1.VMI{
+			VmiJson: vmiJson,
+		},
+		Options: &cmdv1.VirtualMachineOptions{
```
slintes: 2 things here:
1. You already have the correct options type here now, no need to reconsctruct it
2. We now have some code duplication with genericSendVMICmd: can you please add the options arg to that func, and let the other users of that func just pass nil to it?

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -905,6 +906,7 @@ func (l *LibvirtDomainManager) SyncVMI(vmi *v1.VirtualMachineInstance, useEmulat
 		IsBlockPVC:     isBlockPVCMap,
 		DiskType:       diskInfo,
 		SRIOVDevices:   getSRIOVPCIAddresses(vmi.Spec.Domain.Devices.Interfaces),
+		SMBios:         options.VirtualMachineSMBios,
```
slintes: At least in theory options can be nil here, please catch that. It can happen when the request is coming from an older virt-handler. That _should_ not happen, because handler _should_ be updated before launcher, but "should" doesn't mean it will never happen.

pkg/virt-handler/cmd-client/client.go
mfranczy: > We now have some code duplication with genericSendVMICmd: can you please add the options arg to that func, and let the other users of that func just pass nil to it?

It seems I am not the only one who wants to fix it here :) https://github.com/kubevirt/kubevirt/pull/2472#discussion_r310978526 as @vladikr mentioned we should refactor ```MigrateVirtualMachine``` as well.

pkg/virt-config/config-map.go
vatsalparekh: I just confirmed that we can, so we should :)
Changed this.

pkg/virt-handler/cmd-client/client.go
vatsalparekh: > 2 things here:
> 
> 1. You already have the correct options type here now, no need to reconsctruct it
Corrected.
> 2. We now have some code duplication with genericSendVMICmd: can you please add the options arg to that func, and let the other users of that func just pass nil to it?

slintes: @mfranczy ha, missed that :D But I would be fine with skipping MigrateVirtualMachine for now, as it uses another request type (MigrationRequest vs VMIRequest)

vatsalparekh: > 2 things here:
> 
> 1. You already have the correct options type here now, no need to reconsctruct it
> 2. We now have some code duplication with genericSendVMICmd: can you please add the options arg to that func, and let the other users of that func just pass nil to it?

@slintes @mfranczy @vladikr 
Not touching Migration atm, but corrected it with an additional arg in genericSendVMICmd now

pkg/virt-launcher/virtwrap/manager.go
vatsalparekh: I agree with the idea, but we already check that in the converter here https://github.com/kubevirt/kubevirt/pull/2472/files#diff-caee6d384498fde52fde15891ac547aaR691
@slintes Do you think we need additional check here?

slintes: converter is unrelated. I mean that `options` can be nil here, so options.VirtualMachineSMBios will fail during runtime

vladikr: I agree it's probably better to verify. At the moment there is no scenario where we don't populate the options, but it may change in the future.

vatsalparekh: Added

slintes:
nice!

/lgtm
/approve

PR #2475, opened 2019-07-11
virtctl image-upload: add support for creating BlockVolume PVCs

nixpanic:

Add a --block-volume option to create the new PVC with VolumeMode=Block.
This can then be used to provision PVCs that can be used for
live-migration on raw block with RWX access-mode.

For example, here with a Ceph-CSI storageclass named "csi-rbd":

```
  $ virtctl image-upload \
        --access-mode=ReadWriteMany --block-volume \
        --pvc-name=fedora-30-gold-block \
        --pvc-size=6Gi --storage-class=csi-rbd \
        --image-path=Fedora-Cloud-Base-30-1.2.x86_64.raw
```

The `--block-volume` option is a boolean, but maybe there is interest in `--volume-mode=Block` to match the PVC.Spec closer? That feels less user-friendly to me though.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/virtctl/imageupload/imageupload.go
+++ b/pkg/virtctl/imageupload/imageupload.go
@@ -69,6 +69,7 @@ var (
 
 	uploadPodWaitSecs uint = 60
 	accessMode             = "ReadWriteOnce"
+	blockVolume            = false
 	noCreate               = false
 )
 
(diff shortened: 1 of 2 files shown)


rmohr:
> The `--block-volume` option is a boolean, but maybe there is interest in `--volume-mode=Block` to match the PVC.Spec closer? That feels less user-friendly to me though.

If it turns out that we want it, we can still add it and catch invalid mixing on arg validation.

pkg/virtctl/imageupload/imageupload.go
```diff
@@ -374,6 +376,11 @@ func createUploadPVC(client kubernetes.Interface, namespace, name, size, storage
 		pvc.Spec.AccessModes = []v1.PersistentVolumeAccessMode{v1.PersistentVolumeAccessMode(accessMode)}
 	}
 
+	if blockVolume {
+		volMode := v1.PersistentVolumeBlock
+		pvc.Spec.VolumeMode = &volMode
```
rmohr: Can you add a simple test in `tests/imageupload_test.go` to verify that the PVC is created the  way we want it?

mhenriks:
Sorry for sleeping on this.  lgtm

pkg/virtctl/imageupload/imageupload.go
nixpanic: Of course! That is done now too.

PR #2476, opened 2019-07-11
[POC] Spice connection

SchSeba:
This PR is a POC to use spice connection.
The spice connection is using the traditional spice protocol without a websocket connection and a proxy in the virtctl binary

Workflow:
1. Start a vm
2. The relevant virt-handler create a unix to tcp proxy on a specific port from a range and update the spiceHandler data with virt-handler pod ip and the created port.
3. the user runs `virtctl spice <vmi-name>`
4. virt api configure a token for the spice connection (spice password) and also the requested time (for token expiration)
5. virt-api find the spice-service pointing the virt-spice pods and take the port and ip of the created loadbalancer type service
6. the user use the vv file that was created after running the `virtctl spice` command (`remote-viewer <vmi>.vv)
7. a connection to one of the virt-spice is created, the virt-spice get the token and search for the related vmi object (using the token as a label for the vmi object)
8. virt-spice check the expiration and get the spiceHandler data (virt-handler ip and port) for the specific vmi
9. virt-handler proxy the traffic from the tcp connection from the virt-spice to the unix socket related to a qemu process

Still need to add:

- [ ] Functional test
- [ ] Provider with support service type loadbalancer

**Release note**:

```release-note
Add spice connection
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/2476)


pkg/virt-handler/spice-proxy/spice-proxy.go
```diff
@@ -0,0 +1,224 @@
...
+	fdChan         chan net.Conn
+	listener       net.Listener
+}
+
+func NewSpiceProxyManager(podIsolationDetector isolation.PodIsolationDetector) UnixToTcpProxyManager {
+	return &spiceProxyManager{proxies: make(map[string]*spiceProxy), allocatedPort: make(map[int32]bool), podIsolationDetector: podIsolationDetector}
+}
+
+func (spm *spiceProxyManager) StartListener(vmNamespace, vmName string, vmi *v1.VirtualMachineInstance) error {
+	spm.managerLock.Lock()
```
fedepaol: You can embed the lock struct in spiceProxyManager struct without the name and use `spm.Lock()`

pkg/virt-handler/spice-proxy/spice-proxy.go
```diff
@@ -0,0 +1,224 @@
...
+	defer spm.managerLock.Unlock()
+
+	if _, isExist := spm.proxies[uniqueVmName(vmNamespace, vmName)]; isExist {
```
fedepaol: can `isExist` be just `exists`? Or even `ok`? The scope is so narrow that the name does not have to be long

pkg/virt-handler/spice-proxy/spice-proxy.go
```diff
@@ -0,0 +1,224 @@
...
+		return err
+	}
+
+	res, err := spm.podIsolationDetector.Detect(vmi)
+	if err != nil {
+		return err
+	}
+
+	// Get the libvirt connection socket file on the destination pod.
+	spiceSocketFile := fmt.Sprintf("/proc/%d/root/var/run/kubevirt-private/%s/virt-spice", res.Pid(), vmi.UID)
```
fedepaol: You could use path.Join here https://golang.org/pkg/path/#Join

pkg/virt-handler/spice-proxy/spice-proxy.go
rmohr: Normally `exists` for checking if a key exists, `ok` if a type conversion works?

fedepaol: I happened to see ok also for checking if a key exists, see for example https://blog.golang.org/go-maps-in-action .

rmohr: Sure, I would never ask to change it. More a tendency.

fedepaol: Agree, would have preferred `exists` over `isExist` though

fabiand:
@SchSeba can the flow also be achieved with VNC?

pkg/virt-api/rest/subresource.go
```diff
@@ -417,6 +420,66 @@ func (app *SubresourceAPIApp) StopVMRequestHandler(request *restful.Request, res
...
+	}
+
+	if !vmi.IsRunning() {
+		response.WriteError(http.StatusBadRequest, goerror.New(fmt.Sprintf("Unable to connect to VirtualMachineInstance because phase is %s instead of %s", vmi.Status.Phase, v1.Running)))
+		return
+	}
+
+	// should not occur if the vm is handled by the virt-handler and in running state
+	if vmi.Status.SpiceConnection == nil {
+		response.WriteError(http.StatusInternalServerError, goerror.New(fmt.Sprintf("spice connection is not configured")))
```
phoracek: This starts with lower-case, while the previous error starts with upper case, please keep it consistent. Also, if this block should never happen, maybe we should state it in the error message? "spice connection is not configured" doesn't sound that fatal.

api/openapi-spec/swagger.json
```diff
@@ -5644,6 +5679,42 @@
...
+     "port": {
+      "type": "integer",
+      "format": "int32"
+     }
+    }
+   },
+   "v1.SpiceToken": {
+    "description": "SpiceToken represents the object for a spice connection token",
+    "properties": {
+     "exparationTime": {
```
phoracek: expiration

pkg/virt-handler/spice-proxy/spice-proxy.go
```diff
@@ -0,0 +1,224 @@
...
+ ATTENTION: Rerun code generators when interface signatures are modified.
+*/
+
+import (
+	"fmt"
+	"io"
+	"net"
+	"sync"
+
+	v1 "kubevirt.io/client-go/api/v1"
```
phoracek: maybe missing the logic behind this. why do we have our own client-go? why is it empty https://github.com/kubevirt/client-go

pkg/virt-handler/spice-proxy/spice-proxy.go
```diff
@@ -0,0 +1,224 @@
...
+	"kubevirt.io/client-go/log"
+	"kubevirt.io/kubevirt/pkg/virt-handler/isolation"
+)
+
+var (
+	proxyRange = []int{30000, 32000}
```
phoracek: Are we sure 2000 connections will be enough? Not sure about the scale we aim for.

pkg/virt-handler/spice-proxy/spice-proxy.go
slintes: @phoracek we want to make our client-go consumable by 3rd party projects without the need to use the import the complete kubevirt repo.

For that we
- moved the client into the `/staging` dir
- created a symlink from the `/vendor` dir to the `/staging` dir (that's what's being used here)
- will sync `staging` into the empty repo you found, not done yet

pkg/virt-handler/spice-proxy/spice-proxy.go
SchSeba: `2000` ports mean 2000 vms on the same host I can of course change it to higher number if required.

maybe make it configurable via the virt-config?

SchSeba:
@fabiand  thanks for the comment.

> @SchSeba can the flow also be achieved with VNC?

Yes the solution can also work for vnc just need to change to code in the virt-spice proxy to a vnc proxy.

I start with a vnc but change to spice because the only information I can pass in the session for both vnc and spice is the token. 
I use the token to find the requested vm for the connect and the problem with vnc is that the max len of the token field is 8 characters
With spice you have 32 this makes the token harder to discover in a brute force attempt.

api/openapi-spec/swagger.json
SchSeba: thanks! done

pkg/virt-api/rest/subresource.go
SchSeba: it's fatal because `should not occur if the vm is handled `

api/openapi-spec/swagger.json
```diff
@@ -6411,6 +6482,9 @@
    },
    "v1.VirtualMachineInstanceStatus": {
     "description": "VirtualMachineInstanceStatus represents information about the status of a VirtualMachineInstance. Status may trail the actual\nstate of a system.",
+    "required": [
+     "spiceConnection"
```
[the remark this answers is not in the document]
SchSeba: done thanks!

pkg/virt-handler/spice-proxy/spice-proxy.go
SchSeba: done I still prefer the `fmt.sprintf` easy to understand

pkg/virt-handler/spice-proxy/spice-proxy.go
SchSeba: done

pkg/virt-handler/spice-proxy/spice-proxy.go
SchSeba: done

pkg/virt-handler/spice-proxy/spice-proxy.go
fabiand: Couldn't the handler already use a websocket mechanism?
WOuldn#t this allow us to open just a single port?

PR #2484, opened 2019-07-12
Update vendored CDI version to 1.10.6

awels:

Update the version of CDI to 1.10.6

Fixes #

Our client go is now updated as well.
**Release note**:

```release-note
NONE
```


awels:
/assign @rmohr 
This updates to latest released CDI version.

awels:
So it looks like I didn't do something right. I updated the go.mod and ran make deps-update and make generate, what else am I missing. Some of the generated files don't seem updated and that is what is making the unit tests fail. Among other things.

slintes:
looks strange, "build failed" in a unit test? I retriggered travis, let's see if it wass hickup only

awels:
No I can reproduce failing unit tests locally, we updated some of the interfaces to have an extra parameter, but something isn't updating right.

slintes:
ah, I see: we use different version of k8s client-go, CDI uses 1.12.4, KubeVirt 1.13.4. Some functions in the testing package changed in between obviously

slintes:
this came with k8s 1.13.0: https://github.com/kubernetes/client-go/commit/41406bf985e3fa3c5eccd0a7c1fb39e9212340e8

rmohr:
/hold

client-go version issues need to be sorted out.

rmohr:
@awels I think we still use 1.8.1. Would you mind rebasing the PR?  Would be great to update to latest CDI.

awels:
> @awels I think we still use 1.8.1. Would you mind rebasing the PR? Would be great to update to latest CDI.

Looks like some mixture of 1.8.1 and 1.10.1, updated to 1.10.6

awels:
@rmohr I ran make deps-update and make generate, anything else I would need to do? looks like travis is not happy.

slintes:
some unit tests are failing, on quick look this can only be caused by the updated `github.com/emicklei/go-restful` dependency?

```
[Fail] Put REST call with missing Content-Type header [It] should return 415 
/root/go/src/kubevirt.io/kubevirt/pkg/rest/endpoints/put_test.go:113
[Fail] Post REST call with missing Content-Type header [It] should return 414 
/root/go/src/kubevirt.io/kubevirt/pkg/rest/endpoints/post_test.go:81
[Fail] Patch REST call with missing Content-Type header [It] should return 415 
/root/go/src/kubevirt.io/kubevirt/pkg/rest/endpoints/patch_test.go:90
```
(note the description of the 2nd test is wrong, should also be 415)

awels:
Okay reverted that depedency update.

awels:
reverted all deps updates except the ones needed for CDI, I think one unit test will still fail. But not sure why that is.

slintes:
> Also the client-go version mismatch should be solved, as we are on 1.13.4 now in CDI

:+1: 

> reverted all deps updates except the ones needed for CDI

hehe, so the next one running deps update will run into this again :wink: 
but works for me, someone of our team can deal with it.

/hold cancel
/lgtm
/approve
/retest

slintes:
/hold

checking something... (having CDI dependency in main go.mod vs client-go go.mod)

slintes:
We will do a follow for the deps.
Thanks Alex!

/hold cancel
/retest

PR #2498, opened 2019-07-17
Fixed failure of networking tests.

yossisegev:
The default MTU on the pod network interface varies, depending
on the environment on which the cluster runs, and it made some
of the tests to fail.
Now the test covers all the possible MTU values.

### DIFF
--- a/tests/vmi_networking_test.go
+++ b/tests/vmi_networking_test.go
@@ -164,11 +164,6 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 				tests.SkipIfOpenShift("Custom MAC addresses on pod networks are not supported")
 			}
 
-			// assuming pod network is of standard MTU = 1500 (minus 50 bytes for vxlan overhead)
-			expectedMtu := 1450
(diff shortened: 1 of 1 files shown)


SchSeba:
@yossisegev please run the `make` command to make travis happy again :)

yossisegev:
ci please test

tests/vmi_networking_test.go
```diff
@@ -190,9 +188,25 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
-			Expect(strings.Contains(output, expectedMtuString)).To(BeTrue())
+
+			// The MTU of the Pod network varies, depending on the environment in use. We want to
+			// verify that the actual MTU is from the possible range, which is {1500, 1450}, minus
+			//  50 bytes for vxlan overhead.
+			possibleMtus := []int{1400, 1450}
+			mtuMatch := false
+			expectedMtuString := ""
+
+			for _, expectedMtu := range possibleMtus {
```
rmohr: I guess we could find that out once and don't do it every time, but the change is great.

PR #2501, opened 2019-07-17
Telemetry metrics

fromanirh:

Add VMI count metrics. This is to be used first for the openshift telemeter integration (https://github.com/openshift/telemeter/pull/194), but in general we want to expose higher level metrics to be consumed by other components

Fixes N/A

I'm not 100% satisfied by the design, so comments are welcome. Is virt-handler the best place?
Do we want to differentiate between higher level (like this one) and lower level (like, say, the storage I/P metrics)?

```release-note
Add metric to report the VMI count.
```

### DIFF
--- a/pkg/monitoring/vms/prometheus/prometheus.go
+++ b/pkg/monitoring/vms/prometheus/prometheus.go
@@ -50,10 +50,10 @@ var (
 
 	// higher-level, telemetry-friendly metrics
 	vmiCountDesc = prometheus.NewDesc(
-		"kubevirt_vmi_status_count",
-		"VMI status.",
+		"kubevirt_vmi_phase_count",
+		"VMI phase.",
 		[]string{
-			"node", "status",
+			"node", "phase",
 		},
 		nil,
 	)
@@ -376,17 +376,17 @@ func updateNetwork(vmi *k6tv1.VirtualMachineInstance, vmStats *stats.DomainStats
 }
 
 func updateTelemetry(nodeName string, vmis []*k6tv1.VirtualMachineInstance, ch chan<- prometheus.Metric) {
-	statusMap := make(map[string]uint64)
+	phasesMap := make(map[string]uint64)
 
 	for _, vmi := range vmis {
-		statusMap[strings.ToLower(string(vmi.Status.Phase))] += 1
+		phasesMap[strings.ToLower(string(vmi.Status.Phase))] += 1
 	}
 
-	for status, count := range statusMap {
+	for phase, count := range phasesMap {
 		mv, err := prometheus.NewConstMetric(
 			vmiCountDesc, prometheus.GaugeValue,
 			float64(count),
-			nodeName, status,
+			nodeName, phase,
 		)
 		if err != nil {
 			continue
--- a/pkg/monitoring/vms/prometheus/prometheus.go
+++ b/pkg/monitoring/vms/prometheus/prometheus.go
@@ -375,12 +375,18 @@ func updateNetwork(vmi *k6tv1.VirtualMachineInstance, vmStats *stats.DomainStats
 	}
 }
 
-func updateVMIsPhase(nodeName string, vmis []*k6tv1.VirtualMachineInstance, ch chan<- prometheus.Metric) {
+func makeVMIsPhasesMap(vmis []*k6tv1.VirtualMachineInstance) map[string]uint64 {
 	phasesMap := make(map[string]uint64)
 
 	for _, vmi := range vmis {
 		phasesMap[strings.ToLower(string(vmi.Status.Phase))] += 1
 	}
+
+	return phasesMap
+}
+
+func updateVMIsPhase(nodeName string, vmis []*k6tv1.VirtualMachineInstance, ch chan<- prometheus.Metric) {
+	phasesMap := makeVMIsPhasesMap(vmis)
 
 	for phase, count := range phasesMap {
 		mv, err := prometheus.NewConstMetric(
--- a/pkg/monitoring/vms/prometheus/prometheus.go
+++ b/pkg/monitoring/vms/prometheus/prometheus.go
@@ -375,7 +375,7 @@ func updateNetwork(vmi *k6tv1.VirtualMachineInstance, vmStats *stats.DomainStats
 	}
 }
 
-func updateTelemetry(nodeName string, vmis []*k6tv1.VirtualMachineInstance, ch chan<- prometheus.Metric) {
+func updateVMIsPhase(nodeName string, vmis []*k6tv1.VirtualMachineInstance, ch chan<- prometheus.Metric) {
 	phasesMap := make(map[string]uint64)
 
 	for _, vmi := range vmis {
@@ -469,7 +469,7 @@ func (co *Collector) Collect(ch chan<- prometheus.Metric) {
 	scraper := &prometheusScraper{ch: ch}
 	co.concCollector.Collect(socketToVMIs, scraper, collectionTimeout)
 
-	updateTelemetry(co.nodeName, vmis, ch)
+	updateVMIsPhase(co.nodeName, vmis, ch)
 	return
 }
 
--- a/pkg/monitoring/vms/prometheus/prometheus.go
+++ b/pkg/monitoring/vms/prometheus/prometheus.go
@@ -22,6 +22,7 @@ package prometheus
 import (
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -47,6 +48,17 @@ var (
 		nil,
 	)
 
+	// higher-level, telemetry-friendly metrics
+	vmiCountDesc = prometheus.NewDesc(
+		"kubevirt_vmi_status_count",
+		"VMI status.",
+		[]string{
+			"node", "status",
+		},
+		nil,
+	)
+
+	// lower level metrics
 	storageIopsDesc = prometheus.NewDesc(
 		"kubevirt_vm_storage_iops_total",
 		"I/O operation performed.",
@@ -363,6 +375,26 @@ func updateNetwork(vmi *k6tv1.VirtualMachineInstance, vmStats *stats.DomainStats
 	}
 }
 
+func updateTelemetry(nodeName string, vmis []*k6tv1.VirtualMachineInstance, ch chan<- prometheus.Metric) {
+	statusMap := make(map[string]uint64)
+
+	for _, vmi := range vmis {
+		statusMap[strings.ToLower(string(vmi.Status.Phase))] += 1
+	}
+
+	for status, count := range statusMap {
+		mv, err := prometheus.NewConstMetric(
+			vmiCountDesc, prometheus.GaugeValue,
+			float64(count),
+			nodeName, status,
+		)
+		if err != nil {
+			continue
+		}
+		ch <- mv
+	}
+}
+
 func updateVersion(ch chan<- prometheus.Metric) {
 	verinfo := version.Get()
 	ch <- prometheus.MustNewConstMetric(
@@ -436,6 +468,8 @@ func (co *Collector) Collect(ch chan<- prometheus.Metric) {
 	socketToVMIs := newvmiSocketMapFromVMIs(co.virtShareDir, vmis)
 	scraper := &prometheusScraper{ch: ch}
 	co.concCollector.Collect(socketToVMIs, scraper, collectionTimeout)
+
+	updateTelemetry(co.nodeName, vmis, ch)
 	return
 }
 
--- a/tests/infra_test.go
+++ b/tests/infra_test.go
@@ -306,6 +306,19 @@ var _ = Describe("Infrastructure", func() {
 				Expect(key).To(ContainSubstring(fmt.Sprintf("name=\"%s\"", vmi.Name)))
 			}
 		}, 300)
+
+		It("should include VMI telemetry for a running VM", func() {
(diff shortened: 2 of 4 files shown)


pkg/monitoring/vms/prometheus/prometheus.go
```diff
@@ -47,6 +48,17 @@ var (
 		nil,
 	)
 
+	// higher-level, telemetry-friendly metrics
+	vmiCountDesc = prometheus.NewDesc(
+		"kubevirt_vmi_status_count",
+		"VMI status.",
+		[]string{
+			"node", "status",
```
rmohr: I think you meant `phase` and not `status`.

tests/infra_test.go
```diff
@@ -289,11 +289,9 @@ var _ = Describe("Infrastructure", func() {
 			By("Checking the collected metrics")
 			keys := getKeysFromMetrics(metrics)
 			for _, key := range keys {
-				if strings.HasPrefix(key, "kubevirt_vm_metrics_") {
```
rmohr: Why do we not need the if statement anymore?

tests/infra_test.go
```diff
@@ -303,9 +301,22 @@ var _ = Describe("Infrastructure", func() {
...
-				Expect(key).To(ContainSubstring(fmt.Sprintf("namespace=%s", vmi.Namespace)))
-				Expect(key).To(ContainSubstring(fmt.Sprintf("name=%s", vmi.Name)))
+				Expect(key).To(ContainSubstring(fmt.Sprintf("node=\"%s\"", nodeName)))
+				Expect(key).To(ContainSubstring(fmt.Sprintf("namespace=\"%s\"", vmi.Namespace)))
+				Expect(key).To(ContainSubstring(fmt.Sprintf("name=\"%s\"", vmi.Name)))
+			}
+		}, 300)
+
+		It("should include VMI telemetry for a running VM", func() {
+			_, _, _, metrics := newRandomVMIWithMetrics(virtClient, "kubevirt_vmi_")
```
rmohr: Do we now expose kubevirt_vm and kubevirt_vmi metrics?

pkg/monitoring/vms/prometheus/prometheus.go
fromanirh: yep, good catch

tests/infra_test.go
fromanirh: It is now guaranteed (or at least that's the intention) by newRandomVMIWithMetrics, see below. We now take metrics with the user-supplied prefix, which is a stronger guarantee because is going to be more specific than the "kubevirt" prefix we used previously.

tests/infra_test.go
fromanirh: That's open for discussion. I don't have strong opinions here. What do you prefer?

rmohr: imo, if we talks about vmis, it should be about kubevirt_vmi. So if we had `kubevirt_vm` before, then I missed it on the reviews. ;)

fromanirh: No problem renaming the metrics. Is this a breaking change compatibility-wise?

rmohr: Probably better changing it now where most likely no-one uses them yet. Can you ask on kubevirt-dev and #virtualization in the k8s slack? @fabiand I am concerned that we will create a mess on the long run if we are not exactly defining `VMI` related metrics with `vmi`.

fromanirh: Filed https://github.com/kubevirt/kubevirt/issues/2504 to track this issue

fabiand: +1 on doing all metric renames that are necessary.

fromanirh: Discussion started on kubevirt-dev and on slack. Posted trivial https://github.com/kubevirt/kubevirt/pull/2506

pkg/monitoring/vms/prometheus/prometheus.go
```diff
@@ -363,6 +375,26 @@ func updateNetwork(vmi *k6tv1.VirtualMachineInstance, vmStats *stats.DomainStats
...
+	}
+
+	for phase, count := range phasesMap {
+		mv, err := prometheus.NewConstMetric(
+			vmiCountDesc, prometheus.GaugeValue,
+			float64(count),
+			nodeName, phase,
+		)
+		if err != nil {
+			continue
```
qinqon: Can add some logging here ?

tests/infra_test.go
```diff
@@ -303,9 +301,22 @@ var _ = Describe("Infrastructure", func() {
 			keys := getKeysFromMetrics(metrics)
 			for _, key := range keys {
 				// we don't care about the ordering of the labels
-				Expect(key).To(ContainSubstring(fmt.Sprintf("node=%s", nodeName)))
-				Expect(key).To(ContainSubstring(fmt.Sprintf("namespace=%s", vmi.Namespace)))
-				Expect(key).To(ContainSubstring(fmt.Sprintf("name=%s", vmi.Name)))
+				Expect(key).To(ContainSubstring(fmt.Sprintf("node=\"%s\"", nodeName)))
```
qinqon: maybe you can use ` so no need for escape \"

qinqon: You can simplify this taking advance of ContaineSubstring alredy do fmt.Sprintf for you and gomega composition with SatisfyAll
Expect(key).To(SatisfyAll(ContainSubstring(`node="%s"`, nodeName), ContainSubstring...)

tests/infra_test.go
```diff
@@ -303,9 +301,22 @@ var _ = Describe("Infrastructure", func() {
...
+
+		It("should include VMI telemetry for a running VM", func() {
+			_, _, _, metrics := newRandomVMIWithMetrics(virtClient, "kubevirt_vmi_")
+			By("Checking the collected metrics")
+			keys := getKeysFromMetrics(metrics)
+
+			for _, key := range keys {
+				if strings.Contains(key, "phase=\"running\"") {
+					value := metrics[key]
+					Expect(value).To(BeNumerically("==", float64(1.0)))
```
qinqon: To(Equal is enough

tests/infra_test.go
```diff
@@ -303,9 +301,22 @@ var _ = Describe("Infrastructure", func() {
...
-				Expect(key).To(ContainSubstring(fmt.Sprintf("node=%s", nodeName)))
-				Expect(key).To(ContainSubstring(fmt.Sprintf("namespace=%s", vmi.Namespace)))
-				Expect(key).To(ContainSubstring(fmt.Sprintf("name=%s", vmi.Name)))
+				Expect(key).To(ContainSubstring(fmt.Sprintf("node=\"%s\"", nodeName)))
+				Expect(key).To(ContainSubstring(fmt.Sprintf("namespace=\"%s\"", vmi.Namespace)))
+				Expect(key).To(ContainSubstring(fmt.Sprintf("name=\"%s\"", vmi.Name)))
+			}
+		}, 300)
+
+		It("should include VMI telemetry for a running VM", func() {
```
qinqon: Can we check more than one VMIs so we check that the VMIs loop is working fine too ?

tests/infra_test.go
```diff
@@ -443,13 +454,20 @@ func newRandomVMIWithMetrics(virtClient kubecli.KubevirtClient, metricSubstring
...
+		metrics, err = parseMetricsToMap(lines)
 		Expect(err).ToNot(HaveOccurred())
 		return metrics
-	}, 30*time.Second, 2*time.Second).Should(HaveKey(ContainSubstring(metricSubstring)))
+	}, 30*time.Second, 2*time.Second).Should(HaveKey(ContainSubstring(metricPrefix)))
+
+	// troubleshooting helper
+	fmt.Fprintf(GinkgoWriter, "metrics [%s]:\nlines=%s\n%#v\n", metricPrefix, lines, metrics)
+	Expect(len(metrics)).To(BeNumerically(">=", float64(1.0)))
+	Expect(len(metrics)).To(Equal(len(lines)))
```
qinqon: Expect(metrics).To(HaveLen(len(lines))

pkg/monitoring/vms/prometheus/prometheus.go
fromanirh: We can, but is a tough call. The `prometheus.NewConstMetric` is supposed to never fail, so failure is exceptional and would deserve a log. But if this happen repeteadly, we can spam the logs, and this is something we want to avoid.

tests/infra_test.go
fromanirh: ack, on it

tests/infra_test.go
fromanirh: Done

tests/infra_test.go
fromanirh: It is a good idea, but I'd like to extend this to other metrics-related checks, so I'll do in another PR.

fromanirh: Here: https://github.com/kubevirt/kubevirt/pull/2512

pkg/monitoring/vms/prometheus/prometheus.go
qinqon: So maybe debug level ?

tests/infra_test.go
qinqon: ack, thanks!

tests/infra_test.go
```diff
@@ -301,9 +301,11 @@ var _ = Describe("Infrastructure", func() {
 			keys := getKeysFromMetrics(metrics)
 			for _, key := range keys {
 				// we don't care about the ordering of the labels
-				Expect(key).To(ContainSubstring(fmt.Sprintf("node=\"%s\"", nodeName)))
-				Expect(key).To(ContainSubstring(fmt.Sprintf("namespace=\"%s\"", vmi.Namespace)))
-				Expect(key).To(ContainSubstring(fmt.Sprintf("name=\"%s\"", vmi.Name)))
+				Expect(key).To(SatisfyAll(
+					ContainSubstring("node=\"%s\"", nodeName),
```
qinqon: What about `node="%"` so we don't have to escape ?

qinqon: I mean using golang raw string literal so we don't have to escape stuff.

pkg/monitoring/vms/prometheus/prometheus.go
fromanirh: Uhm, good point. This maybe another case on which better to change all the occurrence in one go. If so, I'll file a new PR and link it here. Otherwise I'll add another commit.

tests/infra_test.go
fromanirh: Yes, just slipped out from the past upload. Fixed now. Thanks!

pkg/monitoring/vms/prometheus/prometheus.go
fromanirh: I decided to tackle all the cases at once: https://github.com/kubevirt/kubevirt/pull/2523

fromanirh:
OK, enough retries. I'm pretty much persuaded the failures are unrelated. Failures:
* pull-kubevirt-e2e-k8s-1.11.0:
  - Creating a VirtualMachineInstance with node tainted [test_id:1635]the vmi with tolerations should be scheduled
* pull-kubevirt-e2e-os-3.11.0-crio:
- Networking VirtualMachineInstance with TX offload disabled
- should honor auto ioThreadPolicy
- VM Live Migration Starting a VirtualMachineInstance with an Fedora shared NFS PVC, cloud init and service account
- VM Live Migration Starting a VirtualMachineInstance migration security
* pull-kubevirt-e2e-os-3.11.0-multus: cluster-up failed

fromanirh:
wrong button!

pkg/monitoring/vms/prometheus/prometheus.go
```diff
@@ -363,6 +375,26 @@ func updateNetwork(vmi *k6tv1.VirtualMachineInstance, vmStats *stats.DomainStats
 	}
 }
 
+func updateTelemetry(nodeName string, vmis []*k6tv1.VirtualMachineInstance, ch chan<- prometheus.Metric) {
```
rmohr: I don't think that this is `telemetry` specific. In general these things are or will be valuable for monitoring. No change required.

pkg/monitoring/vms/prometheus/prometheus.go
```diff
@@ -363,6 +375,26 @@ func updateNetwork(vmi *k6tv1.VirtualMachineInstance, vmStats *stats.DomainStats
...
+	phasesMap := make(map[string]uint64)
+
+	for _, vmi := range vmis {
+		phasesMap[strings.ToLower(string(vmi.Status.Phase))] += 1
+	}
+
+	for phase, count := range phasesMap {
```
rmohr: There seem to be no explicit unit tests to cover that summing them per status works.

pkg/monitoring/vms/prometheus/prometheus.go
fromanirh: Not sure I fully understand your point, but I surely agree this is not Telemetry specific, so I'll rename this function, or squash in the calling site.

pkg/monitoring/vms/prometheus/prometheus.go
fromanirh: Same point @qinqon raised. I wanted to fix all the tests running multiple VMIs (https://github.com/kubevirt/kubevirt/pull/2512) but looks like I bited more than I can chew :) I'll quickly fix this test and keep working on #2512 to improve the testsuite later.

rmohr: I am more concerned about the missing unit tests here, not so much about the functional tests.

fromanirh: ack, adding

PR #2517, opened 2019-07-22
Added root go file for client-go for dep dependency manager

slintes:
Added root go file for client-go for `dep` dependency manager

Just the same as we already have for kubevirt's root dir.
Already done on client-go v0.19.0 and tested on HCO

**Release note**:
```release-note
NONE
```


rmohr:
/hold

@slintes I think it needs a `make generate` when looking on the travis error.

slintes:
> @slintes I think it needs a `make generate` when looking on the travis error.

yes, new BUILD.bazel file was missing...

slintes:
Travis is happy now

/hold cancel

PR #2519, opened 2019-07-22
Remove pod/exec privileges from virt-api

ahadas:
This PR mainly removes the `exe`part that executes sock-connector from virt-api. Alternatively, virt-api connects to a REST endpoint exposed by virt-handler via a websocket and virt-handler proxies that to the corresponding unix socket.

In addition, this PR slightly improves the console connection by reading 10226 bytes rather than reading 10240 and then sending via two messages (one of size 10226 and one of size 14). This adds another read on every 731 reads but may cut half of the amount of send calls.

**Release note**:

```release-note
NONE
```


pkg/virt-api/rest/subresource.go
```diff
@@ -107,49 +96,141 @@ func (app *SubresourceAPIApp) streamRequestHandler(request *restful.Request, res
 	defer clientSocket.Close()
 
 	log.Log.Object(vmi).Infof("Websocket connection upgraded")
-	wsReadWriter := &kubecli.BinaryReadWriter{Conn: clientSocket}
 
-	inReader, inWriter := io.Pipe()
-	outReader, outWriter := io.Pipe()
+	tlsConfig, err := app.getConsoleTLSConfig()
```
davidvossel: It looks like this function is doing a "get" on the virt-handler secret for every request. We don't have cert rotation right now, so we really only need to to pull in the secret once.  Once cert rotation comes into play, we'll have some sort of shared code that ensures the tls configs are always up to date with the latest.

@rmohr keep me honest here though, does what I'm saying sound correct based on what you have planned?

pkg/virt-handler/rest/console.go
```diff
@@ -0,0 +1,98 @@
...
+	stream(request, response, t.getUnixSocketPath(request.PathParameter("uid"), "virt-vnc"))
+}
+
+func (t *Console) Serial(request *restful.Request, response *restful.Response) {
+	stream(request, response, t.getUnixSocketPath(request.PathParameter("uid"), "virt-serial0"))
+}
+
+func (t *Console) getUnixSocketPath(vmiID, socketName string) string {
+	pid, _ := t.podIsolationDetector.GetPid(vmiID)
+	return path.Join("proc", strconv.Itoa(pid), "root", "var", "run", "kubevirt-private", vmiID, socketName)
```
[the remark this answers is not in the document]
ahadas: I think it would make sense to change the structure of these directories in the pod and use namespace/name. In my opinion, being able to include the namespace+name in the logs to ease debugging would be the main benefit of this.

But I would prefer not to do it without changing the structure of the directories so if we go in this direction, let's do it separately.

pkg/virt-api/rest/subresource.go
ahadas: Ack, makes sense

ahadas:
Update: I'm struggling with moving `socat` to the virt-handler container.
Without it, tests are failing:
(1) When trying to to connect multiple times through VNC
(2) When invoking serial console multiple times and assuming the new connection terminates the previous one

So I think about implementing the logic of (2) in console.go for both consoles for now (without calling `socat`).
@davidvossel thanks for the feedback. Would the above make sense?

rmohr:
> Update: I'm struggling with moving `socat` to the virt-handler container.

Yes please remove `socat`. Just open the socket directly from virt-handler and copy io.

ahadas:
> > Update: I'm struggling with moving `socat` to the virt-handler container.
> 
> Yes please remove `socat`. Just open the socket directly from virt-handler and copy io.

Ack, so are we ok with limiting the VNC console to one connection per-VMI?

rmohr:
> Ack, so are we ok with limiting the VNC console to one connection per-VMI?

Sounds good. There is not other option anyway.

pkg/virt-handler/rest/console.go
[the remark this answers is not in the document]
davidvossel: >I think it would make sense to change the structure of these directories in the pod and use namespace/name.

yep, Since the kubevirt-private dir is completely isolated within the pod, i'd be fine with changing the directory tree to use namespace/name and making the REST endpoint use namespace/name as well. It just requires a little more work here. 

> But I would prefer not to do it without changing the structure of the directories so if we go in this direction, let's do it separately.

what does "separately" mean here? separate PRs?

davidvossel:
> (2) When invoking serial console multiple times and assuming the new connection terminates the previous one

:+1:  as Roman said, i actually don't think there's another option except to only allow a single connection.

pkg/virt-api/rest/subresource.go
```diff
@@ -420,37 +510,40 @@ func (app *SubresourceAPIApp) StopVMRequestHandler(request *restful.Request, res
...
-				if pod.Name == vmi.Status.MigrationState.TargetPod {
-					return pod.Name, nil
-				}
+	podList, err := app.VirtCli.CoreV1().Pods(namespaceKubevirt).List(selector)
+	if err != nil {
+		return nil, err
+	}
+
+	switch {
+	case len(podList.Items) == 1:
```
cynepco3hahue: nit: you can remove first case because you return it under the default and IMHO `if` conditions more suitable here than `switch`(not critical :smile:)

ahadas:
Ack, so I dropped all use of `socat` and I'll remove it from the image of virt-launcher as well.

pkg/virt-api/rest/subresource.go
```diff
@@ -20,84 +20,92 @@
...
 }
 
 type requestType struct {
 	socketName string
 }
 
-var CONSOLE = requestType{socketName: "serial0"}
-var VNC = requestType{socketName: "vnc"}
+// TODO: make the port configurable
+const consoleTemplateURL = "wss://%s:8185/console/%s"
```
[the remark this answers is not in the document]
ahadas: Done

pkg/virt-api/rest/subresource.go
```diff
@@ -107,49 +115,131 @@ func (app *SubresourceAPIApp) streamRequestHandler(request *restful.Request, res
...
+		return nil, err
+	}
+
+	certPool := x509.NewCertPool()
+	for _, crt := range caCert {
+		certPool.AddCert(crt)
+	}
+
+	// we use the same TLS configuration that is used for live migrations
+	migrationTLSConfig := &tls.Config{
```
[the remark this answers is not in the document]
ahadas: Done

pkg/virt-api/rest/subresource.go
```diff
@@ -107,49 +115,131 @@ func (app *SubresourceAPIApp) streamRequestHandler(request *restful.Request, res
...
 
-	httpResponseCode := http.StatusOK
-	select {
-	case httpResponseCode = <-httpResponseChan:
-	case err := <-copyErr:
-		if err != nil {
-			log.Log.Object(vmi).Reason(err).Error("Error in websocket proxy")
-			httpResponseCode = http.StatusInternalServerError
-		}
+	response.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
ahadas: Remaining that should have been removed, good catch

pkg/virt-api/rest/subresource.go
ahadas: I agree that it's not so significant in this particular case because the evaluation of all case statements is trivial but as I expect the first case to hold in more than 90% of the cases, I prefer not to evaluate the other case statements in those cases (who knows, maybe the migration case will require a query in the future or we'll add another complex case..)

pkg/virt-handler/rest/console.go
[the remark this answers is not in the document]
ahadas: yes, as a separate PR - this one gets too big. I'll post another one for making the port of the console endpoint (8185) configurable so I can do this as part of that PR

pkg/virt-handler/isolation/isolation.go
```diff
@@ -45,6 +45,9 @@ type PodIsolationDetector interface {
 	// It returns an IsolationResult containing all isolation information
 	Detect(vm *v1.VirtualMachineInstance) (*IsolationResult, error)
 
+	// Detects the PID of the VMI with the given UID
+	GetPid(vmiUid string) (int, error)
```
rmohr: If you use on the rest-api namespace/name, you can look up the vmi from the informer cache and use the normal `Detect` flow. I would prefer that, since things have the one canonical entrypoint, no matter if on the api or in the controller loop.

pkg/virt-handler/rest/console.go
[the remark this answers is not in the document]
rmohr: > @rmohr this is why I recommended using uid in the rest call instead of namespace/name. The uid maps directly to the vnc/console socket's directory path.

I would prefer if we pass in namespace/name and use all already prepared flows which we use in the controller function. They expect the VMI object, which can be easily retrieved from the cache. There is nothing wrong with the UID per se, I would prefer the more high-level and consistent  entrypoint via VMI.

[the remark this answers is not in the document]
rmohr: For instance, if we pass the VMI to our helpers, it does not matter if we refactor directory structure or not. You would always pass the VMI to the helpers and get a path back. You don't care about the actual form of the path.

pkg/virt-handler/isolation/isolation.go
ahadas: @rmohr not sure I'm following. by `I would prefer that, since things have the one canonical entrypoint, no matter if on the api or in the controller loop.` you mean that the res-api in virt-handler would match the one of virt-api? if so, then yeah, I agree that it would be better to change the rest-api to namespace/name, but let's do it as a separate PR

and regarding the `Detect` flow, the thing is that I don't need the slice part. that's why I separated out the retrieval of pid. I don't mind to call `Detect` with an option to skip retrieving the slice if you think it's better

rmohr: > but let's do it as a separate PR

why?

rmohr: > and regarding the `Detect` flow, the thing is that I don't need the slice part. that's why I separated out the retrieval of pid. I don't mind to call `Detect` with an option to skip retrieving the slice if you think it's better

Detecting the slice is cheap and we do it all the time in the controller func, I don't think it matters. You just open a pseudo-file in proc.

ahadas: > > but let's do it as a separate PR
> 
> why?

this PR in its current state already includes:
1. Opening the console through virt-handler rather than virt-api executing on virt-launcher
2. Dropping sock-connector
3. Dropping pods/exe priviledges from virt-api
4. Improving copying from/to/between websockets
5. Some refactoring
... and became XXL.
so I think that if we find that it meets the requirements and is stable and not negatively affects the existing code, we should merge it and beautify the code separately. practically, that minimizes rebases on my side and for others.

ahadas: > > and regarding the `Detect` flow, the thing is that I don't need the slice part. that's why I separated out the retrieval of pid. I don't mind to call `Detect` with an option to skip retrieving the slice if you think it's better
> 
> Detecting the slice is cheap and we do it all the time in the controller func, I don't think it matters. You just open a pseudo-file in proc.

honestly, I would prefer to split the code into smaller reusable pieces that can be invoked independently, depending on the desired output. but let's not argue about it, I will call `Detect` if you like

rmohr: > so I think that if we find that it meets the requirements and is stable and not negatively affects the existing code, we should merge it and beautify the code separately. practically, that minimizes rebases on my side and for others.

I don't see how this has a smaller or bigger effect for others. It is a question of luck if you have right now a PR open which will need a rebase after this one is merged, or if you have one open a few days later when we merge the PR.

This change will have immediate effect after the PR is merged. If you don't manage to prepare and merge the follow-up PR until we do a release, virt-handler will strictly speaking have to support both apis. the one with the UID and the one with namespace/name.

Every commit on master is a release candidate and therefore we extra have all testing possibilities on the PR itself.

ahadas: > > so I think that if we find that it meets the requirements and is stable and not negatively affects the existing code, we should merge it and beautify the code separately. practically, that minimizes rebases on my side and for others.
> 
> I don't see how this has a smaller or bigger effect for others. It is a question of luck if you have right now a PR open which will need a rebase after this one is merged, or if you have one open a few days later when we merge the PR.
> 
> This change will have immediate effect after the PR is merged. If you don't manage to prepare and merge the follow-up PR until we do a release, virt-handler will strictly speaking have to support both apis. the one with the UID and the one with namespace/name. Otherwise seamless updates would not work for console and vnc.
> 
> Every commit on master is a release candidate and therefore we extra have all testing possibilities on the PR itself.

I see, so can we drop pods/exec from virt-api? or drop sock-connector? because then I wonder what will happen when old virt-api talks with new virt-handler that launches virt-launcher with no sock-connector?

davidvossel: > If you don't manage to prepare and merge the follow-up PR until we do a release, virt-handler will strictly speaking have to support both apis.

oh yeah... I see the risk here. We have to make a decision about the rest endpoint and execute on it in this PR. I think namespace/name with the informer lookup is the way forward. 

>  see, so can we drop pods/exec from virt-api? or drop sock-connector? because then I wonder what will happen when old virt-api talks with new virt-handler that launches virt-launcher with no sock-connector?

Here's what I suggest. 

* keep socat and sock-connector for now. We can safely remove it the release after next. 
* don't change the kubevirt-private directory tree. Otherwise we risk breaking console/vnc access with existing running VMIs during update. We're pretty much locked in to that directory structure unless we want to go through the hassle of attempting to support two paths... probably not worth it 
* Make the virt-handler vnc/console rest api use namespace/name in the url, then translate the namespace/name to the right kubevirt-private directory tree (which uses uuid) by doing a lookup in the vmi informer cache.

sound reasonable?

ahadas: > If you don't manage to prepare and merge the follow-up PR until we do a release, virt-handler will strictly speaking have to support both apis.
> 
> oh yeah... I see the risk here. We have to make a decision about the rest endpoint and execute on it in this PR. I think namespace/name with the informer lookup is the way forward.

@fabiand , please correct me if I'm wrong here:
IIUC the console that virt-api exposes is intended for installations. For ordinary use by users, VDI-like use, we count on other guest-level tools like RDP.
So I see the that updating to new versions on virt-api/virt-hander that cannot interact with existing virt-launcher pods is something we should avoid. However, I think it is ok then to assume that console connections would be terminated on update and that users would reinstantiate the connection. so I don't see the risk yet, can you please clarify it?

> 
> see, so can we drop pods/exec from virt-api? or drop sock-connector? because then I wonder what will happen when old virt-api talks with new virt-handler that launches virt-launcher with no sock-connector?
> 
> Here's what I suggest.
> 
> keep socat and sock-connector for now. We can safely remove it the release after next.

but do I understand it correctly that we re required to support partial updates where for instance virt-api is not updated and virt-handler is updated? if so, how can we ever change the API between those components non-backward compatible way?

> don't change the kubevirt-private directory tree. Otherwise we risk breaking console/vnc access with existing running VMIs during update. We're pretty much locked in to that directory structure unless we want to go through the hassle of attempting to support two paths... probably not worth it
> Make the virt-handler vnc/console rest api use namespace/name in the url, then translate the namespace/name to the right kubevirt-private directory tree (which uses uuid) by doing a lookup in the vmi informer cache.
> 
> sound reasonable?

yeah, ok.

rmohr: > I see why updating to new versions on virt-api/virt-hander that cannot interact with existing virt-launcher pods is something we should avoid. However, I think it is ok to assume that console connections are terminated on update and that users would reinstantiate the connection. so I don't see the risk yet, can you please clarify it?

General things: We first update all virt-handlers, then virt-controller and finally virt-api. This means that if we have virt-api to virt-handler connections, we need to be backward compatible on the virt-handler api for at least one release, since it is not really predictable how long it takes to update all virt-handlers (or if we even end up with a mixed state for a long time).

Witht he socat approach we actually have a depencency between virt-api, virt-controller and virt-launcher. Now, once virt-controller is updated, it will create new launcher pods, but virt-api will still try to connect the old way. There may be temporary issues which prevent the new virt-api to become ready. until this is the case, the old virt-api is still in place and can't establish new connections for a potentially long time.

> but do I understand it correctly that we re required to support partial updates where for instance virt-api is not updated and virt-handler is updated? if so, how can we ever change the API between those components non-backward compatible way?

Right now we try to support n-1 to n updates. Maybe n-2 to n soon. So we just have to keep internal things for one additional release.

rmohr: > I see, so can we drop pods/exec from virt-api? or drop sock-connector? because then I wonder what will happen when old virt-api talks with new virt-handler that launches virt-launcher with no sock-connector?

You are right, we have to keep it for at least one release. Nit: virt-controller lauchnes virt-launcher, not virt-handler.

ahadas: > I see why updating to new versions on virt-api/virt-hander that cannot interact with existing virt-launcher pods is something we should avoid. However, I think it is ok to assume that console connections are terminated on update and that users would reinstantiate the connection. so I don't see the risk yet, can you please clarify it?
> 
> General things: We first update all virt-handlers, then virt-controller and finally virt-api. This means that if we have virt-api to virt-handler connections, we need to be backward compatible on the virt-handler api for at least one release, since it is not really predictable how long it takes to update all virt-handlers (or if we even end up with a mixed state for a long time).
> Witht he socat approach we actually have a depencency between virt-api, virt-controller and virt-launcher. Now, once virt-controller is updated, it will create new launcher pods, but virt-api will still try to connect the old way. There may be temporary issues which prevent the new virt-api to become ready. until this is the case, the old virt-api is still in place and can't establish new connections for a potentially long time.
>

ok, so by "seamless upgrade" in regards to the console connections we actually mean that on upgrade existing console connections could be terminated (because they go through at least virt-api that would restart) but it should be possible to reinitiate them "immediately", i.e., also during "partial upgrade", where only  some of the components in the cluster were upgraded, right?

Is that a clear requirement that we got from a stakeholder? I'm asking because under the assumption the console that is provided by kubevirt is only used for installations and one-time configuration, we may be able to relax this requirement..  but I guess that the console is not expect to change that often so not sure it would worth the time..

> but do I understand it correctly that we re required to support partial updates where for instance virt-api is not updated and virt-handler is updated? if so, how can we ever change the API between those components non-backward compatible way?
> 
> Right now we try to support n-1 to n updates. Maybe n-2 to n soon. So we just have to keep internal things for one additional release.

Alright, that makes sense.
So we'll keep `socat` and `sock-connector` for now and proceed as proposed by @davidvossel

pkg/virt-handler/rest/console.go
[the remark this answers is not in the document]
ahadas: > For instance, if we pass the VMI to our helpers, it does not matter if we refactor directory structure or not. You would always pass the VMI to the helpers and get a path back. You don't care about the actual form of the path.

Well, but on the other hand it contradicts what others called `principle of least privilege` - a function that seem to require only the UID of the VMI will get a whole VMI. Some of the downsides are that within this function anyone that introduces new code would have access to all the information of the VMI and this function becomes less reusable since it requires the caller to have the VMI at hand.

Yes, it is often that case that such modeling questions involve trade offs. I agree that you pointed out a benefit of passing the whole VMI.

I would prefer to split the `Detect` into smaller functions that require only what they actually need and avoid introducing a VMI informer into virt-handler just for that. But as this can lead to an endless discussion affected by programming style, experience and prioritization I suggest we'll break it here and I'll go with your suggestion :)

rmohr:
> Update: I'm struggling with moving `socat` to the virt-handler container.

Why do you want to move it to virt-handler? It should stay where it is. Just leave `socat` and the script in virt-launcher and `virt-api` can still connect like always. After the next release we can remove it.

pkg/virt-handler/rest/console.go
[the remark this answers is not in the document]
rmohr: It is just a layered approach rest-api -> additional checks on the VMI (does it exist, does it have, does it have vnc configured (vmis may not have a vnc device, I don't expect that you directly go to the file and then guess that there may be no vnc device vased on a non-existing file, and so on.)

This is not about style on low-level functions, it is just a normal layered model which allows more checks and decissions to be done on the right levels where you have the informations you need at hand.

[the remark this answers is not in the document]
ahadas: ok :)

ahadas:
> > Update: I'm struggling with moving `socat` to the virt-handler container.
> 
> Why do you want to move it to virt-handler? It should stay where it is. Just leave `socat` and the script in virt-launcher and `virt-api` can still connect like always. After the next release we can remove it.

That's an old comment, irrelevant now

pkg/virt-handler/isolation/isolation.go
davidvossel: > k, so by "seamless upgrade" in regards to the console connections we actually mean that on upgrade existing console connections could be terminated (because they go through at least virt-api that would restart) but it should be possible to reinitiate them "immediately", i.e., also during "partial upgrade", where only some of the components in the cluster were upgraded, right?


correct. the expectation for now is that console/vnc connections will be broken during the upgrade.

cmd/virt-handler/virt-handler.go
```diff
@@ -279,14 +280,17 @@ func (app *virtHandlerApp) Run() {
...
-	err = http.ListenAndServeTLS(app.ServiceListen.Address(), certStore.CurrentPath(), certStore.CurrentPath(), nil)
-	if err != nil {
-		log.Log.Reason(err).Error("Serving prometheus failed.")
-		panic(err)
-	}
+	// wait for server to exit
+	<-errCh
+}
+
+func (app *virtHandlerApp) runPrometheusServer(errCh chan error, certStore certificate.FileStore) {
```
fromanirh: do we need to actually return the certStore?

ahadas: Thanks @fromanirh :)
Note that this function doesn't return anything, it only gets the certStore as an argument

cmd/virt-handler/virt-handler.go
```diff
@@ -275,18 +289,36 @@ func (app *virtHandlerApp) Run() {
...
+	errCh <- http.ListenAndServeTLS(app.ServiceListen.Address(), certStore.CurrentPath(), certStore.CurrentPath(), nil)
+}
 
-	err = http.ListenAndServeTLS(app.ServiceListen.Address(), certStore.CurrentPath(), certStore.CurrentPath(), nil)
-	if err != nil {
-		log.Log.Reason(err).Error("Serving prometheus failed.")
-		panic(err)
+func (app *virtHandlerApp) runConsoleServer(errCh chan error, port int, consoleHandler *rest.ConsoleHandler) {
+	ws := new(restful.WebService)
+	ws.Route(ws.GET("/namespaces/{namespace}/vmi/{name}/console").To(consoleHandler.SerialHandler))
```
rmohr: Can we use `{namespace}/virtualmachineinstances/{name}` like on the k8s apiserver?

cmd/virt-handler/virt-handler.go
```diff
@@ -275,18 +289,36 @@ func (app *virtHandlerApp) Run() {
...
+	ws.Route(ws.GET("/namespaces/{namespace}/vmi/{name}/vnc").To(consoleHandler.VNCHandler))
+	restful.DefaultContainer.Add(ws)
+	server := &http.Server{
+		Addr:    fmt.Sprintf("%s:%d", app.ServiceListen.BindAddress, port),
+		Handler: restful.DefaultContainer,
+		// we use migration TLS also for console connections (initiated by virt-api)
+		TLSConfig: app.migrationTLSConfig,
 	}
+	errCh <- server.ListenAndServeTLS("", "")
```
rmohr: Can you tie this server into the server which we already use to serve prometheus metrics? There is no need for another webserver.

pkg/virt-config/virt-config.go
```diff
@@ -46,6 +46,7 @@ const (
 	DefaultUseEmulation                             = false
 	DefaultUnsafeMigrationOverride                  = false
 	DefaultPermitSlirpInterface                     = false
+	DefaultConsoleServerPort                        = 8186
```
rmohr: Same as above, no need for an extra weberserver. We just need one for all our http traffic.

pkg/virt-handler/rest/console.go
```diff
@@ -0,0 +1,195 @@
...
+		vncStopChans:         make(map[types.UID](chan struct{})),
+		serialLock:           &sync.Mutex{},
+		vncLock:              &sync.Mutex{},
+		vmiInformer:          vmiInformer,
+	}
+}
+
+func (t *ConsoleHandler) VNCHandler(request *restful.Request, response *restful.Response) {
+	vmi, err := t.getVMI(request)
+	if err != nil {
```
rmohr: If a VMI does not exist, return `404`. If the VMI can't be retrieved from the cache, return `500`.

pkg/virt-handler/rest/console.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+type cleanupOnError func()
+
+func (t *ConsoleHandler) stream(request *restful.Request, response *restful.Response, unixSocketPath string, stopCh chan struct{}, cleanup cleanupOnError) {
+	var upgrader = kubecli.NewUpgrader()
+	clientSocket, err := upgrader.Upgrade(response.ResponseWriter, request.Request, nil)
+	if err != nil {
+		log.Log.Reason(err).Error("Failed to upgrade client websocket connection")
+		response.WriteError(http.StatusBadRequest, err)
```
rmohr: The socket seems to exist here already: That would be `500`.

pkg/virt-handler/rest/console.go
```diff
@@ -0,0 +1,195 @@
...
+		return
+	}
+	defer clientSocket.Close()
+
+	log.Log.Infof("Websocket connection upgraded")
+	log.Log.Infof("Connecting to %s", unixSocketPath)
+
+	fd, err := net.Dial("unix", unixSocketPath)
+	if err != nil {
+		log.Log.Reason(err).Errorf("failed to dial unix socket %s", unixSocketPath)
```
rmohr: Needs to return `500`.

pkg/virt-handler/rest/console.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (t *ConsoleHandler) VNCHandler(request *restful.Request, response *restful.Response) {
+	vmi, err := t.getVMI(request)
+	if err != nil {
+		log.Log.Reason(err).Error("Failed retrieve VMI")
+		response.WriteError(http.StatusBadRequest, err)
+		return
+	}
+	unixSocketPath, err := t.getUnixSocketPath(vmi, "virt-vnc")
```
rmohr: You will have to check if the path exists. If it does not exist, consult the `domain` cache and see if the VMI was reported as down, or as not existent, to return appropriate errors.

pkg/virt-handler/rest/console.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (t *ConsoleHandler) SerialHandler(request *restful.Request, response *restful.Response) {
+	vmi, err := t.getVMI(request)
+	if err != nil {
+		log.Log.Reason(err).Error("Failed retrieve VMI")
+		response.WriteError(http.StatusBadRequest, err)
+		return
+	}
+	unixSocketPath, err := t.getUnixSocketPath(vmi, "virt-serial0")
```
rmohr: You will have to check if the path exists. If it does not exist, consult the domain cache and see if the VMI was reported as down, or as not existent, to return appropriate errors. In theory we should probably even consolidate with virt-handler and not rely on the domain cache, but that would be too much for the PR. We can deal with this very small time window later. Looking into the domain cache is good  enough for now.

pkg/virt-handler/rest/console.go
```diff
@@ -0,0 +1,195 @@
...
+	log.Log.Infof("Connecting to %s", unixSocketPath)
+
+	fd, err := net.Dial("unix", unixSocketPath)
+	if err != nil {
+		log.Log.Reason(err).Errorf("failed to dial unix socket %s", unixSocketPath)
+		return
+	}
+	defer fd.Close()
+
+	log.Log.Infof("Connected to %s", unixSocketPath)
```
rmohr: Also log the object (`Object(vmi`) where appropriate (on all locations where you use `log`), so that it is possible to get related logs  for vmis easily.

tests/vnc_test.go
```diff
@@ -118,7 +122,7 @@ var _ = Describe("[rfe_id:127][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 				vncConnect()
 			})
 
-			It("should allow accessing the VNC device multiple times", func() {
+			It("should allow accesing the VNC device multiple times", func() {
```
rmohr: `accessing`.

cmd/virt-handler/virt-handler.go
ahadas: does the prometheus one use the migration tls configuration as well?

cmd/virt-handler/virt-handler.go
ahadas: concretely, you mean to rename `vmi` to `virtualmachineinstances` ?

cmd/virt-handler/virt-handler.go
rmohr: Right now it uses afaik an untrusted tls certificate. It can as well use an a little less untrusted one.

cmd/virt-handler/virt-handler.go
rmohr: Yep. Also forgot to mention that we should have a version prefix. `v1`. Probably (not related to our public api).

cmd/virt-handler/virt-handler.go
ahadas: so how can we configure a server with two different certificates?

rmohr: Not sure I can follow. Just use the migration certificate for everything and don't generate the other insecure one.

ahadas: ah, so changing the prometheus server to use the migration certificate, right?

rmohr: :+1:

cmd/virt-handler/virt-handler.go
fromanirh: ofc right, I just misread the code.

pkg/virt-handler/rest/console.go
ahadas: are you sure? I don't mind doing that, but I would go with `500` in both cases - after all, the vmi existed at the api-level at the time the user invoked this opeation..

rmohr: It is very unlikely but possible to hit the scenario where it was still ok and there when you hit the api, but disappeared afterwards. Or not?

pkg/virt-handler/rest/console.go
ahadas: can you elaborate -
1. by `domain` you mean libvirt domain? or you mean to query the vmi from vmiinformer?
2. you mean when `Dial`ing the socket fails?

rmohr: Hm, a little bit tricky to do the right thing here. Le'ts keep it as you have it (I meant querying the domain-cache, similar to the vmi domain, just for libvirt domains, the virt-handler has that too).

pkg/virt-handler/rest/console.go
ahadas: I think it may happen, all I'm saying is that in this case the error would be `500` because as a user I invoked the operation on a valid entity - the server later failed to execute the operation

rmohr: From virt-handler perspective it is 404 (not 500). It was also valid from virt-api to forward the connection, based on its information. Looks to me like virt-api should return 404?

rmohr: I think it is important that also for internal REST-communication it is important to send the right error codes. Sending 500 on an action which is perfectly valid will look like a bug.

pkg/virt-handler/rest/console.go
rmohr: I think on the long run we will do a "consolidate" call and directly check the domain state to see why it could not connect and return an appropriate error.

pkg/virt-handler/rest/console.go
ahadas: ok, if virt-api is not expected to forward this code to the client then I agree

tests/vnc_test.go
ahadas: Done

pkg/virt-handler/rest/console.go
ahadas: Done

pkg/virt-handler/rest/console.go
ahadas: ok

pkg/virt-handler/rest/console.go
ahadas: Done

pkg/virt-handler/rest/console.go
ahadas: Done

pkg/virt-handler/rest/console.go
ahadas: Done

pkg/virt-api/rest/subresource.go
```diff
@@ -420,37 +510,40 @@ func (app *SubresourceAPIApp) StopVMRequestHandler(request *restful.Request, res
 	response.WriteHeader(http.StatusAccepted)
 }
 
-func (app *SubresourceAPIApp) findPod(namespace string, vmi *v1.VirtualMachineInstance) (string, error) {
-	fieldSelector := fields.ParseSelectorOrDie("status.phase==" + string(k8sv1.PodRunning))
-	labelSelector, err := labels.Parse(fmt.Sprintf(v1.AppLabel + "=virt-launcher," + v1.CreatedByLabel + "=" + string(vmi.UID)))
-	if err != nil {
-		return "", err
+func (app *SubresourceAPIApp) findVirtHandlerPodForVMI(vmi *v1.VirtualMachineInstance) (*k8sv1.Pod, error) {
+	if !vmi.IsRunning() {
```
rmohr: Can you take the node name from the status of the vmi and do `kubecli.NewVirtHandlerClient(app.VirtCli).ForNode(vmi.Status.NodeName).ConsoleURI(vmi)`? That should pretty much do what this method does. You will have to adjust the connection URI in `ConsoleURI` since it is a left-over which we forgot to remove, but apart from that it should give you waht you need. You can add a `VNCURI(vmi)` method there.

pkg/virt-handler/rest/console.go
```diff
@@ -0,0 +1,196 @@
...
+	var upgrader = kubecli.NewUpgrader()
+	clientSocket, err := upgrader.Upgrade(response.ResponseWriter, request.Request, nil)
+	if err != nil {
+		log.Log.Object(vmi).Reason(err).Error("Failed to upgrade client websocket connection")
+		response.WriteError(http.StatusInternalServerError, err)
+		return
+	}
+	defer clientSocket.Close()
+
+	log.Log.Object(vmi).Infof("Websocket connection upgraded")
```
rmohr: If you want to keep these log lines, I think they should be `V(4)` (debug). In general we handle all errors, so if there is an error the errror log should reveal how fare  we go.

ahadas:
@rmohr @davidvossel thanks for the feedback. Before addressing the comments, I wonder if you could have some insights on a weird scenario I'm investigating:
The vnc functional tests fail when having multiple vnc connections. The first connection is properly established. The second fails. The reason is that "dialing" the unix socket fails because the socket file doesn't exist. I saw some reports that the underlying file [may be removed when the listener is closed ](https://groups.google.com/forum/#!topic/golang-nuts/UtBR4IfgaEw) but didn't find anything related to the client being closed. It works when commenting out [this line](https://github.com/ahadas/kubevirt/blob/console/pkg/virt-handler/rest/console.go#L167). It is apparently unrelated to [this line](https://github.com/ahadas/kubevirt/blob/console/pkg/virt-launcher/virtwrap/cmd-server/server.go#L328). I'm going to dig into go/net to see if the underlying file is unlinked anywhere else, but maybe you could save me that time if you faced this before...

ahadas:
> @rmohr @davidvossel thanks for the feedback. Before addressing the comments, I wonder if you could have some insights on a weird scenario I'm investigating:
> The vnc functional tests fail when having multiple vnc connections. The first connection is properly established. The second fails. The reason is that "dialing" the unix socket fails because the socket file doesn't exist. I saw some reports saying the underlying file [may be removed when the listener is closed ](https://groups.google.com/forum/#!topic/golang-nuts/UtBR4IfgaEw) but didn't find anything related to the client being closed. It works (the tests pass) when commenting out [this line](https://github.com/ahadas/kubevirt/blob/console/pkg/virt-handler/rest/console.go#L167). It is apparently unrelated to [this line](https://github.com/ahadas/kubevirt/blob/console/pkg/virt-launcher/virtwrap/cmd-server/server.go#L328). I'm going to dig into go/net to see if the underlying file is unlinked anywhere else, but maybe you could save me that time if you faced this before...

Or it may be related to the [grpc server handling](https://github.com/grpc/grpc/issues/17122)...

rmohr:
@ahadas I think you are hitting the qemu bug, where qemu (wrongly) deletes the socket after the first connection. @petrkotas solved that by making the qemu socket read-only. Maybe the logic introduced in https://github.com/kubevirt/kubevirt/pull/2171 got lost in the refactoring?

ahadas:
> @ahadas I think you are hitting the qemu bug, where qemu (wrongly) deletes the socket after the first connection. @petrkotas solved that by making the qemu socket read-only. Maybe the logic introduced in #2171 got lost in the refactoring?

QEMU also unlinked it.. interesting
Looks like Petr's fix is still there. I think that I'm facing a different issue because without [closing the the socket](https://github.com/ahadas/kubevirt/blob/console/pkg/virt-handler/rest/console.go#L167) it works - seems it has something do to with the close operation

rmohr:
> QEMU also unlinked it.. interesting

It must be QEMU. `net.Dial` does not unlink sockets. Only `net.Listen` (server side).

rmohr:
I think that the fix in place is a little bit racy. Now that we open the connections through virt-handler, I would do the following: After the check if the socket is there, and before opening the socket, ensure that the socket directory is read-only.

ahadas:
> > QEMU also unlinked it.. interesting
> 
> It must be QEMU. `net.Dial` does not unlink sockets. Only `net.Listen` (server side).

I tried to reproduce it manually in two scenarios:
1. Connecting to the VM via VNC from one terminal and then connecting from another terminal - the code seems to behave normally, virt-handler close the first connect and makes the second one.
2. Connecting to the VM via VNC from one terminal, closing that connection and then connect again - it also works.

I think that if it's QEMU fault, then those scenarios would have failed, right? but they seem to work

ahadas:
> I think that the fix in place is a little bit racy. Now that we open the connections through virt-handler, I would do the following: After the check if the socket is there, and before opening the socket, ensure that the socket directory is read-only.

Ah you say that the connection is established too quicky? ok, that makes sense

rmohr:
I tested it with a go client and a server, just to be really sure. No sockets got removed when the client closed the connection. The symptoms you describe sound exactly like the qemu bug (note that the solution from petr may leave the directory as writeable for up to one second, so unit tests have plenty of time to break things). The bug is that QEMU deletes the socket if a connection gets closed.

cynepco3hahue:
@rmohr IMHO if we have such preparation stuff under the virt-launcher code, we should not move the VMI to the running state until all preparation is done.

ahadas:
> I tested it with a go client and a server, just to be really sure. No sockets got removed when the client closed the connection. The symptoms you describe sound exactly like the qemu bug (note that the solution from petr may leave the directory as writeable for up to one second, so unit tests have plenty of time to break things). The bug is that QEMU deletes the socket if a connection gets closed.

Yeah, it didn't manage to reproduce it in a separate client-server application I wrote either.
You seem to be right - adding a sleep of 5 sec before the first connection solves that - I guess that's the time it takes to ensure the directory is changed to read-only.
So why do we set the VMI as running before those preparations are completed?

rmohr:
The fix should just have ensured on every connection, inside the exec script that the directory is read-only.

ahadas:
> The fix should just have ensured on every connection, inside the exec script that the directory is read-only.

Petr's fix you mean?

rmohr:
> Petr's fix you mean?

Yes.

rmohr:
> @rmohr IMHO if we have such preparation stuff under the virt-launcher code, we should not move the VMI to the running state until all preparation is done.

In general yes. We should however be done when libvirt says the the VMI is running (and there the sockets already exist). The only issue is this bug. I think in this case it is by far the simplest solution to just ensure that the directory is read-only (check if read-only, if not setting it read-only) on every connection.

ahadas:
@rmohr ok, kudos on figuring it out :) but how come that without the changes in this PR the multiple-VNC test passes? is it because it takes more time to trigger the exec script? or because of the use of `socat`?

rmohr:
> @rmohr ok, kudos on figuring it out :) but how come that without the changes in this PR the multiple-VNC test passes? is it because it takes more time to trigger the exec script?

I think that the exec script does not kill `socat` properly. It only kills it when you re-connect, so qemu always saw an open connection and when the second connection attempt happened, the directory was already read-only. I am very glad that we get rid of this bash script soon.

petrkotas:
The fix was only meant as a temporary workaround and is designed in a way it does not block on IO too much.

Moreover, the QEMU 4 does not suffer from these issues and deems the fix deprecated. It is already released upstream. For downstream we are still waiting.

Once the QEMU is upgraded, these issues will go away.

pkg/virt-api/rest/subresource.go
ahadas: Done

pkg/virt-handler/rest/console.go
ahadas: Nah, those were remaining from the time before I wrote proper error-handling - removed

pkg/virt-api/rest/subresource.go
```diff
@@ -20,84 +20,93 @@
...
-			return
-		}
+	if err := validate(vmi); err != nil {
+		response.WriteError(http.StatusBadRequest, err)
+		return
 	}
 
-	cmd := []string{"/usr/share/kubevirt/virt-launcher/sock-connector", fmt.Sprintf("/var/run/kubevirt-private/%s/virt-%s", vmi.GetUID(), requestType.socketName)}
+	once.Do(func() {
+		if tlsConfig, err := app.getConsoleTLSConfig(); err == nil {
```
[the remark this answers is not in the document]
ahadas: Right, changed

davidvossel:
> I think that the exec script does not kill socat properly. It only kills it when you re-connect, so qemu always saw an open connection and when the second connection attempt happened, the directory was already read-only. I am very glad that we get rid of this bash script soon.

yep, this is a pretty reasonable explanation.

The "read-only" check before virt-handler connects solves the issue in that the socket won't disappear, but we'll still have flaky functional tests due to the timing of when read-only is set. 

Like @cynepco3hahue pointed out, can we ensure read-only is set before marking the VMI as Phase: Running?

rmohr:
> Like @cynepco3hahue pointed out, can we ensure read-only is set before marking the VMI as Phase: Running?

As mentioned before, the read-only fix which we had in place did its work asynchronous. There is not need to wait. The socket already exists when the VMI is reported by libvirt as running.

rmohr:
> The "read-only" check before virt-handler connects solves the issue in that the socket won't disappear, but we'll still have flaky functional tests due to the timing of when read-only is set.

The "check" will ensure that it is set. It does not just check and return with an error if it is not-yet read-only.

ahadas:
> > Like @cynepco3hahue pointed out, can we ensure read-only is set before marking the VMI as Phase: Running?
> 
> As mentioned before, the read-only fix which we had in place did its work asynchronous (and left a time window open to let something like this happen). There is not need to wait. The socket already exists when the VMI is reported by libvirt as running.

[Already coded](https://github.com/ahadas/kubevirt/blob/console/pkg/virt-handler/rest/console.go#L144-L153) but in light of the last part I wonder if it's needed to check whether the socket exists - we can try to `Chmod` the parent directory, and fail if it doesn't exist, and if the socket itself doesn't exist then the `Dial` operation will fail, no? (@rmohr)

rmohr:
@ahadas the prometheus metrics tests have issues (Probably because of TLS verification?). Apart from that everything looks good. To be honest I did not think about the fact that we need to allow insecure TLS connections for prometheus right now ...

So back to two servers I guess. Sorry for that.

ahadas:
> @ahadas the prometheus metrics tests have issues (Probably because of TLS verification?). Apart from that everything looks good. To be honest I did not think about the fact that we need to allow insecure TLS connections for prometheus right now ...
> 
> So back to two servers I guess. Sorry for that.

Sure, I thought it's debatable so I implemented it in an independent commit - I can just drop it and modify the one that makes the port configurable at deployment phase to inject the port to virt-handler (and not only to virt-api). On it

ahadas:
interestingly, `make client-python` started to fail although it passes locally...

rmohr:
> interestingly, `make client-python` started to fail although it passes locally...

Let me retrigger travis.

fabiand:
WOW - So many green lanes.

PR #2520, opened 2019-07-22
Check file existence in parse-shasums.sh

slintes:
Check file existence in parse-shasums.sh. File might not exist if images weren't pushed before.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/parse-shasums.sh
+++ b/hack/parse-shasums.sh
@@ -24,6 +24,11 @@ if [[ -z "$push_log_file" ]]; then
     return
 fi
 
+if [[ ! -f "$push_log_file" ]]; then
+    echo "$push_log_file not found: won't use shasums, falling back to tags"
+    return
+fi
+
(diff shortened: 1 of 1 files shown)


slintes:
test failures are about failed downloads, but os-3.11.0 and k8s-1.11.0 are green

/approve

rmohr:
@slintes does this mean that we could code in shasums from the previous run?

slintes:
Mh, good question.
If you run `make manifests` alone, yes. But I think it's not an issue. You might want to create new manifests with "old" shasums, if you just changed something in the manifest templating but not in the binaries/images.
`make cluster-build/sync` is fine, because `hack/cluster-build.sh` always pushes first.

rmohr:
Hm, that is not very reassuring. Let's see if we can improve that in the future. E.g. by directly specifying the right dependency tree in bazel.

PR #2521, opened 2019-07-22
Fix migration ioerror

### LINKED ISSUE #2481: [flaky test] [rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system] VM Live Migration with sata disks [test_id:1853]VM with containerDisk + CloudInit + ServiceAccount + ConfigMap + Secret

>
/kind bug

> /kind enhancement

**What happened**:

This is the most unstable migration test:

Lanes: | pull-kubevirt-e2e-k8s-1.13.3 | pull-kubevirt-e2e-k8s-multus-1.13.3 | pull-kubevirt-e2e-os-3.11.0-crio | pull-kubevirt-e2e-k8s-genie-1.11.1 | pull-kubevirt-e2e-k8s-1.11.0 | pull-kubevirt-e2e-os-3.11.0-multus | pull-kubevirt-e2e-os-3.11.0
-- | -- | -- | -- | -- | -- | -- | --
[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system] VM Live Migration with sata disks [test_id:1853]VM with containerDisk + CloudInit + ServiceAccount + ConfigMap + Secret | 2/25/0 | 0/0/25 | 0/20/0 | 0/0/26 | 3/21/0 | 0/0/27 | 3/22/0
-- | -- | -- | -- | -- | -- | -- | --

**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

vladikr:

Live migration tests of a VMI with configMap fail occasionally with an I/O error.
This is most probably because of an inconsistency between the virtual/actual size of configMap/secret disks on the source and the destination. This may partially happen because the test doesn't' delete 
the created configMaps and secrets.

```release-note
NONE
```

### DIFF
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -1138,11 +1138,9 @@ var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system]
 	})
 
 	Context("with sata disks", func() {
-
-		It("[test_id:1853]VM with containerDisk + CloudInit + ServiceAccount + ConfigMap + Secret", func() {
-			configMapName := "configmap-" + rand.String(5)
-			secretName := "secret-" + rand.String(5)
-
+		var configMapName = "configmap-" + rand.String(5)
+		var secretName = "secret-" + rand.String(5)
+		BeforeEach(func() {
 			config_data := map[string]string{
 				"config1": "value1",
 				"config2": "value2",
@@ -1155,6 +1153,13 @@ var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system]
 
 			tests.CreateConfigMap(configMapName, config_data)
 			tests.CreateSecret(secretName, secret_data)
+		})
+		AfterEach(func() {
+			tests.DeleteConfigMap(configMapName)
+			tests.DeleteSecret(secretName)
+		})
+
+		It("[test_id:1853]VM with containerDisk + CloudInit + ServiceAccount + ConfigMap + Secret", func() {
 
 			vmi := tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskFedora))
 			vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse(fedoraVMSize)


pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -245,7 +245,7 @@ func prepareMigrationFlags(isBlockMigration bool, isUnsafeMigration bool, allowA
 	migrateFlags := libvirt.MIGRATE_LIVE | libvirt.MIGRATE_PEER2PEER
 
 	if isBlockMigration {
-		migrateFlags |= libvirt.MIGRATE_NON_SHARED_INC
+		migrateFlags |= libvirt.MIGRATE_NON_SHARED_DISK
```
fromanirh: could you please add few lines in the commit message about why this flag is needed?

rmohr: @vladikr what is the difference between the two flags?

rmohr:
> This is most probably because of an inconsistency between the virtual/actual size of configMap/secret disks on the source and the destination. This may partially happen because the test doesn't' delete
> the created configMaps and secrets.

I would expect that we solely copy the disks over and don't create them again. We basically drag a disk-snapshot of the configMaps and secrets with us, from the timepoint where the VMI was created. So changing the entities on the cluster should  not have any effect.

vladikr:
> I would expect that we solely copy the disks over and don't create them again. We basically drag a disk-snapshot of the configMaps and secrets with us, from the timepoint where the VMI was created. So changing the entities on the cluster should not have any effect.

Yes, that what I was trying to do at the beginning. The problem is that when we try to block migrate a disk that doesn't exist on the destination side with `libvirt.MIGRATE_NON_SHARED_INC`migration flag, libvirt doesn't allow to create a new disk. With `libvirt.MIGRATE_NON_SHARED_INC` when the fine doesn't exist on dest. libvirt requires a store to be defined.
In this PR I let the dest. create the disk and completely override it during migration. I will look into how to define a store and what are the implications.

pkg/virt-launcher/virtwrap/manager.go
vladikr: > VIR_MIGRATE_NON_SHARED_DISK 	= 	64 	
> 
> Migrate full disk images in addition to domain's memory. By default only
> non-shared non-readonly disk images are transferred. The
> VIR_MIGRATE_PARAM_MIGRATE_DISKS parameter can be used to specify which
> disks should be migrated. This flag and VIR_MIGRATE_NON_SHARED_INC are
> mutually exclusive.
> 
> VIR_MIGRATE_NON_SHARED_INC 	= 	128 	
> 
> Migrate disk images in addition to domain's memory. This is similar to
> VIR_MIGRATE_NON_SHARED_DISK, but only the top level of each disk's
> backing chain is copied. That is, the rest of the backing chain is
> expected to be present on the destination and to be exactly the same as
> on the source host. This flag and VIR_MIGRATE_NON_SHARED_DISK are
> mutually exclusive.
> 

I think in our case it doesn't change much. I can switch it back to `MIGRATE_NON_SHARED_INC` - Right now, I just wanted to see if it makes any difference.

rmohr:
Let's ensure that we synchronize this work with the containerdisk PR. With the changes there libvirt can't override the image in the backing chain.

vladikr:
> Let's ensure that we synchronize this work with the containerdisk PR. With the changes there libvirt can't override the image in the backing chain.

I reverted to libvirt.MIGRATE_NON_SHARED_INC flag. This change is only about fixing the test now.

tests/migration_test.go
```diff
@@ -1126,6 +1124,13 @@ var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system]
 
 			tests.CreateConfigMap(configMapName, config_data)
 			tests.CreateSecret(secretName, secret_data)
+		}, 60)
```
ahadas: the second argument (`60`) can be dropped as the provided function is called synchronously

rmohr: @vladikr could you address that?

vladikr: Sure, done

rmohr:
/hold

I think the right approach is to add a `tests.kubevirt.io/something` label  to the configmaps on creation and delete them after every test in general.

PR #2525, opened 2019-07-23
Better timeout expired message for ObjectEventWatcher

fedepaol:
When a `ObjectEventWatcher` fails watching for a timeout (https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L361) it gives back a very generic error message. 

What we see on the output is something along the line of:
```
/root/go/src/kubevirt.io/kubevirt/tests/vmi_multus_test.go:700
Waited for 182 seconds on the event stream to match a specific event
```

Which does not give much hint of what it is failing.

This PR adds a bit more context specifying the event type and the message we are waiting for while trying to be not too invasive.

**Release note**:

```release-note

NONE

```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -378,7 +378,7 @@ func (w *ObjectEventWatcher) WaitFor(stopChan chan struct{}, eventType EventType
 			return true
 		}
 		return false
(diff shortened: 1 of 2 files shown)


slintes:
good idea :+1: 
let's wait for test results

/lgtm

PR #2528, opened 2019-07-24
Ignore the first 5 sync vmi failed errors

### LINKED ISSUE #2471: SyncVMI sometimes fails with "context deadline exceeded" (not only) during migrations

/kind bug

**What happened**:
Mostly during migrations, but also on other tests (see 1st comment), SyncVMI sometimes fails with "context deadline exceeded". During our CI tests this results in a test failure, because this triggers an unexpected warning event. However, if you ignore this warning, SyncVMI succeeds on the next try.

e.g.:
```
[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system] VM Live Migration Starting a VirtualMachineInstance with auto converge enabled should complete a migration expand_less | 6m7s
-- | --
/root/go/src/kubevirt.io/kubevirt/tests/migration_test.go:431 Unexpected Warning event received: testvmid6j2tfjm72qrwn5wsnsn6p4g2rbt9ft9mdnp9h9n55ln5gnw,a9419b63-a29a-11e9-b527-525500d15501: unknown error encountered sending command SyncVMI: rpc error: code = DeadlineExceeded desc = context deadline exceeded Expected     <string>: Warning not to equal     <string>: Warning /root/go/src/kubevirt.io/kubevirt/tests/utils.go:348
```

No SyncVMI failures, no warning events during migrations

not 100% reproducable
see e.g. cri-o CI test lane on this PR, there it happened several times: https://prow.apps.ovirt.org/pr-history/?org=kubevirt&repo=kubevirt&pr=2459
I also experienced this regularly while working on the NFS migration test.

fedepaol:
In `ObjectEventWatcher` we can raise a flag that will make the test fail upon Warning events.

Sometimes, if the guest takes more than 5 seconds to start, the SyncVMI grpc call from the `virt-handler` to the `virt-launcher` fails with a timeout and raises a warning event. What really happens is that the `virt-handler` will retry and eventually the call will succeed (and the server side of the first call will go on normally), so this specific warning is not something that should make our test fail.

This is described [in this comment](https://github.com/kubevirt/kubevirt/issues/2471#issuecomment-514514248) for #2471 

Here we try to ignore the first 5 occurrencies of such warnings, maintaining the existing behaviour for all the others.

Fixes #2471

[Here](https://github.com/kubevirt/kubevirt/compare/master...fedepaol:filterwarnings?expand=1#diff-a5a178b4c66835110ceff63c61d28c96R319) a bit more of refactor could have been done merging the two branches, wasn't sure if there was much difference in the two error messages and I left the way it was before.

**Release note**:

```release-note
NONE

```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -223,19 +223,23 @@ const (
 	SecretLabel = "kubevirt.io/secret"
 )
 
+const syncVmiTimeoutMessage = "unknown error encountered sending command SyncVMI: rpc error: code = DeadlineExceeded desc = context deadline exceeded"
+
 type ProcessFunc func(event *k8sv1.Event) (done bool)
+
+type FailOnWarningFunc func(event *k8sv1.Event) bool
 
 type ObjectEventWatcher struct {
 	object                 runtime.Object
 	timeout                *time.Duration
-	failOnWarnings         bool
+	failOnWarnings         FailOnWarningFunc
 	resourceVersion        string
 	startType              startType
 	dontFailOnMissingEvent bool
 }
 
 func NewObjectEventWatcher(object runtime.Object) *ObjectEventWatcher {
-	return &ObjectEventWatcher{object: object, startType: invalidWatch}
+	return &ObjectEventWatcher{object: object, startType: invalidWatch, failOnWarnings: neverFailOnWarnings}
 }
 
 func (w *ObjectEventWatcher) Timeout(duration time.Duration) *ObjectEventWatcher {
@@ -243,8 +247,13 @@ func (w *ObjectEventWatcher) Timeout(duration time.Duration) *ObjectEventWatcher
 	return w
 }
 
+func (w *ObjectEventWatcher) MightFailOnWarnings(f FailOnWarningFunc) *ObjectEventWatcher {
+	w.failOnWarnings = f
+	return w
+}
+
 func (w *ObjectEventWatcher) FailOnWarnings() *ObjectEventWatcher {
-	w.failOnWarnings = true
+	w.failOnWarnings = alwaysFailOnWarnings
 	return w
 }
 
@@ -307,10 +316,8 @@ func (w *ObjectEventWatcher) Watch(abortChan chan struct{}, processFunc ProcessF
 		panic(err)
 	}
 
-	f := processFunc
-
-	if w.failOnWarnings {
-		f = func(event *k8sv1.Event) bool {
+	f := func(event *k8sv1.Event) bool {
+		if event.Type == string(WarningEvent) && w.failOnWarnings(event) {
 			msg := fmt.Sprintf("Event(%#v): type: '%v' reason: '%v' %v", event.InvolvedObject, event.Type, event.Reason, event.Message)
 			if event.Type == string(WarningEvent) {
 				log.Log.Reason(fmt.Errorf("unexpected warning event received")).ObjectRef(&event.InvolvedObject).Error(msg)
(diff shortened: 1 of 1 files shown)


rmohr:
@fedepaol nice investigation.

fedepaol:
> > What really happens is that the virt-handler will retry and eventually the call will succeed (and the server side of the first call will go on normally), so this specific warning is not something that should make our test fail.
> 
> this is a bug.
> 
> the VMI should not be scheduled on the node until virt-launcher can process the sync request.

Ehm, you mean a bug of how the tests are launched or a bug of kubevirt? 

In my case (more details on the [issue](https://github.com/kubevirt/kubevirt/issues/2471#issuecomment-514514248)) it took 14 seconds to go from https://github.com/kubevirt/kubevirt/blob/35b40d7a96c23b4ab7effc01aaa24ecf7d2cba8a/pkg/virt-launcher/virtwrap/manager.go#L893 to https://github.com/kubevirt/kubevirt/blob/35b40d7a96c23b4ab7effc01aaa24ecf7d2cba8a/pkg/virt-launcher/virtwrap/manager.go#L924 

which is the mostly (I think) in the `dom.Create()` call.

fromanirh:
> In my case (more details on the [issue](https://github.com/kubevirt/kubevirt/issues/2471#issuecomment-514514248)) it took 14 seconds to go from
> 
> https://github.com/kubevirt/kubevirt/blob/35b40d7a96c23b4ab7effc01aaa24ecf7d2cba8a/pkg/virt-launcher/virtwrap/manager.go#L893
> to
> 
> https://github.com/kubevirt/kubevirt/blob/35b40d7a96c23b4ab7effc01aaa24ecf7d2cba8a/pkg/virt-launcher/virtwrap/manager.go#L924
> 
> which is the mostly (I think) in the `dom.Create()` call.

Makes sense. We had a long story of slow VM start in oVirt. Creatinga VM can indeed be slow even on powerful machines.

rmohr:
> Makes sense. We had a long story of slow VM start in oVirt. Creating a VM can indeed be slow even on powerful machines.

A first step would be to increase the timeout of grpc and increase the number of workers in virt-handler. At least in our CI system we must be in a healthy enough environment that normal flows work without warnings.

fedepaol:
> > Makes sense. We had a long story of slow VM start in oVirt. Creating a VM can indeed be slow even on powerful machines.
> 
> A first step would be to increase the timeout of grpc and increase the number of workers in virt-handler. At least in our CI system we must be in a healthy enough environment that normal flows work without warnings.

Not 100% sure about this. 
If we go for very long timeouts and in production we keep the 5 seconds, it will only run without warnings in our CI but in the real world retries will still happen as @fromanirh said, and this does not necessarily means that this is a sign of bad health.

rmohr:
> Not 100% sure about this.
> If we go for very long timeouts and in production we keep the 5 seconds, it will only run without warnings in our CI but in the real world retries will still happen as @fromanirh said, and this does not necessarily means that this is a sign of bad health.

The goal is to set a sane baseline. This baseline may be too high or too low at the beginning and has to be adjusted according to our needs and real-world scenarios. We have to extra-test if long-startup will be handled gracefully. We can't have it as a base assumption. That will hide issues. This allows to test in a well-known scenario that long-startup is handled gracefully and allows for other tests that if warnings occur there that we have an issue.

fedepaol:
> The goal is to set a sane baseline. This baseline may be too high or too low at the beginning and has to be adjusted according to our needs and real-world scenarios. We have to extra-test if long-startup will be handled gracefully. We can't have it as a base assumption. That will hide issues. This allows to test in a well-known scenario that long-startup is handled gracefully and allows other tests to detect that we have a real issue.

Ok, now I am 100% sure :smiley: , thanks a lot for taking the time to explain.

PR #2531, opened 2019-07-25
[WIP] do not copy configMap, secrets and serviceaccount disks during live migration

vladikr:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/cmd/virt-launcher/virt-launcher.go
+++ b/cmd/virt-launcher/virt-launcher.go
@@ -395,6 +395,25 @@ func main() {
 	domainConn := createLibvirtConnection()
 	defer domainConn.Close()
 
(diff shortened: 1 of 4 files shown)


fabiand:
@vladikr could you please describe why you want to not copy the things?

vladikr:
> @vladikr could you please describe why you want to not copy the things?

@fabiand This was a trial balloon. I'll close it. 
I've started this due to i/o errors we've seen during the live migration of all the read-only disks. Normally it happens because of the dest/src size inconsistency. The two solutions are either to pre-allocate the exact size of the source on the dest. or not to copy and use the generated disk on the dest.
Long story short, our current approach should be fine :)

PR #2536, opened 2019-07-28
eliminate brodge+pod combination from examples

alonSadan:
In this PR we changed example-generators
which included a VMI with a
pod network and a bridge interface combination,
to masquerade interface.

this is regarding the issues discussed in detail here:
https://github.com/kubevirt/kubevirt/pull/2493
we would like the examples to be consistent with
the changes suggested in this PR.

```release-note

```


rmohr:
Can you just do this in the other PR? I don't think you need an extra PR and all related changes would be done in one go.

rmohr:
Ups wrong button.

rmohr:
@alonSadan can we close this?

alonSadan:
@rmohr 
Yes!. sorry for leaving it like this.
this PR: https://github.com/kubevirt/kubevirt/pull/2607 handled it already  .

PR #2537, opened 2019-07-28
Change default binding method to bridge

alonSadan:
In this PR we change the default binding method from bridge to masquerade,
both in the code and the examples.
 
We would prefer masquerade as our default binding method, since live migration
is not allowed with bridge (see @vladikr 's merged PR regarding this: 
https://github.com/kubevirt/kubevirt/pull/2405) ,which means live migration is not possible by default.

Furthermore, it is not possible networking with sidecars when using bridge, which restricts, for example, the use of Istio.  

In order for the examples to be consistent with this implementation , and to avoid the situation of 
a random user copy-pasting the examples and unexpectedly not receiving support
for live migration and Istio, we changed the examples as well from bridge masquerade.

**Release note**:

```release-note
default binding method was changed from bridge to masquerade
```

### DIFF
--- a/pkg/virt-config/config_test.go
+++ b/pkg/virt-config/config_test.go
@@ -86,8 +86,8 @@ var _ = Describe("ConfigMap", func() {
 		table.Entry("is bridge, it should return bridge", "bridge", "bridge"),
 		table.Entry("is slirp, it should return slirp", "slirp", "slirp"),
 		table.Entry("is masquerade, it should return masquerade", "masquerade", "masquerade"),
-		table.Entry("when unset, it should return the default", "", "bridge"),
-		table.Entry("when invalid, it should return the default", "invalid", "bridge"),
+		table.Entry("when unset, it should return the default", "", "masquerade"),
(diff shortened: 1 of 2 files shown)


booxter:
@rmohr I believe it's API breakage. I thought we were holding the change in the past because of that exact reason. Isn't it still a concern? Otherwise, I would of course prefer having the `masquerade` type by default.

rmohr:
> @rmohr I believe it's API breakage. I thought we were holding the change in the past because of that exact reason. Isn't it still a concern? Otherwise, I would of course prefer having the `masquerade` type by default.

Responded here: https://github.com/kubevirt/kubevirt/pull/2493#issuecomment-517018275

alonSadan:
This PR is closed because we preferred to keep the default as it is, and use a knob that enables, not by default, the cluster administrator to block a user from configuring bridge interface on pod network.
the knob is implemented here: 
https://github.com/kubevirt/kubevirt/pull/2493/files

PR #2544, opened 2019-07-30
Make the kubevirt rising timeout longer in cluster-deploy.sh

fedepaol:

While running sriov ci tests, a few runs failed for `timed out waiting for the condition on kubevirts/kubevirt`  during cluster-deploy. This PR makes the timeout longer in order to see if it was just a matter of short time.

**Release note**:

```release-note

NONE

```

### DIFF
--- a/hack/cluster-deploy.sh
+++ b/hack/cluster-deploy.sh
@@ -87,6 +87,6 @@ until _kubectl -n kubevirt get kv kubevirt; do
 done
 
 # wait until KubeVirt is ready
-_kubectl wait -n kubevirt kv kubevirt --for condition=Ready --timeout 6m || (echo "KubeVirt not ready in time" && exit 1)
+_kubectl wait -n kubevirt kv kubevirt --for condition=Ready --timeout 12m || (echo "KubeVirt not ready in time" && exit 1)
 
 echo "Done"
--- a/hack/cluster-deploy.sh
+++ b/hack/cluster-deploy.sh
@@ -21,9 +21,19 @@ set -ex
 
 DOCKER_TAG=${DOCKER_TAG:-devel}
 
+
 source hack/common.sh
 source cluster-up/cluster/$KUBEVIRT_PROVIDER/provider.sh
 source hack/config.sh
+
+function dump_kubevirt {
+  echo "Dump kubevirt state:"
+  _kubectl get all -n kubevirt
+  for operator in $(kubectl -n kubevirt get pods | grep operator | awk '{print $1}'); do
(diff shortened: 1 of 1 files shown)


hack/cluster-deploy.sh
```diff
@@ -87,6 +87,6 @@ until _kubectl -n kubevirt get kv kubevirt; do
 done
 
 # wait until KubeVirt is ready
-_kubectl wait -n kubevirt kv kubevirt --for condition=Ready --timeout 6m || (echo "KubeVirt not ready in time" && exit 1)
```
cynepco3hahue: hm, why it can take 12m? all images should be under the container registry, so we should not have latency problems.

fedepaol: Re 12m, I kept the timeout large. This https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2395/pull-kubevirt-e2e-kind-k8s-sriov-1.14.2/1155957792633262080 is a run that failed for that reason, missing the 6m deadline.

It happened a couple of times here and there so I am not sure I can debug it. Maybe an alternative would be to print out the content of kubevirt namespace in case of failure?

cynepco3hahue: yep it can be useful you can also to print virt-operator log

fedepaol: ok, I'll close this and open another PR.

slintes: can you reproduce this long time locally? I'm wondering if there is something going wrong with the kind cluster.

fedepaol: Nope. For sure, I needed to relax some timeouts so maybe it's just that those machine where we have sriov cards are just less powerful.
Another thing I can do is to time it in this throw away PR and see how long it takes compared to the other providers / machines.

fedepaol:
Ok, here I added both the timing AND the troubleshooting output.
After the tests are ran, will compare the durations and leave only the troubleshooting output in case of failure.

fedepaol:
On kind (pnx17 machine):

```
 Before Tue Jul 30 10:25:48 UTC 2019 
 After Tue Jul 30 10:28:08 UTC 2019  
```

In pull-kubevirt-e2e-k8s-1.11.0:
```
Before Tue Jul 30 10:45:16 UTC 2019 
After Tue Jul 30 10:46:10 UTC 2019 
```

So nothing that could justify missing the 6m deadline. Will close this and opened https://github.com/kubevirt/kubevirt/pull/2547 for adding the debug info.

PR #2546, opened 2019-07-30
Enable sriov support by default

fedepaol:

This pr makes the sr-iov support enabled by default by removing the need for feature gate. 

**Release note**:

```release-note

SR-IOV support enabled by default, removed feature gate.

```

### DIFF
--- a/pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter_test.go
+++ b/pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter_test.go
@@ -1735,36 +1735,6 @@ var _ = Describe("Validating VMICreate Admitter", func() {
 			Expect(len(causes)).To(Equal(0))
 		})
 
-		It("should accept SRIOV interface when feature gate is disabled", func() {
-			vmi := v1.NewMinimalVMI("testvmi")
-			vmi.Spec.Domain.Devices.Interfaces = []v1.Interface{*v1.DefaultBridgeNetworkInterface()}
-			vmi.Spec.Networks = []v1.Network{*v1.DefaultPodNetwork()}
-
-			vmi.Spec.Domain.Devices.Interfaces = append(
-				vmi.Spec.Domain.Devices.Interfaces,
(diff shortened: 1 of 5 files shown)


pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter_test.go
```diff
@@ -1735,7 +1735,7 @@ var _ = Describe("Validating VMICreate Admitter", func() {
 			Expect(len(causes)).To(Equal(0))
 		})
 
-		It("should reject SRIOV interface when feature gate is disabled", func() {
+		It("should accept SRIOV interface when feature gate is disabled", func() {
```
booxter: In my mind, we have no code for SR-IOV in the admitter so it's arguable if we need to have any test cases for sriov here. But I am ok either way.

fedepaol:
> I'd like to better understand why we introduced the feature gate, and why we remove it now. Was this discussed somewhere?

My understanding is that sriov support was kind of experimental and we did not have proper ci around. Also, it's opt-in per se since it won't get activated if a sriov card is not explicitly added to a guest, but I'd also wait for @booxter if he has more info about this

booxter:
@slintes there were two main reasons: 1) no upstream CI for the feature; and 2) SR-IOV interface attached enabled privileged mode for the pod. Both issues are now tackled, hence now we enable it by default.

slintes:
@fedepaol ack, thanks! Works for me.
@rmohr is this ok for you too?
@fedepaol the user-guide needs an update as well.

/lgtm

fedepaol:
> @booxter ack, thanks! Works for me.
> @rmohr is this ok for you too?
> @fedepaol the user-guide needs an update as well.

Will do

> 
> /lgtm

fedepaol:
User guide pr: https://github.com/kubevirt/user-guide/pull/270

rmohr:
> @rmohr is this ok for you too?

Since CI is now in place, no issue with it. :+1:

PR #2548, opened 2019-07-30
Use HCO conditions on KubeVirt CR

slintes:
Use HCO conditions on KubeVirt CR

See https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/docs/conditions.md
Needs userguide update and heads up via mail for renaming `Ready` condition to `Available`, which is being used while waiting for a sucessfull deployment of KubeVirt

**Release note**:
```release-note
Added and renamed some KubeVirt CR conditions for usage by HCO
```

### DIFF
--- a/pkg/virt-operator/util/client.go
+++ b/pkg/virt-operator/util/client.go
@@ -20,6 +20,7 @@
 package util
 
 import (
+	"fmt"
 	"time"
 
 	secv1 "github.com/openshift/api/security/v1"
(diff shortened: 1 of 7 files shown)


pkg/virt-operator/util/client.go
```diff
@@ -74,7 +126,6 @@ func UpdateCondition(kv *virtv1.KubeVirt, conditionType virtv1.KubeVirtCondition
 	}
 
 	kv.Status.Conditions = conditions
-
 }
 
 func getCondition(kv *virtv1.KubeVirt, conditionType virtv1.KubeVirtConditionType) (*virtv1.KubeVirtCondition, bool) {
```
djzager: Is it possible for this to be made public so the HCO can get the conditions?

slintes: I leave it private. It does something you don't need (creating a condition if it does not exist), and I want to be able to change it if needed, without breaking HCO.
I think copying the few lines you need to HCO is good enough :slightly_smiling_face:

slintes:
@davidvossel thx for review! Merge conflict is resolved. Can you please LGTM again? :slightly_smiling_face:

PR #2552, opened 2019-07-31
Add a new reporter which collects cluster info into files

rmohr:

Upload in CI all info into separate files, instead of printing huge traces on every test failure which are only only of limited use.

It logs the cluster state for the first 10 failing tests. Prow will upload the files and make it available for browsing.

For each failing test the the following things are logged:

 * Logs of all pods, including previous logs, starting from the start time of the test
 * Specs of all pods in the cluster
 * Specs of all vmis in the cluster
 * All events starting from the start time of the test

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/reporter/kubernetes.go
+++ b/tests/reporter/kubernetes.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/onsi/ginkgo/config"
@@ -20,6 +21,7 @@ import (
 type KubernetesReporter struct {
 	failureCount int
 	artifactsDir string
+	mux          sync.Mutex
 }
 
 func NewKubernetesReporter(artifactsDir string) *KubernetesReporter {
@@ -44,6 +46,8 @@ func (r *KubernetesReporter) SpecWillRun(specSummary *types.SpecSummary) {
 }
 
 func (r *KubernetesReporter) SpecDidComplete(specSummary *types.SpecSummary) {
+	r.mux.Lock()
+	defer r.mux.Unlock()
 	if r.failureCount > 10 {
 		return
 	}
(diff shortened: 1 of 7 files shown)


tests/vmi_configuration_test.go
```diff
@@ -59,7 +59,7 @@ var _ = Describe("Configurations", func() {
 
 	Context("for CPU and memory limits should", func() {
 
-		It("lead to get the burstable QOS class assigned when limit and requests differ", func() {
+		FIt("lead to get the burstable QOS class assigned when limit and requests differ", func() {
```
fedepaol: Left the F here

tests/vmi_configuration_test.go
```diff
@@ -72,9 +72,10 @@ var _ = Describe("Configurations", func() {
 				}
 				return *vmi.Status.QOSClass
 			}, 10*time.Second, 1*time.Second).Should(Equal(kubev1.PodQOSBurstable))
+			Expect(true).To(BeFalse())
 		})
 
-		It("lead to get the guaranteed QOS class assigned when limit and requests are identical", func() {
+		FIt("lead to get the guaranteed QOS class assigned when limit and requests are identical", func() {
```
fedepaol: And here

tests/vmi_configuration_test.go
```diff
@@ -101,6 +102,7 @@ var _ = Describe("Configurations", func() {
 				}
 				return *vmi.Status.QOSClass
 			}, 10*time.Second, 1*time.Second).Should(Equal(kubev1.PodQOSGuaranteed))
+			Expect(true).To(BeFalse())
```
fedepaol: Also this

rmohr: Good eye. Yes, I left them there to test the reporter :)

tests/reporter/kubernetes.go
```diff
@@ -0,0 +1,221 @@
...
+	if r.artifactsDir != "" {
+		os.RemoveAll(r.artifactsDir)
+	}
+}
+
+func (r *KubernetesReporter) SpecWillRun(specSummary *types.SpecSummary) {
+}
+
+func (r *KubernetesReporter) SpecDidComplete(specSummary *types.SpecSummary) {
+	r.mux.Lock()
```
fedepaol: Why the need for the mutex? Are the specs expected to be executed in parallel?

rmohr: In theory. It was cheap to add. I can also remove it.

rmohr: and might be that ginkgo itself protects it too already.

fedepaol: Yes, my main concern here is that if they run in parallel, the first report passes the mutex, the other ones locks on it and the collected info is a mixture of all the specs blocked on the mutex.

rmohr: So there is only one instance of the reporter and only this one log. Every invocation has a unique file-prefix, to have individual files for each test which fails.

fedepaol: Yes, but the reporter collects all the info of all the vmis, pod, around so if we had multiple specs at the same time the reported logs would collect the infos of all the specs locked on the mutex, isn't it?

fedepaol: *in the same time window

rmohr: ah, absolutely right. I think it is not that bad, because they both see a similar cluster state

PR #2557, opened 2019-08-01
Fix handling error in migration wait loop

petrkotas:
Fix handling error in migration wait loop

Sometimes kube-api returns "unexpected EOF"
while calling GET for any object.

The error is rare and reported in
https://github.com/kubernetes-sigs/kubespray/issues/2322

The fix makes the loop resilient to these errors
by ignoring and retrying.

Also it makes the VM smaller, to enable faster migration.

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/2557)

### DIFF
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -228,19 +228,21 @@ var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system]
 		By("Waiting until the Migration Completes")
 
 		uid := ""
-		Eventually(func() bool {
+		Eventually(func() error {
 			migration, err := virtClient.VirtualMachineInstanceMigration(migration.Namespace).Get(migration.Name, &metav1.GetOptions{})
-			Expect(err).To(BeNil())
+			if err != nil {
+				return err
+			}
 
 			Expect(migration.Status.Phase).ToNot(Equal(v1.MigrationFailed))
 
 			uid = string(migration.UID)
 			if migration.Status.Phase == v1.MigrationSucceeded {
-				return true
+				return nil
 			}
-			return false
+			return fmt.Errorf("Migration is in the phase: %s", migration.Status.Phase)
 
-		}, timeout, 1*time.Second).Should(BeTrue())
+		}, timeout, 1*time.Second).ShouldNot(HaveOccurred(), fmt.Sprintf("migration should succeed after %d s", timeout))
 		return uid
 	}
 	runAndCancelMigration := func(migration *v1.VirtualMachineInstanceMigration, vmi *v1.VirtualMachineInstance, timeout int) string {


petrkotas:
The error seems to be related to kubernetes reported issue. 
https://github.com/kubernetes-sigs/kubespray/issues/2322

petrkotas:
I committed focused test by accident. Sorry for that. Fixed.

tests/migration_test.go
```diff
@@ -432,7 +434,7 @@ var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system]
 
 			It("should complete a migration", func() {
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
-				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse("1Gi")
+				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse("512M")
```
SchSeba: I think you can remove this line if you reduce the memory.

```golang
func NewRandomFedoraVMIWitGuestAgent() *v1.VirtualMachineInstance {
	agentVMI := NewRandomVMIWithEphemeralDiskAndUserdata(ContainerDiskFor(ContainerDiskFedora), GetGuestAgentUserData())
	agentVMI.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse("512M")
	return agentVMI
}
```

petrkotas:
@davidvossel the error is exactly the same as I have linked at the beginning https://github.com/kubernetes-sigs/kubespray/issues/2322. It is sometimes git, sometimes not. I believe it is the error related to kube-api, however it occurs so rarely I do not have enough data to fully understand what is happening.

davidvossel:
> @davidvossel the error is exactly the same as I have linked at the beginning kubernetes-sigs/kubespray#2322. It is sometimes git, sometimes not. I believe it is the error related to kube-api, however it occurs so rarely I do not have enough data to fully understand what is happening.

ah, ok I see. That's a generic connectivity error. If the migration object didn't exist, you'd see something to the effect of "not found" in the message. This error is indicating that the api server isn't accessible for some reason. 

These kinds of generic connectivity errors can really be caused by anything (often outside of our control). I don't have any issues with us building functional tests that are resilient to this type of error.

davidvossel:
@petrkotas Before this gets merged, can you just change anything in the description or commit msg that indicates we're waiting for the migration to appear in etcd. I don't want anyone to be confused about that being something we need to account for when it's not. This is something that should be completely impossible since around k8s 1.9.

petrkotas:
Hi @davidvossel thanks for the review. I have updated both, commit message and PR comment.
You are right, it would be confusing.

tests/migration_test.go
petrkotas: Good catch @SchSeba, thanks.

rmohr:
/hold

This makes me uneasy. While I also don't mind making the tests resilient, consider the following: We have now roughly 360 tests. Summing up all lanes we run more than 1000 tests. Still we have a set of less than 10 tests which fail often enough to make our tests unreliable. These migration tests are part of that.

petrkotas:
@rmohr I am not sure I understand your comment. Code in this PR is designed to make the tests more stable.

One of the reasons the migration failed, was due to the "unexpected EOF" error, which KubeVirt codebase does not have any control of.

What issue are you seeing? Can it be addressed?

rmohr:
> @rmohr I am not sure I understand your comment. Code in this PR is designed to make the tests more stable.
> 
> One of the reasons the migration failed, was due to the "unexpected EOF" error, which KubeVirt codebase does not have any control of.
> 
> What issue are you seeing? Can it be addressed?

What I mean is that there is very likely an issue hidden. We don't see other tests failing often. The chance is hight that this just masks an issue.

davidvossel:
> What I mean is that there is very likely an issue hidden. We don't see other tests failing often. The chance is hight that this just masks an issue.

is this error where the migration test is actually failing though?  I doubt this fixes the flakiness of the test on CI as much as it accounts for issues people may encounter locally on systems with limited resources.  my guess is the apiserver is briefly unresponsive during the migration because of system load.

petrkotas:
@davidvossel and @rmohr this indeed does not fix nor mask flakiness. It simply solves the issue with kube-api. The tests sometimes fail for kube-api error, not our and it makes debugging flaky tests harder.

rmohr:
> @davidvossel and @rmohr this indeed does not fix nor mask flakiness. It simply solves the issue with kube-api. The tests sometimes fail for kube-api error, not our and it makes debugging flaky tests harder.

IMO it does not help nor make things worse. After this PR is  in we are exactly where we were before. Canceling the hold because it will also not destroy anything.

/hold cancel

petrkotas:
/retest
________________________________

petrkotas:
Good morning @slintes sorry for the delay. I have taken care of the rebase.

PR #2559, opened 2019-08-01
Increase cloud init user-data max size to 16384 bytes

pbudds:
The current cloud-init user-data max size is 2048 bytes which is insufficient. Increasing it to 16384 brings it to a more reasonable size that will work for wider use cases.

Fixes #

**Release note**:

```release-note
Cloud-init user-data max supported length will be 16384 bytes.
```


pbudds:
@petrkotas - Thank you for looking at this PR. The use case I'm looking at is to customize a stock ubuntu image using cloud init and if one happens to use "User-Data script" which allows for a lot more customization, the user-data can quickly become quite large.

As for squashing the commits, I'll work on redoing this on a clean branch. Ran into rebase issues on this branch.  Sorry for the noise.

rmohr:
@pbudds we made it deliberately small. The background is a technical reason which would hurt as drastically at scale (every mutation of the object sends the whole object to multiple receivers).

Are you aware of the possibility to use a secret instead (https://kubevirt.io/user-guide/docs/latest/creating-virtual-machines/startup-scripts.html#cloud-init-userdata-as-k8s-secret)?

Let us know if this does  not work for you.

pbudds:
@rmohr - Thank you for the insight. That makes sense. I will try the config you suggested. Seems like it should work for my case. Any reason why the NetworkDataMaxLen was still left at 16K?

rmohr:
Hm, sounds like a bug.

pbudds:
@rmohr  and @petrkotas - Using k8s secret for large user-data works for me. Thank you. I will close this pull request. If you guys agree I can do the needful to change NetworkDataMaxLen to 2K.

petrkotas:
Hi @pbudds we are glad to help. Also we would be more than happy to accept your PR.
Please do the change.

PR #2563, opened 2019-08-03
Added functional test for custom MAC address for SR-IOV

booxter:
**What this PR does / why we need it**: test custom MAC address for SR-IOV interfaces in kubevirt context

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/vmi_multus_test.go
+++ b/tests/vmi_multus_test.go
@@ -634,6 +634,11 @@ var _ = Describe("SRIOV", func() {
 			// fedora requires some more memory to boot without kernel panics
 			vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceName("memory")] = resource.MustParse("1024M")
 
+			return
+		}
+
+		startVmi := func(vmi *v1.VirtualMachineInstance) {
+
 			_, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
 			Expect(err).ToNot(HaveOccurred())
 			tests.WaitUntilVMIReady(vmi, tests.LoggedInFedoraExpecter)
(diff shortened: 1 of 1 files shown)


booxter:
/hold

There is a legit CI failure in the sriov job. I suspect that the order of interfaces is not deterministic (probably depends on VF PCI IDs allocated - can be eth0 or eth1), hence the failure. I probably need to adjust the test case so that it doesn't assume a particular name of the interface.

SchSeba:
@booxter I really don't know

booxter:
I think I can work around the order issue by iterating over interfaces and checking the one that has an expected vendor ID (!= virtio ID since VFs may be of different vendors).

/hold

booxter:
/hold cancel

There was some issue with previous version where I used the first interface name for SR-IOV interface, despite it's clearly always eth1+ as per other test cases. The latest version should be fine.

There still may be a theoretical issue with ordering interface names in the suite but we'll deal with that if it actually bites us, and we'll need to deal with it for the whole suite, not just this test case.

booxter:
@SchSeba @phoracek it should be ready for review. Thanks.

booxter:
@phoracek can we merge it? or do we expect the last job to pass? code wise, there is no change this PR affects it, but maybe if I rebase then there is some fix in newest master that would tackle it. Is it worth a rebase?

slintes:
> @phoracek can we merge it? or do we expect the last job to pass? code wise, there is no change this PR affects it, but maybe if I rebase then there is some fix in newest master that would tackle it. Is it worth a rebase?

A rebase is not needed, tide (the prow component for merging) is testing PRs against latest commit on master. We "only" need to stop merging manually, because that will make tide to start tests all over again because of the new head on master... let we check with the team if we want to wait for tide or merge manually.

PR #2568, opened 2019-08-06
Fix virt-handler restart issue

rmohr:

First selinux does not like if a fcontext is redefined. Check if we already
added our labels and skip adding the label if we did. The issue fixed looks like this:

```
panic: failed to set label for directory /var/run/kubevirt(/.*)?: ValueError: File context for /var/run/kubevirt(/.*)? already defined
```

Second, don't truncate the container-disk binary on virt-handler restarts. Instead of that delete it. This allows running VMIs to continue using the containerDisk binary, while virt-handler can already copy over the new version. The issue fixed looks like this:

```
panic: failed to crate target file: open /var/lib/kubevirt/init/usr/bin/container-disk: text file busy
```

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/cmd/virt-handler/virt-handler.go
+++ b/cmd/virt-handler/virt-handler.go
@@ -423,6 +423,9 @@ func main() {
 
 func copy(sourceFile string, targetFile string) error {
 
+	if err := os.RemoveAll(targetFile); err != nil {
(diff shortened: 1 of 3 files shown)


rmohr:
/cc @stu-gott 

that needs to be part of the upcoming release.

PR #2573, opened 2019-08-06
Implement ServiceMonitor to integrate OpenShift's Prometheus monitoring

stu-gott:
This PR will create a ServiceMonitor resource for Prometheus using the KubeVirt operator. This is enabled by defining the serviceMonitorNamespace (and possibly serviceMonitorAccount) in the KubeVirt custom resource.

Test cases aren't yet written for this, but I wanted feedback in case others disagree with this approach.

**Release note**:
```release-note
KubeVirt can deploy a ServiceMonitor resource to prompt Prometheus to scrape KubeVirt's endpoints.
```


pkg/virt-operator/util/types.go
```diff
@@ -40,6 +40,7 @@ type Stores struct {
 	InstallStrategyJobCache       cache.Store
 	InfrastructurePodCache        cache.Store
 	PodDisruptionBudgetCache      cache.Store
+	ServiceMonitorCache           cache.Store
```
slintes: This store needs to be initialized. And you need to check if the `ServiceMonitor` CRD exists. We do something similar for SCCs here: https://github.com/kubevirt/kubevirt/blob/d6832ae4e1802aa8c35d46c27d73059bdb0ea7ad/pkg/virt-operator/application.go#L166

staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -1220,6 +1220,15 @@ type KubeVirtSpec struct {
 
 	// The ImagePullPolicy to use.
 	ImagePullPolicy k8sv1.PullPolicy `json:"imagePullPolicy,omitempty" valid:"required"`
+
+	// The namespace the Prometheus ServiceMonitor
```
slintes: IIUC this is the namespace of Prometheus, not the ServiceMonitor (which we install in our own namespace atm)?

staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -1220,6 +1220,15 @@ type KubeVirtSpec struct {
...
+	// ServiceMonitor related resources will not be enabled if this is omitted.
+	// For openshift clusters with Prometheus, this is normally: openshift-monitor
+	ServiceMonitorNamespace string `json:"serviceMonitorNamespace,omitempty"`
+
+	// The name of the Prometheus ServiceMonitor service account
+	// Defaults to prometheus-k8s
```
slintes: same here

staging/src/kubevirt.io/client-go/api/v1/types.go
stu-gott: You are correct! That wording/naming can certainly be improved.

pkg/virt-operator/kubevirt_test.go
```diff
@@ -273,6 +277,8 @@ var _ = Describe("KubeVirt Operator", func() {
 		virtClient.EXPECT().SecClient().Return(secClient).AnyTimes()
 		virtClient.EXPECT().ExtensionsClient().Return(extClient).AnyTimes()
 		virtClient.EXPECT().PolicyV1beta1().Return(kubeClient.PolicyV1beta1()).AnyTimes()
+		// FIXME: is this needed?
```
slintes: I think it is neeed :)

pkg/virt-operator/kubevirt_test.go
```diff
@@ -702,6 +717,8 @@ var _ = Describe("KubeVirt Operator", func() {
 		handler, _ := components.NewHandlerDaemonSet(NAMESPACE, config.GetImageRegistry(), config.GetApiVersion(), config.GetImagePullPolicy(), config.GetVerbosity())
 		all = append(all, apiDeployment, apiDeploymentPdb, controller, controllerPdb, handler)
 
+		// FIXME: this will be needed once these tests are stable
+		//all = append(all, components.NewServiceMonitorCR(NAMESPACE, true))
```
slintes: I think this should be something like this:
```
		if config.IsServiceMonitorEnabled() {
			all = append(all, rbac.GetAllServiceMonitor(NAMESPACE, config.GetServiceMonitorNamespace(), config.GetServiceMonitorServiceAccount())...)
			all = append(all, components.NewServiceMonitorCR(NAMESPACE, true))
		}
```

staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -1220,6 +1220,14 @@ type KubeVirtSpec struct {
 
 	// The ImagePullPolicy to use.
 	ImagePullPolicy k8sv1.PullPolicy `json:"imagePullPolicy,omitempty" valid:"required"`
+
+	// The namespace Prometheus is deployed in
+	// Defaults to openshift-monitor
+	monitorNamespace string `json:"monitorNamespace,omitempty"`
```
slintes: field name needs to be uppercase, else it won't be (de-)serialized. Actually this is the reason why a json tag is always added AFAIK: field name needs to be uppercase, best practice for json keys is lowercase. Otherwise json key would default to field (uppercase) field name.

staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -1220,6 +1220,14 @@ type KubeVirtSpec struct {
...
+
+	// The name of the Prometheus service account that needs read-access to KubeVirt endpoints
+	// Defaults to prometheus-k8s
+	monitorAccount string `json:"monitorAccount,omitempty"`
```
slintes: same

stu-gott:
@slintes Functional tests added. I believe this PR should be ready for a critical review now.

slintes:
will review, but there are unusual many func tests failing...

/retest

tests/operator_test.go
```diff
@@ -1152,4 +1152,55 @@ spec:
...
+			}
+		})
+
+		It("Should allow Prometheus to scrape KubeVirt endpoints", func() {
+			coreClient := virtClient.CoreV1()
+
+			By("Obtaining Prometheus' configuration data")
+			secret, err := coreClient.Secrets("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			data := secret.Data["prometheus.yaml"]
```
slintes: good idea to check it there :+1:

fabiand:
Could it be that there is an issue with the PR? As I see that at least #2602 became green in the end.

stu-gott:
I've assumed every test failure on this was something related to this PR, and have been debugging. So far every test failure appears to be due to flakey tests--including the two sha-related failures in operator tests. @slintes can you double check this?

fabiand:
I see a couple of failures with `context deadline exceeded` and this leads (via google) to prometheus. Thus looks related

`The VirtualMachineInstance crashed.` appears with many tests - any ideas why this happens?

slintes:
os-3.11.0 lane still looks bad :disappointed:

slintes:
Wow, this little bug fix [0] solved all problems. It fixes the GetDomainStats function, which is called when Prometheus is scraping KubeVirt. And with this PR we completed the configuration for Prometheus for the first time, so this is the first we hit this bug (at leas in CI). The bug resulted in crashing VMs when is Prometheus is scraping for KubeVirt metrics, and with that in many test failures, which looked unrelated.

/lgtm
/approve

[0] https://github.com/kubevirt/kubevirt/pull/2573/commits/0b480eefc054b88444ca9d197e90a1537c1d8a69

fromanirh:
kudos team and @stu-gott for the good work in this PR!

fabiand:
@slintes @stu-gott - In retrospective: Is there something we can add to make debugging this kind of failures more easy?
Or rather to identify that it is really an issue?

slintes:
> @slintes @stu-gott - In retrospective: Is there something we can add to make debugging this kind of failures more easy?
> Or rather to identify that it is really an issue?

from my side I can say that I concentrated too much on the PR itself and the operator tests only, which didn't fail that often. Once I looked into the virt-launcher logs of other failing tests (which are nicely available since a recent change of Roman via the artifacts link of a job summary), finding the bug was quite easy. So, everything is available, we just need to use it.

stu-gott:
@fabiand This is kind of a special case in the sense that the PR enabled already-merged code--so the scope of what was in play was larger than we realized.

stu-gott:
@slintes Last test run had a single test failure.

PR #2584, opened 2019-08-09
Remove SYS_RESOURCE capability from launcher pod

booxter:
Instead, set the (unlimited) limit for libvirtd from handler pod that
is already privileged.

```release-note
Remove SYS_RESOURCE capability from SR-IOV attached VMI launcher pods
```

### DIFF
--- a/pkg/virt-handler/isolation/isolation.go
+++ b/pkg/virt-handler/isolation/isolation.go
@@ -193,6 +193,8 @@ func (s *socketBasedIsolationDetector) AdjustResources(vm *v1.VirtualMachineInst
 }
 
 // consider reusing getMemoryOverhead()
+// This is not scientific, but neither what libvirtd does is. See details in:
+// https://www.redhat.com/archives/libvirt-users/2019-August/msg00051.html
 func getMemlockSize(vm *v1.VirtualMachineInstance) (int64, error) {
 	memlockSize := resource.NewQuantity(0, resource.DecimalSI)
 
--- a/pkg/virt-handler/isolation/isolation.go
+++ b/pkg/virt-handler/isolation/isolation.go
@@ -40,6 +40,7 @@ import (
 	"golang.org/x/sys/unix"
 
 	ps "github.com/mitchellh/go-ps"
+	"k8s.io/apimachinery/pkg/api/resource"
 
 	v1 "kubevirt.io/client-go/api/v1"
 	"kubevirt.io/client-go/log"
@@ -166,10 +167,14 @@ func (s *socketBasedIsolationDetector) AdjustResources(vm *v1.VirtualMachineInst
 			continue
 		}
 
-		// TODO estimate actual memory size needed
+		// make the best estimate for memory required by libvirt
+		memlockSize, err := getMemlockSize(vm)
+		if err != nil {
+			return err
+		}
 		rLimit := unix.Rlimit{
-			Max: unix.RLIM_INFINITY,
-			Cur: unix.RLIM_INFINITY,
+			Max: uint64(memlockSize),
+			Cur: uint64(memlockSize),
 		}
 		err = prLimit(process.Pid(), unix.RLIMIT_MEMLOCK, &rLimit)
 		if err != nil {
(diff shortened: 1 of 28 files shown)


booxter:
/hold

Need to consider calculating the exact limit size in handler (the formula is not trivial but we can do a reasonable estimate).

slintes:
heads up, we just merged a PR which introduces kubevirt specific SCCs instead of using the privileged SCC. The capability can be removed there as well I guess: https://github.com/kubevirt/kubevirt/blob/548ed6342cc0c3def45f1d76b0dc1abaa6441b93/pkg/virt-operator/creation/components/scc.go#L77

booxter:
Unit test coverage is somewhat down because the code dealing with processes / ulimits is not unit tested. But it should be enough to cover it for regressions with existing functional tests: as long as VMIs still start correctly and have SR-IOV interfaces inside the guest, it should be enough. There is little value in validating that the capability is indeed not present anymore because it's trivial to double check it's not referenced anywhere in the code anymore.

booxter:
@phoracek @SchSeba this PR should be ready to review and merge.

pkg/virt-handler/isolation/isolation.go
```diff
@@ -120,6 +128,59 @@ func (s *socketBasedIsolationDetector) Detect(vm *v1.VirtualMachineInstance) (*I
 	return NewIsolationResult(pid, slice, controller), nil
 }
 
+// standard golang libraries don't provide API to set runtime limits
+// for other processes, so we have to directly call to kernel
+func prLimit(pid int, limit uintptr, rlimit *unix.Rlimit) error {
+	_, _, errno := unix.RawSyscall6(unix.SYS_PRLIMIT64,
```
phoracek: in previous commit you dropped RawSyscall, can we do it also here?

pkg/virt-handler/isolation/isolation.go
```diff
@@ -120,6 +128,59 @@ func (s *socketBasedIsolationDetector) Detect(vm *v1.VirtualMachineInstance) (*I
...
+		0, 0, 0)
+	if errno != 0 {
+		return fmt.Errorf("Error setting prlimit: %v", errno)
+	}
+	return nil
+}
+
+func (s *socketBasedIsolationDetector) AdjustResources(vm *v1.VirtualMachineInstance) error {
+	// bump memlock ulimit for libvirtd
+	res, err := s.Detect(vm)
```
phoracek: could we use a more specific name than result?

pkg/virt-handler/isolation/isolation.go
```diff
@@ -120,6 +128,59 @@ func (s *socketBasedIsolationDetector) Detect(vm *v1.VirtualMachineInstance) (*I
...
+		// TODO estimate actual memory size needed
+		rLimit := unix.Rlimit{
+			Max: unix.RLIM_INFINITY,
+			Cur: unix.RLIM_INFINITY,
+		}
+		err = prLimit(process.Pid(), unix.RLIMIT_MEMLOCK, &rLimit)
+		if err != nil {
+			return fmt.Errorf("failed to set rlimit for memory lock: %v", err)
+		}
+		// we assume a single process should match
```
phoracek: What if we have side cars? Would that add other processes?

pkg/virt-handler/isolation/isolation.go
phoracek: (won't block the PR on this one, if you answer no no no to my comments, this PR is good to go)

pkg/virt-handler/isolation/isolation.go
booxter: no because this syscall is not implemented in standard libraries.

I dropped raw syscall in a separate commit because it was brought to my attention that we have similar code in another place, but instead of reusing the new prLimit function there, it was easier and better to switch to a proper library call.

pkg/virt-handler/isolation/isolation.go
booxter: it's of "IsolationResult" type so I thought it's a good enough name. I am ok with renaming. What would be a better name?

pkg/virt-handler/isolation/isolation.go
booxter: Other side cars running libvirtd process?

phoracek: Ah, my bad, misunderstood

pkg/virt-handler/isolation/isolation.go
phoracek: it is being used only in this block and i don't really know. since the other comments are resolved, i wont bother

pkg/virt-handler/vm_test.go
```diff
@@ -1286,6 +1288,24 @@ func (m *MockGracefulShutdown) TriggerShutdown(vmi *v1.VirtualMachineInstance) {
 	Expect(err).NotTo(HaveOccurred())
 }
 
+type MockIsolationDetector struct{}
```
SchSeba: This should be auto generated?

booxter: OK I've found a way to make it work without manually writing a fake.

pkg/virt-handler/isolation/isolation.go
```diff
@@ -120,6 +129,84 @@ func (s *socketBasedIsolationDetector) Detect(vm *v1.VirtualMachineInstance) (*I
...
+// consider reusing getMemoryOverhead()
+func getMemlockSize(vm *v1.VirtualMachineInstance) (int64, error) {
+	memlockSize := resource.NewQuantity(0, resource.DecimalSI)
+
+	// start with base memory requested for the VM
+	vmiMemoryReq := vm.Spec.Domain.Resources.Requests.Memory()
+	memlockSize.Add(*resource.NewScaledQuantity(vmiMemoryReq.ScaledValue(resource.Kilo), resource.Kilo))
+
+	// allocate 1Gb for VFIO needs
+	memlockSize.Add(resource.MustParse("1G"))
```
booxter: I just realized this line should be avoided if no VFIO devices are attached. Perhaps we can even skip the whole prlimit() call if it's not SR-IOV VMI.

booxter:
/hold

There is a way to optimize the code for non-SRIOV case (by skipping the whole prlimit() business).

booxter:
/hold cancel

Now the ulimit adjustment happens only for VFIO attached VMIs. Also rebased the branch to latest master in hope that some job failures go away.

slintes:
@booxter Travis is red, please run make generate

PR #2590, opened 2019-08-11
Improve prometheus functional tests

ahadas:
Refactor the functional tests of the (prometheus based) metrics and reduce their execution time.
Specifically:
1. Most of the tests rely on having a running VMI. This PR changes the timing of when that VMI is started to be at the beginning (i.e., `BeforeAll`) rather than started at each of these tests.
2. Remove unused timeout parameter for synchronous tests.
3. Remove unused expecter in the test of memory metrics  - it used to serve as a sync point, ensuring that the VM completed the boot process. Alternatively, we now run the portion that writes to the disk using another expecter in the `BeforeAll` phase and so we don't need that expecter anymore.
4. Reduce the scope of functions that are intended for the prometheus metrics tests - this also enables them to access the "shared state" of the aforementioned running VMI, its pod and its metrics URL directly.
5. The functions mentioned above are split in a way they can be used in more places, reducing code duplication within the tests. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/infra_test.go
+++ b/tests/infra_test.go
@@ -58,6 +58,53 @@ var _ = Describe("Infrastructure", func() {
 		var pod *k8sv1.Pod
 		var metricsURL string
 
+		// start a VMI, wait for it to run and return the node it runs on
+		startVMI := func(vmi *v1.VirtualMachineInstance) string {
+			By("Starting a new VirtualMachineInstance")
+			obj, err := virtClient.RestClient().Post().Resource("virtualmachineinstances").Namespace(tests.NamespaceTestDefault).Body(vmi).Do().Get()
+			Expect(err).ToNot(HaveOccurred(), "Should create VMI")
+
+			By("Waiting until the VM is ready")
+			return tests.WaitForSuccessfulVMIStart(obj)
+		}
+
+		// returns metrics from the node the VMI(s) runs on
+		getKubevirtVMMetrics := func() string {
+			stdout, _, err := tests.ExecuteCommandOnPodV2(virtClient,
+				pod,
+				"virt-handler",
+				[]string{
+					"curl",
+					"-L",
+					"-k",
+					metricsURL,
+				})
+			Expect(err).ToNot(HaveOccurred())
+			return stdout
+		}
+
+		// collect metrics whose key contains the given string, expects non-empty result
+		collectMetrics := func(metricSubstring string) map[string]float64 {
+			By("Scraping the Prometheus endpoint")
+			var metrics map[string]float64
+			var lines []string
+
+			Eventually(func() map[string]float64 {
+				out := getKubevirtVMMetrics()
+				lines = takeMetricsWithPrefix(out, metricSubstring)
+				metrics, err = parseMetricsToMap(lines)
+				Expect(err).ToNot(HaveOccurred())
+				return metrics
+			}, 30*time.Second, 2*time.Second).ShouldNot(BeEmpty())
+
+			// troubleshooting helper
+			fmt.Fprintf(GinkgoWriter, "metrics [%s]:\nlines=%s\n%#v\n", metricSubstring, lines, metrics)
+			Expect(len(metrics)).To(BeNumerically(">=", float64(1.0)))
+			Expect(metrics).To(HaveLen(len(lines)))
+
+			return metrics
+		}
+
 		tests.BeforeAll(func() {
 			tests.BeforeTestCleanup()
 			By("Creating the VirtualMachineInstance")
@@ -70,17 +117,11 @@ var _ = Describe("Infrastructure", func() {
 			vmi = tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskAlpine))
 			tests.AppendEmptyDisk(vmi, "testdisk", "virtio", "1Gi")
 
-			By("Starting a new VirtualMachineInstance")
-			obj, err := virtClient.RestClient().Post().Resource("virtualmachineinstances").Namespace(tests.NamespaceTestDefault).Body(vmi).Do().Get()
-			Expect(err).ToNot(HaveOccurred(), "Should create VMI")
-
-			By("Waiting until the VM is ready")
-			nodeName := tests.WaitForSuccessfulVMIStart(obj)
+			nodeName := startVMI(vmi)
 
 			By("Finding the prometheus endpoint")
 			pod, err = kubecli.NewVirtHandlerClient(virtClient).ForNode(nodeName).Pod()
 			Expect(err).ToNot(HaveOccurred(), "Should find the virt-handler pod")
-
 			metricsURL = fmt.Sprintf("https://%s:%d/metrics", pod.Status.PodIP, 8443)
 		})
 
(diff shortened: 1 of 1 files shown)


tests/infra_test.go
```diff
@@ -53,11 +53,36 @@ var _ = Describe("Infrastructure", func() {
...
-	BeforeEach(func() {
-		tests.BeforeTestCleanup()
-	})
-
 	Describe("Prometheus Endpoints", func() {
+		var vmi *v1.VirtualMachineInstance
+		var pod *k8sv1.Pod
+		var metricsURL string
+
+		tests.BeforeAll(func() {
```
fromanirh: In general I don't like to share resources among tests. It breaks the rule that each test should be self-sufficient and isolated. This case however is borderline, we consume various metrics and we change a little, if not at all, the shared resource (the VMI instance), so it could go for the sake of practicality.

tests/infra_test.go
```diff
@@ -448,41 +418,32 @@ func takeMetricsWithPrefix(output, prefix string) []string {
 	return ret
 }
 
-// shortcut helper
-func newRandomVMIWithMetrics(virtClient kubecli.KubevirtClient, metricSubstring string) ([]*v1.VirtualMachineInstance, string, map[string]float64) {
-	return newRandomVMIsWithMetrics(1, "", virtClient, metricSubstring)
-}
-
-func newRandomVMIsWithMetrics(count int, preferredNodeName string, virtClient kubecli.KubevirtClient, metricSubstring string) ([]*v1.VirtualMachineInstance, string, map[string]float64) {
+func newRandomVMIWithMetrics(virtClient kubecli.KubevirtClient, preferredNodeName, metricSubstring string) ([]*v1.VirtualMachineInstance, string, map[string]float64) {
```
fromanirh: the simplification looks OK, but I think this utility function can be kept as is, should we need to add more than a single VMI to the future test cases.

tests/infra_test.go
```diff
@@ -219,12 +219,6 @@ var _ = Describe("Infrastructure", func() {
 		}, 300)
 
 		It("should include the network metrics for a running VM", func() {
-			By("Expecting the VirtualMachineInstance console")
-			expecter, err := tests.LoggedInAlpineExpecter(vmi)
-
-			Expect(err).ToNot(HaveOccurred())
-			defer expecter.Close()
-
```
fromanirh: We added this expecter to act as sync point to make sure the VM completed the boot, and to reduce risk of false negatives. Do you think this risk is mitigated elsewhere? (or, even better, no longer relevant?)

tests/infra_test.go
```diff
@@ -58,6 +58,53 @@ var _ = Describe("Infrastructure", func() {
 		var pod *k8sv1.Pod
 		var metricsURL string
 
+		// start a VMI, wait for it to run and return the node it runs on
+		startVMI := func(vmi *v1.VirtualMachineInstance) string {
```
fromanirh: what's the benefit of defining those functions inline instead of having them free functions like before? Could you please amend the commit message to describe it?

tests/infra_test.go
```diff
@@ -200,49 +235,29 @@ var _ = Describe("Infrastructure", func() {
...
-			lines := takeMetricsWithPrefix(out, "kubevirt")
-			metrics, err := parseMetricsToMap(lines)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(metrics).To(HaveKey(ContainSubstring("kubevirt_vmi_network_")))
-
+			metrics := collectMetrics("kubevirt_vmi_network_")
 			By("Checking the collected metrics")
 			keys := getKeysFromMetrics(metrics)
 			for _, key := range keys {
-				if strings.HasPrefix(key, "kubevirt_vmi_network_") {
```
fromanirh: I'm not sure about possible false negatives here (and yes, this should have been documented with a comment)

tests/infra_test.go
```diff
@@ -57,6 +57,7 @@ var _ = Describe("Infrastructure", func() {
 		var vmi *v1.VirtualMachineInstance
 		var pod *k8sv1.Pod
 		var metricsURL string
+		var cachedMetrics string
```
fromanirh: I'm not sure about caching. It goes against the basic tenet to exercise the flows as much as we can, and adds another shared resource. What's the benefit of using cache? How much time do we save?

tests/infra_test.go
```diff
@@ -141,7 +141,7 @@ var _ = Describe("Infrastructure", func() {
 				&expect.BExp{R: "localhost:~#"},
 				&expect.BSnd{S: "sync\n"},
 				&expect.BExp{R: "localhost:~#"},
-			}, 10*time.Second)
```
fromanirh: In general, OK. I used this value to be overcautious against false negatives on slow CI. I don't have any hard data to back any choice here, so your guess (if not please document why is not a guess) is good as mine.

tests/infra_test.go
ahadas: Right, it is also a common practice in the existing functional tests

tests/infra_test.go
ahadas: Seems like you looked at a middle step in this PR, right?
What bothers me in this function is that it does two things - run the VMI and gather its metrics. The former can be used for running the first VM and the latter seems to be redundant at this point (as we intentionally target 2 VMs due to resource constraints in the CI, right?)
Maybe we can discuss the final step, where this function is removed, and think if it makes sense or not

tests/infra_test.go
ahadas: Ah I see, I didn't think about this reason. So now we have the expecter at the BeforeAll that runs the writing to the disk which renders this redundant I think, right?

tests/infra_test.go
ahadas: Well, the benefit is that it can access the "state" of the virtclient, the pod and the vmi directly and it also limits their scope to the prometheus tests

tests/infra_test.go
ahadas: Can you please explain this? the reason I removed this is because we collect all metrics whose key starts with `kubevirt_vmi_network_` (which fails if there're no such metrics reported) so in light of that, this part seems like a redundant filtering

tests/infra_test.go
ahadas: Yeah, good point, I'm not sure about it myself :) it surely saves time but at the expense of what you have mentioned. I'll need to measure the time saved

tests/infra_test.go
ahadas: Oh that's something I didn't fully tested yet, wanted to check what this duration actually means - if this time includes the time it takes to interact with the pod then yeah, it might be risky to lower it that much

tests/infra_test.go
ahadas: Kept this

tests/infra_test.go
ahadas: Removed the caching

tests/infra_test.go
ahadas: Changed it back. Apparently it's not an execution time but rather a timeout - so considering the `dd` operation is fairly quick, we (probably) never get to this timeout anyway

ahadas:
@fromanirh addressed what appeared to be problematic, could you please follow up on the changes?

tests/infra_test.go
ahadas: Added a comment to the relevant place

tests/infra_test.go
fromanirh: Yes, I was reviewing a commit a time in chronological order. And yes, I rolled two responsabilities in one function. It's probably better to split them.

tests/infra_test.go
fromanirh: A comment is good enough for me.

tests/infra_test.go
fromanirh: I'm not a fan of this, it seems to me the benefits aren't too clear - and this feels... too dynamic. But it', not fighting this too much.

tests/infra_test.go
fromanirh: On second thought it is probably redundant, yes. Let's simplify things.

slintes:
thx for reviewing @fromanirh !

/lgtm
/approve

PR #2592, opened 2019-08-12
Fixed release process

slintes:
- fixed pushing of client-go by using a local branch name that is
different from the tag, they can't be the same (but TRAVIS_TAG and
TRAVIS_BRANCH are the same on a tag)
- moved pushing of CSV manifest after the github release, so it can't
break the release anymore.

**Release note**:
```release-note
NONE
```


slintes:
cherrypick release-0.20

PR #2595, opened 2019-08-12
Qemu bug

ahadas:
Previously, the solution for the disappearing unix sockets for VNC (due
to a QEMU bug) was to have virt-launcher track when the unix socket for
VNC connection was created and then make its parent folder read-only.

In 944e019, we changed virt-handler to make the parent folder of
the unix socket for VNC read-only when a new connection request comes
and we find that the unix socket exists. This resolves a race in which
a VNC connection is established before virt-launcher manages to change
the parent folder of the unix socket read-only.

Here we remove the former solution as the latter renders it redundant.
Upgrade flow is covered since we assume upgrades from the previous
version where console connections are established through virt-handler.

That is cleanup for  #2519 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/virt-launcher/virt-launcher.go
+++ b/cmd/virt-launcher/virt-launcher.go
@@ -292,26 +292,6 @@ func waitForFinalNotify(deleteNotificationSent chan watch.Event,
 	}
 }
 
-// writeProtectPrivateDir waits until the kubevirt private vnc socket exists and than mark its folder as read only
-// this is a workaround preventing QEMU from deleting its sockets prematurely as described in a bug https://bugs.launchpad.net/qemu/+bug/1795100
-// once the QEMU 4.0 is released the need for this workaround goes away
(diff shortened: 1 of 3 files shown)


ahadas:
Oops, sorry - missed a potential issue during upgrade flow: we may have updated virt-handler while virt-api is not updated yet, in that case virt-api initiate the console connection not through virt-handler and so no one will change the folder to read-only... I guess we can remove it only together with removing that `sock-connector` script or when upgrading to QEMU 4 :/

ahadas:
Sorry about that but I think I missed a potential issue during upgrade flow: in a scenario where virt-controller is updated and a VMI is launched + a new console request for that VMI is received before virt-api is updated, the console connection request will go through directly from virt-api to virt-launcher (not through virt-handler) and so in that case the parent folder will not change to read-only.. the same reason we couldn't remove `sock-connector` yet...

slintes:
But this is for the next version, isn't it? The "old" virt-apiserver is already connecting to virt-handler.

PR #2599, opened 2019-08-13
Add polarion test case ids regarding PDB test cases

dhiller:

Adds ids to tests to connect test cases to polarion test plans.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/virt_control_plane_test.go
+++ b/tests/virt_control_plane_test.go
@@ -45,9 +45,6 @@ const (
 
 var _ = Describe("KubeVirt control plane resilience", func() {
 
-	var nodeNames []string
-	var selectedNodeName string
-
 	RegisterFailHandler(Fail)
 
 	virtCli, err := kubecli.GetKubevirtClient()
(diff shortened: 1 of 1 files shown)


tests/virt_control_plane_test.go
```diff
@@ -260,7 +260,7 @@ var _ = Describe("KubeVirt control plane resilience", func() {
 		eventuallyWithTimeout(waitForDeploymentsToStabilize)
 	})
 
-	When("evicting pods of control plane, last eviction should fail", func() {
+	When("[ref_id:2717]evicting pods of control plane, last eviction should fail", func() {
```
dshchedr: I'm not sure if we can use ref_id in "When" block, Is it possible to add "Describe" block that will wrap all this tests?

dhiller: The [documentation for When](https://github.com/onsi/ginkgo/blob/master/ginkgo_dsl.go#L343) says

> Describe, Context and When blocks are functionally equivalent.  The difference is purely semantic -- you typical Describe the behavior of an object
> or method and, within that Describe, outline a number of Contexts and Whens.

That's exactly what I've done here. The outmost is a `Describe` with two nested `When`s. 

So to me this looks like a missing implementation in the "translation" towards Polarion.

Would it be much effort to change the "translation" towards polarion? And is it not possible to look up whether this is already done?

dhiller: When looking at the code I found a couple of places where `Context` contained `ref_id`, so I would guess that `When` should work also.

slintes: Agree, `Describe`, `Context` and `When` should be handled the same way. I think this is only the first time that `When` is used. Actually, before this I also didn't know it exists :wink: 

@dshchedr Would be great if you can verify that this works, and if not please add it to the polarion integration. Thanks!

pkg/virt-operator/kubevirt_test.go
```diff
@@ -1495,7 +1495,7 @@ var _ = Describe("KubeVirt Operator", func() {
 
 		}, 15)
 
-		It("should add resources on create", func() {
+		It("[test_id:2806]should add resources on create", func() {
```
dshchedr: we don't need to add test attributes to unit tests, we can use it only inside kubevirt/tests/ folder

I think it's better to add steps from 2806 to other functional test

dhiller: OK, added.

dhiller: Also I removed the ref_id and test_id from unit test.

tests/virt_control_plane_test.go
```diff
@@ -45,9 +45,6 @@ const (
 
 var _ = Describe("KubeVirt control plane resilience", func() {
```
dshchedr: I think it's better to put ref_id here, don't need to have the same ref_id in each group of tests

pkg/virt-operator/kubevirt_test.go
dshchedr: Thank you!

tests/virt_control_plane_test.go
dhiller: OK, done!

tests/virt_control_plane_test.go
```diff
@@ -56,228 +53,282 @@ var _ = Describe("KubeVirt control plane resilience", func() {
...
+	Context("control plane components check", func() {
+
+		When("control plane pods are running", func() {
+
+			It("[test_id:2806]virt-controller and virt-api pods have a pod disruption budget", func() {
+
+				By("check deployments")
+				deployments, err := deploymentsClient.List(metav1.ListOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				expectedDeployments := []string{"cdi-http-import-server", "virt-operator", "virt-api", "virt-controller"}
```
slintes: it does not hurt, but why are we testing for cdi-http-import-server here? :thinking:

dhiller: Because that was described in the Polarion test case.

slintes: ok. Don't want to discuss this further then :wink:

dhiller: :-D

dshchedr: I think Zhe added it to polarion accidentally.. IMO, you don't need to check "cdi" and "virt-operator", because they are not related to this functional test :)
Please, remove it and I will update testcase in Polarion

dhiller: Done. /cc @slintes

Issue #2438, opened 2019-06-27
VM with Datavolume in Block mode not starting


rollandf:
/kind bug

**What happened**:
Define a VM with a DataVolume in Block mode and the VM failed to start.

The virt launcher pod stays in ContainerCreating state with the following error:

Warning  FailedMount  28s (x2 over 2m43s)  kubelet, node01    Unable to mount volumes for pod "virt-launcher-vm-cirros-vmd4n_default(1eceea32-98b7-11e9-97da-525500d15501)": timeout expired waiting for volumes to attach or mount for pod "default"/"virt-launcher-vm-cirros-vmd4n". list of unmounted volumes=[dv-disk]. list of unattached volumes=[dv-disk infra-ready-mount virt-share-dir libvirt-runtime ephemeral-disks]

**What you expected to happen**:
Define a VM with a DataVolume in Block mode and the VM starts.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- KubeVirt version (use `virtctl version`):
Client Version: version.Info{GitVersion:"v0.17.4", GitCommit:"adfdb8c07830b99fc79d2fd1d004e862ef70979e", GitTreeState:"clean", BuildDate:"2019-06-25T11:44:21Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{GitVersion:"v0.17.4", GitCommit:"adfdb8c07830b99fc79d2fd1d004e862ef70979e", GitTreeState:"clean", BuildDate:"2019-06-25T12:49:03Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}

- Kubernetes version (use `kubectl version`):
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.3", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"clean", BuildDate:"2019-02-01T20:08:12Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.3", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"clean", BuildDate:"2019-02-01T20:00:57Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}

- VM or VMI specifications:
[...]

PR #2612, opened 2019-08-19
add support to detect block mode on a CDI Datavolume

pbudds:
Signed-off-by: Prashanth Buddhala <pbudds@gmail.com>

Currently Kubevirt virt-controller and virt-launcher code are assuming CDI DataVolume to be backed by filesystem PVC only. As a result when block mode is turned ON on the CDI DV, the block device does not get properly appended to "VolumeDevices" of the virt-launcher pod. Add support to check Block mode on the DV and handle it accordingly.

Fixes #

**Release note**:

```release-note
Block mode on a Containerized Data Importer DataVolume does not get detected correctly and as a result VM fails to run.
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/manager.go
+++ b/pkg/virt-launcher/virtwrap/manager.go
@@ -876,7 +876,7 @@ func (l *LibvirtDomainManager) SyncVMI(vmi *v1.VirtualMachineInstance, useEmulat
 	isBlockPVCMap := make(map[string]bool)
 	diskInfo := make(map[string]*containerdisk.DiskInfo)
 	for i, volume := range vmi.Spec.Volumes {
-		if volume.VolumeSource.PersistentVolumeClaim != nil {
+		if volume.VolumeSource.PersistentVolumeClaim != nil || volume.VolumeSource.DataVolume != nil {
 			isBlockPVC, err := isBlockDeviceVolume(volume.Name)
 			if err != nil {
 				logger.Reason(err).Errorf("failed to detect volume mode for Volume %v and PVC %v.",
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -380,9 +380,14 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 			Name:      volume.Name,
 			MountPath: hostdisk.GetMountedHostDiskDir(volume.Name),
 		}
-		if volume.PersistentVolumeClaim != nil {
+		if volume.PersistentVolumeClaim != nil || volume.DataVolume != nil {
 			logger := log.DefaultLogger()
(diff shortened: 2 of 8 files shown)


pbudds:
@petrkotas - Sure. I can try to do that. Just to make sure I understand, would those be the tests that run as part of "make bazel-tests" and "make functest" ?

petrkotas:
Hi @pbudds essentially KubeVirt requires both. 

First one is part of `make test` which runs unit tests and are run in the TravisCI. The relevant files are [template_test.go](https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-controller/services/template_test.go), [converter_test.go](https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/api/converter_test.go).
These are here to check for regression.

Second is part of `make functest` which verifies the overall functionality. The most relevant place for the CDI test is [datavolume_test.go](https://github.com/kubevirt/kubevirt/blob/master/tests/datavolume_test.go).

Feel free to take inspiration from other tests. 

Again thank you for your effort with adding the functionality.

ahadas:
@pbudds , thanks for the PR. Next time please provide the issue it fixes (in this case #2438 ) that could have prevented the similar work on my side (#2636 ).
@petrkotas I think we can avoid duplicate tests here.  In #2617  I add tests with block mode volumes in Ceph. They currently pass because I'm doing a workaround that is described in #2438 - this PR can be tested by dropping that workaround (that's how I tested my PR)

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -876,7 +876,7 @@ func (l *LibvirtDomainManager) SyncVMI(vmi *v1.VirtualMachineInstance, useEmulat
 	isBlockPVCMap := make(map[string]bool)
 	diskInfo := make(map[string]*containerdisk.DiskInfo)
 	for i, volume := range vmi.Spec.Volumes {
-		if volume.VolumeSource.PersistentVolumeClaim != nil {
+		if volume.VolumeSource.PersistentVolumeClaim != nil || volume.VolumeSource.DataVolume != nil {
```
ahadas: Yeah, I've also thought about doing it that way - but this kind of assumes that the name of the PVC will always match the name of the DataVolume, right? it's true that we assume it elsewhere but here it's implicitly defined so it would be harder to detect this if we ever need to.

ahadas:
@pbudds , I merged our work, trying to take the best of both worlds :) could you please take a look at it in #2636 ? note I cherry-picked the stuff that was missing in my PR, changing the one that fixes the test in vmi_test.go (thus added myself as a co-author) and added you as a co-author on my patches

slintes:
As discussed, superseded by https://github.com/kubevirt/kubevirt/pull/2636

But still thanks a lot for your PR, afaik some commits moved over to the new one :)

/close

PR #2614, opened 2019-08-20
Static go build of containerdisk binary

slintes:
The containerdisk binary is used in containerdisk images which are build from scratch, so it needs to be static

**Release note**:
```release-note
NONE
```


slintes:
Travis is green, functests are not affected by this

PR #2626, opened 2019-08-23
update doc about ssh method usage

jichenjc:
I am using the ssh way to test but actually it's `kind`, so we need change doc about this
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -105,19 +105,19 @@ have a completely fresh cluster to play with.
 
 ### Accessing the containerized nodes via ssh
 
-The containerized nodes are named starting from `node01`, `node02`, and so
-forth. `node01` is always the master of the cluster.
-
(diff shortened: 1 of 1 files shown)


slintes:
Hey @jichenjc, thanks for rising this :+1: 
Actually I think we should update the doc to use `cluster-up/ssh.sh <nodename>` for all clusters. Only the node names differ depending on cluster type (kubernetes/openshift, kind, okd). You can get the node names with `cluster-up/kubectl.sh get nodes`. This should also work for kind. Can you give it a try and update the doc with this please?

jichenjc:
@slintes I updated according to your suggestion :)

slintes:
Thanks a lot, looks good now!

/lgtm
/approve

slintes:
tide queue is empty, will merge manually (doc change only)

PR #2628, opened 2019-08-23
Change tests to comply with sriov operator network attachment definitions

fedepaol:

Once https://github.com/kubevirt/kubevirtci/pull/134 gets merged, the way we reference the sriov network attachement definitions must match those provided by the openshift sriov operator.
We don't create network attachement definitions but `SriovNetwork`s and the api endpoint is different.

**Release note**:

```release-note
NONE

```


fedepaol:
/hold

need https://github.com/kubevirt/kubevirtci/pull/134 to get merged in order to regenerate the cluster-up subdir

tests/vmi_multus_test.go
```diff
@@ -593,21 +579,41 @@ var _ = Describe("SRIOV", func() {
...
 		result = virtClient.RestClient().
 			Post().
-			RequestURI(fmt.Sprintf(postUrl, tests.NamespaceTestDefault, "sriov2")).
+			RequestURI(fmt.Sprintf(sriovNetworksEndpoint, "sriov2")).
 			Body([]byte(fmt.Sprintf(sriovConfCRD, "sriov2", tests.NamespaceTestDefault))).
 			Do()
 		Expect(result.Error()).NotTo(HaveOccurred())
 	})
 
+	AfterEach(func() {
```
phoracek: Could you mention and explain this in the commit message please?

tests/vmi_multus_test.go
```diff
@@ -593,21 +579,41 @@ var _ = Describe("SRIOV", func() {
...
+		// Multus tests need to ensure that old VMIs are gone
```
phoracek: Otherwise what? How comes this hasn't been already?

tests/vmi_multus_test.go
```diff
@@ -593,21 +579,41 @@ var _ = Describe("SRIOV", func() {
 
 	tests.BeforeAll(func() {
 		tests.BeforeTestCleanup()
-		// Create two sriov networks referring to the same resource name
+		// Extra cleaning of sriov networks
```
phoracek: Why only now? Please explain in the commit message, this has way more than alignment with the operator.

fedepaol: It's part of the sriov-operator setup.

Now in the test we are creating SriovNetworks CRDs instead of NetworkAttachmentDefinition CRD's, and if we create an already existing one the creation will fail (so the second test will fail)

phoracek: Ok

tests/vmi_multus_test.go
fedepaol: It was when multus + sriov were a single test suite. 
It probably got lost after they got split. If we have dangiling vms (and a lot of tests) we risk to finish the vfs.

tests/vmi_multus_test.go
fedepaol: Changed

booxter:
@fedepaol on second thought, is it correct that our suite now DEPENDS on operator deployment? Can't we still manually create the networks as before without relying on the operator machinery? While it's the recommended way of deployment now, perhaps we should not make it a dependency for the suite if possible. What do you think? (I understand that the resource name now has a different prefix since it's managed by the operator, at least in CI, but perhaps we can live with just this prefix change; or maybe even think about passing the expected prefix / resource name from the environment into the test suite to allow to run the suite in other deployments.

fedepaol:
> @fedepaol on second thought, is it correct that our suite now DEPENDS on operator deployment? Can't we still manually create the networks as before without relying on the operator machinery? While it's the recommended way of deployment now, perhaps we should not make it a dependency for the suite if possible. What do you think? (I understand that the resource name now has a different prefix since it's managed by the operator, at least in CI, but perhaps we can live with just this prefix change; or maybe even think about passing the expected prefix / resource name from the environment into the test suite to allow to run the suite in other deployments.


Given that:
- we want to give each test the freedom to create its own networks (i.e. not moving the network creation in the provider)
- we rely on the operator to setup the test environment

The only option we have in order to soften the dependency is to try to see if the whole thing still works by manually creating the networkattachmentdefinitions (instead of relying on the operator to create them). 

By doing this though, we are circumventing the "canonical" way to use the operator in order to create new networks, so it may break if they change something in the future (i.e. they instantiate the resources only if there is a network referring to them).

On the other hand, I kind of like the "vanilla" approach where the suite would not depend on having the operator deployed so I don't really have a strong opinion.

Let me give it a try, if it works the resource prefix is something we can deal with in a way or another.

booxter:
@fedepaol I don't want my word on this be final. I would like others to comment on whether it's ok to make the suite dependent on using a particular operator. @phoracek @SchSeba

fedepaol:
For the records, it worked as expected by manually creating a network attachment definition that refers the resource created by the operator.

fedepaol:
This latest version creates the network attachment definition manually. 
I think it looks better.

If we are all ok, I'll clean the commit history, otherwise I'll drop the last commit.

fedepaol:
Regardless of the version we are going to ship (I a now tend towards the manual NetworkAttachmentDefinition version, the latest one), the tests revealed another problem:

The mac address feature is not working anymore since the cni plugin shipped with the operator does not have it on board (yet), and I am not sure what is the best approach since we want to use the operator. Maybe disable the tests until the openshift cni plugin gets that feature from upstream?

@booxter what do you think?

fedepaol:
Re: mac tests and using upstream cni plugin, I did a quick local test and it should work by just sed-ing the image name into the operator manifest, so I would proceed in that way.

fedepaol:
The CI run passed all the 3 tests (including the mac address one), this is good to go from my point of view.

https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2628/pull-kubevirt-e2e-kind-k8s-sriov-1.14.2/1166942929336406016

slintes:
expected failure

/override pull-kubevirt-e2e-kind-k8s-sriov-1.14.2

PR #2639, opened 2019-08-27
cpu-manager detection fix

mfranczy:
- It checks a CPU policy from `cpu_manager_state` file

Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=1718944

**Release note**:

```release-note
cpu-manager detection fix
```

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -2,47 +2,83 @@ package cluster
 
 import (
 	"encoding/json"
-	"strconv"
+	"strings"
+
+	secv1 "github.com/openshift/api/security/v1"
+	k8sversion "k8s.io/apimachinery/pkg/version"
+	"k8s.io/client-go/discovery"
 
 	"kubevirt.io/client-go/kubecli"
-
-	"k8s.io/apimachinery/pkg/api/errors"
+	"kubevirt.io/client-go/log"
 )
 
-const OpenShift3Major = 3
+const (
+	OpenShift3Major = 3
+	OpenShift4Major = 4
+	k8sBaseForOKD   = "1.12"
+)
 
-type OpenShiftVersion struct {
-	Major string `json:"major"`
-	Minor string `json:"minor"`
+func IsOnOpenShift(clientset kubecli.KubevirtClient) (bool, error) {
+	apis, err := clientset.DiscoveryClient().ServerResources()
+	if err != nil && !discovery.IsGroupDiscoveryFailedError(err) {
+		return false, err
+	}
+
+	// In case of an error, check if security.openshift.io is the reason (unlikely).
+	// If it is, we are obviously on an openshift cluster.
+	// Otherwise we can do a positive check.
+	if discovery.IsGroupDiscoveryFailedError(err) {
+		e := err.(*discovery.ErrGroupDiscoveryFailed)
+		if _, exists := e.Groups[secv1.GroupVersion]; exists {
+			return true, nil
+		}
+	}
+
+	for _, api := range apis {
+		if api.GroupVersion == secv1.GroupVersion.String() {
+			for _, resource := range api.APIResources {
+				if resource.Name == "securitycontextconstraints" {
+					return true, nil
+				}
+			}
+		}
+	}
+
+	return false, nil
 }
 
-func IsOnOpenShift3() (bool, error) {
-	var osVersion OpenShiftVersion
+func GetKubernetesVersion(clientset kubecli.KubevirtClient) (string, error) {
+	var info k8sversion.Info
 
-	clientset, err := kubecli.GetKubevirtClient()
+	virtClient, err := kubecli.GetKubevirtClient()
 	if err != nil {
-		return false, err
+		return "", err
 	}
 
-	obj, err := clientset.CoreV1().RESTClient().Get().RequestURI("/version/openshift").Do().Raw()
+	response, err := virtClient.RestClient().Get().AbsPath("/version").DoRaw()
 	if err != nil {
-		if errors.IsForbidden(err) || errors.IsNotFound(err) {
-			return false, nil
-		}
-		return false, err
+		return "", err
 	}
 
-	if err := json.Unmarshal(obj, &osVersion); err != nil {
-		return false, err
+	if err := json.Unmarshal(response, &info); err != nil {
+		return "", err
 	}
 
-	v, err := strconv.Atoi(osVersion.Major)
+	curVersion := strings.Split(info.GitVersion, "+")[0]
+	curVersion = strings.Trim(curVersion, "v")
+
+	return curVersion, nil
+}
+
+func GetOpenShiftMajorVersion(clientset kubecli.KubevirtClient) int {
+	k8sVersion, err := GetKubernetesVersion(clientset)
 	if err != nil {
-		return false, err
+		log.Log.Errorf("Unable to detect major OpenShift version: %v", err)
+		return -1
 	}
 
-	if v == OpenShift3Major {
-		return true, nil
+	if k8sVersion >= k8sBaseForOKD {
+		return OpenShift4Major
 	}
-	return false, nil
+	return OpenShift3Major
 }
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -1,0 +1,48 @@
+package cluster
+
+import (
+	"encoding/json"
+	"strconv"
+
+	"kubevirt.io/client-go/kubecli"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+)
+
+const OpenShift3Major = 3
+
+type OpenShiftVersion struct {
+	Major string `json:"major"`
+	Minor string `json:"minor"`
+}
+
+func IsOnOpenShift3() (bool, error) {
+	var osVersion OpenShiftVersion
+
+	clientset, err := kubecli.GetKubevirtClient()
+	if err != nil {
+		return false, err
+	}
+
+	obj, err := clientset.CoreV1().RESTClient().Get().RequestURI("/version/openshift").Do().Raw()
+	if err != nil {
+		if errors.IsForbidden(err) || errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	if err := json.Unmarshal(obj, &osVersion); err != nil {
+		return false, err
+	}
+
+	v, err := strconv.Atoi(osVersion.Major)
+	if err != nil {
+		return false, err
+	}
+
+	if v == OpenShift3Major {
+		return true, nil
+	}
+	return false, nil
+}
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -8,6 +8,9 @@ const ExtensionAPIServerAuthenticationConfigMap = "extension-apiserver-authentic
 const RequestHeaderClientCAFileKey = "requestheader-client-ca-file"
 const VirtShareDir = "/var/run/kubevirt"
 const VirtLibDir = "/var/lib/kubevirt"
+const HostRootMount = "/proc/1/root/"
+const CPUManagerOS3Path = HostRootMount + "var/lib/origin/openshift.local.volumes/cpu_manager_state"
+const CPUManagerPath = HostRootMount + "var/lib/kubelet/cpu_manager_state"
 
 func IsSRIOVVmi(vmi *v1.VirtualMachineInstance) bool {
 	for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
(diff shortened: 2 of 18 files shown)


mfranczy:
/hold

I am testing the fix

mfranczy:
it seems k8s tests are failing, checking that

pkg/virt-handler/vm.go
```diff
@@ -1584,10 +1586,36 @@ func (d *VirtualMachineController) updateNodeCpuManagerLabel() {
 			log.DefaultLogger().Reason(err).Errorf("failed to set a cpu manager label on host %s", d.host)
 			return
 		}
-		if strings.Contains(string(content), "kubelet") && strings.Contains(string(content), "cpu-manager-policy=static") {
+		if !bytes.Contains(content, []byte("kubelet")) {
```
cynepco3hahue: hm I can not connect between PR description and fix, I can see you started to use `bytes` instead of `strings` and I can see you check the kubelet config for the parameter, but how it can fix the test?

mfranczy: thanks for the review :) this is not about fixing the test but https://bugzilla.redhat.com/show_bug.cgi?id=1718944, sometimes cpu-manager-policy might be set in the config file instead of kubelet's cmdline, the idea was to check the kubelet config and check the policy file there, however, this approach is bad because `kubelet.conf` doesn't have always the same location. I am going to change solution here and check `cpu_manager_state` files.

/hold

cynepco3hahue: thanks for the explanation, it was unclear for me, that we are talking about BZ, in this case, I can see you already got the `kubelet` config path from the `kubelet` cmd line, so it always should point you to the right file, no?

mfranczy: yes, it points me to the right file, but before starting virt-handler I would have to use `hostPath` to mount the kubelet.conf, so it's a problem if the file location is not the same, I cannot mount the file later and I cannot assume that the conf file path is static (always the same location).

mfranczy:
for okd4.x discovery, I need to check `ClusterVersion` CRD... PR still in progress. It's more complex than I thought.

pkg/util/util.go
```diff
@@ -8,6 +8,8 @@ const ExtensionAPIServerAuthenticationConfigMap = "extension-apiserver-authentic
 const RequestHeaderClientCAFileKey = "requestheader-client-ca-file"
 const VirtShareDir = "/var/run/kubevirt"
 const VirtLibDir = "/var/lib/kubevirt"
+const CpuManagerOpenShift3Path = "/var/lib/origin/openshift.local.volumes/cpu_manager_state"
```
cynepco3hahue: Will the OKD 4 also have a different path?

mfranczy: yes

cynepco3hahue: Can you please add it under the PR?

cynepco3hahue:
I can see it has some problem with unit tests or goveralls
`The command "if [[ $TRAVIS_REPO_SLUG == "kubevirt/kubevirt" ]]; then make goveralls; else make bazel-tests; fi" exited with 2.`

mfranczy:
@cynepco3hahue thanks for pointing that, probably unit tests, I will fix that when I have a ready solution to discover OKD4 cluster.

pkg/util/util.go
mfranczy: I will :) working on that :)

cynepco3hahue: I found only one path `/sysroot/ostree/deploy/rhcos/var/lib/kubelet/cpu_manager_state`

cynepco3hahue: and I think it is correct one

mfranczy: ah yes, I meant it's different than the path in os-3.11 :). Btw. funny thing... :D I am going to drop all changes here, after speaking with @fabiand and @vladikr I am going to leave the old solution with cpu-manager on os-3.11 and for okd4.1 I will try to get the information about policy from node configmap. The reason why is we don't want to add more hostPath(s) to virt-handler.

cynepco3hahue: I see you will need to check `KubeletConfigs`, for me, it looks pretty complicated because it can have a number of `KubeletConfigs` and each config can be only valid for specific nodes :(

mfranczy: maybe `virt-handler` could check `KubeletConfig` for its node, it sounds right. Do you see any risk here?

mfranczy: I can imagine flow like this:

1. virt-handler checks on which node is running (maybe that info is available already)
2. virt-handler checks node annotation `machineconfiguration.openshift.io/currentConfig`
3. virt-handler gets `machineconfiguration` based on the value from step 2, I see there is the content of kubelet.conf

that should work.

mfranczy:
I thought about updating the `cpumanager` label in node controller, but I think it's good to keep the update mechanism in one place since it's a [workaround](https://github.com/kubevirt/kubevirt/pull/2639/files#diff-27207346391eaa19190c76760eb8851fR1577) :)

@cynepco3hahue @vladikr could you take a look?
@rmohr if you have time then take a look as well.

Commits without vendor changes:

- https://github.com/kubevirt/kubevirt/pull/2639/commits/a2ed7aa947533f77140e2529e2e23add97207a1b
- https://github.com/kubevirt/kubevirt/pull/2639/commits/57dc1aaaa56d1e33b9c8bbec2bf88504fe569b9d

/hold cancel

pkg/virt-handler/vm.go
```diff
@@ -1572,43 +1576,109 @@ func (d *VirtualMachineController) heartBeat(interval time.Duration, stopCh chan
...
-				d.updateNodeCpuManagerLabel()
+				updateNodeCpuManagerLabel()
 			}
 		}, interval, 1.2, true, stopCh)
 	}
 }
 
-func (d *VirtualMachineController) updateNodeCpuManagerLabel() {
+func (d *VirtualMachineController) getCPUManagerLabelFunc() func() {
+	var lastConfigID string
```
cynepco3hahue: hm where do you set value for this variable?
IMHO you should get it from annotation under `dicoverCPUPolicyMachineConf` method and do not send it as a parameter

cynepco3hahue:
@vladikr @fabiand I do not sure that without exposing `cpu_manager_state` to the virt-handler we have a robust way to know if the CPU manager is enabled or disabled, all hack that we have will not work in case if a user will edit `kubelet.conf` on the node(without the usage of any CRD's)

pkg/virt-handler/vm.go
mfranczy: I do that to keep the last id between function calls, to avoid unnecessary API calls, if lastId == currentId (from annotation) then don't do anything more.

pkg/virt-handler/vm.go
```diff
@@ -1572,43 +1576,109 @@ func (d *VirtualMachineController) heartBeat(interval time.Duration, stopCh chan
...
+	machineConfigExists := true
+
+	apiClient := d.clientset.ExtensionsClient().ApiextensionsV1beta1()
```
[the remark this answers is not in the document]
mfranczy: This gonna be called only once, but you're right if someone will create this CRD later `virt-handler` won't be aware of that.

pkg/virt-handler/vm.go
```diff
@@ -1572,43 +1576,109 @@ func (d *VirtualMachineController) heartBeat(interval time.Duration, stopCh chan
...
+		isEnabled := false
+		updateRequired := true
+
+		if !machineConfigExists {
+			// on OpenShift3 cpu manager policy is exposed in proc cmdline
+			isEnabled, err = discoverCPUPolicyProc()
+		} else {
+			// on OpenShift4 cpu manager policy is not exposed anymore in proc cmdline
+			// it's required to check machineconfiguration kind to get kubelet configuration for nodes
+			isEnabled, updateRequired, err = dicoverCPUPolicyMachineConf(d.clientset, &lastConfigID, d.host)
```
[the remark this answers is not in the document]
mfranczy: from this what I understand, I had to prepare a bugfix only for OKD4 and i focused only on that.

mfranczy:
@cynepco3hahue generally I agree with you that we should check `cpu_manager_state` file, that would work for k8s and OpenShift (I had a working solution already), but AFAIK we don't want to add more hostPath(s) to virt-handler. @vladikr and @fabiand please check the PR and share your opinion.

mfranczy:
maybe it's the best to let users take care of CPU label (as we spoke initially).

mfranczy:
@cynepco3hahue @vladikr I changed the approach here as we spoke, please take a look when it's possible.

pkg/virt-handler/vm.go
```diff
@@ -1555,6 +1556,15 @@ func (d *VirtualMachineController) updateDomainFunc(old, new interface{}) {
 }
 
 func (d *VirtualMachineController) heartBeat(interval time.Duration, stopCh chan struct{}) {
+	// This is a temporary workaround until k8s bug #66525 is resolved
```
vladikr: do we need to find the correct path on each heartBeat or can it be done once outside of this function?

mfranczy: it's called only once, outside the loop [here](https://github.com/kubevirt/kubevirt/pull/2639/files/a66097c601e994e4a12b9f0061a56cc069e42479#diff-27207346391eaa19190c76760eb8851fR1568) :)

vladikr: ah, sorry, I was sure that heartBeat is being called from a loop. Sorry :)

pkg/util/cluster/cluster.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+const OpenShift3Major = 3
+
+type OpenShiftVersion struct {
+	Major string `json:"major"`
+	Minor string `json:"minor"`
+}
+
+func IsOnOpenShift3() (bool, error) {
```
rmohr: The `virt-operator` already contains logic which detects reliably if we are on openshift. As far as I know `/version/openshift` does not exist anymore in OCP4.

mfranczy: I need to know the version, OpenShift3 has a different cpu-manager path than OpenShift4 (the same as k8s), the check from virt-operator is not helpful here (it checks only if security context constraints exists which is available on OpenShift4 as well)

rmohr: There is also code in virt-operator which helps you finding that out. Also virt-operator has to distinguish between ok3 and okd4 in some places. Afaik it does something like, "if on openshift and k8s version is bigger than X, then it is okd4`".

mfranczy: ..and you're right about OCP4 `version` endpoint, it doesn't exist, as far as i remember there was an issue to bring it back, not sure about the decision though.

Anyway, I need to know that I am on OpenShift3, k8s and ocp4 have the same file path.

mfranczy: Ok, I will check it.

mfranczy: Well, it's not exactly in `virt-operator` pkg but in tests, but I will get the solution from it and refactor a bit. https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L3436

pkg/virt-handler/vm.go
```diff
@@ -1572,33 +1582,22 @@ func (d *VirtualMachineController) heartBeat(interval time.Duration, stopCh chan
...
-			continue
-		} else if err != nil {
-			log.DefaultLogger().Reason(err).Errorf("failed to set a cpu manager label on host %s", d.host)
-			return
-		}
-		if strings.Contains(string(content), "kubelet") && strings.Contains(string(content), "cpu-manager-policy=static") {
-			isEnabled = true
-			break
-		}
+	if strings.Contains(string(content), "\"policyName\":\"static\"") {
```
rmohr: Do you think you can parse the file and extract the entry from a parsed json (can be anonymous into maps/arrays)?

mfranczy: Yes, it should be possible. I will change.

mfranczy: done

rmohr:
this one may contain a PR-related issue: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2639/pull-kubevirt-e2e-k8s-1.11.0/1166977656105734144

mfranczy:
I am checking what's wrong with the solution for k8s-1.11.

/hold

mfranczy:
/hold cancel

k8s-1.11.0 is fine with the fix

mfranczy:
> this one may contain a PR-related issue: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2639/pull-kubevirt-e2e-k8s-1.11.0/1166977656105734144

@rmohr it makes sense that back in the time it failed for k8s-1.11.0, the solution was completely different and could not work (the solution proposed then [link](https://github.com/kubevirt/kubevirt/commit/d301f7cb879a1402d2542f82b6d618d26d749997)).

PR #2643, opened 2019-08-28
make examples/vmi-migratable.yaml migratable again!

ahadas:
This PR introduces the minimal changes to make the VMI described
in examples/vmi-migratable.yaml migratable as VMIs with only bridged
pod network are not migratable anymore.

**Release note**:

```release-note
Change the VMI `vmi-migratable` in the examples folder to be migratable.
```

### DIFF
--- a/tools/vms-generator/utils/utils.go
+++ b/tools/vms-generator/utils/utils.go
@@ -287,6 +287,10 @@ func addHostDisk(spec *v1.VirtualMachineInstanceSpec, path string, hostDiskType 
 
 func GetVMIMigratable() *v1.VirtualMachineInstance {
 	vmi := getBaseVMI(VmiMigratable)
+	// having no network leads to adding a default interface that may be of type bridge on
+	// the pod network and that would make the VMI non-migratable. Therefore, adding a network.
+	vmi.Spec.Networks = []v1.Network{*v1.DefaultPodNetwork()}
+	vmi.Spec.Domain.Devices.Interfaces = []v1.Interface{*v1.DefaultMasqueradeNetworkInterface()}
 
 	addContainerDisk(&vmi.Spec, fmt.Sprintf("%s/%s:%s", DockerPrefix, imageAlpine, DockerTag), busVirtio)
 	return vmi
--- a/tools/vms-generator/utils/utils.go
+++ b/tools/vms-generator/utils/utils.go
@@ -131,28 +131,24 @@ func addRNG(spec *v1.VirtualMachineInstanceSpec) *v1.VirtualMachineInstanceSpec 
 }
 
 func addContainerDisk(spec *v1.VirtualMachineInstanceSpec, image string, bus string) *v1.VirtualMachineInstanceSpec {
-	spec.Domain.Devices = v1.Devices{
-		Disks: []v1.Disk{
-			{
-				Name: "containerdisk",
-				DiskDevice: v1.DiskDevice{
-					Disk: &v1.DiskTarget{
-						Bus: bus,
-					},
-				},
+	disk := &v1.Disk{
+		Name: "containerdisk",
+		DiskDevice: v1.DiskDevice{
+			Disk: &v1.DiskTarget{
+				Bus: bus,
 			},
 		},
 	}
-	spec.Volumes = []v1.Volume{
-		{
-			Name: "containerdisk",
-			VolumeSource: v1.VolumeSource{
-				ContainerDisk: &v1.ContainerDiskSource{
-					Image: image,
-				},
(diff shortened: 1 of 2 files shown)


ahadas:
> you need to change this in https://github.com/kubevirt/kubevirt/tree/master/tools/vms-generator

Done

ahadas:
> @ahadas you also need to configure the masquerade interface pointing the default network

But the VM can start and migrate without having an interface, it can be networkless, right?
We already have vmi-masquerade for demonstrating that, I want the minimal configuration for making the VM migratable

ahadas:
> @ahadas what happens when we add a network but no interface? does it make the VMI disconnected completely? I don't think this is a useful example

Yes, the VMI will be provided only with the loopback interface, i.e., no external connectivity.
I agree it's not a common case but it can be useful sometimes (as a matter of face, in some virtualization management like oVirt it's provided as default). For instance, in school you may create a pool of VMs that have no external connectivity so students won't be able to communicate with each other or look for solutions in Google :)

And even if there are not many cases where it's useful, we simulate real hardware, right? so network is not like CPU, it is important aspect but should be optional - I can boot my computer without having external connectivity. In k8s where they mainly talk about services it doesn't make much sense to start a workload without external connectivity but in our domain, where I can start a networkless workload, connect to it  via e.g. SPICE that doesn't require external connectivity and it would still be useful - so I think we shouldn't rule out networkless (as well as diskless btw) workloads.

The problem is now with the semantics. It makes the most sense to me to say "networkless workload" is one that no network is defined for but as it was decided to add a default network in this case we'll need at some point to provide users with a different semantics to specify "do not add default network".

@fabiand  ^

SchSeba:
@ahadas I think we can create a new example for a non connected vm.

For this you can use the `AutoattachPodInterface` in the device section then you don't need to configure network and interfaces.

but for the migratable vm I will like it to be connected to the pod network (I think this is the most common use case and expect the user to just copy/paste our example to try kubevirt)

ahadas:
> @ahadas I think we can create a new example for a non connected vm.
> 
> For this you can use the `AutoattachPodInterface` in the device section then you don't need to configure network and interfaces.

Ah ok, I missed that option.

> 
> but for the migratable vm I will like it to be connected to the pod network (I think this is the most common use case and expect the user to just copy/paste our example to try kubevirt)

Yeah, for the migratable VM I agree. The comment above was in the broader scope

ahadas:
> @ahadas you also need to configure the masquerade interface pointing the default network

Done

ahadas:
> @ahadas you also need to configure the masquerade interface pointing the default network

Done

SchSeba:
@slintes I think we can overwrite the tests no?

tools/vms-generator/utils/utils.go
```diff
@@ -291,6 +287,10 @@ func addHostDisk(spec *v1.VirtualMachineInstanceSpec, path string, hostDiskType
 
 func GetVMIMigratable() *v1.VirtualMachineInstance {
 	vmi := getBaseVMI(VmiMigratable)
+	// having no network leads to adding a default interface that may be of type bridge on
+	// the pod network and that would make the VMI non-migratable. Therefore, adding a network.
```
booxter: (no action required) this comment would be clearer as: "Default interface type is bridge is not migratable. Use masquerade."

tools/vms-generator/utils/utils.go
```diff
@@ -131,28 +131,24 @@ func addRNG(spec *v1.VirtualMachineInstanceSpec) *v1.VirtualMachineInstanceSpec
 }
 
 func addContainerDisk(spec *v1.VirtualMachineInstanceSpec, image string, bus string) *v1.VirtualMachineInstanceSpec {
-	spec.Domain.Devices = v1.Devices{
```
booxter: how are these changes related to the PR?

ahadas: Look at `GetVMIMigratable`. With the existing `addContainerDisk` it wouldn't work because the interface would be clear by this fucntion. This prevents that unexpected side-effect of `addContainerDisk`

tools/vms-generator/utils/utils.go
ahadas: but the default interface type is not necessarily a bridge as it is configurable at the cluster level, right? also I could have interpreted the message you propose as "if you add an interface without a type (e.g., with just a name) we set its type to bridge by default and the VMI is not migratable then" - which is not what I actually mean here. so I prefer to be more explicit here

booxter: @ahadas you are right, I forgot we allow to configure it now.

PR #2645, opened 2019-08-29
Add 'migrate' command to virtctl

ahadas:
This PR enables migrating a VMI using `virtctl` by executing: `virtctl migrate <vm>`

**Release note**:

```release-note
Enable triggering live-migration through virtctl
```

### DIFF
--- a/staging/src/kubevirt.io/client-go/kubecli/vm.go
+++ b/staging/src/kubevirt.io/client-go/kubecli/vm.go
@@ -150,3 +150,8 @@ func (v *vm) Stop(name string) error {
 	uri := fmt.Sprintf(vmSubresourceURL, v1.ApiStorageVersion, v.namespace, name, "stop")
 	return v.restClient.Put().RequestURI(uri).Do().Error()
 }
+
+func (v *vm) Migrate(name string) error {
+	uri := fmt.Sprintf(vmSubresourceURL, v1.ApiStorageVersion, v.namespace, name, "migrate")
+	return v.restClient.Put().RequestURI(uri).Do().Error()
+}
--- a/tests/vm_test.go
+++ b/tests/vm_test.go
@@ -753,6 +753,28 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
 
 				})
 
+				It("should migrate a running VM", func() {
+					By("creating a VM with RunStrategyAlways")
+					virtualMachine := newVirtualMachineWithRunStrategy(v1.RunStrategyAlways)
+
+					By("Waiting for VM to be ready")
+					Eventually(func() bool {
+						virtualMachine, err = virtClient.VirtualMachine(virtualMachine.Namespace).Get(virtualMachine.Name, &v12.GetOptions{})
+						Expect(err).ToNot(HaveOccurred())
+						return virtualMachine.Status.Ready
+					}, 360*time.Second, 1*time.Second).Should(BeTrue())
+
+					By("Invoking virtctl migrate")
+					migrateCommand := tests.NewRepeatableVirtctlCommand(vm.COMMAND_MIGRATE, "--namespace", virtualMachine.Namespace, virtualMachine.Name)
+					Expect(migrateCommand()).To(Succeed())
+
+					By("Ensuring the VirtualMachineInstance is migrated")
+					Eventually(func() bool {
+						nextVMI, err := virtClient.VirtualMachineInstance(virtualMachine.Namespace).Get(virtualMachine.Name, &v12.GetOptions{})
+						Expect(err).ToNot(HaveOccurred())
+						return nextVMI.Status.MigrationState != nil && nextVMI.Status.MigrationState.Completed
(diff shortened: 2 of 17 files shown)


ahadas:
> lgtm, but lack unit and functional tests

Done

staging/src/kubevirt.io/client-go/kubecli/vm.go
```diff
@@ -150,3 +153,17 @@ func (v *vm) Stop(name string) error {
 	uri := fmt.Sprintf(vmSubresourceURL, v1.ApiStorageVersion, v.namespace, name, "stop")
 	return v.restClient.Put().RequestURI(uri).Do().Error()
 }
+
+func (v *vm) Migrate(name string) error {
```
cynepco3hahue: I do not sure it is the correct way, in this case, we do not have an endpoint for migration under subresources, it maybe not critical, but nice to have, so the plan:
- implement migration point under subresources
- call to it similar to all other commands

ahadas: That's also possible, maybe even better as other clients could trigger the migration by invoking the migrate operation on the VM. But I think such a new endpoint would render the existing one redundant - and dropping it is a bigger and more controversial change I believe.

The propose change provides the ability to migrate the VM through `virtctl` and limits the modifications only to `virtctl`. How about getting proceeding with it and consider changing/adding the endpoint you propose separately? it would be fairly quick to change `virtctl` to cope with it. In the meantime having the `migrate` functionality can be helpful to users as well as for us for debugging

cynepco3hahue: will work for me, I will not stop this PR on it, and will wait for the following one:)

slintes: I think Artyom is right: I don't remember the reason (maybe @fabiand or @mfranczy do?), but the other virtctl commands use subresources, and I think we should not change this.

> But I think such a new endpoint would render the existing one redundant

which existing one?

fabiand: The reason is simply that for every thing we wanted to do conveniently, the work was don on the server side first, and then exposed via virtctl. To avoid that we build up technical debt.

ahadas: > > But I think such a new endpoint would render the existing one redundant
> 
> which existing one?

Please ignore that part. Talked with Artyom offline earlier and realized I mixed two unrelated things here

ahadas: @cynepco3hahue @slintes this seems to require enabling kubevirt-apiserver to create resources on `virtualmachineinstancemigrations.kubevirt.io`, would it be ok?

slintes: @ahadas yes

ahadas: Done

tests/vm_test.go
```diff
@@ -753,6 +740,28 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
+					By("Invoking virtctl migrate")
+					migrateCommand := tests.NewRepeatableVirtctlCommand(vm.COMMAND_MIGRATE, "--namespace", virtualMachine.Namespace, virtualMachine.Name)
+					Expect(migrateCommand()).To(Succeed())
+
+					By("Ensuring the VirtualMachineInstance is migrated")
+					Eventually(func() bool {
+						nextVMI, err := virtClient.VirtualMachineInstance(virtualMachine.Namespace).Get(virtualMachine.Name, &v12.GetOptions{})
+						Expect(err).ToNot(HaveOccurred())
+						return nextVMI.Status.MigrationState != nil && nextVMI.Status.MigrationState.Completed
+					}, 240*time.Second, 1*time.Second).Should(BeTrue())
```
SchSeba: maybe we need to add a bit more time?

tests/vm_test.go
```diff
@@ -753,6 +740,28 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
 				})
 
+				It("should migrate a running VM", func() {
+					By("creating a VM with RunStrategyAlways")
+					virtualMachine := newVirtualMachineWithRunStrategy(v1.RunStrategyAlways)
+
+					By("Waiting for VM to be ready")
+					Eventually(func() bool {
+						virtualMachine, err = virtClient.VirtualMachine(virtualMachine.Namespace).Get(virtualMachine.Name, &v12.GetOptions{})
+						Expect(err).ToNot(HaveOccurred())
```
SchSeba: the expect will close the `Eventually` here?

ahadas: yeah, I believe it will. it was actually copied from `startVM`

tests/vm_test.go
ahadas: 240 sec is the timeout we use in migration_test.go and that seems to work

fabiand:
Yey.

PR #2646, opened 2019-08-29
Flaky tests: wait for configmaps updated values to be consolidated

fedepaol:

There are a bunch of flaky tests (not only `Slirp Networking` I was targeting) that use `tests.UpdateClusterConfigValue`, which turned out to apply the change and return.
Since the change _might_ not be instantaneous, the tests may sometimes fail introducing flakyness.

With this PR the function waits for the change to be applied before returning.

**Release note**:

```release-note

NONE

```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -3864,8 +3864,8 @@ func GenerateHelloWorldServer(vmi *v1.VirtualMachineInstance, testPort int, prot
 }
 
 // UpdateClusterConfigValueAndWait updates the given configuration in the kubevirt config map and then waits
-// for the given duration, to allow the configuration events to be propagated to the consumers.
-func UpdateClusterConfigValueAndWait(key string, value string, howLong time.Duration) {
+// to allow the configuration events to be propagated to the consumers.
+func UpdateClusterConfigValueAndWait(key string, value string) {
 	virtClient, err := kubecli.GetKubevirtClient()
 	PanicOnError(err)
 	cfgMap, err := virtClient.CoreV1().ConfigMaps(KubeVirtInstallNamespace).Get(kubevirtConfig, metav1.GetOptions{})
@@ -3874,7 +3874,7 @@ func UpdateClusterConfigValueAndWait(key string, value string, howLong time.Dura
 	_, err = virtClient.CoreV1().ConfigMaps(KubeVirtInstallNamespace).Update(cfgMap)
 	ExpectWithOffset(1, err).ToNot(HaveOccurred())
 
-	time.Sleep(howLong)
+	time.Sleep(2 * time.Second)
 }
 
 func WaitAgentConnected(virtClient kubecli.KubevirtClient, vmi *v1.VirtualMachineInstance) {
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -3863,14 +3863,18 @@ func GenerateHelloWorldServer(vmi *v1.VirtualMachineInstance, testPort int, prot
 	Expect(err).ToNot(HaveOccurred())
 }
 
-func UpdateClusterConfigValue(key string, value string) {
+// UpdateClusterConfigValueAndWait updates the given configuration in the kubevirt config map and then waits
+// for the given duration, to allow the configuration events to be propagated to the consumers.
+func UpdateClusterConfigValueAndWait(key string, value string, howLong time.Duration) {
 	virtClient, err := kubecli.GetKubevirtClient()
 	PanicOnError(err)
 	cfgMap, err := virtClient.CoreV1().ConfigMaps(KubeVirtInstallNamespace).Get(kubevirtConfig, metav1.GetOptions{})
-	Expect(err).NotTo(HaveOccurred())
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
 	cfgMap.Data[key] = value
 	_, err = virtClient.CoreV1().ConfigMaps(KubeVirtInstallNamespace).Update(cfgMap)
-	Expect(err).ToNot(HaveOccurred())
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
+
+	time.Sleep(howLong)
 }
 
 func WaitAgentConnected(virtClient kubecli.KubevirtClient, vmi *v1.VirtualMachineInstance) {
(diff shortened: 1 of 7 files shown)


slintes:
good catch, thx!

/lgtm
/approve

tests/utils.go
```diff
@@ -3838,6 +3838,12 @@ func UpdateClusterConfigValue(key string, value string) {
 	cfgMap.Data[key] = value
 	_, err = virtClient.CoreV1().ConfigMaps(KubeVirtInstallNamespace).Update(cfgMap)
 	Expect(err).ToNot(HaveOccurred())
+
+	Eventually(func() string {
+		cfgMap, err := virtClient.CoreV1().ConfigMaps(KubeVirtInstallNamespace).Get(kubevirtConfig, metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		return cfgMap.Data[key]
+	}).Should(BeEquivalentTo(value))
```
fromanirh: Hey, good catch. Do we want to introduce a delay in the Eventually to avoid a busy loop? Also, do we want to add a timeout, and fail if the configMap wasn't updated in a timely way?

fedepaol: `Eventually` has defaults, both for the timeout and for the polling interval (so it's not really a busy loop).

As per [gomega doc](https://onsi.github.io/gomega/)s: 

> The default value for the timeout is 1 second and the default value for the polling interval is 10 milliseconds. You can change these values by passing them in just after your function

They seemed enough to me and to the results of (some of) the CI runs  (at least for changing a config map, we have looser values in other cases), but I see no problem in relaxing them a bit.

booxter:
@fedepaol perhaps I am missing something... Is k8s API asynchronous? (meaning, you may get a result from Update() while the update is still being processed?) First time I hear about this.

Is it possible that the issue is not about Update() returning too early but about the cluster not applying the new option value to all depending components in time?

I have a hard time believing that Update() returns before the state is changed in etcd. Have you been able to reproduce the issue without the patch and failed to reproduce with the fix?

tests/utils.go
fromanirh: ok, LGTM

fedepaol:
> @fedepaol perhaps I am missing something... Is k8s API asynchronous? (meaning, you may get a result from Update() while the update is still being processed?) First time I hear about this.
> 
> Is it possible that the issue is not about Update() returning too early but about the cluster not applying the new option value to all depending components in time?
> 
> I have a hard time believing that Update() returns before the state is changed in etcd. Have you been able to reproduce the issue without the patch and failed to reproduce with the fix?

That's a good point. It was kind of my assumption because all the tests involving `UpdateClusterConfigValue` appear in the dreaded flaky list, and the other case where we use `ConfigMaps.Update` there always is a Sleep(5) afterwards, but looking better when the hook needs the configuration values does not read directly from etcd but invokes the `Informer` which probably does something like you say (like with callbacks).

Need to dig deeper.

/hold

/hold

slintes:
> Is it possible that the issue is not about Update() returning too early but about the cluster not applying the new option value to all depending components in time?

Good point indeed. Didn't think enough about it yesterday. The delay might be in the configmap informer, which needs to pick up the new value from etcd. I have no quick idea on how to check if it is up to date from a functional test. That's probably why Federico spotted sleeps at other places.

/approve cancel

petrkotas:
The issue might be similar to what I reported here #2557. There is known issue within k8s API that sometimes return enexpected EOF. This occurs not only for GET but AFAIK for other calls.

However I am unsure how to approach it, since it mich be related to deeper issues in the k8s.

fedepaol:
> > Is it possible that the issue is not about Update() returning too early but about the cluster not applying the new option value to all depending components in time?
> 
> Good point indeed. Didn't think enough about it yesterday. The delay might be in the configmap informer, which needs to pick up the new value from etcd. I have no quick idea on how to check if it is up to date from a functional test. That's probably why Federico spotted sleeps at other places.
> 
> /approve cancel

If I understood correctly (the code is a bit complex :blush: ), the informer relies on a listwatch built from the rest client which, among the other things, ends up filling the store where the hooks reads from.
So yes, I think @booxter is right. Is more about having the event propagated than having it available for reading. 

Now the quesiton is, is there a way other than Sleep to be sure that the apiserver received the value we just set?

fedepaol:
> The issue might be similar to what I reported here #2557. There is known issue within k8s API that sometimes return enexpected EOF. This occurs not only for GET but AFAIK for other calls.
> 
> However I am unsure how to approach it, since it mich be related to deeper issues in the k8s.

@petrkotas not sure it's direclty related though (or at least I hope it's not) since here we are reading from the `informer`'s store object that is fed by a list watch built from the rest client (which I am not sure how works internally). 

It could be that it's a latency on the propagation more than a lost event.

booxter:
@fedepaol while sleeps are not exactly the most beautiful approach possible, I would be fine if we introduce a new helper that updates the config map, that would do the sleeping, with a comment explaining the current thinking about why we need it, and then reusing it in all other places where we already sleep now.

One other crazy idea I have is somehow using https://kubernetes.io/docs/reference/using-api/api-concepts/#dry-run We would come up with some request that would fail if the option is not yet applied through the cluster, and then spin on it until the dry-run request succeeds. Not sure there's such a request in this particular case though.

tests/utils.go
```diff
@@ -3863,14 +3863,18 @@ func GenerateHelloWorldServer(vmi *v1.VirtualMachineInstance, testPort int, prot
...
 	PanicOnError(err)
 	cfgMap, err := virtClient.CoreV1().ConfigMaps(KubeVirtInstallNamespace).Get(kubevirtConfig, metav1.GetOptions{})
-	Expect(err).NotTo(HaveOccurred())
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
 	cfgMap.Data[key] = value
 	_, err = virtClient.CoreV1().ConfigMaps(KubeVirtInstallNamespace).Update(cfgMap)
-	Expect(err).ToNot(HaveOccurred())
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
+
+	time.Sleep(howLong)
```
slintes: I see you are using 2 and 5 seconds as timeout, why are they different? Wouldn't it be easier to use the same for all and just hardcode it here?

fedepaol: Yes, it was one of the things I was uncertain about.

The current value (when applied) was 5 seconds, which seems like a lot to me so I did not want to put a lot of 5 seconds here and there. On the other hand, I was not sure about lowering the current sleep time where applied.

It's also true that we can put a constant and tune until we find a one size / fits all. Shall I try with 2 seconds?

slintes: > Shall I try with 2 seconds?

yes, let's try that and have a careful look at test results :+1:

slintes:
/lgtm
/approve

retriggering all test to make sure we don't have new flakes (would have been done by tide anyway)

/test all

PR #2648, opened 2019-08-29
adding flag to make the controller threads configurable

shengnuo:
There was a bottleneck with the number of threads when trying to create a big replica set of size 1000. The code was modified to increase the threads from 3 to 10 (or any other greater value) and the performance was improved. Opening this PR so that users can configure according to the underlying hardware they use.
 
**Release notes**:
```release-note
allow configurable threads in virt-controller
```

### DIFF
--- a/pkg/virt-controller/watch/application.go
+++ b/pkg/virt-controller/watch/application.go
@@ -129,6 +129,7 @@ type VirtControllerApp struct {
 	// the channel used to trigger re-initialization.
 	reInitChan chan string
 
+	// number of threads for each controller
 	nodeControllerThreads             int
 	vmiControllerThreads              int
 	rsControllerThreads               int
@@ -440,6 +441,7 @@ func (vca *VirtControllerApp) AddFlags() {
 	flag.StringVar(&vca.containerDiskDir, "container-disk-dir", containerDiskDir,
 		"Base directory for container disk data")
 
+	// allows user-defined threads based on the underlying hardware in use
 	flag.IntVar(&vca.nodeControllerThreads, "node-controller-threads", defaultControllerThreads,
 		"Number of goroutines to run for node controller")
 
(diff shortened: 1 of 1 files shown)


slintes:
> create a big replica set of size 1000

Interesting, do you mind telling us a bit more about how you are using KubeVirt? :slightly_smiling_face:
@fabiand fyi

How do you get the new flags into the controller deployment? Are you not using the operator yet? 

/ok-to-test

cynepco3hahue:
The PR makes sense to me, I just do not sure that we need different number of threads for each controller, maybe use the same number of threads for every controller under the virt-controller?

slintes:
@stu-gott wdyt? New flags aren't exposed on the operator, but they also don't hurt. So if it helps, I'm fine with it.

/lgtm

smiteshNVIDIA:
> > create a big replica set of size 1000
> 
> Interesting, do you mind telling us a bit more about how you are using KubeVirt? 🙂
> @fabiand fyi

We have a use case where we spin up large number of use and throw VMs and we saw that it takes quite long time to schedule them. That's what prompted us to dig deeper. 

> 
> How do you get the new flags into the controller deployment? Are you not using the operator yet?
> 
> /ok-to-test

We are not using operator yet.

smiteshNVIDIA:
> The PR makes sense to me, I just do not sure that we need different number of threads for each controller, maybe use the same number of threads for every controller under the virt-controller?

It really depends on ones use case and available resources to run controllers. e.g. For our use case we thought it's kind of waste of resources to run 10 migrationController thread and we would rather run more count of threads for other controller.

slintes:
> > The PR makes sense to me, I just do not sure that we need different number of threads for each controller, maybe use the same number of threads for every controller under the virt-controller?
> 
> It really depends on ones use case and available resources to run controllers. e.g. For our use case we thought it's kind of waste of resources to run 10 migrationController thread and we would rather run more count of threads for other controller.

makes sense :)
thanks for the feedback!

/approve
/retest

PR #2653, opened 2019-09-02
Relaxes the wait for pods in entrire cluster

petrkotas:
Our test infrastructure waits for the entrire
cluster to become ready. This can create unreasonable
timeout due to large number of pods in OpenShift clusters.

More reasonable approach is to wait only for KubeVirt related
infrastructure.

Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=1747945

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt/2653)

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -437,7 +437,7 @@ func WaitForAllPodsReady(timeout time.Duration, listOptions metav1.ListOptions) 
 		virtClient, err := kubecli.GetKubevirtClient()
 		PanicOnError(err)
 
-		podsList, err := virtClient.CoreV1().Pods(k8sv1.NamespaceAll).List(listOptions)
+		podsList, err := virtClient.CoreV1().Pods(KubeVirtInstallNamespace).List(listOptions)
 		PanicOnError(err)
 		for _, pod := range podsList.Items {
 			for _, status := range pod.Status.ContainerStatuses {


petrkotas:
@dollierp is this change sufficient to make the test start under the test infrastructure?

If not, what are the issues encountered? What components are causing the prolonged wait period?
Thank you for your help figuring out the issue.

dollierp:
IIRC, @AsherShoshan had troubles with hco-operator Pod not ready. Since it is in KubeVirtInstallNamespace I guess this patch won't be sufficient.

@AsherShoshan, could you confirm please?

petrkotas:
In case of troubles if HCO operands, we need to investigate why it fails. The HCO deploys CNV and the tests are testing CNV. I do no feel comfortable disabling check without proper investigation.

rmohr:
Note that also on the current configured CI test-lanes we ensure that all pods in the cluster are ready. I can see that this may not apply to all  cluster if others are using it too. However, for testing I would strongly recommend to setup clusters in a way that all pods become ready, or at least specific pods are explicitly skipped, to avoid running tests on half-working clusters which lead to unpredictable test results.

rmohr:
/hold

blocking until this was more discussed.

petrkotas:
Do we still need this PR? I would close it as it seem to me, that it was rendered unnecessary. We already agreed that the failures were due to unhealthy infrastructure.

slintes:
+1 for closing, since nobody complained about it anymore

petrkotas:
Closing, it can be redone anytime, if required.

PR #2656, opened 2019-09-03
not check failed status when doing live migration hook

### LINKED ISSUE #2652: Avoid more than 1 ongoing live migrations
/kind bug

**What happened**:
It is possible to initiate two live migrations for the same VMI at the same time

The system should have blocked me from initiating an additional live-migration until the first migration is completed (successfully or with failure)

1. Create VMI
2. Post one migration job for the VMI
3. Post another migration job for the VMI before the previous migration is completed

The second migration attempt would probably be rejected by libvirt, but it would be better to block it earlier at the apiserver level

jichenjc:
the live migration hook check should only check completed status , not the failed statue which just indicate a error, sometimes a good live migratino only mark complete but no failed status

Related to #2652 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-api/webhooks/validating-webhook/admitters/migration-create-admitter.go
+++ b/pkg/virt-api/webhooks/validating-webhook/admitters/migration-create-admitter.go
@@ -88,8 +88,7 @@ func (admitter *MigrationCreateAdmitter) Admit(ar *v1beta1.AdmissionReview) *v1b
 	// are already in flight.
 	if vmi.Status.MigrationState != nil &&
 		string(vmi.Status.MigrationState.MigrationUID) != "" &&
-		!vmi.Status.MigrationState.Completed &&
-		!vmi.Status.MigrationState.Failed {
+		!vmi.Status.MigrationState.Completed {
 
 		return webhooks.ToAdmissionResponseError(fmt.Errorf("in-flight migration detected. Active migration job (%s) is currently already in progress for VMI %s.", string(vmi.Status.MigrationState.MigrationUID), vmi.Name))
 	}


pkg/virt-api/webhooks/validating-webhook/admitters/migration-create-admitter.go
```diff
@@ -88,8 +88,7 @@ func (admitter *MigrationCreateAdmitter) Admit(ar *v1beta1.AdmissionReview) *v1b
 	// are already in flight.
 	if vmi.Status.MigrationState != nil &&
 		string(vmi.Status.MigrationState.MigrationUID) != "" &&
-		!vmi.Status.MigrationState.Completed &&
-		!vmi.Status.MigrationState.Failed {
+		!vmi.Status.MigrationState.Completed {
```
slintes: hm, I think this is more tricky... when I see correct, an aborted migration has Completed = false and Failed = false, but starting a new migration would be fine imho. My next idea was to check if EndTimestamp > StartTimestamp, but on a quick look I think EndTimeStamp is not set for failed migrations... @vladikr, can you help please?

ahadas:
I don't think it would be enough to resolve #2652: let's say we get two migration requests at the same time and no migration was executed before, both may pass this part of the validation webhook.

jichenjc:
oh... that's a good suggestion, this seems need some lock introduced..

ahadas:
> oh... that's a good suggestion, this seems need some lock introduced..

👍 
 
> but this should fix part of the issue?

Yeah, I believe it would fix part of the issue.
Another part I'm not so sure about is whether the following race is possible:
1. Request A gets to the apiserver
2. VirtualMachineInstanceMigration is added for request A
3. The apiserver returns positive response
4. Request B gets to the apiserver
5. VirtualMachineInstanceMigration is added for request B
6. The VMI is updated with the status of the migration initiated by request A

PR #2657, opened 2019-09-03
Updated download location for some rpms needed for test images

slintes:
Currently bazel build and so all test lanes fail, because some rpms moved from dl.fedoraproject.org to archives.fedoraproject.org. Fixed download location of those rpms.

This is a quick fix, in the longer run we should probably move to rpms of newer fedora versions.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -270,7 +270,7 @@ http_file(
     name = "qemu-img",
     sha256 = "eadbd81fe25827a9d7712d0d96b134ab834bfab9e7332a8e9cf54dedd3c02581",
     urls = [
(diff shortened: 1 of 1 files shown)


cynepco3hahue:
Or we should move all relevant dependencies to our google storage.
/lgtm
/approve

slintes:
> Or we should move all relevant dependencies to our google storage.

or both :wink: 

> but on the long term the right approach is to switch our base images to something supported for a longer timespan, like centos/ubi. 

works for me :+1:

slintes:
first test lanes are green, will merge manually in order to make other PRs testable again

PR #2662, opened 2019-09-03
When waiting for a vm phase fails, log the name of the vm we are waiting for

fedepaol:
When a wait fails, it does not log the name of the vm we are waiting for, so in tests were we create multiple vms is not easy to track which one failed. Having the name tracked would help looking for the right launcher logs.

```release-note

NONE

```


slintes:
:+1: 

/lgtm
/approve

PR #2663, opened 2019-09-03
Use masquerade for tests with explicit pod interface

booxter:
Bridge has issues with fe: MAC address prefix, so unless a test
case depends on bridge type used, we should avoid it.

```release-note
NONE
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -2161,7 +2161,7 @@ func NewRandomVMIWithMasqueradeInterfaceEphemeralDiskAndUserdata(containerImage 
 }
 
 func AddExplicitPodNetworkInterface(vmi *v1.VirtualMachineInstance) {
-	vmi.Spec.Domain.Devices.Interfaces = []v1.Interface{*v1.DefaultBridgeNetworkInterface()}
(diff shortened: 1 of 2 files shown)


fedepaol:
/lgtm

@booxter you should add release notes: NONE I think

slintes:
@booxter can you please check failed tests? There are some which look unfamiliar to me... thanks.
https://prow.apps.ovirt.org/pr-history/?org=kubevirt&repo=kubevirt&pr=2663

/release-note-none

booxter:
/retest

Couldn't reproduce the failures even when using the exact same providers.

booxter:
Right now the PR fails on build stage (because of git clone failure and because of missing file on a server).

slintes:
missing file was fixed in the late (CEST) evening
/test all

booxter:
/hold

at least windows failure is related to this patch b/c the test case expects the VMI to have the same IP address that was allocated to the launcher pod, which is not the case for `masquerade`.

booxter:
Only the windows test failure seems related. The rest are unrelated (based on code path analysis) or can't be reproduced locally (those in vmi_networking_test.go in one of failed jobs).

booxter:
/retest

all failures seem unrelated

booxter:
@slintes looks finally ready to go CI-wise!

Issue #2664, opened 2019-09-04
Unable to launch virt-handler v0.20.4 on CoreOS with selinux


CoRfr:
/kind bug

**What happened**:
I started kubevirt v0.20.4 using some instructions from https://kubevirt.io/quickstart_minikube/ on a Kubernetes v1.15 cluster running on bare-metal nodes with CoreOS 1883.0.0 (~1yo ...).

**What you expected to happen**:
`virt-handler` fails to start with:
```
{"component":"virt-handler","hostname":"carmd-er-lxtb0cc47a4c27b6","level":"info","pos":"virt-handler.go:187","timestamp":"2019-09-04T06:02:51.362115Z"}
panic: failed to list labels:  

goroutine 1 [running]:
main.(*virtHandlerApp).Run(0xc0001f6000)
	cmd/virt-handler/virt-handler.go:205 +0x1ad6
main.main()
	cmd/virt-handler/virt-handler.go:457 +0x6e
```

**How to reproduce it (as minimally and precisely as possible)**:
I think running kubevirt on CoreOS should be enough, but hard to say.

**Anything else we need to know?**:
It looks like the `selinux/labels.go` is calling `semanage`, and I couldn't find it in the image.

**Environment**:
- KubeVirt version (use `virtctl version`): v0.20.4
- Kubernetes version (use `kubectl version`): v1.15.1
- VM or VMI specifications: N/A
- Cloud provider or hardware configuration: bare-metal x86-64 servers
- OS (e.g. from /etc/os-release): CoreOS 1883.0.0 (w/ ACS patch)
- Kernel (e.g. `uname -a`): 4.18.5-coreos
- Install tools:
- Others:

CoRfr:
Also the created date of the image is not set: `"Created": "1970-01-01T00:00:00Z"`

cynepco3hahue:
@CoRfr thanks for the report, looks like we have wrong log message - https://github.com/kubevirt/kubevirt/blob/0a69f68c14f26af53d21d454dd087e038a68335e/pkg/virt-handler/selinux/labels.go#L119
we do not really print the error, so its hard to say what the problem, I will create the PR with the fix, and can you try again after it will be merged, just to see what the error?

slintes:
> Also the created date of the image is not set: `"Created": "1970-01-01T00:00:00Z"`

Works as designed, that's how bazel produces reproducable images.

CoRfr:
[...]

PR #2665, opened 2019-09-04
Log the error under the SELinux package

cynepco3hahue:
See - https://github.com/kubevirt/kubevirt/issues/2664#issuecomment-527766499

Fixes #

**Release note**:

```release-note
Log errors under virt-handler SELinux package
```

### DIFF
--- a/pkg/virt-handler/selinux/labels.go
+++ b/pkg/virt-handler/selinux/labels.go
@@ -103,20 +103,21 @@ func copyPolicy(policyName string, dir string) (err error) {
 	return nil
 }
 
-func (se *SELinuxImpl) Label(label string, dir string) (err error) {
+// Label sets selinux label on the directory
+func (se *SELinuxImpl) Label(label string, dir string) error {
 	dir = strings.TrimRight(dir, "/") + "(/.*)?"
-	out, err := se.execute("semanage", se.Paths, "fcontext", "-a", "-t", label, dir)
-	if err != nil {
-		return fmt.Errorf("failed to set label for directory %v: %v ", dir, string(out))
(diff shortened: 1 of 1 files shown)


pkg/virt-handler/selinux/labels.go
```diff
@@ -103,32 +103,33 @@ func copyPolicy(policyName string, dir string) (err error) {
...
 
-func (se *SELinuxImpl) Label(label string, dir string) (err error) {
+// Label sets selinux label on the directory
+func (se *SELinuxImpl) Label(label string, dir string) error {
 	dir = strings.TrimRight(dir, "/") + "(/.*)?"
-	out, err := se.execute("semanage", se.Paths, "fcontext", "-a", "-t", label, dir)
-	if err != nil {
-		return fmt.Errorf("failed to set label for directory %v: %v ", dir, string(out))
+	if _, err := se.execute("semanage", se.Paths, "fcontext", "-a", "-t", label, dir); err != nil {
+		return fmt.Errorf("failed to set label for directory %v: %v ", dir, err)
```
slintes: Would it make sense to return both the output of the command and the error?
Same at other places

cynepco3hahue: To be honest I do not sure what the output will include, at least from the issue I saw that it is empty, but I can change the PR to print output as well.

slintes: I'm also not sure if it will ever provide useful info, just want to prevent that we lose info while adding other info... and it doesn't hurt to have both imho :)

slintes:
:+1: 

/lgtm
/approve

PR #2670, opened 2019-09-04
Do not consider SELinux in Permissive mode

### LINKED ISSUE #2664: Unable to launch virt-handler v0.20.4 on CoreOS with selinux

CoRfr:
From the application point-of-view, Permissive mode should be
the same as Disabled mode.

So far there is a dependency for the `semanage` binary to be available on the host if SELinux is not `Disabled`. On CoreOS `semanage` is not available, and SELinux is in `Permissive` mode.
Since `Permissive` mode doesn't seem to have any effect on applications, this patch consider this mode just as if it was `Disabled`.

Fixes #2664 

**Release note**:

```release-note
Removed dependency on host semanage in SELinux Permissive mode
```

### DIFF
--- a/pkg/virt-handler/selinux/BUILD.bazel
+++ b/pkg/virt-handler/selinux/BUILD.bazel
@@ -5,6 +5,7 @@ go_library(
     srcs = ["labels.go"],
     importpath = "kubevirt.io/kubevirt/pkg/virt-handler/selinux",
     visibility = ["//visibility:public"],
+    deps = ["//staging/src/kubevirt.io/client-go/log:go_default_library"],
 )
 
 go_test(
--- a/pkg/virt-handler/selinux/labels.go
+++ b/pkg/virt-handler/selinux/labels.go
@@ -7,6 +7,8 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+
+	"kubevirt.io/client-go/log"
 )
 
 const procOnePrefix = "/proc/1/root"
@@ -54,6 +56,17 @@ func (se *SELinuxImpl) IsPresent() (present bool, err error) {
 		return false, nil
 	}
 	return true, nil
+}
+
+func (se *SELinuxImpl) IsPermissive() (present bool, err error) {
+	out, err := se.execute("getenforce", se.Paths)
+	if err != nil {
+		return false, err
+	}
+	if strings.Contains(string(out), "Permissive") {
+		return true, nil
+	}
+	return false, nil
 }
 
 func lookupPath(binary string, prefix string, paths []string) (string, bool, error) {
@@ -107,6 +120,10 @@ func copyPolicy(policyName string, dir string) (err error) {
 func (se *SELinuxImpl) Label(label string, dir string) error {
 	dir = strings.TrimRight(dir, "/") + "(/.*)?"
 	if out, err := se.execute("semanage", se.Paths, "fcontext", "-a", "-t", label, dir); err != nil {
+		if perm, _ := se.IsPermissive(); perm {
+			log.Log.Warningf("Permissive mode, ignoring 'semanage' failure: out: %q, error: %v", string(out), err)
+			return nil
+		}
 		return fmt.Errorf("failed to set label for directory %v: out: %q, error: %v", dir, string(out), err)
 	}
 	return nil
@@ -117,6 +134,10 @@ func (se *SELinuxImpl) IsLabeled(dir string) (bool, error) {
 	dir = strings.TrimRight(dir, "/") + "(/.*)?"
 	out, err := se.execute("semanage", se.Paths, "fcontext", "-l")
 	if err != nil {
+		if perm, _ := se.IsPermissive(); perm {
+			log.Log.Warningf("Permissive mode, ignoring 'semanage' failure: out: %q, error: %v", string(out), err)
+			return false, nil
+		}
 		return false, fmt.Errorf("failed to list labels: out: %q, error: %v", string(out), err)
 	}
 	if strings.Contains(string(out), dir) {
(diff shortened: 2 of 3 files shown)


CoRfr:
> Thanks for the PR, LGTM.
> But do you mind adding a test case for `Permissive` similar to the one for `Disabled` in labels_test.go?

Yes good idea, done.

slintes:
:+1: 

/ok-to-test

CoRfr:
Just rebased to include fix for #2671

slintes:
> Just rebased to include fix for #2671

thanks, but not really needed (as long as there are no merge conflicts with master). The tests run on a potential merge commit with master :slightly_smiling_face:

/retest

slintes:
(I do not really expect any relevant test failures on this PR, but every test run gives at least some data about flaky tests we need to fix :slightly_smiling_face: )

cynepco3hahue:
I do not have problems with this PR, but will it really fix your problem, does CoreOS always has `Permissive` state?

CoRfr:
I'm not sure if it's always the case, but so far I've checked in multiple environments (bare-metal and deployed through Azure) and it's always `Permissive`.

CoreOS provide some documentation to enforce SELinux: https://coreos.com/os/docs/latest/selinux.html
In that case this PR wouldn't help much, because kubevirt would still rely on `semanage` in that case, which is not shipped by CoreOS.
I'm not an expert in SELinux at all so I wouldn't know if there is an alternative to `semanage` that already ships with CoreOS.

However, I do think that this PR would help most people that runs CoreOS since `Permissive` is the default option.

CoRfr:
Actually I just though that the `Permissive` mode seems to be for people interested in debugging their SELinux environment, so managing the labels, or at least trying to, seems like it would make sense.

I'm thinking that, instead of ignoring SELinux when in `Permissive` mode, we could make the calls to `IsLabeled` & `Label` non-fatal.

CoRfr:
I changed the patch to reflect "try to do things and ignore if Permissive" approach. If you think that's a worst idea I can restore the original patch (https://github.com/CoRfr/kubevirt/commit/e62e2cd583a61db8ae9cdd2af716a97ff0e703dd).

CoRfr:
With this change I have:
```
{"component":"virt-handler","hostname":"carmd-er-lxtb0cc47a4c27b6","level":"info","pos":"virt-handler.go:187","timestamp":"2019-09-11T18:30:06.105254Z"}
{"component":"virt-handler","level":"warning","msg":"Permissive mode, ignoring 'semanage' failure: out: \"\", error: could not find binary semanage","pos":"labels.go:138","timestamp":"2019-09-11T18:30:06.178813Z"}
{"component":"virt-handler","level":"warning","msg":"Permissive mode, ignoring 'semanage' failure: out: \"\", error: could not find binary semanage","pos":"labels.go:124","timestamp":"2019-09-11T18:30:06.189779Z"}
{"component":"virt-handler","level":"warning","msg":"Permissive mode, ignoring 'semanage' failure: out: \"\", error: could not find binary semanage","pos":"labels.go:138","timestamp":"2019-09-11T18:30:06.223830Z"}
{"component":"virt-handler","level":"warning","msg":"Permissive mode, ignoring 'semanage' failure: out: \"\", error: could not find binary semanage","pos":"labels.go:124","timestamp":"2019-09-11T18:30:06.232716Z"}
{"component":"virt-handler","level":"warning","msg":"Permissive mode, ignoring 'semodule' failure: out: \"libsemanage.semanage_create_store: Could not create module store at /var/lib/selinux/mcs. (No such file or directory).\\nlibsemanage.semanage_direct_connect: could not establish direct connection (No such file or directory).\\n/usr/sbin/semodule:  Could not connect to policy handler\\n\", error: exit status 1","pos":"labels.go:172","timestamp":"2019-09-11T18:30:06.279743Z"}
{"component":"virt-handler","level":"info","msg":"Starting collector: node name=carmd-er-lxtb0cc47a4c27b6","pos":"prometheus.go:395","timestamp":"2019-09-11T18:30:07.330122Z"}
```
And it continues as if nothing happened, which is the expected behavior from that patch.

rmohr:
/lgtm
/approve

@CoRfr can you sign your commit?

CoRfr:
@rmohr , just signed the commit.
However, before you merge it, I'm gonna try to see if I can get your PR to work for the next 30 mins or so.

> My main issue with the change is that people are supposed to enable selinux once the app seems to work fine and the logs show no issues in permissive mode. So I am not sure if this goes into the right direction 
In the first version of this commit, I was considering SELinux Permissive == Absent, which would prevent fixing errors as it doesn't try to do things until you enable SELinux ... and then it would fail (at least on CoreOS, given that some stuff that it depends is not there.
I also realized that it might not be the best, so the second version (and latest) of the patch at least tries to do stuff, and just logs the failures as warnings, and continues.

It seems that you're taking a similar approach #2694 while removing the dependency to `semanage`, which is a better approach IMHO since it would allow it to go further in the SELinux setup that with my patch.

> @CoRfr do you know how people are supposed to configure selinux if the python binaries are not there on coreos?
TBH, I don't know much about SELinux on CoreOS or in general. I only read https://coreos.com/os/docs/latest/selinux.html and it seems like they are using `semodule` and `systemctl restart audit-rules`.

rmohr:
> It seems that you're taking a similar approach

Yep, however that was not orignially inteded. I did not know then that semodule does not work on container linux. That is why I thought I would take your PR first, since you spent time on it (and your PR works already) and then bring in the improved version on top of it.

CoRfr:
Sounds good, works for me.
I couldn't get #2694 to work in my env at the moment.

Issue #2672, opened 2019-09-05
Flaky tests: domain is started but the vmi is still scheduled


fedepaol:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>

/kind bug

While looking at this https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2576/pull-kubevirt-e2e-os-3.11.0-crio/1159042117620731904 flaky test report (test [rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:component]Networking Multiple virtual machines connectivity using bridge binding interface should be able to reach [test_id:1539]the Inbound VirtualMachineInstance), I found the following situation:

We have four vmis started, the launcher seems to be ok but the state of the vmi never reaches the running state.

Now, after spending a lot of time on this I feel I am hitting a wall, probably because of my ignorance of some of the parts involved.
I'll try to add as much value as I can from what I think I understood.

I'll focus on the logs of two out of the four vmis.

#### VMI testvmiw8s5wlv9zrkh857x4llpqrv728qbgld92t2q4qfpkpkxm8vs:
`Status.Phase == Failed`

Looking at [the launcher logs](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2576/pull-kubevirt-e2e-os-3.11.0-crio/1159042117620731904/artifacts/pods/2_kubevirt-test-default_virt-launcher-testvmiw8s5wlv9zrkh857x4llpqrv728qbgld92t2q4bkw5m-compute.log), everything seems fine.
The domain is started, and probably the event is passed to the handler.
```
{"component":"virt-launcher","level":"info","msg":"kubevirt domain status: Running(1):Unknown(1)","pos":"client.go:179","timestamp":"2019-08-07T15:09:50.918967Z"}
```

[In the handler](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2576/pull-kubevirt-e2e-os-3.11.0-crio/1159042117620731904/artifacts/pods/2_kubevirt_virt-handler-dmvz7-virt-handler.log) the running event is received, and it's the last dumped state of the domain in the handler:
```
[...]

PR #2674, opened 2019-09-05
Add logs to troubleshoot event handling issues.

fedepaol:

While investigating the issues described in https://github.com/kubevirt/kubevirt/issues/2672 , I found really hard to understand what the problem was. 

What seemed to happen there was that a domain running event reached the virt-handler but (ironically :-) ) it was not handled. for some reason.

This pr adds / extends logs to simplify the troubleshooting, in particular for:
- expired / missed watchdog events (which are supposed to be exceptional and have an high impact)
- When a domain event was received we already had a log in place, but it did not say what domain it was referring to making it harder to troubleshoot when more vmis are on the same node
- every time a worker processes an event it dumps the current phase of the vmi / domain, which are the main drivers to the evolution of the vmi

```release-note

NONE

```

### DIFF
--- a/pkg/virt-handler/cache/cache.go
+++ b/pkg/virt-handler/cache/cache.go
@@ -111,7 +111,7 @@ func (d *DomainWatcher) handleStaleWatchdogFiles() error {
 	}
 
 	for _, domain := range domains {
-		log.Log.Object(domain).Warning("detected expire watchdog for domain")
+		log.Log.Object(domain).Warning("detected expired watchdog for domain")
 		d.eventChan <- watch.Event{Type: watch.Deleted, Object: domain}
 	}
 	return nil
--- a/pkg/virt-handler/cache/cache.go
+++ b/pkg/virt-handler/cache/cache.go
@@ -111,6 +111,7 @@ func (d *DomainWatcher) handleStaleWatchdogFiles() error {
 	}
 
 	for _, domain := range domains {
+		log.Log.Object(domain).Warning("detected expire watchdog for domain")
 		d.eventChan <- watch.Event{Type: watch.Deleted, Object: domain}
 	}
 	return nil
--- a/pkg/virt-handler/vm.go
+++ b/pkg/virt-handler/vm.go
@@ -758,14 +758,14 @@ func (d *VirtualMachineController) defaultExecute(key string,
 
 	log.Log.Infof("Processing event %v", key)
 	if vmiExists {
-		log.Log.Infof("Vmi %s in phase %v", vmi.Name, vmi.Status.Phase)
+		log.Log.Object(vmi).Infof("VMI is in phase: %v\n", vmi.Status.Phase)
 	} else {
-		log.Log.Info("Vmi does not exist")
+		log.Log.Info("VMI does not exist")
 	}
 	if domainExists {
-		log.Log.Infof("Domain %s in status %v, reason: %v", domain.ObjectMeta.Name, domain.Status.Status, domain.Status.Reason)
+		log.Log.Object(domain).Infof("Domain status: %v, reason: %v\n", domain.Status.Status, domain.Status.Reason)
 	} else {
-		log.Log.Info("Vmi does not exist")
+		log.Log.Info("Domain does not exist")
 	}
 
 	domainAlive := domainExists &&
--- a/pkg/virt-handler/vm.go
+++ b/pkg/virt-handler/vm.go
@@ -756,14 +756,16 @@ func (d *VirtualMachineController) defaultExecute(key string,
 	// set true to ensure that no updates to the current VirtualMachineInstance state will occur
 	forceIgnoreSync := false
 
-	log.Log.V(3).Infof("Processing vmi %v, existing: %v\n", vmi.Name, vmiExists)
+	log.Log.Infof("Processing event %v", key)
 	if vmiExists {
-		log.Log.V(3).Infof("vmi is in phase: %v\n", vmi.Status.Phase)
+		log.Log.Infof("Vmi %s in phase %v", vmi.Name, vmi.Status.Phase)
+	} else {
+		log.Log.Info("Vmi does not exist")
 	}
-
-	log.Log.V(3).Infof("Domain: existing: %v\n", domainExists)
 	if domainExists {
-		log.Log.V(3).Infof("Domain status: %v, reason: %v\n", domain.Status.Status, domain.Status.Reason)
+		log.Log.Infof("Domain %s in status %v, reason: %v", domain.ObjectMeta.Name, domain.Status.Status, domain.Status.Reason)
+	} else {
+		log.Log.Info("Vmi does not exist")
 	}
 
 	domainAlive := domainExists &&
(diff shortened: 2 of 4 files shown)


pkg/virt-handler/cache/cache.go
```diff
@@ -111,6 +111,7 @@ func (d *DomainWatcher) handleStaleWatchdogFiles() error {
 	}
 
 	for _, domain := range domains {
+		log.Log.Object(domain).Warning("detected expire watchdog for domain")
```
ahadas: expired?

pkg/virt-handler/vm.go
```diff
@@ -756,14 +756,16 @@ func (d *VirtualMachineController) defaultExecute(key string,
 	// set true to ensure that no updates to the current VirtualMachineInstance state will occur
 	forceIgnoreSync := false
 
-	log.Log.V(3).Infof("Processing vmi %v, existing: %v\n", vmi.Name, vmiExists)
+	log.Log.Infof("Processing event %v", key)
```
ahadas: any reason not to have Object(vmi)?

pkg/virt-handler/vm.go
```diff
@@ -756,14 +756,16 @@ func (d *VirtualMachineController) defaultExecute(key string,
...
 	if vmiExists {
-		log.Log.V(3).Infof("vmi is in phase: %v\n", vmi.Status.Phase)
+		log.Log.Infof("Vmi %s in phase %v", vmi.Name, vmi.Status.Phase)
```
ahadas: how about changing that to VMI?

pkg/virt-handler/vm.go
```diff
@@ -756,14 +756,16 @@ func (d *VirtualMachineController) defaultExecute(key string,
...
+	} else {
+		log.Log.Info("Vmi does not exist")
 	}
-
-	log.Log.V(3).Infof("Domain: existing: %v\n", domainExists)
```
ahadas: I wonder why those logs used to be on V(3), they shouldn't be written too often, right?

pkg/virt-handler/vm.go
fedepaol: Was afraid it would have been too much stuff I think. Going to change it

pkg/virt-handler/vm.go
fedepaol: Should not happen more often that the log `Received Domain Event of type MODIFIED` we are already logging and I think they'll add a lot of useful information.

pkg/virt-handler/vm.go
fedepaol: not sure I understood. Do you mean Vmi -> VMI ?

pkg/virt-handler/vm.go
fedepaol: I'll also do the same for the domain object

pkg/virt-handler/vm.go
ahadas: yes

pkg/virt-handler/vm.go
```diff
@@ -756,14 +756,16 @@ func (d *VirtualMachineController) defaultExecute(key string,
 	// set true to ensure that no updates to the current VirtualMachineInstance state will occur
 	forceIgnoreSync := false
 
-	log.Log.V(3).Infof("Processing vmi %v, existing: %v\n", vmi.Name, vmiExists)
+	log.Log.Infof("Processing event %v", key)
 	if vmiExists {
-		log.Log.V(3).Infof("vmi is in phase: %v\n", vmi.Status.Phase)
+		log.Log.Object(vmi).Infof("VMI exists")
```
ahadas: this seems to reduce the amount of information we expose or is it that `Object(vmi)` would log something related to the status as well?

pkg/virt-handler/vm.go
```diff
@@ -756,14 +756,16 @@ func (d *VirtualMachineController) defaultExecute(key string,
...
+	} else {
+		log.Log.Info("VMI does not exist")
 	}
-
-	log.Log.V(3).Infof("Domain: existing: %v\n", domainExists)
 	if domainExists {
-		log.Log.V(3).Infof("Domain status: %v, reason: %v\n", domain.Status.Status, domain.Status.Reason)
+		log.Log.Object(domain).Infof("Domain exists")
```
ahadas: same question here

pkg/virt-handler/vm.go
fedepaol: My (wrong) understanding was that Object would log the whole object (as %v), but looking at the implementation you are right, thanks!
Will revert the status.

pkg/virt-handler/vm.go
fedepaol: Same as above

pkg/virt-handler/vm.go
```diff
@@ -756,14 +756,16 @@ func (d *VirtualMachineController) defaultExecute(key string,
 	// set true to ensure that no updates to the current VirtualMachineInstance state will occur
 	forceIgnoreSync := false
 
-	log.Log.V(3).Infof("Processing vmi %v, existing: %v\n", vmi.Name, vmiExists)
+	log.Log.Infof("Processing event %v", key)
 	if vmiExists {
-		log.Log.V(3).Infof("vmi is in phase: %v\n", vmi.Status.Phase)
+		log.Log.Object(vmi).Infof("VMI is in phase: %v\n", vmi.Status.Phase)
+	} else {
+		log.Log.Info("VMI does not exist")
```
jellonek: Shouldn't that be .Warning?

pkg/virt-handler/vm.go
```diff
@@ -756,14 +756,16 @@ func (d *VirtualMachineController) defaultExecute(key string,
...
 	}
-
-	log.Log.V(3).Infof("Domain: existing: %v\n", domainExists)
 	if domainExists {
-		log.Log.V(3).Infof("Domain status: %v, reason: %v\n", domain.Status.Status, domain.Status.Reason)
+		log.Log.Object(domain).Infof("Domain status: %v, reason: %v\n", domain.Status.Status, domain.Status.Reason)
+	} else {
+		log.Log.Info("Domain does not exist")
```
jellonek: ditto

fedepaol: I think it could (and will) happen the first time a vmi creation is handled and then the handler performs the SyncVMI call which in result creates the domain and generates another event to be processed, and it's part of the normal workflow so info is fine.

pkg/virt-handler/vm.go
fedepaol: As per my answer below, not 100% sure if this happens only in "exceptional" conditions

pkg/virt-handler/vm.go
jellonek: If it's normal - probably this log will provide only a noise in logs, so maybe it should be logged only on some non standard level (like previously used `V(3)`?)

fedepaol: It's normal the first time a vm gets created, not normal after that. 
I think the log is useful to diagnose exceptional events as those I described in https://github.com/kubevirt/kubevirt/issues/2672 
I would certainly not hide it behind a non standard level.
On top of that, the occourrence of these logs are reasonable enough to keep them in order to help us in bad situations.

PR #2682, opened 2019-09-10
[WIP] investigating migration failures

ahadas:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/virt-handler/vm.go
+++ b/pkg/virt-handler/vm.go
@@ -642,6 +642,22 @@ func (d *VirtualMachineController) migrationOrphanedSourceNodeExecute(key string
 	return nil
 }
 
+func isMigrating(start, stop *v12.Time) bool {
+	now := v12.Now()
+
+	running := false
+
+	if start != nil && (now.After(start.Time) || now.Equal(start)) {
+		running = true
+	}
+
+	if stop != nil && (now.After(stop.Time) || now.Equal(stop)) {
+		running = false
+	}
+
+	return running
+}
+
 func (d *VirtualMachineController) migrationTargetExecute(key string,
 	vmi *v1.VirtualMachineInstance,
 	vmiExists bool,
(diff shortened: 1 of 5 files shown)


pkg/virt-handler/vm.go
```diff
@@ -642,6 +642,22 @@ func (d *VirtualMachineController) migrationOrphanedSourceNodeExecute(key string
 	return nil
 }
 
+func isMigrating(start, stop *v12.Time) bool {
```
vladikr: I've commented this on the original PR. I think we need to make sure that `vmi.Status.MigrationState` isn't `nil`.
Perhaps let's change this to something like the following? :

```suggestion
func isMigrating(vmi *v1.VirtualMachineInstance) bool {                         
                                                                                
    now := v12.Now()                                                            
                                                                                
    running := false                                                            
    if vmi.Status.MigrationState != nil {                                       
        start := vmi.Status.MigrationState.StartTimestamp                       
        stop := vmi.Status.MigrationState.EndTimestamp                          
        if start != nil && (now.After(start.Time) || now.Equal(start)) {        
            running = true                                                      
        }                                                                       
                                                                                
        if stop != nil && (now.After(stop.Time) || now.Equal(stop)) {           
            running = false                                                     
        }                                                                       
    }                                                                           
    return running                                                              
}               
```

tests/migration_test.go
```diff
@@ -390,20 +390,23 @@ var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system]
 			})
 		})
 		Context("with a Cirros disk", func() {
-			It("[test_id:1783]should be successfully migrated multiple times with cloud-init disk", func() {
+			FIt("[test_id:1783]should be successfully migrated multiple times with cloud-init disk", func() {
```
vladikr: I think we should run the whole suite to catch the issue.

pkg/virt-handler/vm.go
ahadas: yeah, saw it on the original PR and I agree but I assume it will be handled there (@petrkotas ?), I just wanted to see how often the problem I was pointed to occurs with petr's fix

tests/migration_test.go
ahadas: ok, does it mean we suspect the problem to happen due to improper cleanup then?

pkg/virt-handler/vm.go
vladikr: OK..

tests/migration_test.go
vladikr: No, it doesn't mean that. I'm not sure what are the conditions that trigger this issue. I've seen the failures when the whole suite was ran.

PR #2683, opened 2019-09-10
Make use of IgnitionFile constant where possible

dollierp:

Hopefully this PR might help with code maintainability.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/ignition/ignition.go
+++ b/pkg/ignition/ignition.go
@@ -72,13 +72,13 @@ func GenerateIgnitionLocalData(vmi *v1.VirtualMachineInstance, namespace string)
 		return err
 	}
 
-	ignitionFile := fmt.Sprintf("%s/%s", domainBasePath, "data.ign")
+	ignitionFile := fmt.Sprintf("%s/%s", domainBasePath, IgnitionFile)
 	ignitionData := []byte(vmi.Annotations[v1.IgnitionAnnotation])
 	err = ioutil.WriteFile(ignitionFile, ignitionData, 0644)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/ignition/ignition.go
```diff
@@ -72,13 +72,13 @@ func GenerateIgnitionLocalData(vmi *v1.VirtualMachineInstance, namespace string)
...
-	ignitionFile := fmt.Sprintf("%s/%s", domainBasePath, "data.ign")
+	ignitionFile := fmt.Sprintf("%s/%s", domainBasePath, IgnitionFile)
 	ignitionData := []byte(vmi.Annotations[v1.IgnitionAnnotation])
 	err = ioutil.WriteFile(ignitionFile, ignitionData, 0644)
 	if err != nil {
 		return err
 	}
 
-	log.Log.V(2).Infof("generated Ignition file %s/data.ign", domainBasePath)
+	log.Log.V(2).Infof("generated Ignition file %s", ignitionFile)
```
ahadas: this seems to change the logged path, is it intentional?

dollierp: The path should be left unchanged. [ignitionFile](https://github.com/kubevirt/kubevirt/blob/8404ebab/pkg/ignition/ignition.go#L75) variable with lower case i is used here, not [IgnitionFile](https://github.com/kubevirt/kubevirt/blob/8404ebab/pkg/ignition/ignition.go#L34) constant.

ahadas: oh, you're right

PR #2687, opened 2019-09-10
Increase amount of memory allocated for container-disk container

booxter:
At least in some environments (e.g. when using sriov kind based
provider for kubevirt deployment), virt-launcher sometimes ends up in
OOMKilled state, and dmesg suggests that container-disk process is
killed:

Memory cgroup out of memory: Kill process 226667 (container-disk)
score 1001 or sacrifice child
Killed process 226431 (container-disk) total-vm:115804kB,
anon-rss:0kB, file-rss:3760kB, shmem-rss:0kB

With slight increase of memory allocated to the container, I no longer
get the launcher pod dying with OOMKilled state.

```release-note
NONE
```


cynepco3hahue:
@booxter Can you please signoff commis?

booxter:
@cynepco3hahue please explain what you mean. Is there a rule now that we should `git commit -s`?

PR #2691, opened 2019-09-11
Cleanup in functional tests

ahadas:
This PR removes two functions from `tests/utils.go` and minimize the generation of random names within the tests.

**Release note**:

```release-note
NONE
```


tests/migration_test.go
```diff
@@ -224,7 +224,7 @@ var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system]
 	runMigrationAndExpectCompletion := func(migration *v1.VirtualMachineInstanceMigration, timeout int) string {
 		By("Starting a Migration")
 		Eventually(func() error {
-			_, err := virtClient.VirtualMachineInstanceMigration(migration.Namespace).Create(migration)
+			migration, err = virtClient.VirtualMachineInstanceMigration(migration.Namespace).Create(migration)
```
vladikr: I think you need to do the same everywhere we create a migration object, otherwise, migration name will remain empty...

ahadas: Done

PR #2696, opened 2019-09-13
Make cluster-deploy issues better diagnosable

rmohr:

Use the test-failure-collector also when deploying kubevirt fails in CI, to collect cluster info. This makes debugging the following scenarios easier:

 * Code-changes lead to a situation where only on some test lanes the kubevirt installation succeeds (allows diagnose without having to rerun locally)
 * The cluster comes up and we can successfully install kubevirt, but some cluster- or test-infra pods never get ready.

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -242,3 +242,13 @@ genrule(
     cmd = "echo '#!/bin/sh\n\ncp $(SRCS) $$1' > \"$@\"",
     executable = 1,
 )
+
+genrule(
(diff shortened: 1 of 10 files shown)


rmohr:
@pkliczewski this is probably loosely related to must-gather. In this case we use again the test code for collecting the data to get unified info, since on k8s I would have to use `kubectl cluster-info dump` and on openshift I would have to use `must-gather`.

PR #2697, opened 2019-09-13
Fix some network/interface mapping validation issues

### LINKED ISSUE #2577: Interfaces set in domain.spec ignored if networks field left as default
**BUG REPORT**

/kind bug

**What happened**:

When creating a VMI (either directly or through a VM or a VMIRS), the VMI ended up with the default values for interfaces, even if something different was supplied. In other words, create a VMI with (say)
~~~
interfaces:
- name: default
  model: e1000
~~~
using `kubectl apply`. As soon as you have done so, get the object you just created with `kubectl get -o yaml`. The interfaces part of the spec is now the default, i.e.  
~~~
interfaces:
- name: default
  bridge: {}
~~~
Note that I have tried a range of different things in interfaces (different model names, multiple interfaces, and so on). In every case the interfaces struct got changed to the default as if not specified.

The controller correctly seems to reject invalid values for interfaces if I try that.

The interfaces field would be respected.

`kubectl apply -f ubuntuvmi.yaml` using the YAML below.

Underlying issue here is that I am trying to run with an image old enough to not support virtio, and was investigating using a standard ubuntu 14.04 cloud image. That needs an interface type which is not virtio but e1000 (or something it supports).

**Environment**:
- KubeVirt version (use `virtctl version`): 0.19.0
- Kubernetes version (use `kubectl version`): 1.12.8
- VM or VMI specifications: See below
Running in an Azure cluster (default AKS configuration) though I do not think that is relevant.

~~~
kind: VirtualMachineInstance
metadata:
  name: ubuntuvmi
  labels:
    billy: bob
spec:
  terminationGracePeriodSeconds: 30
  domain:
    resources:
      requests:
        memory: 1024M
        cpu: 2
    devices:
      disks:
      - name: containerdisk
        disk:
          bus: sata
      - name: emptydisk
        disk:
          bus: sata
      - disk:
          bus: sata
        name: cloudinitdisk
      interfaces:
      - name: default
        model: e1000       
  volumes:
  - name: containerdisk
    containerDisk:
      image: metaswitchglobal.azurecr.io/ubuntu:1404
  - name: emptydisk
    emptyDisk:
      capacity: "2Gi"
  - name: cloudinitdisk
    cloudInitNoCloud:
      userData: |-
        #cloud-config
        password: ubuntu
        chpasswd: { expire: False }
        ssh_pwauth: True
~~~

rmohr:

 * Don't add a default network if a single interface misses the network
configuration
 * Check for unique names in the network section
 * Explicitly say which network does not have an interface if it is
missing

Fixes #2577

**Release note**:

```release-note
Fix some network defaulting and validation issues
```

### DIFF
--- a/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator_test.go
+++ b/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator_test.go
@@ -358,7 +358,7 @@ var _ = Describe("VirtualMachineInstance Mutator", func() {
 		),
 	)
 
-	table.DescribeTable("should set default network interface",
+	table.DescribeTable("should add the default network interface",
 		func(iface string) {
 			expectedIface := "bridge"
 			switch iface {
@@ -390,7 +390,7 @@ var _ = Describe("VirtualMachineInstance Mutator", func() {
 		table.Entry("as slirp", "slirp"),
 	)
 
-	table.DescribeTable("should not default interfaces if", func(interfaces []v1.Interface, networks []v1.Network) {
+	table.DescribeTable("should not add the default interfaces if", func(interfaces []v1.Interface, networks []v1.Network) {
 		vmi.Spec.Domain.Devices.Interfaces = append([]v1.Interface{}, interfaces...)
 		vmi.Spec.Networks = append([]v1.Network{}, networks...)
 		vmiSpec, _ := getVMISpecMetaFromResponse()
@@ -664,5 +664,4 @@ var _ = Describe("VirtualMachineInstance Mutator", func() {
 		}
 		Expect(ok).To(BeTrue())
 	})
-
 })
--- a/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator_test.go
+++ b/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator_test.go
@@ -389,6 +389,27 @@ var _ = Describe("VirtualMachineInstance Mutator", func() {
 		table.Entry("as masquerade", "masquerade"),
 		table.Entry("as slirp", "slirp"),
 	)
+
+	table.DescribeTable("should not default interfaces if", func(interfaces []v1.Interface, networks []v1.Network) {
+		vmi.Spec.Domain.Devices.Interfaces = append([]v1.Interface{}, interfaces...)
+		vmi.Spec.Networks = append([]v1.Network{}, networks...)
+		vmiSpec, _ := getVMISpecMetaFromResponse()
+		if len(interfaces) == 0 {
+			Expect(vmiSpec.Domain.Devices.Interfaces).To(BeNil())
+		} else {
+			Expect(vmiSpec.Domain.Devices.Interfaces).To(Equal(interfaces))
+		}
+		if len(networks) == 0 {
+			Expect(vmiSpec.Networks).To(BeNil())
+		} else {
+			Expect(vmiSpec.Networks).To(Equal(networks))
+		}
+	},
+		table.Entry("interfaces and networks are non-empty", []v1.Interface{{Name: "a"}}, []v1.Network{{Name: "b"}}),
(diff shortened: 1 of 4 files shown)


ahadas:
> the changes look ok to me, but I wonder - what would be our strategy to support networkless VMs (if needed)? with an additional "autoattach default network" parameter like that for graphics?

Oh never mind, now that I have the whole code in front of me - I see the `obj.Spec.Domain.Devices.AutoattachPodInterface`

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter_test.go
```diff
@@ -1722,7 +1762,7 @@ var _ = Describe("Validating VMICreate Admitter", func() {
 			vmi.Spec.Domain.Devices.NetworkInterfaceMultiQueue = &_true
 
 			causes := ValidateVirtualMachineInstanceSpec(k8sfield.NewPath("fake"), &vmi.Spec, config)
-			Expect(len(causes)).To(Equal(1))
+			Expect(len(causes)).To(Equal(2))
```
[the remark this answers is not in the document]
rmohr: I think that I should have given a spec which only triggers the expected cause here. Will change it. @ahadas thx for  the review.

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
```diff
@@ -519,322 +519,331 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
...
-			// verify that selected model is supported
-			if iface.Model != "" {
-				isModelSupported := func(model string) bool {
-					for _, m := range validInterfaceModels {
-						if m == model {
-							return true
-						}
+		// verify that selected model is supported
+		if iface.Model != "" {
+			isModelSupported := func(model string) bool {
```
SchSeba: can we move the creation of the function outside the loops?

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
```diff
@@ -519,322 +519,331 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
...
+	}
+	// Network interface multiqueue can only be set for a virtio driver
+	if vifMQ != nil && *vifMQ && !isVirtioNicRequested {
 
+		causes = append(causes, metav1.StatusCause{
+			Type:    metav1.CauseTypeFieldValueInvalid,
+			Message: "virtio-net multiqueue request, but there are no virtio interfaces defined",
+			Field:   field.Child("domain", "devices", "networkInterfaceMultiqueue").String(),
+		})
+
```
alonSadan: I think this new line is not needed

pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator_test.go
```diff
@@ -389,6 +389,27 @@ var _ = Describe("VirtualMachineInstance Mutator", func() {
 		table.Entry("as masquerade", "masquerade"),
 		table.Entry("as slirp", "slirp"),
 	)
+
+	table.DescribeTable("should not default interfaces if", func(interfaces []v1.Interface, networks []v1.Network) {
```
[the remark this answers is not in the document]
rmohr: addressed

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
rmohr: replaced this and another occurrence with a map.

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
rmohr: addressed

PR #2699, opened 2019-09-15
Check if the service monitor namespace exists

### LINKED ISSUE #2688: V0.21.0 breaks with non open shift deployment
There needs to be a way for the operator to not break if the openshift metrics ns isn't there. Maybe making metrics a configurable yaml file element would be a good idea? 
>
/kind bug
> /kind enhancement

**What happened**:

**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

cynepco3hahue:
Verify that service monitor namespace exists before the creation of the
ServiceMonitor resource.

Fixes #2688

In general, it can be a good idea to create k8s cluster with Prometheus under our providers, just to verify such cases. But it can be done under the other PR.

**Release note**:

```release-note
Verify that service monitor namespace exists before the creation of the ServiceMonitor resource.
```

### DIFF
--- a/pkg/virt-operator/install-strategy/strategy.go
+++ b/pkg/virt-operator/install-strategy/strategy.go
@@ -27,7 +27,9 @@ import (
 
 	promv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/ghodss/yaml"
+	"github.com/golang/glog"
 	secv1 "github.com/openshift/api/security/v1"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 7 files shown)


pkg/virt-operator/install-strategy/strategy.go
```diff
@@ -201,8 +209,14 @@ func GenerateCurrentInstallStrategy(config *operatorutil.KubeVirtDeploymentConfi
...
 	monitorNamespace := config.GetMonitorNamespace()
-	monitorServiceAccount := config.GetMonitorServiceAccount()
-	rbaclist = append(rbaclist, rbac.GetAllServiceMonitor(config.GetNamespace(), monitorNamespace, monitorServiceAccount)...)
+	if addMonitorServiceResources {
+		// TODO: we should check that monitor SA exists in the monitor namespace
+		monitorServiceAccount := config.GetMonitorServiceAccount()
+		rbaclist = append(rbaclist, rbac.GetAllServiceMonitor(config.GetNamespace(), monitorNamespace, monitorServiceAccount)...)
+		strategy.serviceMonitors = append(strategy.serviceMonitors, components.NewServiceMonitorCR(config.GetNamespace(), monitorNamespace, true))
+	} else {
+		glog.Errorf("failed to create service monitor resources because namespace %s does not exist", monitorNamespace)
```
SchSeba: This should be an error?

if we are running on k8s we don't have the MonitorService right?

cynepco3hahue: Nope we only want to publish the error without interrupting the operator flow

rmohr: Hm, better `Warningf`.

PR #2701, opened 2019-09-16
add masquerade binding method to templates

alonSadan:
In this PR we added masquerade binding mehtod
to RHEL7 and Fedora templates, so that those
template VMs will be migratable.
(if no binding method is specified we use bridge
as a default)

**Release note**:

```release-note
Fedora and RHEL7 template VMs were changed to use masquerade binding method
```

### DIFF
--- a/tools/vms-generator/utils/utils.go
+++ b/tools/vms-generator/utils/utils.go
@@ -84,6 +84,7 @@ const (
 )
 
 const windowsFirmware = "5d307ca9-b3ef-428c-8861-06e72d69f223"
+const defaultInterfaceName = "default"
 
 var DockerPrefix = "registry:5000/kubevirt"
 var DockerTag = "devel"
@@ -122,6 +123,22 @@ func getBaseVMI(name string) *v1.VirtualMachineInstance {
 func initFedora(spec *v1.VirtualMachineInstanceSpec) *v1.VirtualMachineInstanceSpec {
 	addContainerDisk(spec, fmt.Sprintf("%s/%s:%s", DockerPrefix, imageFedora, DockerTag), busVirtio)
 	addRNG(spec) // without RNG, newer fedora images may hang waiting for entropy sources
+
+	return spec
+}
+
+func setDefaultNetworkAndInterface(spec *v1.VirtualMachineInstanceSpec, bindingMethod v1.InterfaceBindingMethod, networkSource v1.NetworkSource) *v1.VirtualMachineInstanceSpec {
+	spec.Domain.Devices.Interfaces = []v1.Interface{
+		v1.Interface{
+			Name:                   defaultInterfaceName,
+			InterfaceBindingMethod: bindingMethod},
+	}
+	spec.Networks = []v1.Network{
+		v1.Network{
+			Name:          defaultInterfaceName,
+			NetworkSource: networkSource},
+	}
+
 	return spec
 }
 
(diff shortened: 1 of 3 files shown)


tools/vms-generator/utils/utils.go
```diff
@@ -122,6 +123,27 @@ func getBaseVMI(name string) *v1.VirtualMachineInstance {
 func initFedora(spec *v1.VirtualMachineInstanceSpec) *v1.VirtualMachineInstanceSpec {
 	addContainerDisk(spec, fmt.Sprintf("%s/%s:%s", DockerPrefix, imageFedora, DockerTag), busVirtio)
 	addRNG(spec) // without RNG, newer fedora images may hang waiting for entropy sources
+
+	return spec
+}
+
+func addINterface(spec *v1.VirtualMachineInstanceSpec, bindingMethod v1.InterfaceBindingMethod) *v1.VirtualMachineInstanceSpec {
```
booxter: why both I and N upper case in the name of the function?

booxter: Also, the name of the function doesn't fully reflect its intent: it is named like a function that allows to add any interface, but in reality it's only fitting for "default" interfaces. Please rename accordingly, e.g.: `addDefaultInterface`.

booxter: Also, the function doesn't *add* but *override* the list of interfaces. Please either rename into: `setDefaultInterface` or change the body to `append` to an existing slice (if it is not `nil` of course).

tools/vms-generator/utils/utils.go
```diff
@@ -122,6 +123,27 @@ func getBaseVMI(name string) *v1.VirtualMachineInstance {
...
+	spec.Domain.Devices.Interfaces = []v1.Interface{
+		v1.Interface{
+			Name:                   defaultName,
+			InterfaceBindingMethod: bindingMethod},
+	}
+
+	return spec
+}
+
+func addNetwork(spec *v1.VirtualMachineInstanceSpec, networkSource v1.NetworkSource) *v1.VirtualMachineInstanceSpec {
```
booxter: Same comments from above applied to this function.

tools/vms-generator/utils/utils.go
```diff
@@ -559,8 +580,16 @@ func GetVMCirros() *v1.VirtualMachine {
 
 func GetTemplateFedora() *Template {
 	vm := getBaseVM("", map[string]string{"kubevirt-vm": "vm-${NAME}", "kubevirt.io/os": "fedora27"})
-	initFedora(&vm.Spec.Template.Spec)
-	addNoCloudDiskWitUserData(&vm.Spec.Template.Spec, "#cloud-config\npassword: fedora\nchpasswd: { expire: False }")
+	spec := &vm.Spec.Template.Spec
+	initFedora(spec)
+	addNoCloudDiskWitUserData(spec, "#cloud-config\npassword: fedora\nchpasswd: { expire: False }")
+
+	addINterface(spec, v1.InterfaceBindingMethod{
```
booxter: is there ever a case where we add one but not the other? perhaps combine them into a single function (`setDefaultNetworking` or `setDefaultNetworkAndInterface` or perhaps `appendDefaultNetworkAndInterface`).

tools/vms-generator/utils/utils.go
```diff
@@ -84,6 +84,7 @@ const (
 )
 
 const windowsFirmware = "5d307ca9-b3ef-428c-8861-06e72d69f223"
+const defaultName = "default"
```
booxter: default name of what? please rename into `defaultInterfaceName` or `defaultNetworkName`.

tools/vms-generator/utils/utils.go
```diff
@@ -122,6 +123,22 @@ func getBaseVMI(name string) *v1.VirtualMachineInstance {
...
 	addContainerDisk(spec, fmt.Sprintf("%s/%s:%s", DockerPrefix, imageFedora, DockerTag), busVirtio)
 	addRNG(spec) // without RNG, newer fedora images may hang waiting for entropy sources
+
+	return spec
+}
+
+func setDefaultNetworkAndInterface(spec *v1.VirtualMachineInstanceSpec, bindingMethod v1.InterfaceBindingMethod, networkSource v1.NetworkSource) *v1.VirtualMachineInstanceSpec {
+	spec.Domain.Devices.Interfaces = []v1.Interface{
+		v1.Interface{
+			Name:                   defaultName,
```
booxter: I believe you should have renamed it here too...

alonSadan: oops. fixed

PR #2702, opened 2019-09-16
Check how prow dco plugin looks

dhiller:

This is just a test for the Prow DCO plugin. I added an unsigned commit to see how it looks.

/hold
/cc @rmohr

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,3 +121,5 @@ KubeVirt is distributed under the
    [cockpit]: https://cockpit-project.org/
    [libvirt]: https://www.libvirt.org
    [kubevirt-ansible]: https://github.com/kubevirt/kubevirt-ansible
+
+


dhiller:
Looking good so far, the only thing I need to know is how to make the check mandatory.

dhiller:
Maybe this is the way to do it: kubevirt/project-infra#192

rmohr:
> Looking good so far, the only thing I need to know is how to make the check mandatory.

I think that for tide it is mandatory anyway, but for normal merges, I don't think so.

dhiller:
Ah, ok, then we don't need to make it mandatory like in the PR I just posted, right? Need to brush up reading about Prow config then :)

dhiller:
But shouldn't they also be marked as required then? I manually did that in branch protection settings to make it look like it is now.

rmohr:
> But shouldn't they also be marked as required then? I manually did that in branch protection settings to make it look like it is now.

Afaik, tide only looks on its internal configuration. I think it even tries to merge PRs if a required job fails and it is not listed in the branch protection config.

rmohr:
> Afaik, tide only looks on its internal configuration. I think it even tries to merge PRs if a required job fails and it is not listed in the branch protection config.

But: there exists the branchprotector https://github.com/kubernetes/test-infra/tree/master/prow/cmd/branchprotector (which we did not deploy), which can configure github the way prow sees the configuration.

dhiller:
> > But shouldn't they also be marked as required then? I manually did that in branch protection settings to make it look like it is now.
> 
> Afaik, tide only looks on its internal configuration. I think it even tries to merge PRs if a required job fails and it is not listed in the branch protection config.

I found [this configuration of tide](https://github.com/openshift/release/blob/master/core-services/prow/02_config/_config.yaml#L644), might be the right one?

rmohr:
> I found [this configuration of tide](https://github.com/openshift/release/blob/master/core-services/prow/02_config/_config.yaml#L644), might be the right one?

Great, that looks like it!

rmohr:
/close

Everything works like expected. @dhiller thx!

PR #2713, opened 2019-09-18
Align with upstream network-attachment-definition-client

booxter:
**What this PR does / why we need it**: this PR switches the tree to upcoming upstream client after it exposes Tracker endpoint in its fake client used in kubevirt unit tests.

Fixes #

**Release note**:

```release-note
NONE
```


booxter:
@phoracek @SchSeba please take a look, it's mostly mechanistic updates; I tried to split commits logically to simplify review.

booxter:
@SchSeba I don't think we have an easy way to update one but not another dependency since our `make deps-update` target is global to all dependencies. I think unless there is a strong reason to pin dependencies, it's ok to bump them. (If there is a reason to pin, there should be a `replace` instruction in `go.mod` to make it explicit and enforced by `go mod`.)

PR #2715, opened 2019-09-19
vmi_configuration_test 3 cores - add check for cpu cores, skip if less than expected

dhiller:

The test series "with 3 cores" within `tests/vmi_configuration_test.go` expects to have three cores but never checks it. This is done now in the `BeforeEach`.

Fixes #

I couldn't test this as I don't have free resources locally. Please advise whether this approach could work and how to check for negative and positive.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -140,11 +140,17 @@ var _ = Describe("Configurations", func() {
 
 	Describe("[rfe_id:140][crit:medium][vendor:cnv-qe@redhat.com][level:component]VirtualMachineInstance definition", func() {
 		Context("with 3 CPU cores", func() {
+			availableNumberOfCPUs := tests.GetHighestCPUNumberAmongNodes(virtClient)
+
 			var vmi *v1.VirtualMachineInstance
 
 			BeforeEach(func() {
+				requiredNumberOfCpus := 3
+				Expect(availableNumberOfCPUs).ToNot(BeNumerically("<", requiredNumberOfCpus),
+					fmt.Sprintf("Test requires %d cpus, but only %d available!", requiredNumberOfCpus, availableNumberOfCPUs))
 				vmi = tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskAlpine))
 			})
+
 			It("[test_id:1659]should report 3 cpu cores under guest OS", func() {
 				vmi.Spec.Domain.CPU = &v1.CPU{
 					Cores: 3,


dhiller:
> I am not sure it's ideal to silently skip a test case if a node doesn't have enough CPUs. This will e.g. disable the test case if a CI machine somehow gets less cores available during the test run.
> 
> Which specific problem are you trying to solve with the PR? Is it that you cannot successfully execute the test suite without enough cores in the cluster? Would moving the test cases with specific hardware requirement into a separate suite that you could skip with standard `ginkgo` CLI options be enough to cover your use case? (We would still have the new suite enabled for all relevant CI jobs not to miss a regression.)

I was trying to mitigate test failures on test clusters where there are not enough cpus on any node available. I was told that if the test cluster has no nodes that can provide this number, the VMs would time out in scheduling, causing a test failure. So the most sane thing would be to skip the test as there could not have been done anything sensible otherwise.

However after talking to @mfranczy and @stu-gott I'll rebase my branch on his and use a utility function from that branch as we decided to use Capacity instead.

booxter:
@dhiller there is no doubt that blindly scheduling a 3-core VMI on a cluster that potentially doesn't have them is wrong and it will take a while for the suite to realize that the VMI will never get up and fail. The only question is whether we should skip or fail when allocatable resources cannot satisfy the prerequisites of the test case. I argue that we should still fail, though a lot quicker, to flag something is wrong; and for cluster that we know in advance don't have enough cores, we would provide a way to disable the test case we know won't pass (hence `ginkgo` suggestion).

Just my cents, if others feel different, please proceed as you do.

rmohr:
> we would provide a way to disable the test case we know won't pass (hence `ginkgo` suggestion).

At least that can be achieved via `--skip`. Like e.g. GPU tests are disabled.

booxter:
> > we would provide a way to disable the test case we know won't pass (hence `ginkgo` suggestion).
> 
> At least that can be achieved via `--skip`. Like e.g. GPU tests are disabled.

Yes. But this implies the test case is in a separate suite so that it's easy to skip it but running other test cases nevertheless. Right now the approach in this PR is to skip the test case when the number of cores available to the test case is less than three, regardless of whether we should have expected cores available to it, or the cluster that tests are executed against is not ever capable to serve the requirements of the case.

dhiller:
> > At least that can be achieved via `--skip`. Like e.g. GPU tests are disabled.
> 
> Yes. But this implies the test case is in a separate suite so that it's easy to skip it but running other test cases nevertheless. Right now the approach in this PR is to skip the test case when the number of cores available to the test case is less than three, regardless of whether we should have expected cores available to it, or the cluster that tests are executed against is not ever capable to serve the requirements of the case.

@booxter @rmohr actually grouping the tests together into suites that somehow define what requirements are there to run them sounds very reasonable. But this also sounds like a bigger task as someone needs to gather these requirements, so we might split it into a separate assignment where we first gather the requirements, then group the tests accordingly and put them into suites. For now we might leave it as is. Within said task we should also review test skipping and adjust accordingly so that we only need to select/skip test suites to run. Makes sense?

ahadas:
> However after talking to @mfranczy and @stu-gott I'll rebase my branch on his and use a utility function from that branch as we decided to use Capacity instead.

Shouldn't we want to check `Allocatable` rather than `Capacity`?

dhiller:
> > However after talking to @mfranczy and @stu-gott I'll rebase my branch on his and use a utility function from that branch as we decided to use Capacity instead.
> 
> Shouldn't we want to check `Allocatable` rather than `Capacity`?

@ahadas thought so too at first. We discussed this at another place and decided on using capacity to leave this up to the scheduler.

mfranczy:
before creating a backport PR we have to merge this one https://github.com/kubevirt/kubevirt/pull/2733

PR #2716, opened 2019-09-19
Multus Tests: Get the image to use for the network config daemon from the virt operator pod

fedepaol:

Right now the network setup is done via a daemon that uses the same image as the virt operator (hardcoded from the local registry). 
This won't work in case of external providers, where the local registry may not exist.

In this PR we fetch the image to use directly from the running pod.

Fixes #

**Release note**:

```release-note

NONE

```

### DIFF
--- a/tests/vmi_multus_test.go
+++ b/tests/vmi_multus_test.go
@@ -820,6 +820,7 @@ func pingVirtualMachine(vmi *v1.VirtualMachineInstance, ipAddr, prompt string) {
 // TODO: Once kubernetes-nmstate is ready, we should use it instead
 func configureNodeNetwork(virtClient kubecli.KubevirtClient) {
 
+	// Fetching the kubevirt-operator image from the pod makes this independent from the installation method / image used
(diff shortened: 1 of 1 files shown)


tests/vmi_multus_test.go
```diff
@@ -819,6 +819,13 @@ func pingVirtualMachine(vmi *v1.VirtualMachineInstance, ipAddr, prompt string) {
 // both iptables and newer nftables.
 // TODO: Once kubernetes-nmstate is ready, we should use it instead
 func configureNodeNetwork(virtClient kubecli.KubevirtClient) {
+
+	pods, err := virtClient.CoreV1().Pods(tests.KubeVirtInstallNamespace).List(metav1.ListOptions{LabelSelector: "kubevirt.io=virt-operator"})
```
phoracek: Please add a comment explaining why we do this.

booxter:
/retest

Failed tests are not the ones affected by this change.

PR #2742, opened 2019-09-25
[fix] input device without bus

ksimon1:
When input device bus is not defined, default value is usb. But enabling of usb controller is done before checking input devices, so usb controller is never enabled. This PR fixes it.

Signed-off-by: ksimon1 <ksimon@redhat.com>

**Release note**:

```release-note
 fix issue, when input device doesn't have bus specified and usb controller is not enabled.
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/api/converter.go
+++ b/pkg/virt-launcher/virtwrap/api/converter.go
@@ -517,19 +517,6 @@ func convertFeatureState(source *v1.FeatureState) *FeatureState {
 	return nil
 }
 
-//isUSBDevicePresent checks if exists device with usb bus in vmi
-func isUSBDevicePresent(vmi *v1.VirtualMachineInstance) bool {
-	usbDeviceExists := false
-	for _, input := range vmi.Spec.Domain.Devices.Inputs {
-		if input.Bus == "usb" {
-			usbDeviceExists = true
-			return usbDeviceExists
-		}
-	}
-
-	return usbDeviceExists
-}
-
 func Convert_v1_Features_To_api_Features(source *v1.Features, features *Features, c *ConverterContext) error {
 	if source.ACPI.Enabled == nil || *source.ACPI.Enabled {
 		features.ACPI = &FeatureEnabled{}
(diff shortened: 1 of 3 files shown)


mfranczy:
@ksimon1 unit tests fail https://travis-ci.org/kubevirt/kubevirt/builds/589418919#L1629

pkg/virt-launcher/virtwrap/api/converter_test.go
```diff
@@ -797,6 +797,20 @@ var _ = Describe("Converter", func() {
...
 		})
 
+		It("should not disable usb controller when device with no bus is present", func() {
+			v1.SetObjectDefaults_VirtualMachineInstance(vmi)
+			vmi.Spec.Domain.Devices.Inputs[0].Bus = ""
+			domain := vmiToDomain(vmi, c)
+			disabled := false
+			for _, controller := range domain.Spec.Devices.Controllers {
+				if controller.Type == "usb" && controller.Model == "none" {
+					disabled = !disabled
```
mfranczy: could you just change it to `disabled = true` is more readable

tests/vmi_configuration_test.go
```diff
@@ -579,6 +579,32 @@ var _ = Describe("Configurations", func() {
 				Expect(err).ToNot(HaveOccurred(), "should report number of usb")
 			})
 
+			It("should start the VMI with usb controller when input device doesn't have bus", func() {
+				vmi := tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskAlpine))
+				vmi.Spec.Domain.Devices.Inputs = []v1.Input{
+					{
+						Name: "tablet0",
+						Type: "tablet",
```
mfranczy: should not conversion fail before the start because of not supported `tablet` type [link](https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/api/converter.go#L455)?

mfranczy: ah my bad, it's correct. Ignore the comment above.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -913,10 +903,24 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 			if err != nil {
```
mfranczy: could you move the error check right after line [901](https://github.com/kubevirt/kubevirt/pull/2742/files#diff-caee6d384498fde52fde15891ac547aaR901)?

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -893,30 +893,34 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
...
-	if usbDeviceExists := isUSBDevicePresent(vmi); !usbDeviceExists {
-		// disable usb controller
-		domain.Spec.Devices.Controllers = append(domain.Spec.Devices.Controllers, Controller{
-			Type:  "usb",
-			Index: "0",
-			Model: "none",
-		})
-	}
-
+	isUSBDevicePresent := false
```
vladikr: hmm, there is a function with the same name on line 521
What am I missing here?

vladikr: ah, @ksimon1 can you please remove the function 
`func isUSBDevicePresent(vmi *v1.VirtualMachineInstance) bool {` on line 521
as @mfranczy suggested?

ksimon1:
@mfranczy, @vladikr i don't think these failing test have something in common with this update

PR #2750, opened 2019-09-26
Make the reporter number of failures to dump log for configurable.

fedepaol:

Currently the kubernetes test reporter dumps the information related to the first 10 failing tests. 

With this PR we make that number configurable, so we can investigate a higher number of failures, especially for new lanes where some misconfiguration can cause different errors in different areas we want to tackle concurrently.

Fixes #

**Release note**:

```release-note

NONE

```

### DIFF
--- a/cmd/dump/dump.go
+++ b/cmd/dump/dump.go
@@ -4,7 +4,6 @@ import (
 	"os"
 	"time"
 
-	"github.com/onsi/ginkgo/types"
 	"github.com/spf13/pflag"
 
 	"kubevirt.io/client-go/kubecli"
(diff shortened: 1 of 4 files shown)


tests/tests_suite_test.go
```diff
@@ -32,7 +34,8 @@ import (
 )
 
 func TestTests(t *testing.T) {
-	reporters := []Reporter{reporter.NewKubernetesReporter(os.Getenv("ARTIFACTS"))}
+	maxFails := getMaxFailsFromEnv()
```
rmohr: Can you move this part before you define the flag and use the outcome as the "default" value? Let's follow the usual order:

 1) no env, no arg -> use the default
 2) env, no arg  -> use the env
 3) env, arg -> use the arg (so explicit arg beats env)

fedepaol: Well the thing is, the reporter in the tests is consuming only the env variable while flag is only in the dump executable (which in turn do not read the env variable).

I did that because it felt more natural to be a parameter in the command without having to duplicate all the env parsing logic. If you think that this is confusing and both the tests / the dump command should have the flag AND the env variable I'll change them.

rmohr: I would say, if we have both, keep the expected order. If we only have envs, then I am fine :)

fedepaol: Yes that's the case.
When running the test executable we only have env, when running the dump command we only have the flag.

rmohr: Perfect. Thanks for pointing me to it.

cmd/dump/dump.go
```diff
@@ -16,9 +16,10 @@ func main() {
 	var duration time.Duration
 	pflag.CommandLine.AddGoFlagSet(kubecli.FlagSet())
 	pflag.DurationVarP(&duration, "since", "s", 10*time.Minute, "collection window, defaults to 10 minutes")
+	maxFails := pflag.Int("maxfails", 10, "max failed tests to report, defaults to 10")
```
rmohr: Ups, just realized that this flag does not make any sense at all. The `dump` binary just reused the per-test dump code, but its purpose is to just dump once and then exit. That can be removed and a hardcoded `1` should be good enough.

fedepaol: Yes I wasn't sure what to put there. 
Will change it and hardcode it to 1 with an explanatory comment.

fedepaol: Done, I also did a bit of cleanup so the invocations from dump are more explicit.

rmohr:
Hm, needs a `make generate && make`.

PR #2768, opened 2019-10-01
Set on virt-handler detector install namespace on tests

rmohr:

Tests can't automatically detect and use the install namespace to fetch
virt-handlers via the handler detector. In contrast to our other
components which run in a pod with a service account, the tests need to
explicitly pass the install namespace to the handler detector.

Fixes #

Fixes test runs if kubevirt is not installed in the "kubevirt" namespace.

**Release note**:

```release-note
NONE
```


slintes:
great catch!

/approve

PR #2769, opened 2019-10-01
Introduce the shasum switch for the tests, needed when testing on OKD 4+

petrkotas:

The tests requiering shasums are failing
on crio due to crio not supporting shasums.

This patch introduces the switch to let
the tests know that it shoould not test
shasums on crio.

The switch is used when running the tests
on OKD 4+ cluster using cri-o.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -40,6 +40,7 @@ import (
 
 	v1 "kubevirt.io/client-go/api/v1"
 	"kubevirt.io/client-go/kubecli"
+	"kubevirt.io/client-go/log"
 	cdiv1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
 	"kubevirt.io/kubevirt/pkg/controller"
 	"kubevirt.io/kubevirt/pkg/virt-operator/creation/components"
@@ -395,6 +396,11 @@ var _ = Describe("Operator", func() {
 	}
 
 	ensureShasums := func() {
+		if tests.SkipShasumCheck {
+			log.Log.Warning("Cannot use shasums, skipping")
+			return
+		}
+
 		for _, name := range []string{"virt-operator", "virt-api", "virt-controller"} {
 			deployment, err := virtClient.AppsV1().Deployments(tests.KubeVirtInstallNamespace).Get(name, metav1.GetOptions{})
 			Expect(err).ToNot(HaveOccurred())
@@ -613,6 +619,10 @@ spec:
 
 	Describe("should start a VM", func() {
 		It("using virt-launcher with a shasum", func() {
+
+			if tests.SkipShasumCheck {
+				Skip("Cannot currently test shasums, skipping")
+			}
 
 			By("starting a VM")
 			vmi := tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskCirros))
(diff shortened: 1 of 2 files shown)


petrkotas:
Hi @slintes I have considered putting it in the config. I choose the parameter option, since I am unsure how the downstream tests are passing configuration. The parameter is easier to set up from jenkins.

slintes:
> Hi @slintes I have considered putting it in the config. I choose the parameter option, since I am unsure how the downstream tests are passing configuration. The parameter is easier to set up from jenkins.

can you please verify with them what they prefer? Thanks

petrkotas:
@slintes I have sent an email, but we have to wait until the 20th of October. Vacation time :)

slintes:
Probably parameter is fine, wasn't aware that there so many already, so it doesn't matter. But please rename as mentioned earlier.

petrkotas:
@slintes I have updated the code. It is more readable now. Thanks for the pointer!

cynepco3hahue:
Hm, I do not see where do you send this parameter? In general, I expected that you will check the container runtime and if it runs CRI-O, you will skip tests. Also, I did not see such problems under the OKD4.1, so I unsure if we need to introduce this flag at all.

petrkotas:
Hi @cynepco3hahue thanks for the feedback. The patch is responding to the bug: https://bugzilla.redhat.com/show_bug.cgi?id=1716432 The parameter will be used on the downstream test suite. The downstream testsuite runs crio and does not have the shasums capability.

cynepco3hahue:
@petrkotas Thanks for the clarification.

petrkotas:
Yes the issue is exactly the use of tags. The tests are insisting on images selected via SHA. The downstream, however, uses only tags. Therefore the test fails because there are no shasums at the end of the image.

This fix provides and option to skip the tests insisting on shasums for deployments where tags are used instead of shasums to select the images.

tests/operator_test.go
```diff
@@ -395,6 +396,12 @@ var _ = Describe("Operator", func() {
 	}
 
 	ensureShasums := func() {
+		if tests.SkipShasumCheck {
+			// Currently shasums are not supported on crio.
+			log.Log.Warning("Cannot use shasums when using CRI-o")
```
slintes: this still mentions cri-o. We just want to skip shasums test, and don't care why imho

tests/operator_test.go
```diff
@@ -614,6 +621,10 @@ spec:
 	Describe("should start a VM", func() {
 		It("using virt-launcher with a shasum", func() {
 
+			if tests.SkipShasumCheck {
+				Skip("Cannot currently test shasums for crio")
```
slintes: same

tests/utils.go
```diff
@@ -120,6 +121,7 @@ func init() {
 	flag.StringVar(&PreviousReleaseTag, "previous-release-tag", "", "Set tag of the release to test updating from")
 	flag.StringVar(&PreviousReleaseRegistry, "previous-release-registry", "", "Set registry of the release to test updating from")
 	flag.StringVar(&ConfigFile, "config", "tests/default-config.json", "Path to a JSON formatted file from which the test suite will load its configuration. The path may be absolute or relative; relative paths start at the current working directory.")
+	flag.BoolVar(&SkipShasumCheck, "skip-shasums-check", false, "Is crio used for deployment of the CNV.")
```
slintes: same

petrkotas:
Hi @slintes thanks for the checkup. I did autorename and forgot about the titles and logs.

slintes:
> the reason for this issue not cri-o, it just uses image tags instead of sha hashes

@cynepco3hahue and the reason for using tags is cri-o ;)
https://github.com/cri-o/cri-o/issues/2157

/lgtm
/approve
/cherrypick release-0.20

PR #2779, opened 2019-10-03
Always have compute as the first container in the virt-launcher pod.

fedepaol:
When we have container disk volumes, those containers are put at the beginning of the container list of the launcher pod, before the compute container.
This may be a problem when using the sriov mutating webhook, which parses the requested network attachment definitions of the pod and modifies the pod adding the requested resources to the first container. If the first container is not the compute one, we are requesting resources for the wrong container (see https://github.com/openshift/ose-sriov-dp-admission-controller/blob/e65f11b5eb7da1918b439a60d7fcc814719b1433/pkg/webhook/webhook.go#L325).

In this PR we ensure the compute container is always the first one of the list.

**Release note**:

```release-note

Change the virt-launcher pod to always have the computer container as the first of the pod.

```

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -736,8 +736,8 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 		compute.Env = append(compute.Env, k8sv1.EnvVar{Name: ENV_VAR_LIBVIRT_DEBUG_LOGS, Value: "1"})
 	}
 
-	// Make sure the compute container is always the first since the cni mutating webhook will add the
-	// requested resources to the first container of the pod
+	// Make sure the compute container is always the first since the mutating webhook shipped with the sriov operator
+	// for adding the requested resources to the pod will add them to the first container of the list
 	containers := []k8sv1.Container{compute}
 	containersDisks := containerdisk.GenerateContainers(vmi, "container-disks", "virt-bin-share-dir")
 	containers = append(containers, containersDisks...)
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -678,7 +678,7 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 
 	volumes = append(volumes, k8sv1.Volume{Name: "infra-ready-mount", VolumeSource: k8sv1.VolumeSource{EmptyDir: &k8sv1.EmptyDirVolumeSource{}}})
 
-	containers := containerdisk.GenerateContainers(vmi, "container-disks", "virt-bin-share-dir")
+	containersDisks := containerdisk.GenerateContainers(vmi, "container-disks", "virt-bin-share-dir")
 
 	networkToResourceMap, err := getNetworkToResourceMap(t.virtClient, vmi)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


cynepco3hahue:
> This may be a problem when using the sriov mutating webhook, which parses the requested network attachment definitions of the pod and modifies the pod adding the requested resources to the first container

Why does it add to the first container, can it check the name of the container? It just too magic, to expect that the container that we need is the first one.

fedepaol:
> > This may be a problem when using the sriov mutating webhook, which parses the requested network attachment definitions of the pod and modifies the pod adding the requested resources to the first container
> 
> Why does it add to the first container, can it check the name of the container? It just too magic, to expect that the container that we need is the first one.

I do agree, the problem here is that the network attachment definition is bound to the pod whereas the device is consumed by a container (and the attachment definition does not contain the name of the container it refers to).
It something we don't have control on since it's going to be shipped as part of the sr-iov operator, so right now we are trying to prevent side effects while trying to understand if we can influence in some way the direction of that admission webhook.

cynepco3hahue:
I see, I was sure that SR-IOV operator something that we also working on.

fedepaol:
> I see, I was sure that SR-IOV operator something that we also working on.

My bad, I should have referred to it as the openshift sr-iov operator. We rely on it for a lot of sr-iov related configuration, my understanding is that it's going to be enabled in openshift 4.3 and as part of it they are going to enable the mutating webhook.

pkg/virt-controller/services/template.go
```diff
@@ -720,25 +718,29 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
+		compute.Env = append(compute.Env, k8sv1.EnvVar{Name: varName, Value: resourceName})
 	}
 
 	if _, ok := vmi.Labels[debugLogs]; ok {
-		container.Env = append(container.Env, k8sv1.EnvVar{Name: ENV_VAR_LIBVIRT_DEBUG_LOGS, Value: "1"})
+		compute.Env = append(compute.Env, k8sv1.EnvVar{Name: ENV_VAR_LIBVIRT_DEBUG_LOGS, Value: "1"})
 	}
 
-	containers = append(containers, container)
+	// Make sure the compute container is always the first since the cni mutating webhook will add the
```
booxter: "which cni mutating webhook?" is a legit question, considering that the hook you refer to is not even under kubevirt organization. this should be explained in the comment with more details.

booxter:
AFAIU there is no reason to bump versions of dependencies in this PR. Are there?

fedepaol:
> AFAIU there is no reason to bump versions of dependencies in this PR. Are there?

Nope & thanks for pointing out. It must have been vscode. I normally skip go.mod / go.sum but they probably sneaked in this time as a result of a git commit --all. Will remove them

fedepaol:
/hold

Changed the timeout to see what happens to sriov lane.

fedepaol:
/test pull-kubevirt-e2e-kind-k8s-sriov-1.14.2

let's see if it was a coincidence

fedepaol:
/test pull-kubevirt-e2e-kind-k8s-sriov-1.14.2
one more time

fedepaol:
Just realized this was on hold
/hold cancel

rmohr:
@fedepaol it still contains the timeout change, right?

fedepaol:
> @fedepaol it still contains the timeout change, right?

ops, you are right!

fedepaol:
@rmohr removed, thanks!

rmohr:
@fedepaol want to add a release note?

fedepaol:
> @fedepaol want to add a release note?

@rmohr done!

PR #2780, opened 2019-10-03
Add Prometheus alert rules for virt-api

danielBelenky:
Adding basic pre-defined Prometheus alert rules with KubeVirt deployment will allow cluster admins to monitor their KubeVirt deployment using OpenShift's cluster monitoring solution.

This PR defines alert rules only for virt-api and only for the "up" condition, which means that the api-server is up (and in this case means it's also ready).

**Release note**:
Introducing basic Prometheus alert rules for virt-api.
```release-note
Add Prometheus alert rules for virt-api.
```

### DIFF
--- a/pkg/virt-operator/creation/components/crds.go
+++ b/pkg/virt-operator/creation/components/crds.go
@@ -19,13 +19,20 @@
 package components
 
 import (
+	"fmt"
+
 	"github.com/coreos/prometheus-operator/pkg/apis/monitoring"
 	promv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	corev1 "k8s.io/api/core/v1"
 	extv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	virtv1 "kubevirt.io/client-go/api/v1"
+)
+
+const (
+	KUBEVIRT_PROMETHEUS_RULE_NAME = "prometheus-kubevirt-rules"
 )
 
 func newBlankCrd() *extv1beta1.CustomResourceDefinition {
@@ -243,7 +250,7 @@ func NewServiceMonitorCR(namespace string, monitorNamespace string, insecureSkip
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: monitorNamespace,
-			Name:      "kubevirt",
+			Name:      KUBEVIRT_PROMETHEUS_RULE_NAME,
 			Labels: map[string]string{
 				"openshift.io/cluster-monitoring": "",
 				"prometheus.kubevirt.io":          "",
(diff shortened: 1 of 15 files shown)


pkg/virt-operator/creation/components/crds.go
```diff
@@ -270,6 +271,65 @@ func NewServiceMonitorCR(namespace string, monitorNamespace string, insecureSkip
...
+							Annotations: map[string]string{
+								"summary": "All VirtAPI servers are down.",
+							},
+						},
+						{
+							Alert: "SingleVirtAPI",
+							Expr:  intstr.FromString("num_of_running_virt_api_servers == 1"),
+							For:   "60m",
+							Annotations: map[string]string{
+								"summary": "A single VirtAPI server is up for the last hour.",
```
rmohr: Maybe better: `Only one virt-api server is up for the last hour`.

pkg/virt-operator/creation/components/crds.go
```diff
@@ -270,6 +271,65 @@ func NewServiceMonitorCR(namespace string, monitorNamespace string, insecureSkip
...
+						{
+							Alert: "VirtAPIDown",
+							Expr:  intstr.FromString("num_of_running_virt_api_servers == 0"),
+							For:   "5m",
+							Annotations: map[string]string{
+								"summary": "All VirtAPI servers are down.",
+							},
+						},
+						{
+							Alert: "SingleVirtAPI",
```
rmohr: Should this be a `Record`?

danielBelenky: We're recording `num_of_running_virt_api_servers`... I don't think that we need to record the alert itself...
Or you meant to something else?

pkg/virt-operator/install-strategy/delete.go
```diff
@@ -217,6 +218,26 @@ func DeleteAll(kv *v1.KubeVirt,
...
 	}
 
+	// delete PrometheusRules
+	objects = stores.PrometheusRuleCache.List()
+	for _, obj := range objects {
+		if prometheusRule, ok := obj.(*promv1.PrometheusRule); ok && prometheusRule.DeletionTimestamp == nil {
+			if key, err := controller.KeyFunc(prometheusRule); err == nil {
+				expectations.PrometheusRule.AddExpectedDeletion(kvkey, key)
+				err := prometheusClient.MonitoringV1().PrometheusRules(prometheusRule.Namespace).Delete(prometheusRule.Name, deleteOptions)
+				if false && err != nil {
```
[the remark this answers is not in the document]
danielBelenky: That's a trailing error from the block that deals with `serviceMonitor` objects.
I'll fix the `serviceMonitor` block as well in a different PR

pkg/virt-operator/creation/components/crds.go
```diff
@@ -270,6 +271,65 @@ func NewServiceMonitorCR(namespace string, monitorNamespace string, insecureSkip
...
+			},
+		},
+		Spec: promv1.PrometheusRuleSpec{
+			Groups: []promv1.RuleGroup{
+				{
+					Name: "kubevirt.rules",
+					Rules: []promv1.Rule{
+						{
+							Record: "num_of_running_virt_api_servers",
+							Expr:   intstr.FromString("sum(up{pod=~'virt-api-.+'})"),
```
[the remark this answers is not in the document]
danielBelenky: Good point. Thanks.

danielBelenky:
> Other than that, the only question I have is whether or not we should make enabling these default kubevirt related prometheus alerts something that admins can toggle on/off.

Admins can silence alerts. IMO, there shouldn't be a configuration for `kubevirt` deployment w/o the alerts as they supposed to be the indication for a proper deployment over time.

rmohr:
@danielBelenky the PR looks now pretty good to me. Only a functional test is missing. Something like checking if the expected rules are created.

pkg/virt-operator/creation/components/crds.go
```diff
@@ -272,6 +279,72 @@ func NewServiceMonitorCR(namespace string, monitorNamespace string, insecureSkip
...
+					{
+						Record: "num_of_allocatable_nodes",
+						Expr:   intstr.FromString("count(count (kube_node_status_allocatable) by (node))"),
+					},
+					{
+						Alert: "LowVirtVPICount",
+						Expr:  intstr.FromString("(num_of_allocatable_nodes > 1) and (num_of_running_virt_api_servers < 2)"),
+						For:   "60m",
+						Annotations: map[string]string{
+							"summary": "More than one VirtAPI should be running if more than one worker node.",
```
rmohr: More than one VirtAPI should be running if more than one worker node[s exist].

pkg/virt-operator/creation/components/crds.go
```diff
@@ -272,6 +279,72 @@ func NewServiceMonitorCR(namespace string, monitorNamespace string, insecureSkip
...
+					{
+						Alert: "VirtAPIDown",
+						Expr:  intstr.FromString("num_of_running_virt_api_servers == 0"),
+						For:   "5m",
+						Annotations: map[string]string{
+							"summary": "All VirtAPI servers are down.",
+						},
+					},
+					{
+						Alert: "SingleVirtAPI",
```
rmohr: Thinking about it `SingleVirtAPI` looks a little bit redundant compared to `LowVirtAPICount`. What do you think?

pkg/virt-operator/creation/components/crds.go
```diff
@@ -272,6 +279,72 @@ func NewServiceMonitorCR(namespace string, monitorNamespace string, insecureSkip
...
+						Annotations: map[string]string{
+							"summary": "Only one virt-api server is up for the last hour.",
+						},
+					},
+					{
+						Record: "num_of_allocatable_nodes",
+						Expr:   intstr.FromString("count(count (kube_node_status_allocatable) by (node))"),
+					},
+					{
+						Alert: "LowVirtVPICount",
```
rmohr: `LowVirtAPICount`.

fabiand:
ping?

pkg/virt-operator/install-strategy/create.go
```diff
@@ -1827,6 +1900,12 @@ func SyncAll(kv *v1.KubeVirt,
 		return false, err
 	}
 
+	// create/update PrometheusRules
```
rmohr: We will have to remove no longer existing objects after `// -------- CLEAN UP OLD UNUSED OBJECTS --------`. Otherwise we will leave old objects in place.

danielBelenky: Fixed.

PR #2786, opened 2019-10-07
Enable libvirt debug logs under our e2e tests

cynepco3hahue:

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -1558,6 +1558,12 @@ func NewRandomVMIWithNS(namespace string) *v1.VirtualMachineInstance {
 	t := defaultTestGracePeriod
 	vmi.Spec.TerminationGracePeriodSeconds = &t
 
+	// Enable by default debug logs
+	// TODO: part of tests can override it
+	vmi.Labels = map[string]string{
+		"debugLogs": "true",
+	}
+
 	// To avoid mac address issue in the tests change the pod interface binding to masquerade
 	// https://github.com/kubevirt/kubevirt/issues/1494
 	vmi.Spec.Domain.Devices = v1.Devices{Interfaces: []v1.Interface{{Name: "default",


tests/utils.go
```diff
@@ -1558,6 +1558,12 @@ func NewRandomVMIWithNS(namespace string) *v1.VirtualMachineInstance {
 	t := defaultTestGracePeriod
 	vmi.Spec.TerminationGracePeriodSeconds = &t
 
+	// Enable by default debug logs
+	// TODO: part of tests can override it
+	vmi.Labels = map[string]string{
```
petrkotas: Hi @cynepco3hahue why do you exactly need this?

cynepco3hahue:
Currently, we do not have any libvirt debug logs under the CI runs, it makes very hard to debug QEMU crashes(or libvirt audit failures).

petrkotas:
@cynepco3hahue thanks for thinking about ways to improve our test infrastructure. It indeed needs more love.

I agree that debugLogs from QEMU and Libvirt are useful. I used them for debugging the migration errors and without them, it would be very hard to figure the issue.

What I am afraid of is the amount of verbosity Libvirt produces when debugLogs are enabled. In general, when I was not looking for Libvirt relevant issue,  the logs were not helpful. Moreover, when enabled it made it very hard to find any information within the logs. I spent hours looking into the logs and finding the right error. Therefore I do not see the value of enabling it for every test by default.

To leverage the power of debugLogs, the test infrastructure has to be able to analyze errors automatically and provide the summary. Otherwise, the information will make the life of normal debugging way harder.

Do we have any infrastructure in place for such an automatic analysis or is there a plan for it?

Thanks again for figuring out how out CI can be better.

slintes:
> What I am afraid of is the amount of verbosity Libvirt produces when debugLogs are enabled

I'm afraid of this as well. I think temporarely enabling debug logs only for the VMIs of the tests where you need more logs is better than enabling them for all.

cynepco3hahue:
@petrkotas @slintes We can have a failure under each test because it just some flaky test, also you have logs only for failures and for a short period of time, so it should not be a problem with the disk space. Regarding the usefulness, I believe you still can find relevant errors, just by error word, but you will have more information about why the error happened.

petrkotas:
@cynepco3hahue thanks for taking the care. I am still afraid the logs will be cluttered. It is not about the disk space. The issue is with finding the right information.

I have consulted with my colleagues from oVirt. They agree the debugLogs are not useful in general as they have too much information which is usually not relevant.

We already have an option to enable the debugLogs where needed for each test.

In case of fully automatic log analytic, I would be more open to this change. Otherwise it will make our life harder when finding the errors.

PR #2770, opened 2019-10-01
Add okd-4.2

dhiller:
Signed-off-by: Daniel Hiller <daniel.hiller.1972@gmail.com>

**What this PR does / why we need it**:

Adds provider okd-4.2 for cluster-up. See kubevirt/kubevirtci#150

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:
```release-note
NONE
```

PR #2623, opened 2019-08-22
Stabilize the OKD 4.1 lane

slintes:
**What this PR does / why we need it**:

Fixes:
- before tests, wait until nodes will be tagged by virt-handler
- increase the total timeout for tests run
- use a newer OKD image, that includes local provisioner, master has more disk space and a worker has more disk space and more memory.
- try to pull images via ssh until it will succeed
- change `[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:component]VirtualMachine [rfe_id:273]with oc/kubectl as ordinary OCP user trough test service account should fail without right rights [It] [test_id:2914]should create VM via command line` error message
- skip one multus tests that not supported under the OKD
- create tests PV's always on the same schedulable node to avoid the situation when we create PV's on different nodes and trying to start VMI with both of them
- increase timeout for the operator test and virt-handler crash test

**Release note**:
```release-note
Stabilize the OKD 4.1 lane
```

PR #2790, opened 2019-10-08
Okd 4.2

dhiller:

This PR contains changes for `okd-4.2` test lane. This closes #2770.

Fixes #

**Release note**:

```release-note
Add okd-4.2
```


dhiller:
@slintes @cynepco3hahue et al, I've rebased my PR on the okd 4.1 stabilisation PR #2623. We'll see how it goes. Looking good so far (at least for the okd 4.2 lane)

automation/test.sh
```diff
@@ -54,6 +54,10 @@ fi
 
 export KUBEVIRT_NUM_NODES=2 
 
+if [[ "$TARGET" =~ okd-.* ]]; then
+  export KUBEVIRT_NUM_NODES=3
```
slintes: Is there a difference from okd 4.1 to 4.2, as in one counts master as node and the other one not?

dhiller: No, the problem is that for OKD 4.2 there are two workers required, otherwise the cluster will not come up. I was told that this would also be the case for 4.1, so I made this general.

dhiller: Erm, as I think about what you wrote: yes. :-$

dhiller:
@slintes now, what do you think about the latest [test results](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2790/pull-kubevirt-e2e-okd-4.2/1181841598330179584)?

slintes:
> @slintes now, what do you think about the latest

well... not great, but good enough? Not sure.
I would say: let's merge, run always, but keep optional for now. That way we get some stats for flaky tests, but it does not block tide yet

@rmohr @cynepco3hahue  wdyt?

rmohr:
Optional sounds good to me.

rmohr:
> Optional sounds good to me.

Taking it back. I don't think that we have the capacity in CI to run it if we don't stop running one or two other lanes.

dhiller:
OK, IIUC I'll wait for merge #2623, then rebase this one. OTOH we can merge  kubevirt/kubevirtci#150 then, so that I can use the git hash from master.

@slintes @rmohr @cynepco3hahue makes sense?

dhiller:
/test pull-kubevirt-e2e-okd-4.2

AFAIU I need to rebase my kubevirtci branch on master also, will do that. In the meanwhile still running the tests to see whether it gets better.

dhiller:
Closing this PR as thjere is no way of provisioning the OKD-4.2 cluster any more.

PR #2792, opened 2019-10-09
Revert sriov to only one node setup

fedepaol:

Hoping this is temporary, we revert the sriov lane to use just one note to reduce the scope of changes in order to understand ci failures.

```release-note
NONE
```


slintes:
@fedepaol do you want to merge this, or only check CI?

fedepaol:
@slintes I'd say let's see if the kind lane gets through, if it does let's merge this so I can investigate the issue in a separate PR without making tide's life difficult

fedepaol:
Ran it a couple of times, seems to be fine.
@slintes this is good to go

PR #2797, opened 2019-10-09
fix file close panic when error happens in OpenFile

beautytiger:

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/api/converter.go
+++ b/pkg/virt-launcher/virtwrap/api/converter.go
@@ -130,10 +130,10 @@ func checkDirectIOFlag(path string) bool {
 	// check if fs where disk.img file is located or block device
 	// support direct i/o
 	f, err := os.OpenFile(path, syscall.O_RDONLY|syscall.O_DIRECT, 0)
-	defer f.Close()
 	if err != nil && !os.IsNotExist(err) {
 		return false
 	}
+	defer f.Close()
 	return true
 }
 


pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -130,10 +130,10 @@ func checkDirectIOFlag(path string) bool {
 	// check if fs where disk.img file is located or block device
 	// support direct i/o
 	f, err := os.OpenFile(path, syscall.O_RDONLY|syscall.O_DIRECT, 0)
-	defer f.Close()
 	if err != nil && !os.IsNotExist(err) {
 		return false
 	}
+	defer f.Close()
```
slintes: Don't we still try to close the file now if `err != nil && os.IsNotExist(err)`?

beautytiger: Yes, I stated wrong previously, f.Close will return err if f is nil, not panic.
So, always try to call f is OK.
I just follow a pattern like [this](https://github.com/kubevirt/kubevirt/blob/304dc581aa00f36e19f71667e13f816381474777/pkg/virt-launcher/virtwrap/util/libvirt_helper.go#L331), when err is nil, just leave a nil f is ok.

slintes: Sorry for the late follow up.
The difference to the pattern (which is good) you mentioned is, that we have a 2nd condition on the `err != nil` check. So we can still have an error here, and `f` being nil, which still would result in a panic.

beautytiger: Yes, to strictly follow that pattern, below code is better
```go
	f, err := os.OpenFile(path, syscall.O_RDONLY|syscall.O_DIRECT, 0)
	if err == nil {
		defer f.Close()
		return true
	} else if os.IsNotExist(err) {
		return true
	}
	return false
```
But it's a little verbose and a little harder to read.
I think keep it simple is ok, besides, close nil file will not panic, that's my fault.
Thanks.

slintes: oh... sorry! I somehow missed that you can call methods on nil pointers. Yes, agree, code is fine then. Thanks for the heads up!

PR #2812, opened 2019-10-15
Add test id for kubevirt ready and available condition

petrkotas:
Add test for kubevirt ready and created condition.

Test was already present. Added simple test function to link the test to test id.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/operator_test.go
+++ b/tests/operator_test.go
@@ -659,6 +659,13 @@ spec:
 		ensureShasums()
 	})
 
+	It("[testid:1746]should have created and available condition", func() {
+		kv := getCurrentKv()
+
+		By("vyrifying that created and available condition is present")
+		waitForKv(kv)
+	})
+
 	Describe("should start a VM", func() {
 		It("using virt-launcher with a shasum", func() {
 


tests/operator_test.go
```diff
@@ -657,7 +657,14 @@ spec:
 
 		// make sure virt deployments use shasums again after each test
 		ensureShasums()
-	})
+    })
+
+    Describe("[testid:1746]should provide created and ready condition", func() {
```
slintes: `ready` is `available` in the meanwhile

tests/operator_test.go
```diff
@@ -657,7 +657,14 @@ spec:
...
+        kv := getCurrentKv()
+
+        By("Creating KubeVirt Object Created and Ready Condition")
```
slintes: Nothing is created here. You just verify conditions on the existing KubeVirt CR (which is ok, no need to create a new one). And ready -> available again.

tests/operator_test.go
```diff
@@ -659,6 +659,13 @@ spec:
 		ensureShasums()
 	})
 
+	Describe("[testid:1746]should have created and ready condition", func() {
+		kv := getCurrentKv()
```
rmohr: This has to be in an `It`

slintes: ah, good catch
and again, there is no `ready` condition anymore, it was renamed to `available`... please change it at all places where you mention ready

PR #2813, opened 2019-10-16
virtctl image-upload should create DataVolumes instead of PVCs

mhenriks:

DataVolume use should be encouraged.  So let's create a DataVolume when uploading.

The tool will still support uploading to a PVC created out of band, though.

Along the way, I was able to clean up the code that waits for uploadserver pods to be ready.  Rather than DIYing it, rely on an annotation set by the upload controller.

I also added exclusions/ignores for .history folder created by VS Code.

Fixes #

**Release note**:

```release-note
virtctl image-upload creates DataVolumes instead of PVCs
```


pkg/virtctl/imageupload/imageupload.go
```diff
@@ -107,15 +112,15 @@ func NewImageUploadCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
 	}
 	cmd.Flags().BoolVar(&insecure, "insecure", insecure, "Allow insecure server connections when using HTTPS.")
 	cmd.Flags().StringVar(&uploadProxyURL, "uploadproxy-url", "", "The URL of the cdi-upload proxy service.")
-	cmd.Flags().StringVar(&pvcName, "pvc-name", "", "The destination PVC.")
+	cmd.Flags().StringVar(&pvcName, "pvc-name", "", "The destination DataVolume/PVC name.")
```
rmohr: This does not look right. How can `pvc-name` be the datavolume or the PVC?

mhenriks: PVCs created by the datavolume controller always have the same name as the DataVolume (can't be different).   Also, I still want it to be possible to upload to PVCs created out of band (so can't get rid of `pvc-name`).  Finally, I didn't want to break any existing scripts that use `virtctl image-upload`.  I thought about adding another arg called `datavolume-name` that is basically an alias to `pvc-name`.  But it seemed confusing to me.  Oh yeah, and it is the PVC name that will be used in any VM/VMI definition later on.  So, for all those reasons, I think it is best to keep `pvc-name`. 

What makes the most sense to you?

rmohr: > PVCs created by the datavolume controller always have the same name as the DataVolume (can't be different)

I was not aware of this and it sounds dangerous, if this is not somehow verified via owner references or such (but I guess it is verified?).

> What makes the most sense to you?

Since providing a name is mandatory, I would suggest a similar semantic to what we have on e.g. `kubectl expose`: `virtctl upload pvc/myvmc ...` and `virtctl upload dv/mydv`.

We could deprecate `--pvc-name` and maybe remove it later.

Would that make sense?

mhenriks: @rmohr please take another look

pkg/virtctl/imageupload/imageupload.go
```diff
@@ -99,39 +105,84 @@ func NewImageUploadCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
...
+	cmd.Flags().UintVar(&uploadPodWaitSecs, "wait-secs", 60, "Seconds to wait for upload pod to start.")
 	cmd.SetUsageTemplate(templates.UsageTemplate())
 	return cmd
 }
 
 func usage() string {
-	usage := `  # Upload a local disk image to a newly created PersistentVolumeClaim:
-  {{ProgramName}} image-upload --uploadproxy-url=https://cdi-uploadproxy.mycluster.com --pvc-name=upload-pvc --pvc-size=10Gi --image-path=/images/fedora28.qcow2`
+	usage := `  # Upload a local disk image to a newly created DataVolume:
+  {{ProgramName}} image-upload dv upload-dv --uploadproxy-url=https://cdi-uploadproxy.mycluster.com --size=10Gi --image-path=/images/fedora30.qcow2
```
rmohr: The `uploadproxy-url` is optional, right? Might make sense to also have a minimum example without it.

pkg/virtctl/imageupload/imageupload.go
```diff
@@ -99,39 +105,84 @@ func NewImageUploadCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
...
+
+  # Upload a local disk image to an exististing DataVolume
+  {{ProgramName}} image-upload dv upload-dv --no-create --uploadproxy-url=https://cdi-uploadproxy.mycluster.com --image-path=/images/fedora30.qcow2
+
+  # Upload a local disk image to an exististing PersistentVolumeClaim
```
rmohr: What is not enitrely clear to me, on what the expected usage of `--no-create` is. Will we by default always just override, or do we want to fail, if a DV or PVC already exists, and `--no-create` is not supplied?

pkg/virtctl/imageupload/imageupload.go
```diff
@@ -99,39 +105,84 @@ func NewImageUploadCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
...
 	return usage
 }
 
 type command struct {
 	clientConfig clientcmd.ClientConfig
 }
 
+func parseArgs(args []string) error {
+	if len(size) > 0 && len(pvcSize) > 0 && size != pvcSize {
+		return fmt.Errorf("--pvc-size deprecated, use --size")
```
rmohr: I think that `deprecated` normally means that you warn people, but you don't block them yet, so that they have a transition time.

pkg/virtctl/imageupload/imageupload.go
```diff
@@ -99,39 +105,84 @@ func NewImageUploadCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
...
+	}
+
+	if len(args) != 2 {
+		return fmt.Errorf("expecting two args")
+	}
+
+	switch strings.ToLower(args[0]) {
+	case "dv":
+	case "pvc":
+		noCreate = true
```
rmohr: Does this mean that a PVC now needs to always exist?

pkg/virtctl/imageupload/imageupload.go
mhenriks: good idea

pkg/virtctl/imageupload/imageupload.go
mhenriks: the message only gets displayed in the unlikely case that both `--pvc-size` and `--size` are specified and their values are different.  `--pvc-size` by itself is fine.

pkg/virtctl/imageupload/imageupload.go
mhenriks: for the new invocation, yes.  So, if using `virtctl image-upload pvc foo` foo must exist.  But using the deprecated invocation `virtctl image-upload --pvc-name=foo` foo does not have to exist unless `--no-create` is specified.

pkg/virtctl/imageupload/imageupload.go
mhenriks: By default, it will create a dv/pvc if it does not already exist.  It will also use an existing dv/pvc if it does exist and has not already been uploaded to.  `--no-create` is for cases when the user wants to explicitly fail when the dv/pvc does not exist.

pkg/virtctl/imageupload/imageupload.go
mhenriks: updated help text

mhenriks:
@rmohr please take another look.

rmohr:
/override pull-kubevirt-e2e-os-3.11.0-crio

we don't run this suite anymore.

PR #2832, opened 2019-10-22
increase cpu-manager test timeout to 3 minutes

mfranczy:
It increases the cpu-manager test timeout to 3 minutes, the heartbeat is the place where we update a cpu-manager label and it happens that between calls we have 2 minutes break, as a result we have a flaky test.  The reason why it takes so long is under investigation.

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -1602,7 +1602,7 @@ var _ = Describe("Configurations", func() {
 					n, err := virtClient.CoreV1().Nodes().Get(node.Name, metav1.GetOptions{})
 					Expect(err).ToNot(HaveOccurred())
 					return n.Labels[v1.CPUManager]
-				}, 2*time.Minute, 2*time.Second).Should(Equal("false"))
+				}, 3*time.Minute, 2*time.Second).Should(Equal("false"))
 			})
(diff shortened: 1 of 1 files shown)


slintes:
The failing tests don't have the mentioned test failing. But let's try again.

/test all
/lgtm
/approve

dhiller:
On a related note: test fails in the OKD 4.2 lane: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2790/pull-kubevirt-e2e-okd-4.2/1186668276843810816

dhiller:
Though it fails on an expectation that there is only one node having the cpumanager=false label attached. I guess this is related to having two workers?

mfranczy:
@dhiller from weekly reports I saw it failed on a timeout, it means the CPU manager label was not updated in 2 minutes. Didn't check this against OKD4. It's not about two workers.

take a look at
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2771/pull-kubevirt-e2e-k8s-1.15.1-ceph/1184643435949199361

mfranczy:
@dhiller it seems you're right, it's about two nodes having `cpumanager=false`

PR #2840, opened 2019-10-24
make template VMs to use networkInterfaceMultiqueue

alonSadan:
This PR changes the template VM- Fedora and RHEL7 to be with
`networkInterfaceMultiqueue`, in order
to receive better network performances.
benefits and reasoning behind `networkInterfaceMultiqueue`
can be found here:
https://www.linux-kvm.org/page/Multiqueue#Multiqueue_virtio-net
The Windows template is still **without** `networkInterfaceMultiqueue`
since the CPU model it uses is `e1000` which does not support `networkInterfaceMultiqueue`.
**Release note**:

```release-note
template Fedora and RHEL7 VMs are now configured with network interface multiqueue
```

### DIFF
--- a/tools/vms-generator/utils/utils.go
+++ b/tools/vms-generator/utils/utils.go
@@ -83,9 +83,10 @@ const (
 	imageCirros = "cirros-container-disk-demo"
 	imageFedora = "fedora-cloud-container-disk-demo"
 )
-
 const windowsFirmware = "5d307ca9-b3ef-428c-8861-06e72d69f223"
 const defaultInterfaceName = "default"
+const enableNetworkInterfaceMultiqueueForTemplate = true
+const EthernetAdaptorModelToEnableMultiqueue = "virtio"
 
 var DockerPrefix = "registry:5000/kubevirt"
 var DockerTag = "devel"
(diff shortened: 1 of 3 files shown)


alonSadan:
@booxter  since you reviewed the last templates PR , would be happy if you can take a look:)

booxter:
@alonSadan two questions:

1) the first question that comes up when I see this patch is - if multiqueue is so great, why do we not just enable it for all templates? I believe there is a proper answer to this question, but I don't see it captured in code comments nor in git commit message. Please capture it somewhere for reviewers and posterity.

2) the patch sets an empy `model` for default interface in golang code. (Though it doesn't seem to reflect in templates in any way.) My question is: why do we need to set the model to empty string now?

alonSadan:
Hi @booxter.
1. I saw only three templates : Windows, Fedora and RHEL7. do you suggest that maybe I should add multiqueue also to the examples (what is the difference actually between the templates and the examples?)?
In the windows template it is not possible to add multiqueue since the  current cpu model (`e1000`) does not support multiqueue. Only VMs with `virtio` cpu model can support multiqueue.
I'll add that note in the commit message.

2. the default  cpu model is `virtio` , so when you leave the model empty it will set the model to `virtio`. But it will defently be clearer if I'll wirte explicitly `virtio`. I'll change that.

Thanks.

tools/vms-generator/utils/utils.go
```diff
@@ -83,9 +83,10 @@ const (
 	imageCirros = "cirros-container-disk-demo"
 	imageFedora = "fedora-cloud-container-disk-demo"
 )
-
 const windowsFirmware = "5d307ca9-b3ef-428c-8861-06e72d69f223"
 const defaultInterfaceName = "default"
+const enableNetworkInterfaceMultiqueueForTemplate = true
+const CPUModelToEnableMultiqueue = "virtio"
```
booxter: it's not CPU model, it's ethernet adaptor model. (e1000 is an Ethernet driver; same for virtio - in this context.)

alonSadan: Thanks for the clarification. Done

alonSadan:
@vladikr could you please take a look as well?

PR #2843, opened 2019-10-27
Move glog package from pkg/staging to staging/src/github.com/golang

cynepco3hahue:
For consistenty move glog package from the `pkg/staging` to `staging/src/github.com/golang` directory.

Fixes #

**Release note**:

```release-note
NONE
```


staging/src/github.com/golang/glog/go.mod
```diff
@@ -0,0 +1,9 @@
+module kubevirt.io/kubevirt/staging/src/github.com/golang/glog
```
slintes: hm, shouldn't this be `github.com/golang/glog`?

cynepco3hahue: good catch! it was autogenerated by `mod init`

PR #2844, opened 2019-10-27
Rewrite container-disk main on C to prevent memory overconsumption

cynepco3hahue:
- add packages for static compilation under the builder container
- rewrite main.go on C

Fixes #

**Release note**:

```release-note
Rewrite container-disk main.go file on C to prevent memory overconsumption
```


cynepco3hahue:
/hold
Looks like this PR broke migration tests.

cynepco3hahue:
The operator test is broken, needs to find a way on C to parse word arguments.

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,142 @@
...
+    FILE *probe;
+    probe = fopen(READINESS_PROBE_FILE, "w");
+    if (probe == NULL) {
+        printf("failed to create readiness probe\n");
+        exit(1);
+    }
+    fclose(probe);
+
+    pthread_t socket_check_thread;
+    pthread_create(&socket_check_thread, NULL, socket_check, (void *)copy_path);
```
rmohr: I think that you just forgot to remove the `pthread` related code.

rmohr:
@slintes FYI, if we are going to merge this, the binary will have to be compiled fully static always. That is cruicial when then building outside of bazel.

slintes:
I still dislike the idea of adding a new language to the stack at all, just
for solving a memory issue of one binary...

Roman Mohr <notifications@github.com> schrieb am Do., 31. Okt. 2019, 08:10:

rmohr:
> I still dislike the idea of adding a new language to the stack at all, just for solving a memory issue of one binary...

Me too, but the overhead is huge. Minimum 40 mb per PR to make it stable right now. That's why I proposed the idea. It is similar to the `pause` container in k8s. @cynepco3hahue can we change our controller to assign less memory, so that we see the improvement?

cmd/container-disk-v2alpha/main.c
cynepco3hahue: not really, in the end, we need it, `accept` block the code until it receives any requests, and socket remove does not unblock it, so we need socket check to exit from the main function and release all resources.

cynepco3hahue:
> @cynepco3hahue can we change our controller to assign less memory, so that we see the improvement?

Sure, I will check how much memory does it consume and will set the limit accordingly.

cynepco3hahue:
Results in memory consumption

### C code
Memory consumption is stable, always stay the same.
- k8s-1.13 docker
```
CONTAINER           CPU %               MEM USAGE / LIMIT    MEM %               NET I/O             BLOCK I/O           PIDS
7e8442775578        0.81%               56 KiB / 38.14 MiB   0.14%               0 B / 0 B           0 B / 0 B           2
```
- okd-4.1 crio

### GO
Memory consumption is not stable and always shows different values

- k8s-1.13 docker
```
CONTAINER           CPU %               MEM USAGE / LIMIT       MEM %               NET I/O             BLOCK I/O           PIDS
148b3a55d8d6        2.85%               2.547 MiB / 38.14 MiB   6.68%               0 B / 0 B           0 B / 0 B           12
```
After an hour the consumption started to jump up to 3.5MiB it still far away, but to be honest it pretty interesting why do we have consumption increase at all.

I still need to check the consumption under the cri-o, but I encountered the bug - https://bugzilla.redhat.com/show_bug.cgi?id=1666358

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,142 @@
...
+#define READINESS_PROBE_FILE "/healthy"
+
+static void *socket_check(void *arg) {
+    struct stat st = {0};
+    char *copy_path;
+    copy_path = (char *)arg;
+
+    for (;;) {
+        sleep(1);
+        if (stat(copy_path, &st) == -1) {
```
stu-gott: This is going to cause constant (albeit small) load on the system. Would it be possible to use something along the lines of inotify or select to block instead of constantly polling?

cmd/container-disk-v2alpha/main.c
stu-gott: This feels like an anti-pattern. Can't we just `listen` on the socket, then use `select` to block until a connection is made or the socket is deleted? It would remove the need for the socket_check_thread entirely I think.

cmd/container-disk-v2alpha/main.c
cynepco3hahue: I just re-used code from the https://github.com/kubevirt/kubevirt/blob/a6aa0981230f97b9996f57f73691e7376b9fddf1/cmd/container-disk-v2alpha/main.go#L78

I do not have the intention to change the code under this PR, but we can leave it for the following PR.

cmd/container-disk-v2alpha/main.c
cynepco3hahue: I do not sure I am following you, I can see that `select` can wait for number of FD to be ready with some specified timeout, in our case we have `accept` that blocked until it receives any request and the tread is checking the socket file, to exit the main when the virt-handler will clean the socket. Maybe do you have some examples of select usage that will apply to our use case?

stu-gott: On line 131, instead of blindly doing an `accept`, which will block. do a `select`--which can also block indefinitely. When select returns, one of two things is true:

1) the socket has been closed, and exit(1) is the next step
2) an incoming connection has been initiated. `accept` is guaranteed not to block.

Thus "select -> check -> accept" alleviates the need for a separate thread that simply monitors the health of the socket. the main thread will be woken up immediately if it becomes unhealthy.

rmohr:
/hold

I am in favor of this PR, but not everyone is. Let's take the time to discuss the impact next week a little bit more.

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,142 @@
...
+        exit(1);
+    }
+    fclose(probe);
+
+    pthread_t socket_check_thread;
+    pthread_create(&socket_check_thread, NULL, socket_check, (void *)copy_path);
+
+    int data_socket;
+    for (;;) {
+        data_socket = accept(fd, NULL, NULL);
```
rmohr: I  don't think that we have to `accept` connections. I would have to try it out, but `listen` above should already be sufficient. If that' the case, we could just live with an interrupt handler instead of a thread.

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,142 @@
...
+
+    struct sockaddr_un address;
+    /*
+    * For portability clear the whole structure, since some
+    * implementations have additional (nonstandard) fields in
+    * the structure.
+    */
+    memset(&address, 0, sizeof(struct sockaddr_un));
+    address.sun_family = AF_UNIX;
+    strcat(copy_path, ".sock");
```
mhenriks: Is this safe?  We do not appear to allocate the memory for `copy_path`.  Sure it has the capacity?

cynepco3hahue: hm, good question, in most examples I can see no allocation, but here https://linux.die.net/man/3/strcat I can see that can be used buffer overruns attack, so the best way will be to allocate memory and to free it after.

cynepco3hahue:
@stu-gott @mhenriks @rmohr I can continue with the PR, once we will have some agreement if we want to add the complexity of C code to the go project.

cmd/container-disk-v2alpha/main.c
rmohr: @cynepco3hahue did not fully check the signal handling, but the VMIs could be started and stopped successfully without the need to accept connections on the socket:

```diff
diff --git a/cmd/container-disk-v2alpha/main.c b/cmd/container-disk-v2alpha/main.c
index b0fd3d855..8f24f82d9 100644
--- a/cmd/container-disk-v2alpha/main.c
+++ b/cmd/container-disk-v2alpha/main.c
@@ -1,7 +1,6 @@
 #include <ctype.h>
 #include <getopt.h>
 #include <libgen.h>
-#include <pthread.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -11,10 +10,18 @@
 #include <sys/types.h>
 #include <sys/un.h>
 #include <unistd.h>
+#include <signal.h>
 
 #define LISTEN_BACKLOG 50
 #define READINESS_PROBE_FILE "/healthy"
 
+char *copy_path = NULL;
+
+void sig_handler(int signo) {
+    unlink(copy_path);
+    exit(0);
+}
+
 static void *socket_check(void *arg) {
     struct stat st = {0};
     char *copy_path;
@@ -30,7 +37,6 @@ static void *socket_check(void *arg) {
 }
 
 int main(int argc, char **argv) {
-    char *copy_path = NULL;
     char *copy_path_dir;
     bool health_check = false;
 
@@ -98,6 +104,11 @@ int main(int argc, char **argv) {
     strcat(copy_path, ".sock");
     strcpy(address.sun_path, copy_path);
 
+   if (signal(SIGTERM, sig_handler) == SIG_ERR) {
+     printf("failed to register SIGTERM callback");
+     exit(1);
+   }
+
     int fd = socket(AF_UNIX, SOCK_STREAM, 0);
     if (fd == -1) {
         printf("failed to create socket on %s\n", copy_path);
@@ -123,20 +134,5 @@ int main(int argc, char **argv) {
     }
     fclose(probe);
 
-    pthread_t socket_check_thread;
-    pthread_create(&socket_check_thread, NULL, socket_check, (void *)copy_path);
-
-    int data_socket;
-    for (;;) {
-        data_socket = accept(fd, NULL, NULL);
-        if (data_socket == -1) {
-            if (stat(copy_path, &st) == -1) {
-                printf("socket %s does not exist anymore\n", copy_path);
-                exit(0);
-            }
-            printf("failed to accept connection\n");
-            unlink(copy_path);
-            exit(1);
-        }
-    }
+    socket_check((void *)copy_path);
 }
```

Should make the binary smaller and simpler, since no threads are required.

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,142 @@
+#include <ctype.h>
```
rmohr: What do we need ctype for? Can we get rid of libstdc++?

cmd/container-disk-v2alpha/main.c
cynepco3hahue: I removed pthread, now let's see if CI likes it:)

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,139 @@
...
+
+        /* Detect the end of the options. */
+        if (c == -1) {
+            break;
+        }
+
+        switch (c) {
+            case 'c':
+                copy_path = optarg;
+                copy_path_dir = strdup(copy_path);
```
rmohr: Let's use `strndup`.

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,139 @@
...
+                dirname(copy_path_dir);
```
rmohr: I am not sure, bit I think that this does nothing?

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,139 @@
...
+    struct sockaddr_un address;
+    /*
+    * For portability clear the whole structure, since some
+    * implementations have additional (nonstandard) fields in
+    * the structure.
+    */
+    memset(&address, 0, sizeof(struct sockaddr_un));
+    address.sun_family = AF_UNIX;
+    strcat(copy_path, ".sock");
+    strcpy(address.sun_path, copy_path);
```
rmohr: Let's use `strncopy`.

rmohr: Let's do it like this `strncpy(addr.sun_path, copy_path, sizeof(addr.sun_path) - 1);` Note that the size for `sun_path` is `108`. Any path longer than this may cause issues.

hack/builder/Dockerfile
```diff
@@ -17,6 +17,8 @@ RUN dnf install -y dnf-plugins-core && \
         sudo \
         gcc \
         gcc-c++ \
+        glibc-static \
+        libstdc++-static.x86_64 \
```
rmohr: I don't think that we need anything from c++ here. Can it be that you can already get rid of this dependency?

cmd/container-disk-v2alpha/main.c
danielBelenky: From the man page, it seems like we will have a leak here because `copy_path` is statically allocated...
Also, consider using `strncat` even though right now we just cat a constant string.
```
DESCRIPTION
       The  strcat()  function  appends  the src string to the dest string, overwriting the terminating
       null byte ('\0') at the end of dest, and then adds a terminating null byte.  The strings may not
       overlap,  and  the  dest  string  must  have  enough space for the result.  If dest is not large
       enough, program behavior is unpredictable; buffer overruns are a favorite avenue  for  attacking
       secure programs.
```

cmd/container-disk-v2alpha/main.c
cynepco3hahue: done

cmd/container-disk-v2alpha/BUILD.bazel
```diff
@@ -1,17 +1,9 @@
...
-
-go_binary(
+cc_binary(
     name = "container-disk",
-    embed = [":go_default_library"],
-    pure = "on",
+    srcs = ["main.c"],
+    linkopts = [
+        "-static",
+        "-pthread",
```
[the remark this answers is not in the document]
cynepco3hahue: done

cmd/container-disk-v2alpha/main.c
cynepco3hahue: done

cmd/container-disk-v2alpha/main.c
cynepco3hahue: done

cmd/container-disk-v2alpha/main.c
cynepco3hahue: done

cmd/container-disk-v2alpha/main.c
cynepco3hahue: done

hack/builder/Dockerfile
cynepco3hahue: I tried to compile without it and it did not work, I think the problem that bazel does not have separate rules for C only, it has only `cc_binary`, so it compiled as C++ with C++ flags and one of the flags is `-lstd++` and it requires `libstdc++-static.x86_64`

cmd/container-disk-v2alpha/BUILD.bazel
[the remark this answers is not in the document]
cynepco3hahue: done

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,139 @@
...
+
+static void *socket_check(void *arg) {
+    struct stat st = {0};
+    char *copy_path;
+    copy_path = (char *)arg;
+
+    for (;;) {
+        sleep(1);
+        if (stat(copy_path, &st) == -1) {
+            printf("socket %s does not exist anymore\n", copy_path);
```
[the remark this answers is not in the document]
cynepco3hahue: done

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,139 @@
...
+            exit(0);
+        }
+    }
+
+    if (stat(copy_path_dir, &st) == -1) {
+        if (mkdir(copy_path_dir, 0777) != 0) {
+            printf("failed to create disk directory %s\n", copy_path_dir);
+            exit(1);
+        }
+    }
```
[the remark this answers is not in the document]
cynepco3hahue: done

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,139 @@
...
+                exit(1);
+            default:
+                abort();
+        }
+    }
+
+    struct stat st = {0};
+    if (health_check) {
+        if (stat(READINESS_PROBE_FILE, &st) == -1) {
+            printf("readiness probe %s does not exist\n", READINESS_PROBE_FILE);
```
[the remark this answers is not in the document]
cynepco3hahue: done

stu-gott:
I think we should proceed with this PR. It appears that the builder container needs to be rebuilt.

vladikr:
We need to try running it on OCP as well. To see how much memory does it take there and how does it behave with podman.

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,141 @@
...
+
+    struct sockaddr_un address;
+    /*
+    * For portability clear the whole structure, since some
+    * implementations have additional (nonstandard) fields in
+    * the structure.
+    */
+    memset(&address, 0, sizeof(struct sockaddr_un));
+    address.sun_family = AF_UNIX;
+    strncat(copy_path, ".sock", 5);
```
rmohr: can you check the returned `char*` to see if the data was not truncated?

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,141 @@
...
+    strncpy(address.sun_path, copy_path, sizeof(address.sun_path));
```
rmohr: can you check the returned char* to see if the data was not truncated?

pkg/container-disk/container-disk.go
```diff
@@ -126,14 +126,14 @@ func GenerateContainers(vmi *v1.VirtualMachineInstance, podVolumeName string, bi
 			if vmi.IsCPUDedicated() || vmi.WantsToHaveQOSGuaranteed() {
 				resources.Limits = make(kubev1.ResourceList)
 				resources.Limits[kubev1.ResourceCPU] = resource.MustParse("10m")
-				resources.Limits[kubev1.ResourceMemory] = resource.MustParse("40M")
+				resources.Limits[kubev1.ResourceMemory] = resource.MustParse("5M")
```
rmohr: We have repeatedly OOM reports. Lets see if we can reduce that later (since it requires additional validation), by now just leaving it at `40M` and merge this earlier.

hack/dockerized
```diff
@@ -3,7 +3,7 @@ set -e
 
 source $(dirname "$0")/common.sh
 
-KUBEVIRT_BUILDER_IMAGE="kubevirt/builder@sha256:7585e27b5a5a019f5749eaa1ef29da3278878eefc862b45d6e1c2e343d9f8859"
+KUBEVIRT_BUILDER_IMAGE="kubevirt/builder@sha256:e6a76ec71f2de87c3bc648bae87ab43a113540bf57da69f07f76bd0edfd500ab"
```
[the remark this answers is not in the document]
cynepco3hahue: done

rmohr:
/lgtm
/approve
/hold

@davidvossel @slintes @vladikr I would prefer to merge this PR soon. We could also increase the memory limit for the go container, but I would prefer a permanent solution without memory limits which try to cope with unpredictable memory spikes which in addition bind a lot of memory.

I know that this may cause some overhead, due to introducing another language to the stack, but I personally still think it is worth it. Your oppinions?

slintes:
I still don't like it, but I see that it helps. So, as long as there are enough maintainers who feel comfortable with maintaining this (I'm not), I won't block.

vladikr:
/lgtm

I also would like to see this merged. However, I'm still puzzled about why the memory consumption of the container disk varies between different systems. Hope this change will reduce the memory consumption in openshift as well.

rmohr:
/hold

Ok, in the meantime something seems to have changed. I see

```
'Warning' reason: 'SyncFailed' failed to detect socket for containerDisk disk0: dial unix /var/run/kubevirt/container-disks/dbdc7abd-6d04-11ea-b15e-525500d15501/disk_0.sock: connect: no such file or directory"
```

all over the place.

cynepco3hahue:
@rmohr yep, it pretty strange, I tried some tests locally and it passed, probably some race :(

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,157 @@
...
+        error_log("failed to create socket on %s\n", copy_path);
+        exit(1);
+    }
+
+    if (bind(fd, (struct sockaddr*)(&address), sizeof(struct sockaddr_un)) == -1) {
+        error_log("failed to bind socket %s\n", copy_path);
+        exit(1);
+    }
+
+    if (listen(fd, LISTEN_BACKLOG) == -1) {
```
davidvossel: don't we need to be accepting incoming connections? I'm surprised the net.Dial() command in isolation.go works when we don't call accept() on the server side. maybe it does somehow? 

i'm not sure what implications this might have without an accept though. For example, what happens to LISTEN_BACKLOG without an accept?

rmohr: I was actually under the impression that just `listen` is enough, because I assumed that the SYN  queue will be emptied when the client side closes the connection. Actually it sums up. So @cynepco3hahue can we please `accept` connections after `listen`?

rmohr: > Besides the code comment, I'm not completely sure of what we solve here. If memory is this inconsistent where we can't guarantee such a simple execution won't go over form 3mb to >40mb, then what hope do we have to accurately calculate the VMI overhead. This is really strange.

Right now we have an overhead of `128MB` for virt-launcher and libvirt. libvirt does not need much, so there is enough room for virt-laucher, since it seems that we did not see many reports. On the other hand reports may just not come in too often, because people don't use too many VMs with QoS guaranteed.

> 
> did we profile the process to understand how in the world this OOMs with a 40mb limit? This seems crazy. What environments are we seeing this in? is something like Transparent Hugepages enabled?

We saw it e.g. in the past on Azure, when we still had a smaller memory limit. Pushing it to 40 solved it there, but now we see it on some openshift environments (and even there only on some nodes). We should definitely not reserve more memory for something simple like this, but I also don't want to always guess for different go versions on different system. I would prefer to eliminate the problem for the containerDisk completely and maybe in the future reduce the `40M` limit.

rmohr:
/hold
/lgtm cancel

We need to accept the calls after listen.

cmd/container-disk-v2alpha/main.c
rmohr: > So @cynepco3hahue can we please accept connections after listen

That probably involves introducing a thread again. Like you had it in the beginning :/

rmohr:
@cynepco3hahue @davidvossel updated the PR with a non-blocking solution, to avoid threading and releated dependencies.

cmd/container-disk-v2alpha/main.c
```diff
@@ -128,6 +179,9 @@ int main(int argc, char **argv) {
         exit(1);
     }
 
+    /* make the socket non-blocking */
+    fcntl(fd, F_SETFL, O_NONBLOCK);
```
cynepco3hahue: nice!

rmohr:
hm, looks like our prow and openshift-ci could not update the github status. Will retrigger all tests.

rmohr:
/hold cancel

unit test added too.

cmd/container-disk-v2alpha/main_test.go
```diff
@@ -0,0 +1,42 @@
...
+		defer os.RemoveAll(dir)
+		cmd := exec.Command(containerDiskBinary, "-c", filepath.Join(dir, "testsocket"))
+		Expect(cmd.Start()).To(Succeed())
+
+		time.Sleep(1 * time.Second)
+		for i := 0; i < 200; i++ {
+			conn, err := net.Dial("unix", filepath.Join(dir, "testsocket.sock"))
+			Expect(err).ToNot(HaveOccurred())
+			conn.Close()
+			time.Sleep(25 * time.Millisecond)
```
davidvossel: actually, we don't want the sleep do we? we actually want to know if it can handle then all at once.

In fact, I'd make this test do all 200 Dials in a separate goroutine and use a waitgroup to wait on them exit.

cmd/container-disk-v2alpha/main.c
```diff
@@ -0,0 +1,204 @@
...
+     * ther is a connection, we close it immediatley and immediately try to
+     * read the next connection until there are no more connections in the
+     * queue. Once the queue is empty, we fall back to sleep for a second.
+     *
+     * If within that second more than 50 connections enter the queue, then we
+     * clearly have a bug in virt-handler.
+     */
+    for (;;) {
+        if (!connReceived) {
+            sleep(1);
```
davidvossel: what's the purpose of the sleep?  Is it to keep this loop from spinning because accept is occurring on a non-blocking socket?

if so, lets use a select or poll on the fd to only call accept when there's something to process

rmohr: afaik poll only allows me to wait on multiple file descriptors, but still blocks. We also want to check if the socket file disappeared, since we use it as a further shutdown signal for the container-disk. I don't know if you remember it but we had on many occasions forever terminating pods, because the `compute` container was terminated but the container-disk pod never received the signal. This additional check allows to transmit the shutdown signal also via virt-handler in ddition.

cmd/container-disk-v2alpha/main_test.go
rmohr: As it is right now, we sleep for 1 second (because we also want to check if the socket still exists). The rate-limit I added here ensures that we can cope with an initial delay of a second, until it finally starts accepting all the connections in a row.

rmohr: It still ensures that we definitely process socket connections way way faster than they are ever allowed to come in from virt-handler.

cmd/container-disk-v2alpha/main.c
davidvossel: oh yeah, i forgot about that. makes sense, and that's something I'll have to remember to add back to my hostpath PR as well. I removed the removal of the socket file.

cmd/container-disk-v2alpha/main_test.go
davidvossel: as long as dial has a timeout > a few seconds, we should be able to flood connections in. right?

rmohr:

the usual ppc64 issue.

PR #2845, opened 2019-10-27
Added a test ID to a slirp test.

yossisegev:
This PR maps a single slirp test by adding a test ID to it, so execution of this test can be tracked in our test tracking system.


yossisegev:
ci please test

yossisegev:
This PR can be abandoned, because the test case it relates to should be deleted.

PR #2870, opened 2019-11-04
Pin QEMU emulator threads to a dedicated pCPU

vladikr:
Currently, KubeVirt supports an assignment of dedicated physical CPUs to VMIs virtual CPUs.
This provides improved latency and better overall performance to the workloads running in the VMI guests. However, there are additional threads of the QEMU process, such as QEMU main event loop, async I/O operation completion, migration data transfer and more, that are also using the same physical CPUs during the VMI execution.

In order to enhance the real-time support in KubeVirt, the QEMU emulator threads need to be "isolated" - pinned to a dedicated physical CPU that is not used by any of the VMI virtual CPUs.
This will require allocating an additional dedicated CPU on the pod exclusively for the emulator thread.  

**Release note**:
```release-note
NONE
```


pkg/virt-controller/services/template.go
```diff
@@ -613,6 +613,18 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 				resources.Limits[k8sv1.ResourceCPU] = cpuRequest
 			}
 		}
+		// allocate 1 more pcpu if IsolateEmulatorThread request
+		if vmi.Spec.Domain.CPU.IsolateEmulatorThread {
+			emulatorThreadCpu := resource.NewQuantity(1, resource.BinarySI)
+			limits := resources.Limits[k8sv1.ResourceCPU]
+			limits.Add(*emulatorThreadCpu)
+			resources.Limits[k8sv1.ResourceCPU] = limits
+			if cpuRequest, ok := resources.Requests[k8sv1.ResourceCPU]; ok {
```
slintes: I was wondering why we need to check if `cpuRequest` exists, and see that we don't set it a few lines above in case `vcpus != 0`. But I don't understand why we don't set it. Don't we have to always make sure that `limit == request`, in order to enable CPU pinning?

vladikr: Yes, we absolutely need `limits == requests`, however, if we only set the limits, requests will be set equal to limits by k8s. Last example in [1]

```
spec:
  containers:
  - name: nginx
    image: nginx
    resources:
      limits:
        memory: "200Mi"
        cpu: "2"
```

> This pod runs in the Guaranteed QoS class because only limits are specified and requests are set equal to limits when not explicitly specified. And the container’s resource limit for the CPU resource is an integer greater than or equal to one.

That's why I check here if requests are set.

[1] https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/

slintes: Ah, smart k8s, thanks for the heads up :+1: 
But I still think we have an issue (not caused by this PR): since we don't do anything about cpu requests when `vcpus != 0`, it can be different than the limit we set, which would disable CPU pinning?

vladikr: We shouldn't have a problem there, I think. We only set the limits so k8s will set the requests.
We also have several validations in the validation webhook[1] to make sure that we reject any inconsistent entries. 

[1] https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go#L340

slintes: the requests might have been set already by copying from the VMI earlier in this method.
but maybe I'm missing something...

vladikr: We can definitely add another validation here, but this should be prevented by the webhook.
Let's say we have both requests and vcpus submitted and don't have the same value, it would be blocked by [1]

https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go#L415

fabiand:
@rmohr @stu-gott do you mind taking a look?

PR #2895, opened 2019-11-22
Increase timeout for expose tests

dhiller:

The tests regarding expose, namely test_id:153[14567], always failed
once in a while. Suspicion is that they get brittle whenever more
load on a node occurs. Tried to check that using stress-ng, and
could reproduce it with `stress-ng --cpu 0 --cpu-load 70 ...`
on local test system.
Current timeout values don't show any failures any more when
running with FMeasure(20) and mentioned load simulation. Will run a set of repeated tests over the weekend.

Fixes #

**Release note**:

```release-note
NONE
```


dhiller:
First test with cpu load of 70% revealed one test failure in test_id:1534 and one in test_id:1537.

PR #2893, opened 2019-11-21
Update to libvirt 5.6.0

rmohr:
**What this PR does / why we need it**:

Update kubevirt to libvirt 5.6.0.

* Libvirt 5.6.0 wants to set XATTRS when `remember_owner` is set to 1. Since we don't want to grant SYS_CAP_ADMIN privileges, disable `remember_owner`.
 * There is an issue with `dynamic_ownership`. It stopped working for us. Taking this as an opportuinity to clean up the code and set the ownership in a consistent fashion for all disks.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

`dynamic_ownership` should work in dependent of XATTRS. We will investigate this further, but for now just set the ownership of disks ourselves in a consistent way.

 https://github.com/kubevirt/libvirt/pull/39 prepared kubevirt/libvirt.

**Release note**:
<!--  Write your release note:

```release-note
Update to libvirt-5.6.0
```

PR #2909, opened 2019-11-26
Prepare kubevirt for libvirt 5.6.0 without switching

rmohr:

Prepare kubevirt for libvirt 5.6.0 without actually switching.

* Libvirt 5.6.0 wants to set XATTRS when `remember_owner` is set to 1. Since we don't want to grant SYS_CAP_ADMIN privileges, disable `remember_owner`.
 * There is an issue with `dynamic_ownership`. It stopped working for us. Taking this as an opportuinity to clean up the code and set the ownership in a consistent fashion for all disks.

Fixes #

`dynamic_ownership` should work in dependent of XATTRS. We will investigate this further, but for now just set the ownership of disks ourselves in a consistent way.

 https://github.com/kubevirt/libvirt/pull/39 prepared kubevirt/libvirt.

#2893 actually switches to libvirt 5.6.0. This switch is done separately, since the 5.6 version there has slirp disabled. This can be sorted out separtely.

**Release note**:
<!--  Write your release note:

```release-note
NONE
```

### DIFF
--- a/cmd/virt-launcher/virt-launcher.go
+++ b/cmd/virt-launcher/virt-launcher.go
@@ -325,7 +325,7 @@ func main() {
 	log.InitializeLogging("virt-launcher")
 
 	if !*noFork {
-		exitCode, err := ForkAndMonitor("qemu-system", *ephemeralDiskDir, *containerDiskDir)
+		exitCode, err := ForkAndMonitor("qemu-kvm", *ephemeralDiskDir, *containerDiskDir)
 		if err != nil {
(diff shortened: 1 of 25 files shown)


rmohr:
/cc @stu-gott @mhenriks 

since we need the code backported with or without libvirt 5.6, I created this PR, since I can't provide immediately a solution to the disabled slirp features in #2893

rmohr:
/hold

nevermind, have to still sort out the exec name issue. Can't solve it today anymore.

PR #2913, opened 2019-11-28
Update workspace libvirt5.6.0

qinqon:

This PR change the bazel WORKSPACE to point to libvirt  5.6.0 and also fix the slirp test it does not work with fedora31 containers https://github.com/kubevirt/kubevirt/pull/2893#issuecomment-559466752

Fixes #

**Release note**:

```release-note
Update to libvirt 5.6
```

### DIFF
--- a/tests/vmi_slirp_interface_test.go
+++ b/tests/vmi_slirp_interface_test.go
@@ -101,13 +101,12 @@ var _ = Describe("Slirp Networking", func() {
 		Expect(err).ToNot(HaveOccurred())
 		// :0050 is port 80, 0A is listening
 		Expect(strings.Contains(output, "0: 00000000:0050 00000000:0000 0A")).To(BeTrue())
-
 		By("return \"Hello World!\" when connecting to localhost on port 80")
 		output, err = tests.ExecuteCommandOnPod(
 			virtClient,
 			vmiPod,
 			vmiPod.Spec.Containers[0].Name,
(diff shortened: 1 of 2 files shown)


rmohr:
/lgtm
/approve

In case the tests all pass, let's just directly take it. :+1:

tests/vmi_slirp_interface_test.go
```diff
@@ -101,13 +101,12 @@ var _ = Describe("Slirp Networking", func() {
...
 		// :0050 is port 80, 0A is listening
 		Expect(strings.Contains(output, "0: 00000000:0050 00000000:0000 0A")).To(BeTrue())
-
 		By("return \"Hello World!\" when connecting to localhost on port 80")
 		output, err = tests.ExecuteCommandOnPod(
 			virtClient,
 			vmiPod,
 			vmiPod.Spec.Containers[0].Name,
-			[]string{"curl", "-s", "--retry", "30", "--retry-delay", "30", "127.0.0.1"},
+			[]string{"nc", "127.0.0.1", "80", "--recv-only"},
```
fabiand: Why is this change needed? Just being curious

rmohr: We by accident used curl here, although we only expect a tcp packet to arive. CURL recently disabled support for http 0.9 by default. As a consequence it did not consider the guest response as a valid response.

I also had issues connecting from within the guest to the outside world, but it looks like this was not reproducible for @qinqon. So that is hopefully it.

PR #2916, opened 2019-12-02
Fix deletion of ServiceMonitor CR

danielBelenky:
We could never detect a failed deletion because the condition was
prefixed with a `false` boolean.

In order to properly handle the deletion of service monitors

Fixes #

**Release note**:

```release-note

```


stu-gott:
Thanks for catching this @danielBelenky

PR #2920, opened 2019-12-03
introduce a flag to enable/disable  managing sc

tareqalayan:
This PR introduces a flag which enables/disable managing the storage classes within the tests.
With this PR we can disable the creation and deletion of storage class in a case we have storage classes defined beforehand. 

**Release note**:

```release-note
NONE
```


rmohr:
@tareqalayan please sign your commit.

PR #2921, opened 2019-12-04
Return proper errors on aggregated-api subresources

rmohr:

Return a Status object in order to avoid that the apiserver overrides
all our custom errors.

This allows us to finally use all return codes which we need and also
improves the error reporting on the client side.

Fixes #

**Release note**:

```release-note
Start returning more reasonable return codes in our aggregated api subresources
```


pkg/virt-api/rest/subresource_test.go
```diff
@@ -21,13 +21,17 @@ package rest
...
 	"flag"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/onsi/ginkgo/extensions/table"
+	"k8s.io/apimachinery/pkg/api/errors"
```
slintes: nit: import ordering

pkg/virt-api/rest/subresource_test.go
```diff
@@ -947,13 +942,17 @@ var _ = Describe("VirtualMachineInstance Subresources", func() {
 	Context("Pausing", func() {
 		It("Should pause a running, not paused VMI", func() {
 
+			backend.AppendHandlers(
+				ghttp.CombineHandlers(
```
slintes: duplicate with line 104ff?

pkg/virt-api/rest/subresource_test.go
```diff
@@ -995,20 +988,21 @@ var _ = Describe("VirtualMachineInstance Subresources", func() {
...
 
-			Expect(response.Error()).To(HaveOccurred())
-			Expect(response.StatusCode()).To(Equal(http.StatusForbidden))
-
+			ExpectStatusErrorWithCode(recorder, http.StatusConflict)
 		})
 
 		It("Should unpause a running, paused VMI", func() {
-
+			backend.AppendHandlers(
```
slintes: duplicate with line 104ff?

pkg/virt-api/rest/subresource.go
```diff
@@ -618,18 +620,18 @@ func (app *SubresourceAPIApp) StopVMRequestHandler(request *restful.Request, res
...
-	validate := func(vmi *v1.VirtualMachineInstance) (error, int) {
-		if vmi == nil || vmi.Status.Phase != v1.Running {
-			return fmt.Errorf("VMI is not running"), http.StatusForbidden
+	validate := func(vmi *v1.VirtualMachineInstance) *errors.StatusError {
+		if vmi.Status.Phase != v1.Running {
+			return errors.NewConflict(v1.Resource("virtualmachineinstance"), vmi.Name, fmt.Errorf("VM is not running"))
 		}
 		if vmi.Spec.LivenessProbe != nil {
-			return fmt.Errorf("Pausing VMIs with LivenessProbe is currently not supported"), http.StatusForbidden
+			return errors.NewForbidden(v1.Resource("virtualmachineinstance"), vmi.Name, fmt.Errorf("Pausing VMIs with LivenessProbe is currently not supported"))
```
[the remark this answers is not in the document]
rmohr: Yep should be conflict. :+1:

rmohr:
> @rmohr The few test failures don't seem random here, you might have to change the error msg that we're expecting on those tests

:+1: looking.

pkg/virt-api/rest/subresource_test.go
rmohr: fixed.

pkg/virt-api/rest/subresource_test.go
rmohr: fixed.

pkg/virt-api/rest/subresource_test.go
rmohr: fixed.

pkg/virt-api/rest/subresource.go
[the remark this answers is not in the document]
rmohr: Oh, on second thought, I did this intentonally. The reason was that the call would never succeed for as long as you don't recreate the VMI with a liveness probe.

I am unsure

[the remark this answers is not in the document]
slintes: +1 for forbidden here
technically you can pause a VM with LivenessProbe, we just decided that it doesn't make sense with the current implementation of the LivenessProbe, so we forbid it.

pkg/virt-api/rest/subresource_test.go
slintes: I still see the same handler at line 106?

rmohr: now really gone.

pkg/virt-api/rest/subresource.go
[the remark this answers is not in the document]
rmohr: @mhenriks what do you think?

PR #2923, opened 2019-12-05
Virt controller alerts

danielBelenky:
In order to allow our users to monitor their kubevirt deployment,
we add a set of default alert rules which will be picked by the
monitoring operator in OKD (supported in 4.x and up).
    
Because it's not enough to have virt-controller up but we need
one of them to win the leader election, we expose a dedicated
metric that marks the controller that won the election and
another metric that indicates that the controller is ready to
step in and replace the current leader.

This PR is rebased on top of #2780 , so you see unrelated files in this PR. #2780 should be merged first as it adds support for alert rules.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/virt-controller/watch/application.go
+++ b/pkg/virt-controller/watch/application.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/emicklei/go-restful"
 	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	flag "github.com/spf13/pflag"
 	k8sv1 "k8s.io/api/core/v1"
@@ -74,6 +75,20 @@ const (
 
 var (
 	containerDiskDir = filepath.Join(util.VirtShareDir, "/container-disks")
+
+	leaderGauge = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "leading_virt_controller",
+			Help: "Indication for an operating virt-controller.",
+		},
+	)
+
+	readyGauge = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "ready_virt_controller",
+			Help: "Indication for a virt-controller that is ready to take the lead.",
+		},
+	)
 )
 
 type VirtControllerApp struct {
@@ -140,6 +155,11 @@ type VirtControllerApp struct {
 }
 
 var _ service.Service = &VirtControllerApp{}
+
+func init() {
+	prometheus.MustRegister(leaderGauge)
+	prometheus.MustRegister(readyGauge)
+}
 
 func Execute() {
 	var err error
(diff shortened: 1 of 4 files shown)


pkg/virt-controller/watch/application.go
```diff
@@ -74,6 +75,20 @@ const (
...
+		prometheus.GaugeOpts{
+			Name: "leading_virt_controller",
+			Help: "Indication for an operating virt-controller.",
+		},
+	)
+
+	readyGauge = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "ready_virt_controller",
+			Help: "Indication for a virt-controller that is ready to take the lead.",
```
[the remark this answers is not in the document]
danielBelenky: Ahh, you're talking about something different.
I think that the `leader` is still `ready`...
Otherwise, it will make the alert complex: in case you have just one controller running, we'll have to check that there is at least 1 ready to not fire the alert.

pkg/virt-controller/watch/application.go
```diff
@@ -303,6 +323,8 @@ func (vca *VirtControllerApp) Run() {
 						vca.vmControllerThreads, vca.migrationControllerThreads, vca.evacuationControllerThreads,
 						vca.disruptionBudgetControllerThreads)
 
+					leaderGauge.Set(1)
```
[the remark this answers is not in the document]
rmohr: Does not matter since the above calls are async.

pkg/virt-operator/creation/components/crds.go
```diff
@@ -331,6 +331,98 @@ func NewPrometheusRuleSpec(ns string) *promv1.PrometheusRuleSpec {
...
+					{
+						Alert: "VirtControllerDown",
+						Expr:  intstr.FromString("num_of_running_virt_controllers == 0"),
+						For:   "5m",
+						Annotations: map[string]string{
+							"summary": "No running virt-controller was detected for the last 5 min.",
+						},
+					},
+					{
+						Alert: "SingleVirtController",
```
rmohr: I guess this is now also redundant when looking at `LowVirtControllersCount` which seems to be an improved version of `SingleVirtController`.

pkg/virt-operator/creation/components/crds.go
```diff
@@ -331,6 +331,98 @@ func NewPrometheusRuleSpec(ns string) *promv1.PrometheusRuleSpec {
...
+					{
+						Alert: "VirtControllerRESTErrorsHigh",
+						Expr:  intstr.FromString("(vec_by_virt_controllers_failed_client_rest_requests_in_last_hour / vec_by_virt_controllers_all_client_rest_requests_in_last_hour) >= 0.05"),
+						For:   "5m",
+						Annotations: map[string]string{
+							"summary": "More than 5% of the rest calls failed in virt-controller for the last hour",
+						},
+					},
+					{
+						Alert: "VirtControllerRESTErrorsBurst",
```
rmohr: This is an excellent error. We had issues in clusters where the pods were fine but for unclear reasons outside of the scope of virt-controller it could not reach the control plane.

tests/infra_test.go
```diff
@@ -175,6 +175,47 @@ var _ = Describe("Infrastructure", func() {
 			metricsURL = fmt.Sprintf("https://%s:%d/metrics", pod.Status.PodIP, 8443)
 		})
 
+		It("should find one leading virt-controller and two ready", func() {
```
danielBelenky: @rmohr I know that a better approach would be to add a unit-test that checks that the controller actually increments/decrements the corresponding metric with a mock but it would require writing something similar to the framework we have for virt-operator but for virt-controller... I think that we should do it in the future, but it will require much more work and I don't think it should be done in this PR.

dhiller: Would you mind open an issue so that we don't forget the task?

danielBelenky: Done.

PR #2934, opened 2019-12-12
Packaged SELinux policy

stu-gott:
Packaging rules for virt-launcher's SELinux policy.

**Release note**:
```release-note
Packaged virt-launcher's SELinux policy as an RPM.
```


stu-gott:
hack/build-rpms.sh is still a WIP -- running into environment issues.

cmd/virt-handler/kubevirt_container.cil
```diff
@@ -0,0 +1,10 @@
+(block kubevirt_container
```
vladikr: Just wondering. Did you change the name to `kubevirt_container` because the `container` policy already exists?
I left this one to match the `container` definition in [1]
Perhaps we can somehow import it from [1] at some point... 

[1] https://github.com/containers/udica/blob/master/udica/templates/base_container.cil

cmd/virt-handler/virt_launcher.cil
```diff
@@ -8,7 +8,6 @@
     (allow process sysfs_t (file (write)))
     (allow process tmp_t (dir (write add_name open getattr setattr read link search remove_name reparent lock ioctl)))
     (allow process tmp_t (file (setattr open read write create getattr append ioctl lock)))
-    (allow process container_share_t (dir (write add_name)))
-    (allow process container_share_t (file (create setattr write )))
```
vladikr: I remember we've discussed this, so I'm just curious.
Don't these have any effect or do we re-label the directories?

vladikr:
:+1: overall it looks good to me. I have just one question about the `kubevirt_container` policy.

cmd/virt-handler/kubevirt_container.cil
stu-gott: yes, exactly. "container" comes from udica, so we would be unable to install the RPM on systems with that.

cmd/virt-handler/virt_launcher.cil
stu-gott: I was unable to find any directories labelled container_share_t. If that does occur in situations I'm unaware of, then the correct course of action is to relabel because container_share_t is never supposed to be write-able.

cmd/virt-handler/kubevirt_container.cil
vladikr: oh, ok :)

stu-gott:
no changes with the last push--I simply forgot to sign the last commit.

rmohr:
@stu-gott can this be closed?

stu-gott:
We have chosen a different direction entirely. This PR is not needed.

PR #2945, opened 2019-12-20
virt-operator should honor image prefix from previous version

stu-gott:
While investigating a fix for this BZ:

https://bugzilla.redhat.com/show_bug.cgi?id=1782979

I observed that virt-operator will store `imagePrefix` in the deployment config, but then won't reference that value when performing an update (and can end up looking for the previous version in the wrong place as a result).

**Release note**:
```release-note
Fix virt-operator to honor imagePrefix of previous image when performing an update.
```


pkg/virt-operator/util/config.go
```diff
@@ -127,8 +130,33 @@ func GetTargetConfigFromKV(kv *v1.KubeVirt) *KubeVirtDeploymentConfig {
...
-	return getConfig(kv.Status.ObservedKubeVirtRegistry, kv.Status.ObservedKubeVirtVersion, kv.Namespace, getKVMapFromSpec(kv.Spec))
+func GetObservedConfigFromKV(kv *v1.KubeVirt) (*KubeVirtDeploymentConfig, error) {
+	additionalProperties := getKVMapFromSpec(kv.Spec)
+
+	imagePrefix, found, err := getImagePrefixFromDeploymentConfig(kv.Status.ObservedDeploymentConfig)
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to load observed config from kubevirt custom resource: %v", err)
+	}
+	if found {
```
slintes: I'm wondering if we can/should always write this to the map? :thinking: 
That would prevent having to use the workaround described in the bz.
Otherwise nice PR, I totally forgot that we already know the prefix when we discussed this :+1:

stu-gott: Thanks Marc! imagePrefix has been written to the deploymentConfig ever since it was introduced of course. If we choose to make the value a first class citizen in it's own right (which is fine!) we'll still need the ability to pull the value from the saved config as a fallback for the foreseeable future.

I suggest ImagePrefix in the custom resource should be a pointer to a string, that way we can distinguish between "there isn't a value recorded for this (because the virt-operator that wrote this is too old)" and "the prefix is the empty string".

because this PR gives us just enough leverage to address the BZ in the description, I wonder if your suggestion makes sense as a follow-up PR.

slintes: > imagePrefix has been written to the deploymentConfig ever since it was introduced 

yes, and that is why I think that it is save to say: if it wasn't found, it is just empty, and can be used as just being empty.
Or in other words: we can __always__ use imagePrefix from observed deploymentConfig for getting the InstallStrategy of the observed version, instead of making the switch between found prefix and the prefix of the current running operator.
With the bz in mind: you would not have to manually add an empty prefix to the observed deploymentConfig in order to test updates from prod env (no prefix) to test env (with prefix).
Feel free to ping me via direct email in case I can further help to get this out to QE during shutdown.

stu-gott: Thanks! that observation makes a lot of sense. I agree that's the best approach.

slintes:
this should do it :+1: 

/lgtm

rmohr:
@stu-gott some unit tests are broken.

stu-gott:
I agree with the updated approach. Thanks for looking at that @slintes!

slintes:
> I agree with the updated approach. Thanks for looking at that @slintes!

yw
/lgtm

PR #2949, opened 2019-12-25
Add bump kubevirtci script

oshoval:
Script would update hack/config-default.sh,
and then its possible to run make generate in order to fetch
newest kubevirtci cluster-up folder

```release-note
NONE
```

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -148,6 +148,9 @@ Typical cases where code regeneration should be triggered are:
 
  We have a check in our CI system, so that you don't miss when `make generate` needs to be called.
 
+ * Another case is when kubevirtci is updated, in order to vendor cluster-up run `hack/bump-kubevirtci.sh` and then
+   `make generate`
+
 ### Testing
 
 After a successful build you can run the *unit tests*:
--- a/hack/bump-kubevirtci.sh
+++ b/hack/bump-kubevirtci.sh
@@ -1,0 +1,4 @@
+#/bin/bash
+val=$(git ls-remote https://github.com/kubevirt/kubevirtci | grep HEAD | awk '{print $1}')
+sed -i "/^[[:blank:]]*kubevirtci_git_hash[[:blank:]]*=/s/=.*/=\"${val}\"/" hack/config-default.sh
+git --no-pager diff hack/config-default.sh | grep "kubevirtci_git_hash"


alonSadan:
looks good to me

oshoval:
@rmohr 
updated according request, thx

oshoval:
> I suppose you forgot to run `make generate` and commit the changes afterwards somehow. Therefore the Travis build complains.
> 
> /lgtm cancel

tried make generate, it didnt change a thing locally
(its logical because i didnt touch any existing files, just added new one in this PR)
rebased, hope it would help somehow

oshoval:
@dhiller 
travis passed, rebase fixed it, thanks

was great i think if we could bypass sometimes all those lanes, when the fix is something like adding new standalone fix / README and such ?

dhiller:
> was great i think if we could bypass sometimes all those lanes, when the fix is something like adding new standalone fix / README and such ?

We can override the job results manually if needed with [`/override <test-lane>`](https://prow.k8s.io/command-help) AFAIR

Other than that the tests act as a safety net, so I wouldn't recommend it in general, just as you already mentioned only for small changes in README, docs etc.

dhiller:
> > was great i think if we could bypass sometimes all those lanes, when the fix is something like adding new standalone fix / README and such ?
> 
> We can override the job results manually if needed with [`/override <test-lane>`](https://prow.k8s.io/command-help) AFAIR
> 
> Other than that the tests act as a safety net, so I wouldn't recommend it in general, just as you already mentioned only for small changes in README, docs etc.

We could also automate this by using `run_if_changed: <regex>` but I fear that the expression would be really unreadable. Haven't looked at it in detail, though.

oshoval:
> > > was great i think if we could bypass sometimes all those lanes, when the fix is something like adding new standalone fix / README and such ?
> > 
> > 
> > We can override the job results manually if needed with [`/override <test-lane>`](https://prow.k8s.io/command-help) AFAIR
> > Other than that the tests act as a safety net, so I wouldn't recommend it in general, just as you already mentioned only for small changes in README, docs etc.
> 
> We could also automate this by using `run_if_changed: <regex>` but I fear that the expression would be really unreadable. Haven't looked at it in detail, though.

yep , maybe we just need a flag /override all, that even will be ran manually only just when  we feel right, (for example by one of the maintainers) ?

oshoval:
for sure not connected, can we override please? or run retest ?

pull-kubevirt-e2e-k8s-1.16.2
---

Summarizing 1 Failure:

[Fail] Configurations [rfe_id:898][crit:medium][vendor:cnv-qe@redhat.com][level:component]New VirtualMachineInstance with all supported drives [It] [test_id:1020]should not create the VM with wrong PCI adress 
/root/go/src/kubevirt.io/kubevirt/tests/vmi_configuration_test.go:1569

Ran 407 of 485 Specs in 11406.697 seconds
FAIL! -- 406 Passed | 1 Failed | 1 Pending | 77 Skipped
--- FAIL: TestTests (11406.72s)

PR #2952, opened 2019-12-26
virt-operator to enforce namespace labels

danielBelenky:

As part of the default KubeVirt installation, we deploy a set of alert rules for OpenShift's cluster-monitoring system to pick. This process requires a specific label on our namespace to be set.
This PR gives `virt-operator` the ability to patch our namespace and set the required labels.

**Release note**:

```release-note
virt-operator is now enforcing labels on the deployment namespace.
```

### DIFF
--- a/pkg/controller/virtinformers.go
+++ b/pkg/controller/virtinformers.go
@@ -41,6 +41,8 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 
+	corev1 "k8s.io/api/core/v1"
+
 	kubev1 "kubevirt.io/client-go/api/v1"
 	"kubevirt.io/client-go/kubecli"
 	"kubevirt.io/client-go/log"
@@ -159,6 +161,9 @@ type KubeInformerFactory interface {
 	// Fake ServiceMonitor informer used when Prometheus is not installed
 	DummyOperatorServiceMonitor() cache.SharedIndexInformer
 
+	// The namespace where kubevirt is deployed in
+	Namespace() cache.SharedIndexInformer
+
 	K8SInformerFactory() informers.SharedInformerFactory
 }
 
@@ -222,6 +227,22 @@ func (f *kubeInformerFactory) getInformer(key string, newFunc newSharedInformer)
 	f.informers[key] = informer
 
 	return informer
+}
+
+func (f *kubeInformerFactory) Namespace() cache.SharedIndexInformer {
+	return f.getInformer("namespaceInformer", func() cache.SharedIndexInformer {
+		lw := cache.NewListWatchFromClient(f.clientSet.CoreV1().RESTClient(), "namespaces", k8sv1.NamespaceAll, fields.Everything())
+		return cache.NewSharedIndexInformer(
+			lw,
+			&corev1.Namespace{},
+			f.defaultResync,
+			cache.Indexers{
+				"namespace_name": func(obj interface{}) ([]string, error) {
+					return []string{obj.(*corev1.Namespace).GetName()}, nil
+				},
+			},
+		)
+	})
 }
 
 func (f *kubeInformerFactory) VMI() cache.SharedIndexInformer {
--- a/pkg/virt-operator/install-strategy/create.go
+++ b/pkg/virt-operator/install-strategy/create.go
@@ -1511,6 +1511,58 @@ func addOrRemoveSSC(targetStrategy *InstallStrategy,
 	return nil
 }
 
+func syncKubevirtNamespaceLabels(kv *v1.KubeVirt, stores util.Stores, clientset kubecli.KubevirtClient) error {
+
+	targetNamespace := kv.ObjectMeta.Namespace
+	obj, exists, err := stores.NamespaceCache.GetByKey(targetNamespace)
+	if err != nil {
+		log.Log.Errorf("Failed to retrieve kubevirt namespace from store. Error: %s", err.Error())
+		return err
+	}
+	if !exists {
+		return fmt.Errorf("Could not find namespace in store. Namespace key: %s", targetNamespace)
+	}
+
+	cachedNamespace := obj.(*corev1.Namespace)
+	// Prepare namespace metadata patch
+	targetLabels := map[string]string{
+		"openshift.io/cluster-monitoring": "true",
+	}
+	cachedLabels := cachedNamespace.ObjectMeta.Labels
+	labelsToPatch := make(map[string]string)
+	for targetLabelKey, targetLabelValue := range targetLabels {
+		cachedLabelValue, ok := cachedLabels[targetLabelKey]
+		if ok && cachedLabelValue == targetLabelValue {
+			continue
+		}
+		labelsToPatch[targetLabelKey] = targetLabelValue
+	}
+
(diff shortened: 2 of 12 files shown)


vatsalparekh:
Apart from the `go.mod` change, Looks fine to me

go.mod
```diff
@@ -91,3 +86,5 @@ replace (
 	k8s.io/client-go => k8s.io/client-go v0.0.0-20190228174230-b40b2a5939e4
 	kubevirt.io/client-go => ./staging/src/kubevirt.io/client-go
 )
+
+go 1.13
```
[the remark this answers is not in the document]
danielBelenky: :+1:  I'll remove it. Thanks

pkg/virt-operator/install-strategy/create.go
```diff
@@ -1509,6 +1509,61 @@ func addOrRemoveSSC(targetStrategy *InstallStrategy,
 	return nil
 }
 
+func syncKubevirtNamespaceLabels(kv *v1.KubeVirt,
+	stores util.Stores,
+	clientset kubecli.KubevirtClient,
+	expectations *util.Expectations) error {
```
enp0s3: expectations passed but not used in the function

pkg/virt-operator/install-strategy/create.go
```diff
@@ -1509,6 +1509,61 @@ func addOrRemoveSSC(targetStrategy *InstallStrategy,
...
+
+	targetNamespace := kv.ObjectMeta.Namespace
+	obj, exists, err := stores.NamespaceCache.GetByKey(targetNamespace)
+	if err != nil {
+		log.Log.Errorf("Failed retrieving kubevirt namespace from store. Error: %s", err.Error())
```
enp0s3: Consider using "Failed to retrieve kubevirt namespace..."

pkg/virt-operator/install-strategy/create.go
danielBelenky: Nice catch. Fixed :+1:

pkg/virt-operator/install-strategy/create.go
```diff
@@ -1511,6 +1511,58 @@ func addOrRemoveSSC(targetStrategy *InstallStrategy,
...
+	labelsPatch, err := json.Marshal(labelsToPatch)
+	if err != nil {
+		log.Log.Errorf("Failed to marshal namespace labels: %s", err.Error())
+		return err
+	}
+
+	log.Log.Infof("Patching namespace %s with %s", targetNamespace, labelsPatch)
+	_, err = clientset.CoreV1().Namespaces().Patch(
+		targetNamespace,
+		types.MergePatchType,
```
rmohr: I think that this will replace all labels just with the target labels: https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/#use-a-json-merge-patch-to-update-a-deployment. The only sane way to patch is to either use strategic merge patch (if that is supported the right way on labels) or using a json patch with `test` and `replace`.

enp0s3: @rmohr Can you please elaborate on how can I determine whether a resource supports strategic merge patch vs test+replace?

rmohr: test/replace is always supported. Here an example where we use it: https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-controller/watch/vmi.go#L393

rmohr: for strategic merge, probably best to just try it out ... Otherwise the openapi spec should also contain hints.

rmohr: but I am not really sure. Maybe the merge patch works exactly like we want on maps. Just check.

danielBelenky: @rmohr JSON merge patch works in this case because the labels are a dictionary. If we were patching arrays we would need a strategic merge patch instead.

pkg/controller/virtinformers.go
```diff
@@ -224,6 +229,22 @@ func (f *kubeInformerFactory) getInformer(key string, newFunc newSharedInformer)
 	return informer
 }
 
+func (f *kubeInformerFactory) Namespace() cache.SharedIndexInformer {
+	return f.getInformer("namespaceInformer", func() cache.SharedIndexInformer {
+		lw := cache.NewListWatchFromClient(f.clientSet.CoreV1().RESTClient(), "namespaces", k8sv1.NamespaceAll, fields.Everything())
```
omeryahud: If you are patching the namespace, why would you care about all of its fields?
Can you not just select the namespace's labels?

staging/src/kubevirt.io/client-go/go.mod
```diff
@@ -40,3 +40,5 @@ replace (
 	k8s.io/client-go => k8s.io/client-go v0.0.0-20190228174230-b40b2a5939e4
 	kubevirt.io/containerized-data-importer => kubevirt.io/containerized-data-importer v1.10.6
 )
+
+go 1.13
```
omeryahud: Should this stay here? I think it is added by the IDE

danielBelenky: I need to disable it :/ thanks.

pkg/controller/virtinformers.go
danielBelenky: The `fields` selector allows us to filter resources we want to watch by different fields. For example, watch just namespaces with a certain label applied, or a namespace with a certain name. We will store the entire object anyway.

omeryahud: I see, thanks for the explanation!

pkg/virt-operator/kubevirt_test.go
```diff
@@ -1257,6 +1279,8 @@ var _ = Describe("KubeVirt Operator", func() {
 			addKubeVirt(kv)
 			customConfig := getConfig(defaultConfig.GetImageRegistry(), "custom.tag")
 
+			fakeNamespaceModificationEvent()
```
omeryahud: Why did you trigger fakeNamespaceModificationEvent and shouldExpectNamespacePatch in multiple tests instead of testing it in a separate test?

danielBelenky: The different tests are testing different scenarios. You can see the description in the `It("...")` blocks.

danielBelenky: The different tests are testing different scenarios. You can see the description in the It("...") blocks.

omeryahud: Does it really matter if the operator works asynchronously?
Unless I'm missing something, it looks like you're testing the same functionality in multiple tests.

pkg/virt-operator/kubevirt_test.go
```diff
@@ -1336,6 +1362,8 @@ var _ = Describe("KubeVirt Operator", func() {
 			addPodsAndPodDisruptionBudgets(defaultConfig)
 			makeApiAndControllerReady()
 			makeHandlerReady()
+			fakeNamespaceModificationEvent()
```
rmohr: I am always a little bit cautious with controllers. Can we have one unit tests where we check that we don't patch if the label is already present? That shows that we do definitely not hot-looping which can sometimes be hard to track.

PR #2954, opened 2020-01-02
remove @ahadas from code-reviewers

ahadas:
Remove @ahadas  from the list of core reviewers.

**Release note**:

```release-note
NONE
```


rmohr:
@ahadas the generated code blocked the merge since it failed because of the new year switch. Should go in now sooner or later.

PR #2827, opened 2019-10-21
Guest agent subresource

petrkotas:
**What this PR does / why we need it**:

Add support for returning guest agent information from the VM.

The guest agent data is returned from the api endpoint
```
/apis/subresources.kubevirt.io/v1alpha3/namespaces/{namespace}/virtualmachineinstances/{name}/guestosinfo
```

Guest agent poller in virt-launcher has been modified to support multiple pollers at once. The data is stored in cache (`sync.Map`) object to allow for multiple queries without stressing guest agent.
Data is poller is queried in following groups:

* sys data: INTERFACES, OSINFO, TIMEZONE, HOSTNAME; queried every: 120s
* Agent: AGENT_VERSION; queried every: 300s
* Filesystem: FILESYSTEM; queried every: 300s
* User list: USERS: queried every: 10s

Intervals for groups can be changed via virt-launcher params: 
* `qemu-agent-sys-interval`, 
* `qemu-agent-file-interval`, 
* `qemu-agent-user-interval`, 
* `qemu-agent-version-interval` 

Practicaly it should not be necessary. Timeouts were chosen from oVirt to be the same as the default ones on oVirt clusters. 

Virt-handler now has new cmdServer command to call launcher to query guest agent data.

Virt-api has new subresource endpoint to query virt-handler for the guest agent data.

The data same looks like:
```json
{
    "fsInfo": {
        "disks": [
            {
                "diskName": "vda1",
                "fileSystemType": "ext4",
                "mountPoint": "/",
                "totalBytes": 0,
                "usedBytes": 0
            }
        ]
    },
    "guestAgentVersion": "2.11.2",
    "hostname": "testvmi6m5krnhdlggc9mxfsrnhlxqckgv5kqrwcwpgr5mdpv76grrk",
    "metadata": {
        "creationTimestamp": null
    },
    "os": {
        "id": "fedora",
        "kernelRelease": "4.18.16-300.fc29.x86_64",
        "kernelVersion": "#1 SMP Sat Oct 20 23:24:08 UTC 2018",
        "machine": "x86_64",
        "name": "Fedora",
        "prettyName": "Fedora 29 (Cloud Edition)",
        "version": "29 (Cloud Edition)",
        "versionId": "29"
    },
    "timezone": "UTC, 0"
}
```

Items FSInfo and UserList are capped to max capacity of 10 items, as a precaution for VMs with thousands of users. The support for full query will be added in follow PR.

**Special notes for your reviewer**:

**Release note**:

```release-note
[...]

PR #2956, opened 2020-01-02
Refactor agentpoller to allow further commands

petrkotas:

Introduce concept of multiple pollers for agent.
Pollers can be simply added as additional worker and configured separately.

Introduce async cache for storing the agent data, this cache can be polled without penalty multiple times from different goroutines.

The goal of the cache is to enable async storage from multiple pollers.
It also enabled async access for future guest agent subresource. 
The subresource will directly access cached data and will not require 
waiting for the poller.

/cc @phoracek 

This is the preparation PR for introducing the guest agent data as subresource.
Original PR: #2827 

**Release note**:

```release-note
NONE
```


petrkotas:
@phoracek I did the refactoring as you requested, would you mind looking at this PR? Thank you.

petrkotas:
@stu-gott would you please take a look at the refactoring PR? It would really help. Thank you.

pkg/virt-launcher/virtwrap/agent-poller/agent_poller_test.go
```diff
@@ -29,15 +29,77 @@ import (
...
-			agentPoller := AgentPoller{}
-			agentPoller.domainData = &DomainData{
-				// net2 only present in DomainData
-				aliasByMac: map[string]string{"0a:58:0a:f4:00:51": "ovs", "02:00:00:b0:17:66": "net1", "02:11:11:b0:17:66": "net2"},
-			}
+			// agentPoller := AgentPoller{}
+			// agentPoller.domainData = &DomainData{
+			// 	// net2 only present in DomainData
+			// 	aliasByMac: map[string]string{"0a:58:0a:f4:00:51": "ovs", "02:00:00:b0:17:66": "net1", "02:11:11:b0:17:66": "net2"},
+			// }
```
vatsalparekh: forgot to remove comments from here?

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"regexp"
+
+	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
+)
+
+// GuestOsInfo is the response from 'guest-get-osinfo'
+type GuestOsInfo struct {
```
phoracek: Please keep this PR limited to change from simple to parallel poller. Are these new structures essential for the PR?

pkg/virt-launcher/notify-client/client.go
```diff
@@ -197,8 +197,9 @@ func eventCallback(c cli.Connection, domain *api.Domain, libvirtEvent libvirtEve
 				events <- event
 			}
 		}
+		log.Log.Infof("7) OSINFO IN EVENT CALLBACK: %v", osInfo)
```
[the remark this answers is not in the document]
petrkotas: It is because I find it easier to locate caps in the logs. Thanks, I missed this log.

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
petrkotas: It is only moved from `agent_poller.go`. Not introduced.

pkg/virt-launcher/notify-client/client.go
```diff
@@ -240,11 +242,16 @@ func (n *Notifier) StartDomainNotifier(domainConn cli.Connection, deleteNotifica
...
-			case agentUpdate := <-agentUpdateChan:
-				interfaceStatuses = agentUpdate.InterfaceStatuses
-				guestOsInfo = agentUpdate.OsInfo
-				domainName := agentUpdate.DomainName
-				eventCallback(domainConn, util.NewDomainFromName(domainName, vmiUID), libvirtEvent{}, n, deleteNotificationSent,
+			case agentUpdate := <-agentStore.AgentUpdated:
+				interfaceStatuses = agentUpdate.DomainInfo.Interfaces
+				guestOsInfo = agentUpdate.DomainInfo.OSInfo
+				if domainCache != nil && interfaceStatuses != nil {
+					// do merge here
```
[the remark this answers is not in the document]
petrkotas: Thanks, removed this one.

pkg/virt-launcher/virtwrap/agent-poller/agent_poller_test.go
petrkotas: You are right, deleted!

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
```diff
@@ -0,0 +1,153 @@
...
+	return re.FindStringSubmatch(agentReply)[1]
+}
+
+// parseGuestOSInfo parse agent reply string, extract guest os info
+// and converts the response to API domain guest os info
+func parseGuestOSInfo(agentReply string) api.GuestOSInfo {
+	guestOSInfo := GuestOsInfo{}
+	response := stripAgentResponse(agentReply)
+
+	json.Unmarshal([]byte(response), &guestOSInfo)
```
ksimon1: What about error handling here?

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
```diff
@@ -0,0 +1,153 @@
...
+	return resultInfo
+}
+
+// parseInterfaces parse agent reply string, extract network interfaces
+// and converts the response to API domain list of interfaces
+func parseInterfaces(agentReply string) []api.InterfaceStatus {
+	interfaces := []Interface{}
+	response := stripAgentResponse(agentReply)
+
+	json.Unmarshal([]byte(response), &interfaces)
```
ksimon1: same here

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
petrkotas: Excellent eye, I have totally missed the error check.

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
petrkotas: The same.

petrkotas:
Hi @phoracek @ksimon1 and @vatsalparekh I have made the changes that have been requested. Would you please take another look?

ksimon1:
LGTM

pkg/virt-launcher/virtwrap/agent-poller/agent_poller.go
```diff
@@ -31,265 +30,183 @@ import (
...
+		updated = true
+	}
 
-// IP for json unmarshalling
-type IP struct {
-	IP     string `json:"ip-address"`
-	Type   string `json:"ip-address-type"`
-	Prefix int    `json:"prefix"`
-}
+	s.store.Store(key, value)
```
mhenriks: Another thread could theoretically store the same key (and value) between calls to Load/Store here.  Which would lead to multiple events.  Probably not a big deal but a single mutex around both calls could protect from that.  Not currently an issue because only one poller thread.

pkg/virt-launcher/virtwrap/agent-poller/agent_poller.go
```diff
@@ -31,265 +30,183 @@ import (
...
-}
 
-func (p *AgentPoller) pollQemuAgentForOsInfo(domainName string) (string, error) {
-	cmdResult, err := p.Connection.QemuAgentCommand("{\"execute\":\"guest-get-osinfo\"}", domainName)
-	return cmdResult, err
-}
+	// this have to be done via configuration passed in, for now this is OK
+	p.workers = append(p.workers, PollerWorker{
+		CallTick:      qemuAgentPollerInterval,
+		AgentCommands: []AgentCommand{GET_INTERFACES, GET_OSINFO},
```
mhenriks: Would it make sense to have these as a parameter to `CreatePoller`?

I also assume that if there are ever multiple pollers, they will not be handling the same commands?

pkg/virt-launcher/virtwrap/agent-poller/agent_poller_test.go
```diff
@@ -29,15 +29,71 @@ import (
...
+					IPs:           []string{"1.2.3.4/24", "fe80::ff:1111:2222/64"},
+					InterfaceName: "eth5",
+				})
+			Expect(interfaceStatuses).To(Equal(expectedStatuses))
+		})
+
+		It("should merge QEMU info and agent info", func() {
+			// agentPoller := AgentPoller{}
+
+			json_input := "{\"return\":[{\"name\":\"lo\",\"ip-addresses\":[{\"ip-address-type\":\"ipv4\",\"ip-address\":\"127.0.0.1\",\"prefix\":8},{\"ip-address-type\":\"ipv6\",\"ip-address\":\"::1\",\"prefix\":128}],\"hardware-address\":\"00:00:00:00:00:00\"},{\"name\":\"eth0\",\"ip-addresses\":[{\"ip-address-type\":\"ipv4\",\"ip-address\":\"10.244.0.81\",\"prefix\":24},{\"ip-address-type\":\"ipv6\",\"ip-address\":\"fe80::858:aff:fef4:51\",\"prefix\":64}],\"hardware-address\":\"0a:58:0a:f4:00:51\"},{\"name\":\"eth1\",\"ip-addresses\":[{\"ip-address-type\":\"ipv6\",\"ip-address\":\"fe80::ff:feb0:1766\",\"prefix\":64}],\"hardware-address\":\"02:00:00:b0:17:66\"}, {\"name\":\"eth5\",\"ip-addresses\":[{\"ip-address-type\":\"ipv4\",\"ip-address\":\"1.2.3.4\",\"prefix\":24},{\"ip-address-type\":\"ipv6\",\"ip-address\":\"fe80::ff:1111:2222\",\"prefix\":64}],\"hardware-address\":\"02:00:00:22:11:11\"}]}"
```
dhiller: This is really hard to read. Being able to read this as plain multiline json without lots of escaped double quotes would really help me to have a more clear view of what's going on here. Maybe you could convert the test data into a multi line string?

pkg/virt-launcher/virtwrap/agent-poller/agent_poller_test.go
```diff
@@ -87,5 +142,51 @@ var _ = Describe("Qemu agent poller", func() {
...
+					Version:       "1.0.0",
+					PrettyName:    "TestGuestOSName 1.0.0",
+					VersionId:     "1.0.0",
+					KernelVersion: "1.1.0",
+					Machine:       "x86_64",
+					Id:            "testguestos",
+				}
+
+				agentStore.Store(GET_OSINFO, fakeInfo)
+				Expect(agentStore.AgentUpdated).To(Receive(Equal(AgentUpdatedEvent{
```
dhiller: This check is redundant (see above). Do you do this because the channel has to be polled? Or can we just skip the event somehow?

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
```diff
@@ -0,0 +1,162 @@
...
+					Name: alias,
+				},
+			)
+		}
+	}
+	return interfaceStatuses
+}
+
+// calculateInterfaceStatusesFromAgentJson does the conversion from agent info to api domain interfaces
+func calculateInterfaceStatusesFromAgentJson(agentResult []Interface) []api.InterfaceStatus {
```
dhiller: nit: this is just a conversion, so the method should be named `convert...`

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
```diff
@@ -0,0 +1,162 @@
...
+		log.Log.Errorf("Cannot parse interface reply: %s", err.Error())
+		return []api.InterfaceStatus{}
+	}
+
+	resultInterfaces := calculateInterfaceStatusesFromAgentJson(interfaces)
+
+	return resultInterfaces
+}
+
+// MergeAgentStatusesWithDomainData merge QEMU interfaces with agent interfaces
```
dhiller: nit: s/merge/merges/g

pkg/virt-launcher/virtwrap/agent-poller/agent_poller.go
```diff
@@ -31,265 +30,183 @@ import (
...
-	Name          string `json:"name"`
-	KernelRelease string `json:"kernel-release"`
-	Version       string `json:"version"`
-	PrettyName    string `json:"pretty-name"`
-	VersionId     string `json:"version-id"`
-	KernelVersion string `json:"kernel-version"`
-	Machine       string `json:"machine"`
-	Id            string `json:"id"`
+// NewAgentAstore creates new agent store
+func NewAgentAstore() AgentStore {
```
dhiller: Why is it `AgentAstore`? My guess: you mean `AsyncAgentStore`? Then either you should write `NewAgentAStore`, but `NewAsyncAgentStore` would be more clear.

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
```diff
@@ -0,0 +1,162 @@
...
+
+// parseGuestOSInfo parse agent reply string, extract guest os info
+// and converts the response to API domain guest os info
+func parseGuestOSInfo(agentReply string) api.GuestOSInfo {
+	guestOSInfo := GuestOsInfo{}
+	response := stripAgentResponse(agentReply)
+
+	err := json.Unmarshal([]byte(response), &guestOSInfo)
+	if err != nil {
+		log.Log.Errorf("Cannot parse guestosinfo reply: %s", err.Error())
```
dhiller: If the error was returned, this would enable better unit testing by being able to check for error conditions.

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
```diff
@@ -0,0 +1,162 @@
...
+
+// parseInterfaces parse agent reply string, extract network interfaces
+// and converts the response to API domain list of interfaces
+func parseInterfaces(agentReply string) []api.InterfaceStatus {
+	interfaces := []Interface{}
+	response := stripAgentResponse(agentReply)
+
+	err := json.Unmarshal([]byte(response), &interfaces)
+	if err != nil {
+		log.Log.Errorf("Cannot parse interface reply: %s", err.Error())
```
dhiller: If the error was returned, this would enable better unit testing by being able to check for error conditions.

pkg/virt-launcher/virtwrap/agent-poller/agent_poller.go
dhiller: Hmm, wouldn't that actually be what we want? I mean, if multiple changes arise one after another, we wouldn't want to skip a change, or would we? Asked another way, would we want to prohibit changes occuring nearly at the same time, and if so, why?

mhenriks: I think we may want to prevent getting two consecutive events for the same key+value.  And that is currently possible.

pkg/virt-launcher/virtwrap/agent-poller/agent_poller_test.go
petrkotas: This check have to be present. The second poll is required to see whether the event has not been fired, and this is done via additional poll.

pkg/virt-launcher/virtwrap/agent-poller/agent_poller_test.go
petrkotas: You are right, fixed that in the whole test file. Thanks for pointing it up, I was so used to it, that it did not occurred to me.

pkg/virt-launcher/virtwrap/agent-poller/agent_poller.go
petrkotas: Enabling the full configuration does not make sense. There will be only 4 polling workers. Each worker have its own special group (inherited from oVirt). Next is taken from PR: #2827 

* sys data: INTERFACES, OSINFO, TIMEZONE, HOSTNAME; queried every: 120s
* Agent: AGENT_VERSION; queried every: 300s
* Filesystem: FILESYSTEM; queried every: 300s
* User list: USERS: queried every: 10s

Enabling user to configure the choice of distribution does not make sense.
The calls are expensive and timeouts are chosen by trial in oVirt to not stress guest agent.

Also. These pollers are mutually exclusive and does not affect each other. This is the reason, why I selected the `async.Map`. Taken from the documentation:
```
The Map type is optimized for two common use cases: 
(1) when the entry for a given key is only ever written once 
but read many times, as in caches that only grow, or 
(2) when multiple goroutines read, write, and overwrite entries for
disjoint sets of keys. In these two cases, use of a Map may 
significantly reduce lock contention compared to 
a Go map paired with a separate Mutex or RWMutex. 
``` 
Case present in this PR and PR: #2827 is case (2).

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
```diff
@@ -0,0 +1,160 @@
...
+					Mac:  mac,
+					Name: alias,
+				},
+			)
+		}
+	}
+	return interfaceStatuses
+}
+
+// calculateInterfaceStatusesFromAgentJson does the conversion from agent info to api domain interfaces
```
dhiller: Please adjust method name here also.

pkg/virt-launcher/virtwrap/agent-poller/agent_poller_test.go
```diff
@@ -28,16 +28,161 @@ import (
 
 var _ = Describe("Qemu agent poller", func() {
 	Context("recieving a reply from the agent", func() {
+		JSONInput := `{
```
dhiller: Much better! :+1:

pkg/virt-launcher/virtwrap/agent-poller/agent_poller.go
petrkotas: Generating two consecutive call require two poll workers with the same configuration polling the same guest agent endpoint. Multiple pollers with the same guest agent endpoint is not the goal. There will be multiple workers, indeed. However each one will have different set of guest agent endpoints, and each will have different timeout.

There will be 4 workers, taken from PR: #2827 

* sys data: INTERFACES, OSINFO, TIMEZONE, HOSTNAME; queried every: 120s
* Agent: AGENT_VERSION; queried every: 300s
* Filesystem: FILESYSTEM; queried every: 300s
* User list: USERS: queried every: 10s

Enabling user to configure the choice of distribution does not make sense.
The calls are expensive and timeouts are chosen by trial in oVirt to not stress guest agent.

Furthermore timeouts can be tuned in PR: #2827 via parameter to virt-launcher, but I would suggest to not touch it.

The rule of thumb is, timeouts are selected to reflect how often the data changes on the host.

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
```diff
@@ -0,0 +1,162 @@
...
+	IP     string `json:"ip-address"`
+	Type   string `json:"ip-address-type"`
+	Prefix int    `json:"prefix"`
+}
+
+// stripAgentResponse use regex to strip the wrapping item and returns the
+// embeded object.
+// It is a workaround so the amount of copy paste code is limited
+func stripAgentResponse(agentReply string) string {
+	re := regexp.MustCompile(`{\s*\"return\":\s*([{\[][\s\S]*[}\]])\s*}`)
```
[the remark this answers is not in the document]
petrkotas: Thanks, fixed that.

petrkotas:
Thank you @mhenriks and @dhiller for thorough review. It helped a lot to polish the details. I believe I have addressed the comments. Once you have a time, would be able to do a quick look again? Thank you!

pkg/virt-launcher/virtwrap/agent-poller/agent_poller_test.go
dhiller: I understand that the polling is required. I'm just complaining about the duplicate check. Well, that's just nitpicking so nevermind.

petrkotas:
Regarding my IDE, it is VSCode and it does not have spell check by default. However, thanks to you @dhiller I searched for an extension and I have a spellcheck.

pkg/virt-launcher/virtwrap/agent-poller/agent_poller.go
```diff
@@ -31,265 +30,182 @@ import (
...
+// AgentUpdatedEvent fire up when data is changes in the store
+type AgentUpdatedEvent struct {
+	Type       AgentCommand
+	DomainInfo api.DomainGuestInfo
 }
 
-// Result for json unmarshalling
-type Result struct {
-	Interfaces []Interface `json:"return"`
+// AgentStore stores the agent data converted to api domain objects
```
dhiller: Please adjust comment also

pkg/virt-launcher/virtwrap/agent-poller/agent_poller.go
```diff
@@ -31,265 +30,182 @@ import (
...
-type ParsedGuestOsInfo struct {
-	Name          string `json:"name"`
-	KernelRelease string `json:"kernel-release"`
-	Version       string `json:"version"`
-	PrettyName    string `json:"pretty-name"`
-	VersionId     string `json:"version-id"`
-	KernelVersion string `json:"kernel-version"`
-	Machine       string `json:"machine"`
-	Id            string `json:"id"`
+// NewAgentAstore creates new agent store
```
dhiller: Please adjust comment.

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
```diff
@@ -0,0 +1,160 @@
...
+					Name: alias,
+				},
+			)
+		}
+	}
+	return interfaceStatuses
+}
+
+// calculateInterfaceStatusesFromAgentJson does the conversion from agent info to api domain interfaces
+func convertInterfaceStatusesFromAgentJson(agentResult []Interface) []api.InterfaceStatus {
```
dhiller: Another nit: s/Json/JSON/

pkg/virt-launcher/virtwrap/agent-poller/agent_parser.go
```diff
@@ -0,0 +1,160 @@
...
+			Mac:           ifc.MAC,
+			Ip:            interfaceIP,
+			IPs:           interfaceIPs,
+			InterfaceName: ifc.Name,
+		})
+	}
+	return interfaceStatuses
+}
+
+func extractIps(ipAddresses []IP) (string, []string) {
```
dhiller: nit: s/Ips/IPs/

pkg/virt-launcher/notify-client/client.go
```diff
@@ -177,7 +177,7 @@ func eventCallback(c cli.Connection, domain *api.Domain, libvirtEvent libvirtEve
 			domain.Spec = *spec
 		}
 
-		log.Log.Infof("kubevirt domain status: %v(%v):%v(%v)", domain.Status.Status, status, domain.Status.Reason, reason)
+		log.Log.V(5).Infof("kubevirt domain status: %v(%v):%v(%v)", domain.Status.Status, status, domain.Status.Reason, reason)
```
dhiller: I understand that default for verbosity is 2. Why do you reduce the log verbosity here? I think that in general it would help debugging this in case anything goes wrong here?

pkg/virt-launcher/notify-client/client.go
```diff
@@ -254,7 +256,6 @@ func (n *Notifier) StartDomainNotifier(domainConn cli.Connection, deleteNotifica
 	}()
 
 	domainEventLifecycleCallback := func(c *libvirt.Connect, d *libvirt.Domain, event *libvirt.DomainEventLifecycle) {
-		log.Log.Infof("DomainLifecycle event %d with reason %d received", event.Event, event.Detail)
```
dhiller: Why is the logging removed here?

pkg/virt-launcher/notify-client/client.go
```diff
@@ -272,7 +273,7 @@ func (n *Notifier) StartDomainNotifier(domainConn cli.Connection, deleteNotifica
 	}
 
 	agentEventLifecycleCallback := func(c *libvirt.Connect, d *libvirt.Domain, event *libvirt.DomainEventAgentLifecycle) {
-		log.Log.Infof("GuestAgentLifecycle event state %d with reason %d received", event.State, event.Reason)
```
dhiller: Why is the verbosity reduced here?

petrkotas:
Ouu. Thanks @dhiller the logs you discovered were actually casualties of my debug log cleanup. I switched to robot phase and just deleted or raised verbosity for every log I touched.

Thanks!

PR #2958, opened 2020-01-03
No namespaced owners on non-namespaces resources

rmohr:

The operator does not need the owner references for anything. Stop
setting them on all objects.

The background why we do this change is that we were setting some owner
references on non-namespaces objects and the owner was namespaced. This
is not allowed and can result in strange side effects.

Fixes #

**Release note**:

```release-note
Avoid unexpected uninstalls in certain scenarios because of wrong owner references
```


dankenigsberg:
Please take care of the upgrade path, too. We need to drop the owner of it happens to already exist due to an older installation.

rmohr:
> Please take care of the upgrade path, too. We need to drop the owner of it happens to already exist due to an older installation.

Yep.

rmohr:
> Please take care of the upgrade path, too. We need to drop the owner of it happens to already exist due to an older installation.

Addressed. :+1:

dankenigsberg:
Thanks. Before this pr, the CRDs you tweak here would have been deleted on deletion of the KubeVirt cr.

This pr changes this. Can you add a word about why it's fine?

rmohr:
> Thanks. Before this pr, the CRDs you tweak here would have been deleted on deletion of the KubeVirt cr.
> 

They still get deleted and it is expected that they get deleted. We only remove invalid owner references to avoid collisions with the garbage collector.

pkg/virt-operator/install-strategy/create.go
```diff
@@ -131,8 +128,13 @@ func createLabelsAndAnnotationsPatch(objectMeta *metav1.ObjectMeta) ([]string, e
 	if err != nil {
 		return ops, err
 	}
+	ownerRefBytes, err := json.Marshal(objectMeta.OwnerReferences)
+	if err != nil {
+		return ops, err
+	}
 	ops = append(ops, fmt.Sprintf(`{ "op": "add", "path": "/metadata/labels", "value": %s }`, string(labelBytes)))
 	ops = append(ops, fmt.Sprintf(`{ "op": "add", "path": "/metadata/annotations", "value": %s }`, string(annotationBytes)))
+	ops = append(ops, fmt.Sprintf(`{ "op": "add", "path": "/metadata/ownerReferences", "value": %s }`, ownerRefBytes))
```
danielBelenky: If I understand it correctly, we're now setting `ownerReferences` to `null` for objects w/o the reference (which are all our objects now...) wouldn't it cause NPE in K8S' validation?
https://github.com/kubernetes/kubernetes/blob/de1cfa9bc16a08116b8f6b56929d3b1ea39fd1e3/staging/src/k8s.io/apimachinery/pkg/api/validation/objectmeta.go#L83

rmohr: @danielBelenky first it is serialized and deserialized, second it is possible to iterate over a `nil` list. It has the same behaviour like iterating over an empty list. So there are no issues to expect. But nice that you looked up the validation code :+1:

rmohr:
> In general this PR does not have changes to unit tests that I would expect. So I guess there are none. Maybe we can add a unit test to document the behaviour?

@dhiller added tests. :+1:

tests/operator_test.go
```diff
@@ -1138,6 +1140,40 @@ spec:
...
+			crd, err := virtClient.ExtensionsClient().ApiextensionsV1beta1().CustomResourceDefinitions().Get("virtualmachineinstances.kubevirt.io", metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(crd.ObjectMeta.OwnerReferences).To(HaveLen(0))
+		})
+
+		It("should remove owner references on non-namespaces resources when updating a resource", func() {
+			By("adding an owner reference")
+			origCRD, err := virtClient.ExtensionsClient().ApiextensionsV1beta1().CustomResourceDefinitions().Get("virtualmachineinstances.kubevirt.io", metav1.GetOptions{})
+			crd := origCRD.DeepCopy()
+			crd.OwnerReferences = []metav1.OwnerReference{*metav1.NewControllerRef(&v1.KubeVirt{ObjectMeta: metav1.ObjectMeta{Name: "kubevirt", UID: "a185f8c3-3f38-4b89-a8cc-80f3731f7ff9"}}, v1.KubeVirtGroupVersionKind)}
```
dhiller: Is the UID a real ID or some randomly generated one?

rmohr: A random UID.

PR #2975, opened 2020-01-09
Pull in a rules_go version wich works with ginkgo

rmohr:

Pull in a rules_go version wich works reliable with ginkgo and resolve
grpc dependency issues.

The pulled in rules_go version includes https://github.com/bazelbuild/rules_go/pull/2331. Without this fix, some go files were never linked to the test binary and some ginkgo unit tests were never executed.

We can now switch to bazel as the default unit test runner. As a next step the unit tests can be run in a prow job and the bazel cache will also be used by them.

Fixes #

**Release note**:

```release-note
NONE
```


hack/kubevirt-builder/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM kubevirt/builder@sha256:0cc33e4e2a04aa4ec44a9c38c37ff926a8b2ca67181200e31c48298bff79c4b2
+FROM kubevirt/builder@sha256:e7935549daa12e222696a2e9715e820ab6ef5b1dd0f109431856fb144928b497
```
dhiller: I can't follow here. Why are we updating the builder image when all we change there is `VERSION` in the `version.sh`?

rmohr: because I rebuilt and pushed the builder image to update bazel to something >= 1.2.0.

dhiller: That was what I was suspecting, therefore also the `.bazelversion` file for `bazelisk` now. I just found it odd that a change in bazel version was not reflected in the builder Dockerfile.

dhiller: But IIRC we agreed on just using the latest version when building new images, right?

dhiller: Maybe we should also create post commit jobs that build and push the images?

rmohr: > Maybe we should also create post commit jobs that build and push the images?

hm, the issue is that we normally need it in the PR already.

PR #2976, opened 2020-01-10
Add uninstall strategies to virt-operator

rmohr:

It is surprising for some that if you delete kubevirt, all related workloads (VMs, VMIs, VMIRSs) are deleted too. Especially when removing the operator via the marketplace UI it was not clear for people that this can happen.

This PR adds a safeguard to the install process, by introducing a `spec.uninstallStrategy` field. If unspecified or set to `RemoveWorkloads`, the uninstall process force-deletes just everything, like it was the case before. If set to `BlockUninstallIfWorkloadsExist`, a validating webhook denies the deletion of the KubeVirt CR until either all workloads are removed from the cluster explicitly, or the `uninstallStrategy` is changed.

Fixes #

**Release note**:

```release-note
Add `spec.uninstallStrategy` to the KubeVirt CRD, to optionally deny uninstall requests for as long as kubevirt workloads are in the cluster.
```


rmohr:
/cc @stu-gott 
/cc @dhiller 

could you have a look?

/cc @slintes if you find the time any feedback would be valuable

rmohr:
The webhook is right now always created. I would need feedback if we should undeploy if we don't have the corresponding uninstall policy set. The background is that a DELETE webhook ties us even deeper into the k8s cluster: If a webhook wants to be informed about deletes, it also is involved when namespaces are removed (delete requests to all resources in the namespace)

So we will effectively block namespace deletes, even if no kubevirt resource is in the namespace, just because the CRD is defined in the cluster.

fabiand:
Let me try to grasp it: The webhook is a risk to cluster operations, as it needs our operator, thus it is desirable to only have it enabled if needed.

Thus:

1. It sounds like we should install the webhook only if it's needed (if we require "zero CRs")
2. We should remove the webhook whenever KubeVirt is getting uninstalled

To me this sounds like a few conditions are needed, but it sounds like a safe approach.

fabiand:
Because this PR actually looks good, I'd propose that we defer the more fine grained webhook handling.

pkg/virt-operator/application.go
```diff
@@ -297,3 +351,58 @@ func (app *VirtOperatorApp) AddFlags() {
...
 	app.AddCommonFlags()
 }
+
+func (app *VirtOperatorApp) getSelfSignedCert() error {
+	var ok bool
+
+	caKeyPair, _ := triple.NewCA("kubevirt.io")
+	keyPair, _ := triple.NewServerKeyPair(
+		caKeyPair,
+		"kubevirt-operator-webhook."+app.operatorNamespace+".pod.cluster.local",
```
slintes: I see we use this at other places as well, so I would not block on this for this PR, but be aware that `cluster.local` is a cluster setting which can be modified at cluster creation. A starting point would be to at least use a const everywhere (e.g. from `pkg/util/net/dns/resolveconf.go`). And in a next step find out how to get the actual domain from the running cluster.

rmohr: That is a very good point. I wonder if it can also be looked up somewhere from within the cluster. I guess it would be at least possible to read it from the pods resolv.conf.

Thanks for looking.

rmohr: Created a follow up: https://github.com/kubevirt/kubevirt/issues/2998

pkg/virt-operator/application.go
```diff
@@ -297,3 +351,58 @@ func (app *VirtOperatorApp) AddFlags() {
...
+		},
+		Type: "Opaque",
+		Data: map[string][]byte{
+			certBytesValue:        cert.EncodeCertPEM(keyPair.Cert),
+			keyBytesValue:         cert.EncodePrivateKeyPEM(keyPair.Key),
+			signingCertBytesValue: cert.EncodeCertPEM(caKeyPair.Cert),
+		},
+	}
+	_, err := app.clientSet.CoreV1().Secrets(app.operatorNamespace).Create(secret)
+	if errors.IsAlreadyExists(err) {
```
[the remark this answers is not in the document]
rmohr: We had the before. The background is that we avoid this way race-condition. Otherwise we would have to do GET, CREATE (here we are again racing with others), GET.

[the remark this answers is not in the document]
dhiller: Makes sense. Thanks for the explanation!

rmohr:
/hold cancel

thx for the reviews.

PR #2980, opened 2020-01-13
Add QAT feature support for Kubevirt

leyao-daily:
Add QAT test file and the example file for k8s to apply.

Do some code changes to enable QAT vf devices added to Kubevirt vms. The device is assigned by pci-passthrough and get the device id from intel/qat-device-plugin with dpdk mode. And then it creates appropriate hostdev devices in libvirt VM xml and start a vm with QAT vf devices.
It will make use of the intel QAT device and accelerates and compresses cryptographic workloads by offloading the data to hardware capable of optimizing those functions.

 `fixes #<2965>`

**Release note**:

```release-note
Make Kubevirt support the QAT device. It can request QAT vf devices from QAT device plugin and assign it to a Kubevirt vf.
```


rmohr:
@leyao-daily thanks for the PR. Very nice. I am not sure if I will be able to review it this week.

 @phoracek @danielBelenky or @dhiller does one of you have some capacity?

phoracek:
@rmohr what is the policy while testing this kind of enhancements? Do we require real end to end tests with the real hardware? That may be problematic. For the record, I'm fine with hardware-less tests and leaving real testing to the community.

rmohr:
Is there a way to test with something like emulated QAT devices? It might also be sufficient to test with a device which can be emulated and has similar characteristics regarding to pass-through ...

leyao-daily:
> Is there a way to test with something like emulated QAT devices? It might also be sufficient to test with a device which can be emulated and has similar characteristics regarding to pass-through ...

Test the whole function of QAT is complex, to test it, we need build dpdk and use its own test case in a kubevirt vm with QAT vfs. And we also need build VPP IPSec to test the connection using QAT vfs. Maybe there is a way to emulate such devices, but I have no idea. I have QAT devices and the base testings are successful.

fabiand:
@mhenriks @phoracek please help to get this PR reviewed - it's a great community contribution.

automation/test.sh
```diff
@@ -1,3 +1,28 @@
+Skip to content
```
rmohr: @leyao-daily I think that was not intentinal.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -1007,6 +1008,17 @@ func Convert_v1_VirtualMachine_To_api_Domain(vmi *v1.VirtualMachineInstance, dom
 		}
 	}
 
+	// Append HostDev to libvirt DomXML if QAT is required
+	if util.IsQATVMI(vmi) {
```
rmohr: This looks like at the end we only do ordinary device passthrough ...

Maybe we could test it with a mock pci device which gets added to the pod instead of the QAT device? This way we could in the future test all pic passthrough devices.

What do others think?

automation/test.sh
leyao-daily: sorry. solved.

pkg/virt-launcher/virtwrap/api/converter.go
leyao-daily: Yes, It is an ordinary device passthrough: get the QAT vf device id and add it to libvirt domain as a hostdev.

fabiand:
Would it actually be possible/useful to then change the PR to cover PCI
devices in general?

That might also be easier to test ...

leyao-daily:
> Would it actually be possible/useful to then change the PR to cover PCI devices in general? That might also be easier to test ...

I think the gap is that k8s gets the device id from specific device plugin and the resource name it advertise. When changing it in general, it is hard to know which device plugin should k8s ask for when a pod need specific device.

phoracek:
Maybe the API should be based around the resource name? To keep it in the Kubernetes API. One would request arbitrary resource exposed by a device plugin. It would not work with all exposed resources, but as long as users request only compatible devices it should be fine. One thing I not sure about is how to limit the access. AFAIK there is no security mechanism limiting user to access only a subset of resources, is there?

Similar to the network API, you would request a resource (specifying the name exposed by DP) and then set which address should it be mounted to.

rmohr:
> Would it actually be possible/useful to then change the PR to cover PCI devices in general? That might also be easier to test .

That would probably the right path. I think we can add a generic device plugin which exposes artibrary resources and always backs it by a dummy PCI device, created via the `pci_endpoint_test` module: https://www.kernel.org/doc/html/latest/PCI/endpoint/pci-test-howto.html

leyao-daily:
> > Would it actually be possible/useful to then change the PR to cover PCI devices in general? That might also be easier to test .
> 
> That would probably the right path. I think we can add a generic device plugin which exposes artibrary resources and always backs it by a dummy PCI device, created via the `pci_endpoint_test` module: https://www.kernel.org/doc/html/latest/PCI/endpoint/pci-test-howto.html

It may be a pretty good way. But I found this pci_endpoint is only supported by linux kernel version 5.2.0+?

leyao-daily:
I have finished tested on my loca server with QAT cards.
root@ubuntu:~# lspci -d:37c8
0000:3d:00.0 Co-processor: Intel Corporation C62x Chipset QuickAssist Technology (rev 04)
0000:3f:00.0 Co-processor: Intel Corporation C62x Chipset QuickAssist Technology (rev 04)
0000:da:00.0 Co-processor: Intel Corporation C62x Chipset QuickAssist Technology (rev 04)

And test assigning 0, 1 and more QAT vf to a Kubevirt VMI, which are all passed.

I compiled dpdk and ran the QAT tests in the Kubevirt vm, all test cases passed.

leyao-daily:
@rmohr @jerry7z @stu-gott 
Is there any more advices to push this PR or make it suitable to be merged?  QAT is in the OpenNESS backlog and this patch will be utilized in OpenNESS if merged then.
Please review and label it to make the tide and ci test continue running. Thanks a lot.

leyao-daily:
@mhenriks @phoracek What else should I do to complete the PR process?

leyao-daily:
@vladikr So I cloud change "qat" to a generic device description and change this PR as a generic PCI device passthrough API, then maybe we can assign a generic pci device to VM like:
```
...
spec:
  domain:
    devices:
      disks:
      - disk:
          bus: virtio
        name: containerdisk
      **pcidevices:
      - deviceName: xxxx.xxx.xx/qatxxx
        name: qat
     - deviceName: xxxx.xxx.xx/other-pci-device
        name: generic-device**
      rng: {}
    cpu:
...
```
The gaps I think are listed:
1. In kubevirt, it gets the device id by getEnvAddressListByPrefix().. and different pci deive has differnet Prefix, this can not be generic.
2.In such arch, like spec.domain.devices.pcidevices...it may be confusing to request 2 and more pci devices into VM. How to make the workload work 
...

fabiand:
@leyao-daily thank you for your contribution!

It would be really great if we can make this PR generic, and I think that @vladikr is willing and has the time to collaborate with you on the design of the API and sorting out the (great!) issues you already raised.
Would oyu be willing to do a quick round on this to a) find a generic approach and b) see that we finish this on a short timeline?

leyao-daily:
Fine. According to design of Kubevirt and Device Plugin, I think the most important point is to get the device ID. In Kubevirt, we can get the ID from the container environment variables. There, the question is that we need find a generic approach to tell Kubevirt the prefix of the variable name.
So in the VMI configuration, maybe the format can be:
```
...
spec:
  domain:
    devices:
      ...
      pcidevices:
      - deviceName: xxxx.xxx.xx/qatxxx
        envName: xxxx
        count: 1 (or more)
        name: qat
     - deviceName: xxxx.xxx.xx/otherxxxx
       envName: xxxxx
       count: 1(or more)
       name: xxxxxx（generic-device）
    - deviceName: ...
      ...
...
```
Then we can get the device ID by getEnvAddressListByPrefix()....

And we store all these device ID in a devID list, we don't need seperate them into specific segments, because they are generic pci device. What we should do is add them to hostdevs in libvirtxml.

@fabiand @vladikr Are there any potential issues in this approach? Thanks.

vladikr:
@leyao-daily Thank you for collaborating with us on this!

When it comes to generalization, I think the main challenge is to correlate the PCI addresses exposed by the device plugin(s) with the API object.
Some devices may require additional configuration, such as setting `rom bar`, or anything else.

With the suggested approach it's not clear to me how the `envName` will get correlated to the env variable set by a device plugin. I think the already existing `name` filed can be used for this purpose, but I still don't see how it solves the problem.

Alternatively, KubeVirt already has a pattern for correlating an API object to a device plugin env variable, that is used with SRIOV. [1]
I'm suggesting that generic PCI passthrough can easily follow this pattern.

In essence, this is a correlation of two sets of env variables, `PCIDEVICE_<resourceName>="address(es),..."` set by the device plugin for a requested device and `KUBEVIRT_RESOURCE_NAME_<API_object_name>=<resourceName>` set by the virt-controller for the virt-launcher[2] (of course the code here will need to be adjusted for the new `pcidevices` items)

Following your example, this make sense to me:
...
spec:
  domain:
    devices:
      ...
      pcidevices:
      - deviceName: xxxx.xxx.xx/qatxxx
        count: 1 (or more)
        name: qat
      - deviceName: xxxx.xxx.xx/otherxxxx
        count: 1(or more)
        name: xxxxxx（generic-device）
	rombar: true
     - deviceName: ...
        ...
...

What do you think?

[1] https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/manager.go#L1015

[2] https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-controller/services/template.go#L820

leyao-daily:
@vladikr Thanks for your suggestion. After a quick look on manager,go, I agree with you, it seems that Kubevirt hold the device ID well. I may do some tests to re-verify the feasibility.

About the additional configuration, the rombar:true makes sense to me. I prefer your example.
Step by step, firstly I will try a generic pci device assignment which is no need for additional configuration.
Then figure out what and how to assign a device with potential additional configuration.

fabiand:
In your example @vladikr  what does count stand for?

leyao-daily:
@fabiand 
The word "count" stands for the number of pci devices we want. If we want 2 qat and 1 fpga, we may configure it as following:
```
spec:
  domain:
    devices:
      ...
      pcidevices:
      - deviceName: xxxx.xxx.xx/qatxxx
        count: 2
        name: qat
      - deviceName: xxxx.xxx.xx/fpgaxxxx
        count: 1
        name: fpga
```

leyao-daily:
@vladikr @fabiand 
Hi, after checking with the API of SRIOV workload in Kubevirt, I find the hard code in its workflow, it use **PCIDEVICE* _INTEL_COM_INTEL_SRIOV as its envVarPrefix to find the device id that sriov device plugin set in the container runtime envrionment. And there is no stardard env name for current device plugins (QAT device plugin set the env name: QAT0), so we should use a envNamePrefix to store this name and send to kubevirt to finish the pci pssthrough. The api should be:
```
...
spec:
  domain:
    devices:
      ...
      pcidevices:
      - deviceName: xxxx.xxx.xx/qatxxx
        envPrefix: QAT
        count: 1 (or more)
        name: qat
     - deviceName: xxxx.xxx.xx/otherxxxx
       envPrefix: xxxxx
       count: 1(or more)
       name: xxxxxx（generic-device）
    - deviceName: ...
      ...
...
```

vladikr:
@leyao-daily Thank you for looking into this.
We already have a `deviceName`: for example `deviceName: intel.com/qat`
Would the device plugin be able to expose a variable name with this deviceName?
Such as `PCIDEVICE_INTEL_COM_QAT`

if this is possible, you can then loop over all `devices.pcidevices` and use the `resourceNameToEnvvar` function[1] to convert the deviceName to an environment variable that the device plugin has created.
Something along these lines:

```
func getGenericPCIDeviceAddresses(pcidevices []v1.PCIDevices) map[string][]string {
	resourceToAddressesMap := map[string][]string{}
	for _, device := range pcidevices {
		resourceToAddressesMap[device.deviceName] = []string{}
		varName := resourceNameToEnvvar(deviceName)
		pciAddrString, isSet := os.LookupEnv(varName)
		if isSet {
		    addrs := strings.Split(pciAddrString, ",")
		    naddrs := len(addrs)
		    if naddrs > 0 {
		         if addrs[naddrs-1] == "" {
				addrs = addrs[:naddrs-1]
			 }
		    }
		    resourceToAddressesMap[device.deviceName] = addrs
		 } else {
			log.DefaultLogger().Warningf("%s not set for PCI device %s", varName, device.deviceName)
	         }
	}
	return resourceToAddressesMap
}
```

What do you think?

leyao-daily:
@vladikr 
Unluckily, different device plugin will expose different env value.
In sriov: PCIDEVICE_Intel_...
In QAT: QAT
In GPU: GPU_PASSTHROUGH_DEVICES
There is no standard name for them. And we need a field to store this envName to find the device id  .

vladikr:
@leyao-daily Yes, I understand that :)
However, we need to establish this standard and I think this is what we are trying to do with this PR.
SRIOV already has a good implementation.
We should now provide a way to plugin other generic devices. After this, we can see how to make changes in the GPU  plugin to follow the new interface.
I think first we need to start with QAT. 
I'm wondering, can the QAT device plugin create a variable name PCIDEVICE_<resource name> ?

Thanks

leyao-daily:
@vladikr In my investigation, there is no standard for device plugin about the env name.
Both in SRIOV and QAT device plugin, the env name is hard code. PCIDEVICE in SRIOV and QAT in QAT.
https://github.com/intel/sriov-network-device-plugin/blob/master/pkg/resources/server.go#L321
https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/cmd/qat_plugin/dpdkdrv/dpdkdrv.go#L44-#L46

So to be more generic, we may need add a env name keyword in spec.domain.devices.pcidevices:
I think it's hard to request all other potential device plugins to set the env name in a standard.  

And about the name of spec.domain.devices.pcidevices, maybe we could change it to pci? 
What do you think? Thanks.

leyao-daily:
root@ubuntu:~# kubectl exec -ti virt-launcher-vmi-qat-nkc7q -c compute env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=vmi-my
TERM=xterm
QAT1=0000:3f:01.3
QAT0=0000:3d:02.4
KUBERNETES_SERVICE_PORT=443
KUBERNETES_SERVICE_PORT_HTTPS=443
...

root@ubuntu:~# kubectl exec -ti virt-launcher-vmi-sriov-tpm27 -c compute env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=vmi-sriov
TERM=xterm
PCIDEVICE_INTEL_COM_INTEL_SRIOV=0000:41:06.6
KUBEVIRT_RESOURCE_NAME_sriov-net=intel.com/intel_sriov
KUBERNETES_SERVICE_PORT_HTTPS=443

vladikr:
> @vladikr In my investigation, there is no standard for device plugin about the env name.
> Both in SRIOV and QAT device plugin, the env name is hard code. PCIDEVICE in SRIOV and QAT in QAT.
> https://github.com/intel/sriov-network-device-plugin/blob/master/pkg/resources/server.go#L321
> https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/cmd/qat_plugin/dpdkdrv/dpdkdrv.go#L44-#L46
> 
> So to be more generic, we may need add a env name keyword in spec.domain.devices.pcidevices:
> I think it's hard to request all other potential device plugins to set the env name in a standard.
> 
> And about the name of spec.domain.devices.pcidevices, maybe we could change it to pci?
> What do you think? Thanks.

@leyao-daily Thanks.
Yes, I understand that the variables are currently may be different. 
I'd prefer to change the device pluggings to name the variables with the same prefix to align with kubevirt.
Just think about how the API user would consume this if he will have to know the variable each device plugin exposes. 

There is no concern with the SRIOV device plugin, it exposes the resources with the correct PCIDEVICE_<resource> prefix so it will be easily discoverable.
How difficult is it to change the variable name in the QAT device plugin? 

Thanks

rmohr:
> Just think about how the API user would consume this if he will have to know the variable each device plugin exposes.

Would it make sense if kubevirt admins can define mappings somewhere? We could assume consistent variable naming in general, but fall back to different names if needed then. My main concern with my own suggestion would be that maybe not only the variable name is different, but also the content format may vary slightly.

leyao-daily:
@vladikr @rmohr 
I agree with @rmohr, because we can not easily change the env name that the device plugin exposed. We may need create PR for each device plugin to change the env name they set.

We can define the mappings somewhere like a guild. And then use the name in VMI spec.domain.device.pci... It make sense to me.
And the contents of the env in my investigation are a standard format xxxx:xx:xx.x

leyao-daily:
```
...
spec:
  domain:
    devices:
      ...
      pcis:
      - deviceName: xxxx.xxx.xx/qatxxx
        envPrefix: QAT
        count: 1 (or more)
        name: qat
     - deviceName: xxxx.xxx.xx/otherxxxx
       envPrefix: xxxxx
       count: 1(or more)
       name: xxxxxx（generic-device）
    - deviceName: ...
      ...
...
```
So I will work on this design. And we can define mappings somewhere. Looking forward to your responses. Thanks.

leyao-daily:
Okay. We can continue dicussing the API design in this PR. And I am looking forward to all your advices.

leyao-daily:
@fabiand @vladikr @rmohr 
Is there any updates on it? Maybe we can discuss in mail or on the comminity meeting if my network runs well. :)

leyao-daily:
@dankenigsberg 
QAT is intel quickassist technology. Intel QuickAssist Technology is developed by Intel and runs on the Intel Architecture to provide security and compression acceleration capabilities to improve performance and efficiency. It will offload the workloads like cryptography and compression from the CPU to hardware. Server, networking, big data, and storage applications use Intel QuickAssist to offload compute-intensive operations, such as:
- Symmetric cryptography functions, including cipher operations and authentication operations
- Public key functions, including RSA, Diffie-Hellman, and elliptic curve cryptography,
- Compression and decompression functions, including DEFLATE
And one particular user case is it can accelerate IPSec. In my test, it double the bandwidth of the network translation. (from 500M to 1.0G)

In detail, it is a hardware that can be a PCIe card or a Chipset on board.
- QAT provides security (encryption) HW acceleration and compression HW acceleration
- QAT makes use of a set of APIs to abstract out the hardware, so the same application can run on multiple generations of QAT hardware
- Customers can also make use of patches that QAT has provided to popular open source software, so they can minimize or eliminate their effort to learn the API

leyao-daily:
And it now can be used by kubernetes pod together with the QAT Device Plugin. When setup QAT DP with DPDK mode and it will exposed the resources to node with resource name: qat.intel.com/generic

Then in pod yaml, we can add to:
```
resources:
      requests:
        cpu: 4000m
        memory: 8192Mi
        qat.intel.com/generic: '2'
      limits:
        cpu: 4000m
        memory: 8192Mi
        qat.intel.com/generic: '2'
```
Then we can utilize the device in POD (accelerate CNF).

gabrielecerami:
@leyao-daily: PR needs rebase.

<details>

</details>

peng-isi:
@leyao-daily,  I am new to kubevirt. Right now,  I am doing something very similar to what you did to support QAT. I followed your method to modify the original source code. The modified code can be compiled. But when I tried to run vmi, I got the following error " The request is invalid: : spec.domain.devices.ibs in body is a forbidden property", where "ib"  is the feature I tried to add to Kubevirt. 

It seems to me that my addition to the "swagger.json" file is not integrated with API.  After modifying the source code, I just run the "make && make push" commands. Do I need to do extra work to make the changes to swagger.json visible to KubeVirt API?

leyao-daily:
> @leyao-daily, I am new to kubevirt. Right now, I am doing something very similar to what you did to support QAT. I followed your method to modify the original source code. The modified code can be compiled. But when I tried to run vmi, I got the following error " The request is invalid: : spec.domain.devices.ibs in body is a forbidden property", where "ib" is the feature I tried to add to Kubevirt.
> 
> It seems to me that my addition to the "swagger.json" file is not integrated with API. After modifying the source code, I just run the "make && make push" commands. Do I need to do extra work to make the changes to swagger.json visible to KubeVirt API?

Some codes are auto-generated by "make generate".
PS: Kubevirt is working on a generic device plugin mechanism now, with which we can easily assign our device into VMs.
https://github.com/kubevirt/kubevirt/pull/3937

peng-isi:
@leyao-daily thanks for your prompt response. I just need to modify the source files and run "make generate"? After that, I run "make && make push".  Please confirm.

leyao-daily:
> @leyao-daily thanks for your prompt response. I just need to modify the source files and run "make generate"? After that, I run "make && make push". Please confirm.

Yes, hope it works :)

peng-isi:
@leyao-daily many thanks!

fabiand:
Hey!
Finally #3937 got merged which is introducing a more generic DP API, please rebase this PR and leverage the new RP, then QAT should be ready to go (roughly).

rmohr:
> Hey!
> Finally #3937 got merged which is introducing a more generic DP API, please rebase this PR and leverage the new RP, then QAT should be ready to go (roughly).

I think it is even better. I think that QAT devices can now be passed through via the generic device-passthrough mechanism from @vladikr and we don't even need explicit code in the project for it. Maybe we can instead add an example to our user-guide on how to pass it through? @vladikr to keep me honest.

leyao-daily:
> Hey!
> Finally #3937 got merged which is introducing a more generic DP API, please rebase this PR and leverage the new RP, then QAT should be ready to go (roughly).

Excellent! I have noticed the patch has been merged some days ago. And I am waiting for the 0.36.0 release, I think the new API will be released in that version :) Meantime I will test the new API with QAT device. Thanks

rmohr:
> Excellent! I have noticed the patch has been merged some days ago. And I am waiting for the 0.36.0 release, I think the new API will be released in that version :) Meantime I will test the new API with QAT device. Thanks

You can use our developer builds for testing: https://kubevirt.io/user-guide/#/installation/installation?id=installing-the-daily-developer-builds. Hope that helps.

leyao-daily:
> > Excellent! I have noticed the patch has been merged some days ago. And I am waiting for the 0.36.0 release, I think the new API will be released in that version :) Meantime I will test the new API with QAT device. Thanks
> 
> You can use our developer builds for testing: https://kubevirt.io/user-guide/#/installation/installation?id=installing-the-daily-developer-builds. Hope that helps.

I will try it. Another question is that it seems in the API, we don't need kubernetes device plugin expose the device info?

rmohr:
> I will try it. Another question is that it seems in the API, we don't need kubernetes device plugin expose the device info?

I think that you only have to add something like this to the KubeVirt CR:

```
configuration:
  permittedHostDevices:
    pciHostDevices:
    - pciVendorSelector: "10DE:1EB8" #real vendor selector here
      resourceName: "intel.com/qat" # name of your liking
```

and you can then attach it like this to your VM:


```
spec:
  domain:
    devices:
      hostDevices:
      - deviceName: intel.com/qat
        name: myquatdevice
```

virt-handler will detect the device, start a device plugin and so on.

vladikr:
> > Hey!
> > Finally #3937 got merged which is introducing a more generic DP API, please rebase this PR and leverage the new RP, then QAT should be ready to go (roughly).
> 
> I think it is even better. I think that QAT devices can now be passed through via the generic device-passthrough mechanism from @vladikr and we don't even need explicit code in the project for it. Maybe we can instead add an example to our user-guide on how to pass it through? @vladikr to keep me honest.

I'm late to the party :)
Exactly. As @rmohr mentioned we shouldn't need any additional code. We can simply specify that the device is allowed in the cluster and virt-handler will start a device plugin for it.

User guide PR is on review: https://github.com/kubevirt/user-guide/pull/368

rmohr:
/close

passing through host devices works now out of the box with kubevirt. Here an example which also includes QAT: https://kubevirt.io/user-guide/virtual_machines/host-devices/#listing-permitted-devices

@leyao-daily as far as I know you also verified that it works together with @vladikr 

Feel free to reopen if that is not the case.

PR #2982, opened 2020-01-13
Make VMI non-migratable when incompatible CPU

petrkotas:
During migration compatible CPU model have to be used.
This is hard when cpu-passthrough or host-model is used
as a choice for CPU.

This issue can be ovecomed in a short term by blocking
migration for host-model or cpu-passthrough.

Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=1760028

**Release note**:

```release-note
NONE
```


petrkotas:
Hi @MarSik would you please have time to take a look at this PR as it is related to the part you have been working on? I understand you have you workload, so once you are OK with it. Thank you.

pkg/virt-handler/vm.go
```diff
@@ -1266,6 +1287,27 @@ func (d *VirtualMachineController) isPreMigrationTarget(vmi *v1.VirtualMachineIn
 	return false
 }
 
+// checkCPUForMigration is a hotfix for bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1760028
+// to disable migration when host-model od cpu-pass-through is used.
+// The check here relies on the fact the first defaulting is done in mutating webhook
+// where the default is set based od cluster-wide config, which still can be empty.
```
danielBelenky: s/od/on

pkg/virt-handler/vm.go
```diff
@@ -1266,6 +1287,27 @@ func (d *VirtualMachineController) isPreMigrationTarget(vmi *v1.VirtualMachineIn
...
+// If no CPU is setup in VMI and the empty CPU gets here it is passed over to virt-launcher.
```
danielBelenky: s/setup in/set on

pkg/virt-handler/vm.go
```diff
@@ -439,36 +439,57 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
...
-			vmi.Status.Conditions = append(vmi.Status.Conditions, liveMigrationCondition)
+			migrate = false
 		}
 
-		// Set VMI Migration Method
-		if isBlockMigration {
-			vmi.Status.MigrationMethod = v1.BlockMigration
-		} else {
-			vmi.Status.MigrationMethod = v1.LiveMigration
+		if migrate {
```
danielBelenky: Just a personal preference here so feel free to keep it as is, but I'd rather checking `if !migrate` and return here. Then, the entire block below will not be indented.
But again, just a personal preference...

staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -233,6 +233,8 @@ const (
 	VirtualMachineInstanceReasonDisksNotMigratable = "DisksNotLiveMigratable"
 	// Reason means that VMI is not live migratioable because of it's network interfaces collection
 	VirtualMachineInstanceReasonInterfaceNotMigratable = "InterfaceNotLiveMigratable"
+	// Reason means that VMI is not live migratioable because of it's network interfaces collection
```
danielBelenky: I think that the comment here is wrong...

pkg/virt-handler/vm.go
petrkotas: @danielBelenky thanks for the review :). I rather would leave the code as is.

petrkotas:
Hi @danielBelenky and @ksimon1 would you have time to do recheck for this PR? I see the tests are failing for following issues, which are unrelated.

```
k8s-1.14.6: [rfe_id:3423][vendor:cnv-qe@redhat.com][level:component]oc/kubectl get vm/vmi tests should verify set of wide columns for [test_id:3422]virtualmachineinstance
k8s-1.16.2: [rfe_id:3423][vendor:cnv-qe@redhat.com][level:component]oc/kubectl get vm/vmi tests should verify set of wide columns for [test_id:3422]virtualmachineinstance
```

pkg/virt-handler/vm.go
```diff
@@ -1266,6 +1287,27 @@ func (d *VirtualMachineController) isPreMigrationTarget(vmi *v1.VirtualMachineIn
 	return false
 }
 
+// checkCPUForMigration is a hotfix for bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1760028
+// to disable migration when host-model od cpu-pass-through is used.
```
dhiller: nit: s/od/or/

pkg/virt-handler/vm_test.go
```diff
@@ -377,6 +377,9 @@ var _ = Describe("VirtualMachineInstance", func() {
 		It("should update from Scheduled to Running, if it sees a running Domain", func() {
 			vmi := v1.NewMinimalVMI("testvmi")
 			vmi.UID = testUUID
+			vmi.Spec.Domain.CPU = &v1.CPU{
```
dhiller: I have seen that new integration tests have been added, which is good. I'd also add new unit tests to make the logic verifiable more easily, so as to i.e. support refactorings without having to run the integration tests to verify.

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1157,53 @@ var _ = Describe("VirtualMachineInstance", func() {
 			Expect(blockMigrate).To(BeTrue())
 			Expect(err).To(Equal(fmt.Errorf("cannot migrate VMI with non-shared HostDisk")))
 		})
+
+		It("should block migrate for empty CPU", func() {
```
dhiller: I'm fine with these tests. :+1:

However, I have two *optional* suggestions:
* you could reduce duplication if you'd convert these tests into a `Table`, maybe adding the `Model` as an additional parameter to the `NewMinimalVMI` method and 
* you should remove the optional description in the `... .To(` clause, because they do not differ from the description in the `It(...`.

staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -233,6 +233,8 @@ const (
 	VirtualMachineInstanceReasonDisksNotMigratable = "DisksNotLiveMigratable"
 	// Reason means that VMI is not live migratioable because of it's network interfaces collection
 	VirtualMachineInstanceReasonInterfaceNotMigratable = "InterfaceNotLiveMigratable"
+	// Reason means that VMI is not live migratioable because of it's CPU configuration
```
dhiller: s/migratioable/migratable/

pkg/virt-handler/vm.go
```diff
@@ -1266,6 +1287,27 @@ func (d *VirtualMachineController) isPreMigrationTarget(vmi *v1.VirtualMachineIn
...
+// The check here relies on the fact the first defaulting is done in mutating webhook
+// where the default is set based on cluster-wide config, which still can be empty.
+// If no CPU is set on VMI and the empty CPU gets here it is passed over to virt-launcher.
+// For empty CPU the default host-model is used when converting VMI to libvirt domain as CPU
+//
+// Therefore the fix checks if the CPU here is either empty, or set to host-model or pass-through
+// and marks VMi as non-migratable if so.
```
dhiller: s/VMi/VMI/

vladikr:
I know I'm late, but I would still like to understand why do we want to block live-migration in this case.

Users who know that their infrastructure is homogeneous and want to take the most advantage of their CPUs, shouldn't even worry about this.
However, with this change, we will block them without providing a way to override this.

Alternatively, we could use libvirt's baselineCPU(), to migrate the VMI with the most common flags. This will prevent the migration from failing in many cases but may reduce the guest's efficiency.
It's not ideal and It's tricky to implement due to our migration model, however, doable.

@fabiand @rmohr ^^

petrkotas:
Hi @vladikr, I understand your doubts about this patch. I can assure you that this is a hotfix (as stated in the comments and in the commit message). The solution you are proposing has indeed been discussed in the attached bugzila: https://bugzilla.redhat.com/show_bug.cgi?id=1760028 and we agreed that it is the best option. However it would take the longest to implement.

The hotfix has been selected as it solved current user issues and proper fix will be done in following releases.

Does this answer your question?

vladikr:
It's not clear to me what are we gaining from blocking the migration? Perhaps a release note or a document would be a better option to explain the current system limitations?

vladikr:
For example, https://docs.openstack.org/nova/rocky/admin/configuration/hypervisor-kvm.html#host-pass-through

petrkotas:
The choice for disabling the migration has been thoroughly discussed in the Bugzilla. 

In short, `host-passthrough` is no-go since migration can fail due to the fact the CPU can differ on each host, migration requires the same CPU otherwise it will fail terribly. The second option, `host-model` makes checking for CPU compatibility hard. Checking for compatibility is not as simple as calling `baselineCPU` since we first need to do a scan of our nodes where kubevirt can be deployed. The solution has already been decided in Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1760028

Summary of the discussion is: it is better to block migration rather than fail it. This saves users their data and workload.

rmohr:
@petrkotas could you add a feature-gate which would allow these types of unsafe migrations and don't allow them by defautl? The name could be something like `AllowUnsafeMigration` or `AllowHostModelCPUMigrations`.

fabiand:
+1 to allow host-model migrations with a featuregate

petrkotas:
Hi @rmohr and @dhiller I believe I have implemented the suggested changes and the tests are passing now. Would you mind having another look please?

tests/kubectl_test.go
```diff
@@ -74,6 +74,6 @@ var _ = Describe("[rfe_id:3423][vendor:cnv-qe@redhat.com][level:component]oc/kub
 		Expect(resultFields[len(resultFields)-verifyPos]).To(Equal(expectedData))
 	},
 		table.Entry("[test_id:3421]virtualmachine", "get", "vm", "wide", []string{"NAME", "AGE", "RUNNING", "VOLUME", "CREATED"}, 1, "true"),
-		table.Entry("[test_id:3422]virtualmachineinstance", "get", "vmi", "wide", []string{"NAME", "AGE", "PHASE", "IP", "NODENAME", "LIVE-MIGRATABLE"}, 1, "True"),
+		table.Entry("[test_id:3422]virtualmachineinstance", "get", "vmi", "wide", []string{"NAME", "AGE", "PHASE", "IP", "NODENAME", "LIVE-MIGRATABLE"}, 1, "False"),
```
dhiller: I can't remember whether I've seen this last time, but it does look like the `expectedData` is comparing case sensitive? So just out of interest: Does the difference matter? I suppose it does? So therefore the test data in the table looks odd in using once `true` (lowercase) and once `False` (uppercase)?

tests/migration_test.go
```diff
@@ -437,6 +454,8 @@ var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system]
 
 			It("[test_id:3237]should complete a migration", func() {
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
+				vmi = makeMigratableVMI(vmi)
```
dhiller: I don't understand: why do we create a new random vm and then do a deep copy and set one field? Why don't we just set the field?

Same goes for the other lines where this is done?

tests/migration_test.go
```diff
@@ -1138,6 +1173,86 @@ var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system]
 
 			})
 		})
+
+		Context("without CPU properly set", func() {
+
+			table.DescribeTable("should block migration with feature-gate CPUMigrationGate enabled", func(cpuModel string) {
```
dhiller: Good use for table here!

nit: I think this one and the following table could be collapsed into one with an additional `boolean` controlling whether to enable or to disable the feature gate. This would further reduce duplicate code I think.

iranzo:
@petrkotas can we get a PR updating the limitations section for live migration when this gets merged?

petrkotas:
Hi, @iranzo sure thing, thank you for pointing the doc to me!

tests/kubectl_test.go
petrkotas: I relate to this finding. I only updated this test to make it pass.
The returning values are indeed inconsistent and it would deserve the fix.

This would be nice followup PR.

dhiller: Ack. Would be great if you created an issue for it so we don't forget.

tests/migration_test.go
petrkotas: Ah thanks @dhiller I forgot about this, it is not required. I copied snippet from different path and edited it. I removed it.

tests/migration_test.go
petrkotas: I agree! I have reduced the code duplicity. I like the two tables more because it is more verbose what it is tested there. Therefore I reduced the duplicity by creating sub-functions.

tests/kubectl_test.go
petrkotas: I created issue #3072 it will make a great starting task.

tests/migration_test.go
```diff
@@ -54,7 +54,7 @@ const (
 	fedoraVMSize      = "256M"
 )
 
-var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system] VM Live Migration", func() {
+var _ = FDescribe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system] VM Live Migration", func() {
```
dhiller: Please remove the focus here.

tests/migration_test.go
```diff
@@ -1138,6 +1173,76 @@ var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system]
...
+				tests.EnableFeatureGate("MigrationCPU")
+
+				vmi, migration := prepareMigration(cpuModel)
+
+				By("Starting a Migration")
+				migration, err = virtClient.VirtualMachineInstanceMigration(migration.Namespace).Create(migration)
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("CPUNotMigratable"))
+
+				disposeVMI(vmi)
```
dhiller: optional nit: put this into an [`AfterEach`](http://onsi.github.io/ginkgo/#extracting-common-setup-beforeeach), and then remove this line and 1238 altogether.

tests/migration_test.go
petrkotas: :man_facepalming: I really need to write checked for focus tests.
Thanks, removed!

tests/migration_test.go
petrkotas: We can do this in further cleaning. The tests deserve more cleanup. I will create issues covering this topic.

tests/migration_test.go
```diff
@@ -1138,6 +1173,76 @@ var _ = Describe("[rfe_id:393][crit:high[vendor:cnv-qe@redhat.com][level:system]
...
+			)
+
+			table.DescribeTable("should not block migration", func(cpuModel string) {
+				tests.DisableFeatureGate("MigrationCPU")
+
+				vmi, migration := prepareMigration(cpuModel)
+
+				By("Starting a Migration")
+				migrationUID := runMigrationAndExpectCompletion(migration, migrationWaitTime)
+				Expect(err).ToNot(HaveOccurred())
```
dhiller: Please remove this line. It's obsolete.

petrkotas:
For the reference https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2982/pull-kubevirt-e2e-k8s-1.15.1-ceph/1232826982346002432

petrkotas:
For the reference https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2982/pull-kubevirt-e2e-k8s-1.15-ceph/1230817132627890176

petrkotas:
For the reference: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/2982/pull-kubevirt-e2e-k8s-1.15.1-ceph/1235002071275540480

davidvossel:
@petrkotas can you reproduce the 1.15-ceph failures locally? If so can you isolate what change in your pr is directly responsible for the failure?

petrkotas:
Hi @davidvossel can I run ceph locally? I am probably not aware of an option to do it. Thank you.

fabiand:
Let's do it as follows:
1. passthrough blocks migraion
2. host model should not (because this is helpful in homogenous clusters where perf is needed)

@petrkotas is this the policy today in this PR?

petrkotas:
Hi @fabiand no, the feature blocks both. It was agreed upon that hostmodel is also problematic. on clusters with different hosts it will result in failures.

fabiand:
Right.
The difference is:

1. host-passthrough MUST be blocked
2. host-model CAN be blocked

Thus, what about changing the PR to

1. always block host-passthrough
2. change feature flag to a kubevirt-config item and allow it to enable or disable live migration with host-model

This allows us to have live migration, even with high performance workloads, the user can opt-in in case that he has a homogenous environment. We'd document this config item in the user-guide.

fabiand:
@vladikr do you have a thought on this?

fabiand:
My 2 ct:
1. Host passthrough should be blocked - no question
2. Hst model has the problem in kubevirt that even if host model is used, kubevrit does not ensure that the same "base model" is used on the target node

Hus for now it's probably safest to block for host model as well.

@vladikr wdyt?

stu-gott:
@fabiand that is indeed tricky. It would have been nice if a mutating webhook could translate a host model CPU into a general CPU model specification, but unfortunately the order is wrong. Host model isn't known until after the virt-launcher pod is scheduled.

From my POV, we probably should use the same rules for both host model and host passthrough.

vladikr:
> @fabiand that is indeed tricky. It would have been nice if a mutating webhook could translate a host model CPU into a general CPU model specification, but unfortunately the order is wrong. Host model isn't known until after the virt-launcher pod is scheduled.

But, migration is happening after the VMI has started, so we could know the model and use this when we migrate.
However, this feature needs to be developed.
We probably shouldn't do it as part of this PR but also I fear that there will a lot of arguments later if we will block the host-model now. 

From my POV, we should block passthrough. Open an issue about migration with host-model and start working on it.
Also, rebase this PR :)

fabiand:
wfm - that actually seems to align with how others like OSP are dealing with it.

stu-gott:
Rebased.

vladikr:
@stu-gott : `runVMIAndExpectLaunchWithIgnoreWarningArg declared and not used`

stu-gott:
Oops! thanks, I overlooked that that function was moved to tests/utils.go

stu-gott:
/hold

Looks like one test might need to be fixed

xpivarc:
/hold
/cc @dhiller  

CI hits timeout the second time.
Is this PR related?

dankenigsberg (changes requested):

pkg/virt-handler/vm.go
```diff
@@ -1782,6 +1817,37 @@ func (d *VirtualMachineController) isPreMigrationTarget(vmi *v1.VirtualMachineIn
...
 
+// checkCPUForMigration is a hotfix for bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1760028
+// to disable migration when cpu-pass-through is used.
+// For cpu-pass-through migration is always blocked
+func (d *VirtualMachineController) checkCPUForMigration(vmi *v1.VirtualMachineInstance) error {
+	if vmi.Spec.Domain.CPU == nil {
+		return fmt.Errorf("cannot migrate VMI with null CPU configuration")
+	}
+	if vmi.Spec.Domain.CPU.Model == v1.CPUModeHostPassthrough {
+		return fmt.Errorf("cannot migrate VMI with current CPU configuration: %s", vmi.Spec.Domain.CPU.Model)
```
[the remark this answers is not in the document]
xpivarc: This currently makes it unable to migrate. Do you think that failed or aborted migration would be better? I think that is something this PR wanted to address.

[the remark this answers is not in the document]
dankenigsberg: My point is that this VM *is* and *should remain* migratable when the source and destination CPUs are precise.
Failed/aborted are bad and should be addressed, but not by debilitating a feature that is important to performance-savvy users. half-baked idea: disallow migration of host-paththrough on non-homogeneous clusters.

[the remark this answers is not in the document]
xpivarc: The best solution would probably take a whole new PR as described by bugzila and PR. I would suggest to feature-gate  `host-paththrough`so admin can choose to allow migrations with `host-paththrough` if he knows that cluster is homogenous. WDYT?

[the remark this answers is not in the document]
dankenigsberg: `migratable-host-pathrough` feature gate looks fine to me, assuming it is exposed in HCO cr.

tests/migration_test.go
```diff
@@ -1320,6 +1358,87 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
...
+				confirmVMIPostMigration(vmi, migrationUID)
+
+				disposeVMI(vmi)
+			},
+				table.Entry("with empty cpu", ""),
+				table.Entry("with host-model cpu", "host-model"),
+			)
+
+			It("should always block migration with cpu-passthrough", func() {
+				tests.DisableFeatureGate("MigrationCPU")
```
xpivarc, line 1429: This dirty context of other tests.

pkg/virt-handler/vm.go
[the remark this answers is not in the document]
stu-gott: The basic problem with this is that "homegenous" is actually really hard to attain. Even this slightest mismatch in CPU flags invalidates passthrough. More info is available in the discussion here: https://bugzilla.redhat.com/show_bug.cgi?id=1760028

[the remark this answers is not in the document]
dankenigsberg: > Even this slightest mismatch in CPU flags invalidates passthrough

Understood. I am fine with (currently) putting the burden of verifying this on the cluster-admin to verify this and set `migratable-host-passthrough`.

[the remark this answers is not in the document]
fabiand: If it's an admin who makes the decision, then it should be on the KubeVirt CR.

If it's a CPU model, then we leave it to the user to do the decision.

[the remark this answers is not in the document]
dankenigsberg: I don't understand you @fabiand .

My only request is to keep allowing live migration of host-passthorough. I am fine if this is allowed only if explicitly requested in HCO CR (KubeVirt CR is not enough).

xpivarc:
/cc @stu-gott @vladikr @davidvossel 
Tests seem to be unrelated. Would appreciate feedback on this(if we want to move with this as is or change something).

pkg/virt-handler/vm.go
```diff
@@ -2278,6 +2248,81 @@ func (d *VirtualMachineController) setVMIGuestTime(vmi *v1.VirtualMachineInstanc
...
+		vmi.Status.Conditions = append(vmi.Status.Conditions, liveMigrationCondition)
+		liveMigrationPossible = false
+	}
+
+	if liveMigrationPossible {
+		liveMigrationCondition := v1.VirtualMachineInstanceCondition{
+			Type:   v1.VirtualMachineInstanceIsMigratable,
+			Status: k8sv1.ConditionTrue,
+		}
+		vmi.Status.Conditions = append(vmi.Status.Conditions, liveMigrationCondition)
```
vladikr, line 2316: Are we sure we want to pile these up? the code is looking for one `VirtualMachineInstanceIsMigratable` condition when we decide to migrate. I don't think it's useful.
These conditions are static and will stay with the VMI for its lifetime. Perhaps we could just expand the `Reason` field or maybe just reason is enough.

vladikr (commented):
I still honestly think that this is not the right solution. Long term we should expose the CPU model and the relevant flags then schedule according to those, instead of blindly blocking. 
Regardless, comment inline.

pkg/virt-handler/vm.go
enp0s3: @vladikr Wouldn't it be useful to know all the reasons why the VM is not live-migratable? It will make me as a user to solve all the issues at once instead of going back and forth. IIUC VMI lifetime will end anyway if an attempt to to make the VMI live-migratable will take place, thus the conditions will reset.

fabiand:
High level comment:
On a homogeneous cluster OR if CPUNodeDiscovery is used, then host-model is safe to migrate.
Thus it must not block migration (speak: not featuregated).

host-passthrough is also migratable, but under much stricter constraints. But because this is required for certain workloads we should allow this, but make it harder to use (speak: featuregated)

pkg/virt-config/feature-gates.go
```diff
@@ -34,6 +34,11 @@ const (
 	SnapshotGate          = "Snapshot"
 	HostDiskGate          = "HostDisk"
 	VirtIOFSGate          = "ExperimentalVirtiofsSupport"
+	// 	CPUMigrationGate allows users to block migrations for virtual machines with host-model
+	CPUMigrationGate = "MigrationCPU"
```
dankenigsberg, line 38: this name is too opaque for me - I cannot tell what it means. How about `MigratableHostModel` ?

dankenigsberg (changes requested):
Please modify the PR text (and summary?) to mention that migration is STILL possible if the feature gates are enabled.

pkg/virt-config/feature-gates.go
enp0s3: @dankenigsberg Completely agree. I haven't started the re-factoring. It was only the rebase. Now I will update the PR code itself according to the discussion

pkg/virt-handler/vm.go
enp0s3: @vladikr @fabiand I can concatenate the error and reason strings into a single string and produce a single condition. Will that be user-friendly enough?

enp0s3:
Due to many changes I would like to close this PR and relate to #4323

PR #2991, opened 2020-01-15
Add kubevirt-builder/Dockerfile.ci

sallyom:
Dockerfile that will work for CI builds
@dhiller 
This PR needs to merge before https://github.com/dhiller/release/pull/2

I recommend merging this then the release PR (that I set the job to `always_run: false` so that to run it you'd `/test e2e` from a test PR here, then once the tests are passing you can submit a PR to release to change to `always_run: true`. 

This is pretty close to what you'll need to run the kubevirt hack/functests.sh in CI
The Dockerfile creates an image that has the tests built, src code - in CI it will be built w/ the PR's src code.  The CI job will use a stock CI template `openshift/release/ci-operator/templates/openshift/installer/cluster-launch-installer-custom-test-image.yaml` where the test image is the image built by this added Dockerfile.  

To Test This:
```console
$ podman pull quay.io/sallyom/kubevirt-test:test  
$ podman run --rm -it -v /path/to/a/local/kubeconfig:/tmp/kubeconfig:Z quay.io/sallyom/kubevirt-test:test sh
then in container:
# export KUBECONFIG=/tmp/kubeconfig
// have to install oc here, but in CI oc,kubectl are installed in the test container
# mkdir /tmp/oc; cd /tmp/oc
# curl -L -O https://mirror.openshift.com/pub/openshift-v4/clients/oc/4.2/linux/oc.tar.gz
# tar xzvf oc.tar.gz
# ln -s /tmp/oc/oc kubectl
# export PATH=$PATH:/tmp/oc/
# cd /go/src/kubevirt.io/kubevirt
# ./hack/ci/test.sh
```
**NOTE:** the image is huge, so after you're done remove it w/ something like: 
podman volume prune
podman system-prune -a 

```release-note
NONE
```

### DIFF
--- a/hack/functests.sh
+++ b/hack/functests.sh
@@ -23,6 +23,8 @@ DOCKER_TAG=${DOCKER_TAG:-devel}
 DOCKER_TAG_ALT=${DOCKER_TAG_ALT:-devel_alt}
 
 export ARTIFACTS=${ARTIFACTS:-_out/artifacts}
+# without this the tests fail because kubeconfig is not set
+export KUBECONFIG=${KUBECONFIG}
 
 source hack/common.sh
 source hack/config.sh
@@ -45,4 +47,4 @@ if [[ ${KUBEVIRT_PROVIDER} == os-* ]] || [[ ${KUBEVIRT_PROVIDER} == okd-* ]]; th
     oc=${kubectl}
 fi
 
-${TESTS_OUT_DIR}/tests.test -kubeconfig=${kubeconfig} -container-tag=${docker_tag} -container-tag-alt=${docker_tag_alt} -container-prefix=${functest_docker_prefix} -image-prefix-alt=${image_prefix_alt} -oc-path=${oc} -kubectl-path=${kubectl} -gocli-path=${gocli} -test.timeout 420m ${FUNC_TEST_ARGS} -installed-namespace=${namespace} -previous-release-tag=${previous_release_tag} -previous-release-registry=${previous_release_registry} -deploy-testing-infra=${deploy_testing_infra}
+${TESTS_OUT_DIR}/tests.test -kubeconfig=${KUBECONFIG} -container-tag=${docker_tag} -container-tag-alt=${docker_tag_alt} -container-prefix=${functest_docker_prefix} -image-prefix-alt=${image_prefix_alt} -oc-path=${oc} -kubectl-path=${kubectl} -gocli-path=${gocli} -test.timeout 420m ${FUNC_TEST_ARGS} -installed-namespace=${namespace} -previous-release-tag=${previous_release_tag} -previous-release-registry=${previous_release_registry} -deploy-testing-infra=${deploy_testing_infra}
--- a/hack/ci/Dockerfile.ci
+++ b/hack/ci/Dockerfile.ci
@@ -1,0 +1,129 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
+
+# Install packages
+# dnf copr enable -y @virtmaint-sig/virt-preview with yum?
+# python3-pip-9.0.3-5.el7.noarch \
+RUN yum install -y yum-plugin-copr && \
+    yum copr enable -y vbatts/bazel epel-7 && \
+    yum -y install \
+        libvirt-devel \
+        bazel \
+        cpio \
+        patch \
+        make \
+        git \
+        mercurial \
+        sudo \
+        gcc \
+        gcc-c++ \
+        glibc-devel \
+        findutils \
+        qemu-img \
+        protobuf-compiler \
+        python3-devel \
+        python3-pip \
+        redhat-rpm-config \
+        jq \
+        unzip && \
+    yum clean all && rm -rf /var/cache/yum/*
+
+ENV GIMME_GO_VERSION=1.12.8
+
+RUN mkdir -p /gimme && curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | HOME=/gimme bash >> /etc/profile.d/gimme.sh
+
+ENV GOPATH="/go" GOBIN="/usr/bin"
+
+# Install persistent go packages
+RUN \
+    mkdir -p /go && \
+    source /etc/profile.d/gimme.sh && \
+    # Install goimports
+    go get -u -d golang.org/x/tools/cmd/goimports && \
+    cd $GOPATH/src/golang.org/x/tools/cmd/goimports && \
+    git checkout release-branch.go1.12 && \
+    go install && \
+    # Install mvdan/sh
+    git clone https://github.com/mvdan/sh.git $GOPATH/src/mvdan.cc/sh && \
+    cd $GOPATH/src/mvdan.cc/sh/cmd/shfmt && \
+    git checkout v2.5.0 && \
+    go get mvdan.cc/sh/cmd/shfmt && \
+    go install && \
+    go get -u -d k8s.io/code-generator/cmd/deepcopy-gen && \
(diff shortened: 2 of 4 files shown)


hack/kubevirt-builder/Dockerfile.ci
```diff
@@ -0,0 +1,131 @@
...
+        protobuf-compiler \
+        python3-devel \
+        python3-pip \
+        redhat-rpm-config \
+        jq \
+        unzip && \
+    yum clean all && rm -rf /var/cache/yum/*
+
+# install gradle
+RUN yum install -y java-1.8.0-openjdk-devel && \
```
dhiller: what do we need gradle for?

cluster-up/hack/config.sh
```diff
@@ -2,7 +2,7 @@ unset docker_prefix master_ip network_provider kubeconfig manifest_docker_prefix
 
 KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-${PROVIDER}}
 
-source ${KUBEVIRTCI_PATH}hack/config-default.sh
+source ${KUBEVIRTCI_PATH}/hack/config-default.sh
```
dhiller: Why is this change required? FYI: The content of the folder `cluster-up` is taken from `kubevirt/kubevirtci`. Any errors should be fixed there.

hack/builder/entrypoint.sh
```diff
@@ -3,5 +3,5 @@ set -e
 set -o pipefail
 
 source /etc/profile.d/gimme.sh
-export GOPATH="/root/go"
+export GOPATH="/go"
```
dhiller: Why is this change required? This file is not used in Dockerfile.ci, is it? Note: it IS used in another Dockerfile.

hack/config.sh
```diff
@@ -2,8 +2,8 @@ unset binaries docker_images docker_tag docker_tag_alt image_prefix image_prefix
     namespace image_pull_policy verbosity \
     csv_version package_name bazel_py
 
-source ${KUBEVIRT_PATH}hack/config-default.sh
-source ${KUBEVIRT_PATH}cluster-up/hack/config.sh
+source hack/config-default.sh
```
dhiller: Why is this change required?

hack/kubevirt-builder/Dockerfile.ci
```diff
@@ -0,0 +1,131 @@
...
+    go get -u github.com/rmohr/go-swagger-utils/swagger-doc && \
+    go get -u github.com/onsi/ginkgo/ginkgo
+
+RUN pip3 install j2cli && pip3 install operator-courier==1.3.0
+
+WORKDIR /tmp/kubevirt
+COPY . .
+
+# update git
+COPY hack/kubevirt-builder/wandisco-git.repo /etc/yum.repos.d/wandisco-git.repo
```
dhiller: why can't we use default git here?

hack/kubevirt-builder/ci-test.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
dhiller: We might integrate this into the Makefile later on or move it into `test/` folder. There is another test script already.

hack/kubevirt-builder/Dockerfile.ci
sallyom: I dunno, it was in the build Dockerfile so I included it here (but have to install like so in centos)  I'll remove it and test, see if we need it

cluster-up/hack/config.sh
sallyom: oh ok!

hack/builder/entrypoint.sh
sallyom: ok, so we don't need entrypoint.sh at all then. I'll remove this change : )

hack/config.sh
sallyom: these paths were wrong when I tested, this cleared up the issue, but I'll test again, make sure it's still needed.

sallyom:
current failure: 
```
Failure [120.886 seconds]
[BeforeSuite] BeforeSuite 
/tmp/kubevirt/tests/tests_suite_test.go:48

  Timed out after 120.001s.
  Both KVM devices and vhost-net devices are required for testing, but are not present on cluster nodes
  Expected
      <bool>: false
  to be true

  /tmp/kubevirt/tests/utils.go:741
```
Is this indicative that the _OpenShift cluster_ is running on hosts where nested virt is not enabled? Or, something else? ie, where/when does this suite pass?  I have tests setup to run in gcp, w/ our default gcp setup.

hack/kubevirt-builder/ci-test.sh
sallyom: for now, I've moved all files necessary for CI to an isolated dir, `hack/ci`

hack/kubevirt-builder/Dockerfile.ci
sallyom: default git is old in Centos 7 (1.8.3.1), so there are build failures when 'git fetch-ing` during `go build`

hack/kubevirt-builder/Dockerfile.ci
sallyom: I've confirmed that gradle is not necessary, I've removed it

dhiller:
@sallyom you need to run `make generate` once and commit the changes. It applies the right format to your script file and makes the Travis build error go away.

dhiller:
Thanks for your great work so far!

/lgtm

hack/functests.sh
```diff
@@ -45,4 +47,4 @@ if [[ ${KUBEVIRT_PROVIDER} == os-* ]] || [[ ${KUBEVIRT_PROVIDER} == okd-* ]]; th
     oc=${kubectl}
 fi
 
-${TESTS_OUT_DIR}/tests.test -kubeconfig=${kubeconfig} -container-tag=${docker_tag} -container-tag-alt=${docker_tag_alt} -container-prefix=${functest_docker_prefix} -image-prefix-alt=${image_prefix_alt} -oc-path=${oc} -kubectl-path=${kubectl} -gocli-path=${gocli} -test.timeout 420m ${FUNC_TEST_ARGS} -installed-namespace=${namespace} -previous-release-tag=${previous_release_tag} -previous-release-registry=${previous_release_registry} -deploy-testing-infra=${deploy_testing_infra}
+${TESTS_OUT_DIR}/tests.test -kubeconfig=${KUBECONFIG} -container-tag=${docker_tag} -container-tag-alt=${docker_tag_alt} -container-prefix=${functest_docker_prefix} -image-prefix-alt=${image_prefix_alt} -oc-path=${oc} -kubectl-path=${kubectl} -gocli-path=${gocli} -test.timeout 420m ${FUNC_TEST_ARGS} -installed-namespace=${namespace} -previous-release-tag=${previous_release_tag} -previous-release-registry=${previous_release_registry} -deploy-testing-infra=${deploy_testing_infra}
```
dhiller: This change breaks the tests somehow. IDK yet why, but maybe we can just remove this here? And then do this in the test script itself before calling this one? Or why wouldn't that work?

hack/functests.sh
```diff
@@ -23,6 +23,8 @@ DOCKER_TAG=${DOCKER_TAG:-devel}
 DOCKER_TAG_ALT=${DOCKER_TAG_ALT:-devel_alt}
 
 export ARTIFACTS=${ARTIFACTS:-_out/artifacts}
+# without this the tests fail because kubeconfig is not set
+export KUBECONFIG=${KUBECONFIG}
```
dhiller: And maybe we can just push this out into the test script also if required?

sallyom:
closing, was merged w/ this: https://github.com/kubevirt/kubevirt/pull/3037 : )

PR #2993, opened 2020-01-16
Changed the logic to only allow specific interface types for migrations

vatsalparekh:

Changed the logic for which interfaces and networks should be allowed for migrations
From not allowing only bridge, changed it to allow only the types we support
Fixes #

**Release note**:

```release-note
Changed the logic for which interfaces and networks should be allowed for migrations
```

### DIFF
--- a/pkg/virt-handler/vm.go
+++ b/pkg/virt-handler/vm.go
@@ -1272,17 +1272,9 @@ func (d *VirtualMachineController) checkNetworkInterfacesForMigration(vmi *v1.Vi
 		networks[network.Name] = network.DeepCopy()
 	}
 	for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
-		if networks[iface.Name].Pod != nil {
-			if iface.Masquerade != nil {
-				continue
-			}
+		if iface.Masquerade == nil && networks[iface.Name].Pod != nil {
+			return fmt.Errorf("cannot migrate VMI which does not use masquerade to connect to the pod network")
 		}
-		if networks[iface.Name].Multus != nil {
-			if iface.Bridge != nil {
-				continue
-			}
-		}
-		return fmt.Errorf("cannot migrate VMI with the given interface & network, only Masquerade with a Pod network allowed for migration")
 	}
 	return nil
 }
--- a/pkg/virt-handler/vm.go
+++ b/pkg/virt-handler/vm.go
@@ -1272,9 +1272,13 @@ func (d *VirtualMachineController) checkNetworkInterfacesForMigration(vmi *v1.Vi
 		networks[network.Name] = network.DeepCopy()
 	}
 	for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
-		if iface.Bridge != nil && networks[iface.Name].Pod != nil {
-			return fmt.Errorf("cannot migrate VMI with a bridge interface connected to a pod network")
+		if networks[iface.Name].Pod != nil {
+			if iface.Masquerade != nil {
+				continue
+			}
+			return fmt.Errorf("cannot migrate VMI with the given interface & network, only Masquerade with a Pod network allowed for migration")
 		}
+		return fmt.Errorf("cannot migrate VMI with the given interface & network, only Masquerade with a Pod network allowed for migration")
 	}
 	return nil
 }
--- a/pkg/virt-handler/vm.go
+++ b/pkg/virt-handler/vm.go
@@ -1276,7 +1276,11 @@ func (d *VirtualMachineController) checkNetworkInterfacesForMigration(vmi *v1.Vi
 			if iface.Masquerade != nil {
 				continue
 			}
-			return fmt.Errorf("cannot migrate VMI with the given interface & network, only Masquerade with a Pod network allowed for migration")
+		}
+		if networks[iface.Name].Multus != nil {
+			if iface.Bridge != nil {
+				continue
+			}
 		}
 		return fmt.Errorf("cannot migrate VMI with the given interface & network, only Masquerade with a Pod network allowed for migration")
 	}
--- a/pkg/virt-handler/vm_test.go
+++ b/pkg/virt-handler/vm_test.go
@@ -1152,7 +1152,7 @@ var _ = Describe("VirtualMachineInstance", func() {
 			Expect(err).To(Equal(fmt.Errorf("cannot migrate VMI with non-shared HostDisk")))
 		})
 
-		It("should allow to migrate with adding default networks", func() {
+		It("should not allow to migrate with adding default network configuration", func() {
 			vmi := v1.NewMinimalVMI("testvmi")
 			err := controller.checkNetworkInterfacesForMigration(vmi)
 			Expect(err).ToNot(HaveOccurred())
@@ -1178,7 +1178,7 @@ var _ = Describe("VirtualMachineInstance", func() {
 			}
 
 			err := controller.checkNetworkInterfacesForMigration(vmi)
-			Expect(err).ToNot(HaveOccurred(), "should block migration for bridging interface")
+			Expect(err).ToNot(HaveOccurred())
 		})
 	})
 	Context("VirtualMachineInstance controller gets informed about interfaces in a Domain", func() {
--- a/pkg/virt-handler/vm_test.go
+++ b/pkg/virt-handler/vm_test.go
@@ -1155,7 +1155,30 @@ var _ = Describe("VirtualMachineInstance", func() {
 		It("should allow to migrate with adding default networks", func() {
 			vmi := v1.NewMinimalVMI("testvmi")
 			err := controller.checkNetworkInterfacesForMigration(vmi)
-			Expect(err).To(BeNil())
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should not block migration for bridge interface assigned to multus network", func() {
+			vmi := v1.NewMinimalVMI("testvmi")
+			interfaceName := "interface_name"
+
+			vmi.Spec.Networks = []v1.Network{
+				v1.Network{
+					Name:          interfaceName,
+					NetworkSource: v1.NetworkSource{Multus: &v1.MultusNetwork{}},
+				},
+			}
+			vmi.Spec.Domain.Devices.Interfaces = []v1.Interface{
+				v1.Interface{
+					Name: interfaceName,
+					InterfaceBindingMethod: v1.InterfaceBindingMethod{
+						Bridge: &v1.InterfaceBridge{},
+					},
+				},
+			}
+
+			err := controller.checkNetworkInterfacesForMigration(vmi)
+			Expect(err).ToNot(HaveOccurred(), "should block migration for bridging interface")
 		})
 	})
 	Context("VirtualMachineInstance controller gets informed about interfaces in a Domain", func() {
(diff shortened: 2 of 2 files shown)


pkg/virt-handler/vm.go
```diff
@@ -1272,9 +1272,10 @@ func (d *VirtualMachineController) checkNetworkInterfacesForMigration(vmi *v1.Vi
 		networks[network.Name] = network.DeepCopy()
 	}
 	for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
-		if iface.Bridge != nil && networks[iface.Name].Pod != nil {
-			return fmt.Errorf("cannot migrate VMI with a bridge interface connected to a pod network")
+		if iface.Masquerade != nil || iface.Bridge != nil {
+			continue
 		}
+		return fmt.Errorf("cannot migrate VMI with a bridge interface connected to a pod network")
```
phoracek: This does not match the logic. If there is a bridge interface, it is skipped `continue` in the previous block. Also, you don't check for the network type `Pod`, so this would be raised for secondary interfaces too.

pkg/virt-handler/vm.go
```diff
@@ -1272,9 +1272,10 @@ func (d *VirtualMachineController) checkNetworkInterfacesForMigration(vmi *v1.Vi
 		networks[network.Name] = network.DeepCopy()
 	}
 	for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
-		if iface.Bridge != nil && networks[iface.Name].Pod != nil {
-			return fmt.Errorf("cannot migrate VMI with a bridge interface connected to a pod network")
+		if iface.Masquerade != nil && networks[iface.Name].Pod != nil {
+			continue
 		}
+		return fmt.Errorf("cannot migrate VMI with the given interface & network, only Masquerade with a Pod network allowed for migration")
```
phoracek: This would return error for secondary (Multus) interfaces with both Masquerade and Bridge. I think you need to have nested ifs:

```
if pod != nil:
    if masquerade:
        continue
    return error
```

Also, would be the default network configured at this point if none was provided by the user? If the user did not specify any network, the list may be empty (unless we add defaults), defaulting to bridge connection to pod.

vatsalparekh: Corrected the nested if and added a test to check this func against a VMI without any networks defined.

rmohr: virt-handler will always see all networks on the VMI. If there is no network, then there is none. The default network is added to the VMI at creation time.

pkg/virt-handler/vm.go
```diff
@@ -1272,9 +1272,13 @@ func (d *VirtualMachineController) checkNetworkInterfacesForMigration(vmi *v1.Vi
 		networks[network.Name] = network.DeepCopy()
 	}
 	for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
-		if iface.Bridge != nil && networks[iface.Name].Pod != nil {
-			return fmt.Errorf("cannot migrate VMI with a bridge interface connected to a pod network")
+		if networks[iface.Name].Pod != nil {
```
rmohr: @vatsalparekh we check if a VMI is migrateable when it gets processed in virt-controller. If it is not migrateable, we add a `VirtualMachineInstanceIsMigratable` condition which is set to false. If that is present creating  migration objects for thes VMI will be rejected.

You will have to move this check to virt-controller.

rmohr: sorry it happens in virt-handler, but anyway, instead of blocking the migration at this place, we have to ensure that it never gets the condition that it is migrateable assigned by virt-handler.

rmohr: Ok, forget me previous comments. I was completely on the wrong track. Still @phoracek is bridge not also used to get a network connection to an external network with its own DHCP? Wouldn't we then not also block migrations for these?

rmohr: Ok, so the code does something different than intended.

pkg/virt-handler/vm.go
rmohr: @vatsalparekh I think this line would return an error for every secondary interface.

vatsalparekh: @rmohr removed the first error msg now

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1151,35 @@ var _ = Describe("VirtualMachineInstance", func() {
 			Expect(blockMigrate).To(BeTrue())
 			Expect(err).To(Equal(fmt.Errorf("cannot migrate VMI with non-shared HostDisk")))
 		})
+
+		It("should allow to migrate with adding default networks", func() {
```
vatsalparekh: @rmohr do you think this is correct and necessary? I'm not sure

phoracek: I don't understand the sentence. Maybe "should allow to migrate with the default network configuration"?
 
Anyway, this should fail. The default binding is bridge and we don't allow to migrate it with that. So there is probably a missing validation step in the code.

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1151,35 @@ var _ = Describe("VirtualMachineInstance", func() {
...
+				v1.Interface{
+					Name: interfaceName,
+					InterfaceBindingMethod: v1.InterfaceBindingMethod{
+						Bridge: &v1.InterfaceBridge{},
+					},
+				},
+			}
+
+			err := controller.checkNetworkInterfacesForMigration(vmi)
+			Expect(err).ToNot(HaveOccurred(), "should block migration for bridging interface")
```
phoracek: you meant to say "should not block migration ..."?

pkg/virt-handler/vm.go
```diff
@@ -1272,9 +1272,17 @@ func (d *VirtualMachineController) checkNetworkInterfacesForMigration(vmi *v1.Vi
...
+			if iface.Masquerade != nil {
+				continue
+			}
+		}
+		if networks[iface.Name].Multus != nil {
+			if iface.Bridge != nil {
```
phoracek: We don't care about binding in secondary network. All should be migratable. You can `continue`, no matter what the binding is.

pkg/virt-handler/vm_test.go
vatsalparekh: Changed this with an additional logic to check if networks and interfaces are not empty.

pkg/virt-handler/vm_test.go
vatsalparekh: Removed

pkg/virt-handler/vm.go
vatsalparekh: Corrected

pkg/virt-handler/vm.go
```diff
@@ -1271,10 +1271,21 @@ func (d *VirtualMachineController) checkNetworkInterfacesForMigration(vmi *v1.Vi
 	for _, network := range vmi.Spec.Networks {
 		networks[network.Name] = network.DeepCopy()
 	}
+	// if no networks or interfaces are found, return error
+	// this can be in the case of a new minimal VMI
+	if len(networks) == 0 || len(vmi.Spec.Domain.Devices.Interfaces) == 0 {
+		return fmt.Errorf("cannot migrate VMI, network or interfaces not found")
```
phoracek: That is not the case. There is the implicitly set configuration with interface connected to the pod network via bridge binding. Please mention in the comment that the default binding (when nothing else is specified) is bridge.

Also, there is also a flag that users can use to really have *no* interfaces at all. Then we should not block the migration. I don't recall what the flag is named though. Maybe @rmohr can help here?

pkg/virt-handler/vm.go
```diff
@@ -1271,10 +1271,21 @@ func (d *VirtualMachineController) checkNetworkInterfacesForMigration(vmi *v1.Vi
...
+	}
 	for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
-		if iface.Bridge != nil && networks[iface.Name].Pod != nil {
```
rmohr: @vatsalparekh how about just this one code change at all:

```go
if iface.Masquerade == nil && networks[iface.Name].Pod != nil {
    return fmt.Errorf("cannot migrate VMI which does not use masquerade to connect to the pod network")
}
```

pkg/virt-handler/vm.go
```diff
@@ -1271,10 +1271,21 @@ func (d *VirtualMachineController) checkNetworkInterfacesForMigration(vmi *v1.Vi
 	for _, network := range vmi.Spec.Networks {
 		networks[network.Name] = network.DeepCopy()
 	}
+	// if no networks or interfaces are found, return error
+	// this can be in the case of a new minimal VMI
+	if len(networks) == 0 || len(vmi.Spec.Domain.Devices.Interfaces) == 0 {
```
rmohr: Migrating without network connection is definitely allowed.

phoracek: We discussed it with Vatsal earlier on this PR. Is seemed like we don't get the default interface (pod via bridge binding) at this point, so "no ifaces" can mean either default or none. However, I found out we do set the defaults in admission webhook. So yes, we can just let 0 interfaces go here. Sorry @vatsalparekh for the confusion before.

rmohr: yep, defaults are set in the mutating webhook. So at this stage you will always see the final configuration.

vatsalparekh: Removed this check now

pkg/virt-handler/vm.go
vatsalparekh: changed to this now

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1151,35 @@ var _ = Describe("VirtualMachineInstance", func() {
 			Expect(blockMigrate).To(BeTrue())
 			Expect(err).To(Equal(fmt.Errorf("cannot migrate VMI with non-shared HostDisk")))
 		})
+
+		It("should not allow to migrate with adding default network configuration", func() {
```
rmohr: can  you rename this in a more understandable thing like `should block  migrations if the pod network is connected via  a binding method different than masquerade`?

Apart from that looks good.

vatsalparekh: @rmohr the func `checkNetworkInterfacesForMigration` will not return any error here because the list of interfaces will be empty.
Should we just need to change the test tittle here or return something different from the function?

rmohr: Not sure i can follow. Just ensure that you have a pod nework with the bridge binding set, so that you test that it gets rejected ...

vatsalparekh: @rmohr Changed it

rmohr:
/lgtm

@phoracek ?

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1151,51 @@ var _ = Describe("VirtualMachineInstance", func() {
 			Expect(blockMigrate).To(BeTrue())
 			Expect(err).To(Equal(fmt.Errorf("cannot migrate VMI with non-shared HostDisk")))
 		})
+
+		It("should block migrations if the pod network is connected via a binding method different than masquerade?", func() {
```
phoracek: I don't think the question mark should be there, but I'm fine leaving it there for fun.

vatsalparekh: Corrected it now :smile:

PR #2996, opened 2020-01-16
Expose kubevirt-configurations field

alonSadan:
We expose a new api-type
which holds all kubevirt configurations. HCO can now vendor this new type
and, and use it to put values in virt-ConfigMap.
Fixes #

This pr goes hand in hand with https://github.com/kubevirt/hyperconverged-cluster-operator/pull/391.
I will update the pr in HCO accordingly.
**Release note**:

```release-note
Add KubevirtConfigurations, MigrationConfig, DeveloperConfigurations and NetworkConfigurations to API-types
```


pkg/virt-config/config-map.go
```diff
@@ -44,22 +44,25 @@ import (
...
-	useEmulationKey                   = "debug.useEmulation"
-	ImagePullPolicyKey                = "dev.imagePullPolicy"
-	MigrationsConfigKey               = "migrations"
-	CpuModelKey                       = "default-cpu-model"
-	CpuRequestKey                     = "cpu-request"
-	MemoryOvercommitKey               = "memory-overcommit"
-	LessPVCSpaceTolerationKey         = "pvc-tolerate-less-space-up-to-percent"
-	NodeSelectorsKey                  = "node-selectors"
-	NetworkInterfaceKey               = "default-network-interface"
+	FeatureGatesKey                   = "featureGates"
```
rmohr: Maybe I miss something, but this would probably break all existing installations which have setting on kubevirt.

rmohr: We need to support both for some time. If no kubevirt yaml is presented in the config map, the config should still be taken as-is from the config map with the old names.

alonSadan: Thank you @rmohr :)
I hope now it's OK.

alonSadan:
Hey @rmohr Can you take a second look when you have time?
Thanks :)

alonSadan:
This PR is obsolete since the desired change is already done here: https://github.com/kubevirt/kubevirt/pull/3455

PR #3001, opened 2020-01-19
Revoke permission to GET secrets from virt-launcher

### LINKED ISSUE #2967: virt-handler daemonset can GET Secrets in all namespaces making it cluster root
**Is this a BUG REPORT or FEATURE REQUEST?**

/kind enhancement

**What happened**:

The kubevirt operator deploys a handful of manifests to the cluster. Many of the created RBAC rules are powerful, but ideally you could isolate the dangerous ones to dedicated nodes. Unfortunately, the virt-handler DaemonSet's ClusterRole lets it GET Secrets in all namespaces, so it can dump any other ServiceAccountSecret. It doesn't seem possible to deploy kubevirt without this.

https://github.com/kubevirt/kubevirt/blob/v0.24.0/manifests/generated/rbac-kubevirt.authorization.k8s.yaml.in#L309-L315

I wouldn't expect a daemonset, deployed to every instance, to have credentials which can root the cluster. This means compromising any node is game-over.

N/A

Is there a reason that this component needs this power? Could it be contained to a single namespace instead?

**Environment**:

N/A

danielBelenky:
In order to avoid giving virt-handler permission to GET all secrets
across all namespaces, virt-controller will now renders an additional
secret volume mount to virt-launcher's pod. This allows virt-launcher
to read the secret directly from the disk instead of having virt-launcher
to GET it.

Fixes #2967 

You don't see any functional test changes in this PR because those already exists from the previous implementation :)
Unit tests were adjusted to the new implementation...

**Release note**:
```release-note
Revoke permission to GET secrets from virt-launcher
```

### DIFF
--- a/tests/vmi_lifecycle_test.go
+++ b/tests/vmi_lifecycle_test.go
@@ -313,8 +313,26 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
 		})
 
 		Context("with user-data", func() {
+
+			findLauncherForVMI := func(vmi *v1.VirtualMachineInstance) *k8sv1.Pod {
+				By("Finding a launcher for VMI: " + vmi.Name)
+				launchers, err := virtClient.
+					CoreV1().
+					Pods(tests.NamespaceTestDefault).
+					List(metav1.ListOptions{LabelSelector: "kubevirt.io=virt-launcher"})
+				Expect(err).To(BeNil(), "Should list virt-launchers")
+				var launcher k8sv1.Pod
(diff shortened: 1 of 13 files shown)


tests/vmi_lifecycle_test.go
```diff
@@ -312,9 +312,27 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
 			)
 		})
 
-		Context("with user-data", func() {
+		FContext("with user-data", func() {
```
rmohr: `Context`

PR #3002, opened 2020-01-19
Added tests for vm/vmi resource watch

omeryahud:
Added 2 tests to make sure `oc get vm -w` and `oc get vmi -w` work correctly

**Release note**:
```release-note
NONE
```


tests/vm_watch_test.go
```diff
@@ -0,0 +1,334 @@
...
+
+		vmiStatus = readVMIStatus(stdout, vmiStatus, readTimeout)
+		Expect(vmiStatus.phase).To(Equal("Running"))
+
+		// Restart the VMI
+		err = virtCli.VirtualMachine(vm.ObjectMeta.Namespace).Restart(vm.ObjectMeta.Name)
+		Expect(err).To(BeNil())
+
+		vmiStatus = readVMIStatus(stdout, vmiStatus, readTimeout)
+		Expect(vmiStatus.phase).To(Equal("Failed"))
```
danielBelenky: Do we really enter a FAILED state when we restart? is it a known bug or expected behavior?

tests/vm_watch_test.go
```diff
@@ -0,0 +1,334 @@
...
+	var err error
+
+	if kubeconfig := os.Getenv("KUBECONFIG"); len(kubeconfig) > 0 {
+		virtCli, err = kubecli.GetKubevirtClientFromFlags("", kubeconfig)
+	}
+
+	tests.PanicOnError(err)
+
+	// Define a timeout for read opeartions in order to prevent test hanging
+	readTimeout := 20 * time.Second
```
danielBelenky: Maybe those should be constants at the top of the test?

omeryahud: I agree, fixed.

tests/vm_watch_test.go
omeryahud: The test plan specifies 'Failed' as an expected result, I will look into it.

omeryahud:
> Do we really need to check the output speifically from `kubectl`? Is it possible to use the go-client with a watch for this test and check for the status of the objects from there? Eventually, kubectl is reading the status from the same place and I'm not sure we should test `kubectl` iteself ...

Do you have an example of how to watch VM/VMI resources with `GetKubevirtClient()`?
I've tried to look for it but couldn't.

tests/vm_watch_test.go
omeryahud: I checked and `Failed` is a valid status since the VM was forcefully restarted

omeryahud:
> Do we really need to check the output speifically from `kubectl`? Is it possible to use the go-client with a watch for this test and check for the status of the objects from there? Eventually, kubectl is reading the status from the same place and I'm not sure we should test `kubectl` iteself ...

Agreed, will implement the `Watch()` logic like `client-go` does and use that in the test instead of `exec`ing oc/kubectl binary

tests/vm_watch_test.go
enp0s3: @omeryahud You can add description to the assertion if it is needed. Next to the "HaveOccured()", the second argument of ToNot can be a string that describes the assertion reason.

fabiand:
@ILpinto @dshchedr up for a review?

tests/vm_watch_test.go
```diff
@@ -0,0 +1,337 @@
...
+	var virtCli kubecli.KubevirtClient
+	var err error
+
+	if kubeconfig := os.Getenv("KUBECONFIG"); len(kubeconfig) > 0 {
+		virtCli, err = kubecli.GetKubevirtClientFromFlags("", kubeconfig)
+	}
+
+	tests.PanicOnError(err)
+
+	It("Should update vm 'RUNNING' status when using 'kubectl get vm -w'", func() {
```
ILpinto: missing test id

tests/vm_watch_test.go
```diff
@@ -0,0 +1,337 @@
...
+
+		vmStatus = readVMStatus(stdout, vmStatus, readTimeout)
+		Expect(vmStatus.running).To(Equal("false"))
+
+		By("Deleting the VM")
+		err = virtCli.VirtualMachine(vm.ObjectMeta.Namespace).Delete(vm.ObjectMeta.Name, &v1.DeleteOptions{})
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	It("Should update vmi 'PHASE' when using 'kubectl get vmi -w'", func() {
```
ILpinto: missing test id

tests/vm_watch_test.go
```diff
@@ -0,0 +1,337 @@
...
+	. "kubevirt.io/kubevirt/tests"
+)
+
+const (
+	// Define a timeout for read opeartions in order to prevent test hanging
+	readTimeout     = 20 * time.Second
+	processWaitTime = 10 * time.Second
+)
+
+var _ = Describe("VmWatch", func() {
```
ILpinto: missing rfe

tests/vm_watch_test.go
omeryahud: Fixed

tests/vm_watch_test.go
omeryahud: Fixed

tests/vm_watch_test.go
omeryahud: Fixed

tests/vm_watch_test.go
omeryahud: @enp0s3 Done, thanks
@danielBelenky We have discussed this offline

tests/vm_watch_test.go
```diff
@@ -0,0 +1,330 @@
...
+		Expect(stdOut).ToNot(BeNil())
+
+		stdErr, err := cmd.StderrPipe()
+		Expect(err).ToNot(HaveOccurred(), "stderr should have been redirected")
+		Expect(stdErr).ToNot(BeNil())
+
+		return cmd, stdOut, stdErr
+	}
+
+	It("[test_id:3468]Should update vm status with the proper columns using 'kubectl get vm -w'", func() {
```
vatsalparekh: @omeryahud the reason for failure here can be that the PR for this might not have made it back to previous versions.
https://github.com/kubernetes/kubernetes/pull/76161

Need to check specific version, but I think it should have made in 1.16 and onwards
You should skip the tests for the versions before that.

omeryahud: Thanks, working on it now

PR #3003, opened 2020-01-20
Move containerdisks to their own BUILD file to not download them when just building

rmohr:

Avoid pulling in all testing and demo container dependencies if one just
wants to build kubevirt and run unittests.

Before we had the containerdisk definitions at the same location like
the containerdisk binary and  we were just pulling them in on every
"make build".

Now the containerdisks are only pulled in when we explicitliy want to
build and/or push containers.

Fixes #

**Release note**:

```release-note
NONE
```


rmohr:
/override pull-kubevirt-e2e-k8s-1.16.2

repeatedly fails on kubectl output tests which are flaky and should be fixed in #3011.

PR #3005, opened 2020-01-20
Remove --deploy-testing-infra test flag.

maya-r:
This flag helps setup missing bits in the environment if they're not
already there, and remove them in the end.

The teardown code, in its current state, is quite dangerous to have.
If those resources were created in other methods, they will still
be removed.

As an alternative, don't have this flag and assume anyone testing
already has an environment setup. Anyone setting up a cluster using
this repository will apply all the manifests already: they are
applied in hack/cluster-deploy.sh.

**Release note**:
```release-note
Removed --deploy-testing-infra and --path-to-testing-infra-manifests flags to the testsuite
```

### DIFF
--- a/hack/functests.sh
+++ b/hack/functests.sh
@@ -45,4 +45,4 @@ if [[ ${KUBEVIRT_PROVIDER} == os-* ]] || [[ ${KUBEVIRT_PROVIDER} =~ (okd|ocp)-* 
     oc=${kubectl}
 fi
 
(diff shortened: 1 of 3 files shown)


danielBelenky:
What happens when I run `make functests` twice now? I think that the code that you just removed took care of scaling down and back up...

maya-r:
It's scaling to 0 (saving the current value),
deploying the manifests,
scaling back to the previous value.

Leaving the scaling as-is should have the same effect

rmohr:
The code was intended to allow running the tests standalone from a pod. Under which circumstances was there an issue? Also why is it dangerous? Isn't it just test infra?

maya-r:
The documentation in tests/README.md suggested always running the tests this way
It's dangerous because it will delete CDI CRD in the end (and all the associated data) on a cluster that already has CDI installed.

For a bit we thought that the ownerReferences bug isn't fixed because CDI CRD disappared in test runs.

PR #3006, opened 2020-01-20
Let the operator remove virt-controller generated traces

### LINKED ISSUE #1491: Uninstall/cleanup procedures

rmohr:

virt-api registers apiservices and webhooks which it does not delete. On
uninstall during the operator this lead to incomplete uninstalls which
could block future namespace deletions.

While letting still virt-api creating all these objects, let
virt-operator now manage their removal. The control is still not fully
handed over to virt-operator since there are still people using the
manifest-only installation path without the operator.

Fixes #1491 

**Release note**:

```release-note
Uninstall via the operator catches now all artifacts. Deleted namespaces should no longer be stuck forever after uninstalling kubevirt correctly.
```

### DIFF
--- a/pkg/virt-api/api.go
+++ b/pkg/virt-api/api.go
@@ -759,6 +759,7 @@ func (app *virtAPIApp) createValidatingWebhook() error {
 			return err
 		}
 	}
+
 	webHooks := app.validatingWebhooks()
 
 	if registerWebhook {
@@ -766,7 +767,8 @@ func (app *virtAPIApp) createValidatingWebhook() error {
 			ObjectMeta: metav1.ObjectMeta{
 				Name: virtWebhookValidator,
 				Labels: map[string]string{
-					v1.AppLabel: virtWebhookValidator,
+					v1.AppLabel:       virtWebhookValidator,
+					v1.ManagedByLabel: v1.ManagedByLabelOperatorValue,
 				},
 			},
 			Webhooks: webHooks,
@@ -775,6 +777,10 @@ func (app *virtAPIApp) createValidatingWebhook() error {
 			return err
 		}
 	} else {
+
+		// Ensure that we have the operator label attached, so that the operator can manage the resource later
+		// This is part of a soft transition from behing controlled by the apiserver, to being controlled by the operator
+		webhookRegistration.Labels[v1.ManagedByLabel] = v1.ManagedByLabelOperatorValue
 
 		// update registered webhook with our data
 		webhookRegistration.Webhooks = webHooks
@@ -894,6 +900,7 @@ func (app *virtAPIApp) createMutatingWebhook() error {
 			return err
 		}
 	}
+
 	webHooks := app.mutatingWebhooks()
 
 	if registerWebhook {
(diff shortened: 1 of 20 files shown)


rmohr:
/hold

Functional test is there but some unit tests are missing.

pkg/virt-api/api.go
```diff
@@ -776,6 +778,10 @@ func (app *virtAPIApp) createValidatingWebhook() error {
 		}
 	} else {
 
+		// Ensure that we have the operator label attached, so that the operator can manage the resource later
+		// This is part of a soft transition from behing controlled by the apiserver, to being controlled by the operator
+		webhookRegistration.Labels[v1.ManagedByLabel] = v1.ManagedByLabelOperatorValue
```
danielBelenky: Are webhooks namespaced? won't we hit the namespaced owner which owns a non-namespaced resource issue?

rmohr: We only add a label, no owner references.

danielBelenky: Right! :smile:

tests/operator_test.go
```diff
@@ -873,6 +873,18 @@ spec:
 			By("Sanity Checking Deployments infrastructure is deleted")
 			sanityCheckDeploymentsDeleted()
 
+			By("ensuring that namespaces can be successfully created and deleted")
```
rmohr: This should in future prevent that we leave anything behind which can block namespace management.

pkg/virt-operator/creation/rbac/operator.go
```diff
@@ -231,6 +231,18 @@ func NewOperatorClusterRole() *rbacv1.ClusterRole {
 				},
 				Resources: []string{
 					"validatingwebhookconfigurations",
+					"mutatingwebhookconfigurations",
```
danielBelenky: I'm not sure that this should be a ClusterRole... can it be just a role in the deployment namespace?

rmohr: I don't think it changes anything regarding to security and we would have to maintain an extra role for the virt-operator service account.

danielBelenky: It does allow kubevirt to create `mutatingwebhookconfigurations` across the entire cluster unnecessarily. Though it's not a security issue by it's own, some cluster admin may wonder why this is needed.

With that said, since we currently do not maintain a role, I agree that it shouldn't be part of this PR

rmohr: webhook configurations are not namespaced.

rmohr:
/approve

@danielBelenky thx for the review.

pkg/virt-operator/creation/rbac/operator.go
danielBelenky: :+1: good point. :smile: 
so forget my comment.

PR #3011, opened 2020-01-21
Fix flaky oc tests

vatsalparekh:

Fix the flaky oc tests
Fixes #

**Release note**:

```release-note
fix flaky oc tests
```

### DIFF
--- a/tests/kubectl_test.go
+++ b/tests/kubectl_test.go
@@ -15,7 +15,7 @@ import (
 var _ = Describe("[rfe_id:3423][vendor:cnv-qe@redhat.com][level:component]oc/kubectl get vm/vmi tests", func() {
 	tests.FlagParse()
 
-	var k8sClient string
+	var k8sClient, result string
 	var vm *v1.VirtualMachine
 
 	virtCli, err := kubecli.GetKubevirtClient()
@@ -33,10 +33,15 @@ var _ = Describe("[rfe_id:3423][vendor:cnv-qe@redhat.com][level:component]oc/kub
 	})
 
 	table.DescribeTable("should verify set of columns for", func(verb, resource string, expectedHeader []string) {
-
-		result, _, _ := tests.RunCommand(k8sClient, verb, resource)
-		Expect(result).ToNot(BeNil())
+		result, _, err = tests.RunCommand(k8sClient, verb, resource)
+		// due to issue of kubectl that sometimes doesn't show CRDs on the first try, retry the same command
+		if err != nil {
+			result, _, err = tests.RunCommand(k8sClient, verb, resource)
+		}
+		Expect(len(result)).ToNot(Equal(0))
 		resultFields := strings.Fields(result)
+		// Verify that only Header is not present
+		Expect(resultFields).ShouldNot(Equal(expectedHeader))
 		columnHeaders := resultFields[:len(expectedHeader)]
 		// Verify the generated header is same as expected
 		Expect(columnHeaders).To(Equal(expectedHeader))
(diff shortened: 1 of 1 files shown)


tests/kubectl_test.go
```diff
@@ -37,6 +37,8 @@ var _ = Describe("[rfe_id:3423][vendor:cnv-qe@redhat.com][level:component]oc/kub
 		result, _, _ := tests.RunCommand(k8sClient, verb, resource)
 		Expect(result).ToNot(BeNil())
 		resultFields := strings.Fields(result)
+		// Verify that only Header is not present
+		Expect(resultFields).ShouldNot(Equal(expectedHeader))
```
dhiller: I don't understand how this is going to fix the test. Can you elaborate please?

vatsalparekh: This is not fixing anything atm, this should just give us the output of `resultFields` if it's simply equal to `expectedHeader` instead of a plain panic on the next line

rmohr:
@vatsalparekh a more efficient way to provoke the errors is to change the tests this way and run them locally on your cluster:

```
Measure("should do something very often", func(b Benchmarker) {
// test here
}, 100) // run it 100 times or until it fails the first time
```

Since you use a table you have to restructure the tests a little bit. But I really recommend debugging this on your machine and then post the fix.

vatsalparekh:
@rmohr @dhiller I did try it with `Measure`, and I don't see any failures here
```
• [MEASUREMENT]
[rfe_id:3423][vendor:cnv-qe@redhat.com][level:component]oc/kubectl get vm/vmi tests
/root/go/src/kubevirt.io/kubevirt/tests/kubectl_test.go:15
  oc tests
  /root/go/src/kubevirt.io/kubevirt/tests/kubectl_test.go:35

  Found no measurements
```

tests/kubectl_test.go
rmohr: Can you do something like `Expect(len(resultFields)).Should(BeNumerically(">=", len(expectedHeader)))`?

vatsalparekh: Added, but `>` is better I think

tests/kubectl_test.go
```diff
@@ -50,9 +56,15 @@ var _ = Describe("[rfe_id:3423][vendor:cnv-qe@redhat.com][level:component]oc/kub
 	table.DescribeTable("should verify set of wide columns for", func(verb, resource, option string, expectedHeader []string, verifyPos int, expectedData string) {
 
 		result, _, _ := tests.RunCommand(k8sClient, verb, resource, "-o", option)
-
-		Expect(result).ToNot(BeNil())
+		// due to issue of kubectl that sometimes doesn't show CRDs on the first try, retry the same command
+		if err != nil {
+			result, _, err = tests.RunCommand(k8sClient, verb, resource)
```
slintes: You are missing the `-o, option` args option here.
Also, wouldn't an Eventually block make sense, that would prevent such mistakes?

vatsalparekh: @slintes good catch for the `-o`, corrected it now
About `Eventually`, the error should ideally occur only first time on a fresh cluster, so only one retry here makes sense, after that more tries would be some bug.

slintes: "ideally" sounds like "eventually", doesn't it? ;) You can retry once with it, and it removes code duplication. But I won't block on this.

rmohr:
Why did the tests pass without the `-o` option?

vatsalparekh:
@rmohr Because the test with `-o` is not the first one to do a get request for CRD, in the flaky tests also, that second one doesn't seem to fail either

PR #3013, opened 2020-01-22
Adding support to specify app armor profile for virt-launcher pod

visheshtanksale:
Signed-off-by: Vishesh Tanksale <vtanksale@nvidia.com>

These changes add ability to run virt-launcher pod under specific app armor profile. It leverages the existing support for app armor profile on k8s
Fixes #

**Release note**:

```release-note
AppArmor profiles can be applied to compute container in virt-launcher. 
```

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -888,6 +888,11 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 	}
 
 	for k, v := range vmi.Annotations {
+		if strings.Contains(k, "container.apparmor.security.beta.kubernetes.io") {
+			annotationsList[k] = v
+			continue
+		}
+
 		if strings.Contains(k, "kubernetes.io") || strings.Contains(k, "kubevirt.io") {
 			// skip kubernetes and kubevirt internal annotations
 			continue
--- a/pkg/virt-controller/services/template_test.go
+++ b/pkg/virt-controller/services/template_test.go
@@ -1589,7 +1589,26 @@ var _ = Describe("Template", func() {
 			Expect(pod.Spec.Containers[0].Command).To(ContainElement("--less-pvc-space-toleration"), "command arg key should be correct")
 			Expect(pod.Spec.Containers[0].Command).To(ContainElement("42"), "command arg value should be correct")
 		})
+		Context("App armor annotation", func() {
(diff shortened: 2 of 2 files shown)


danielBelenky:
Hello :smile:, Thank you contributing.

/ok-to-test
/lgtm

danielBelenky:
@visheshtanksale hi, can AppArmor be used to actually **extend** the permissions on virt-launcher? if so, it potentially introduces a hidden security issue where granting users the ability to create VMIs will let them also extend the permissions on the launcher pod. WDYT?

fabiand:
@visheshtanksale do you have a peer who could help reviewing this PR?

visheshtanksale:
@danielBelenky This cannot be actually used to extend permissions to virt-launcher pod. It will only let the user make a decision about under which app-armor profile the container in virt-launcher pod would run. Currently, in kubernetes user can specify app-armor profile for any of the pods more details [here](https://kubernetes.io/docs/tutorials/clusters/apparmor/). But because user does not create a virt-launcher pod directly we have to add this change

visheshtanksale:
@fabiand My peers have already reviewed this internally. I want to have some opinion from the community members.

rmohr:
@visheshtanksale the implementation sounds reasonable

> @danielBelenky This cannot be actually used to extend permissions to virt-launcher pod. It will only let the user make a decision about under which app-armor profile the container in virt-launcher pod would run. Currently, in kubernetes user can specify app-armor profile for any of the pods more details [here](https://kubernetes.io/docs/tutorials/clusters/apparmor/). But because user does not create a virt-launcher pod directly we have to add this change

@visheshtanksale see https://kubernetes.io/docs/tutorials/clusters/apparmor/#restricting-profiles-with-the-podsecuritypolicy for more details. PSPs allow restrcting the profiles. One of the reasons is that we give pods already some extra privileges already. So having e.g. NET_ADMIN on the pod (which we still need to add) plus a too generous appamor profile could lead to unwanted privilege escalation. Let's just play save.

danielBelenky:
@visheshtanksale maybe you could take a similar approach to what is done in https://github.com/kubevirt/kubevirt/pull/3056 and make this profile configurable from kubevirt's global config? This will ensure that only the cluster admin can define the profile for virt-launcher? or at least let admins to choose a whitelisted set of profiles that users can choose if more flexibility makes sense.

fabiand:
+1!

We are eagerly awaiting a patch! ;)

PR #3014, opened 2020-01-22
Add unit tests to cover the migration blocking for bridge network.

petrkotas:

Add unit tests to cover the migration blocking for bridge network.

/cc @dhiller @danielBelenky 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-handler/vm_test.go
+++ b/pkg/virt-handler/vm_test.go
@@ -835,7 +835,8 @@ var _ = Describe("VirtualMachineInstance", func() {
 		}, 3)
 	})
 
-	Context("check VMI disks for migration", func() {
+	Context("check if migratable", func() {
+
 		var testBlockPvc *k8sv1.PersistentVolumeClaim
 
 		BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1151,53 @@ var _ = Describe("VirtualMachineInstance", func() {
...
+						Bridge: &v1.InterfaceBridge{},
+					},
+				},
+			}
+
+			err := controller.checkNetworkInterfacesForMigration(vmi)
+			Expect(err).To(HaveOccurred(), "should block migration for bridging interface")
+		})
+
+		It("should not block migration for bridge interface assigned to pod", func() {
```
dhiller: s/bridge/masquerade/

petrkotas:
Thanks for the feedback!

@dhiller I have refactored the code using `Context` to make it more readable. I do not like the idea of table. The variables I would  need to pass would still be ugly.

@phoracek added the multus check.

phoracek:
Could you submit the storage changes separately? I don't really see how is that refactoring related to "blocking for bridge".

petrkotas:
Ahh sorry for that, I was trying to save PR.

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1152,82 @@ var _ = Describe("VirtualMachineInstance", func() {
...
+							Masquerade: &v1.InterfaceMasquerade{},
+						},
+					},
+				}
+
+				err := controller.checkNetworkInterfacesForMigration(vmi)
+				Expect(err).ToNot(HaveOccurred(), "should not block migration for masquerade binding method")
+			})
+
+			It("should not block migration for bridge interface assigned to multus network", func() {
```
phoracek: "should not block migration for bridge binding assigned to a multus network"

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1152,82 @@ var _ = Describe("VirtualMachineInstance", func() {
...
+				vmi := v1.NewMinimalVMI("testvmi")
+				interface_name := "interface_name"
+
+				vmi.Spec.Networks = []v1.Network{
+					v1.Network{
+						Name:          "pod",
+						NetworkSource: v1.NetworkSource{Pod: &v1.PodNetwork{}},
```
phoracek: you can drop the "pod" network from the test, it is not the subject of this case and there is no `Interface` using it in the following section

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1152,82 @@ var _ = Describe("VirtualMachineInstance", func() {
...
+							Bridge: &v1.InterfaceBridge{},
+						},
+					},
+				}
+
+				err := controller.checkNetworkInterfacesForMigration(vmi)
+				Expect(err).To(HaveOccurred(), "should block migration for bridge binding")
+			})
+
+			It("should not block migration for bridge interface assigned to pod", func() {
```
phoracek: "should not block migration for *marquerade* binding assigned to the pod network"

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1152,82 @@ var _ = Describe("VirtualMachineInstance", func() {
 			Expect(blockMigrate).To(BeTrue())
 			Expect(err).To(Equal(fmt.Errorf("cannot migrate VMI with non-shared HostDisk")))
 		})
+
+		Context("with network configuration", func() {
+			It("should block migration for bridge interface assigned to pod", func() {
```
phoracek: "should block migration for bridge binding assigned to the pod network"

PR #3019, opened 2020-01-22
Fix Prometheus alert rules and add CI for it

danielBelenky:
Fixed PromQL syntax error and added CI to ensure that we catch issues
like this in the future.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,14 @@ builder-publish:
 olm-verify:
 	hack/dockerized "./hack/olm.sh verify"
 
+current-dir := $(shell pwd)
+
+build-prom-spec-dumper:
+	hack/dockerized "go build -o rule-spec-dumper ./hack/prom-rule-ci/rule-spec-dumper.go"
+
+prom-rules-verify: build-prom-spec-dumper
+	./hack/prom-rule-ci/verify-rules.sh ${current-dir}/rule-spec-dumper
+
 olm-push:
 	hack/dockerized "DOCKER_TAG=${DOCKER_TAG} CSV_VERSION=${CSV_VERSION} QUAY_USERNAME=${QUAY_USERNAME} \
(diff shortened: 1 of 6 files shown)


danielBelenky:
forgot to run `make` to add license to the new script -,- will update

hack/prom-rule-ci/verify-rules.sh
```diff
@@ -0,0 +1,35 @@
...
+# Copyright 2020 Red Hat, Inc.
+#
+
+readonly PROM_IMAGE="docker.io/prom/prometheus:v2.15.2"
+
+function main() {
+    local prom_spec_dumper="${1:?}"
+    local target_file
+
+    target_file="$(mktemp --tmpdir -u tmp.prom_rules.XXXXX)"
```
rmohr: I guess we would also have to clean them up?

danielBelenky: Done

Makefile
```diff
@@ -120,6 +120,14 @@ builder-publish:
 olm-verify:
 	hack/dockerized "./hack/olm.sh verify"
 
+current-dir := $(shell pwd)
+
+build-prom-spec-dumper:
```
dhiller: (optional) suggestion: Maybe put this into a separate Makefile in subdir `/hack/prom-rule-ci`?

danielBelenky: I don't have a particular objection about moving it there. Why you think it will be better in a separate Makefile?

danielBelenky: (though I wouldn't want to block this PR from getting merged on it right now because it fixes a bug)

dhiller: I don't want to block the merge, I just didn't have time yet to have a look. And furthermore I don't yet understand what it does honestly.

danielBelenky: @dhiller it ensures that we don't have syntax or configuration errors in our prometheus rules.
`promtool` is shipped with the official Prometheus container, but it can only read files (no option to work from stdin directly), so first, we extract the prometheus rules for kubevirt and dump them to a file. Then, we pass it to `promtool` for linting.
Maybe in the future, we can also leverage this for unt-testing the rules as well...

PR #3021, opened 2020-01-24
WIP: Install cnao at ocp provider

qinqon:
Signed-off-by: Quique Llorente <ellorent@redhat.com>

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,7 @@ release-announce: .release-functest
 
 cluster-up:
 	./cluster-up/up.sh
+	./hack/install-cnao.sh
 
 cluster-down:
 	./cluster-up/down.sh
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -96,6 +96,7 @@ release-announce: .release-functest
 
 cluster-up:
 	./cluster-up/up.sh
+	./hack/install-cnao.sh
```
phoracek: Images should be pre-pulled in kubevirtci provider. In kubevirtci/cluster-up, there should be a flag enabling installation of these manifests. That way we get stable snapshot of our released components that would be available on demand. With kubevirt we would install them, with cnao, they would stay dormant.

PR #3025, opened 2020-01-27
virtctl: If we have global flags, display them in usage

maya-r:
Rather than suggesting another command that the user could run
(it's more work)

**Release note**:
```release-note
Improve virtctl usage message
```


rmohr:
I may be wrong but I think that we have `virtctl options` like kubectl. It should also be mentioned on the usage pages. The reason is that we don't want to overload every command with repetitive and well known options.

danielBelenky:
removing the `approved` label until we can sort out Roman's comment... (though hold was set)

maya-r:
@rmohr, what do you think about this additional rationale?

rmohr:
> @rmohr, what do you think about this additional rationale?

I can understand that the very first time you may miss `options`, or may not know that `-n|--namespace` exists, but displaying the "uinteresting" options always does not sound very appealing either. I have the feeling that mostly people which did not work with k8s before would have this issue, and they may have the same issue with `kubectl`.

PR #3033, opened 2020-01-29
Increase memory limit for kubevirt-infra pod

### LINKED ISSUE #3029: OOMKiller shutting down kubevirt-infra sidecar causing VM shutdown
**BUG REPORT**:

/kind bug

**What happened**:

Running on Azure, sequence of events was the following.

- kubevirt-infra container died with OOMKiller
- VM running in hypervisor did clean shutdown
- VMI vanished

Neither of the first two.

- The OOMKiller is likely because the pod spec for the kubevirt-infra container has slightly too low a limit on memory. Azure appears to be a bit trigger happy enforcing this limit.
- The clean shutdown is rather odd. I'd guess that the underlying logic is that with kubevirt-infra having blown up, the handler decides that the VMI is sick, sends it a shutdown signal and blows it away. That is arguably correct behaviour given the initial error.

The repro scenario appears to be to run on Azure (or somebody else would have noticed it!), and have a readiness probe on some port. The time from VMI creation to stopping is very uneven. I've seen it vary from a few minutes (so VM doesn't even always finish booting) to a few hours. There's some correlation in timing which I think is some Azure artifact.

I'm happy to test a fix or attempt to repro with a debug build.

I just realised I am not on the latest kubevirt version; will repro on 0.25 today.

If you can tell me at how to change the memory limit on the sidecar, I'd be happy to test if just raising it slightly helps; I'm 90% sure that would make the problem go away.

**Environment**:
- KubeVirt version (use `virtctl version`): 0.22.0
- VM or VMI specifications: 
- Cloud provider or hardware configuration: AKS, standard cluster
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Install tools: N/A
- Others: N/A

plwhite:

The OOM Killer on Azure is occasionally killing the kubevirt-infra sidecar, causing VMIs to shutdown without warning at random intervals. This fix resolves that problem by increasing the memory limit.

Fixes #3029

I have manually tested on an Azure platform that I am testing with, and this appears to completely resolve the issue (random shutdowns reduced from every hour or two to none in a couple of days, and still running).

**Release note**:
NONE

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -870,7 +870,7 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 			Resources: k8sv1.ResourceRequirements{
 				Limits: map[k8sv1.ResourceName]resource.Quantity{
 					k8sv1.ResourceCPU:    resource.MustParse("1m"),
-					k8sv1.ResourceMemory: resource.MustParse("12Mi"), // this is the minimum memory limit for cri-o!
+					k8sv1.ResourceMemory: resource.MustParse("40Mi"),
 				},
 			},
 			Command:        []string{"/usr/bin/tail", "-f", "/dev/null"},


plwhite:
Fixes bug #3029

danielBelenky:
/retest

thanks for the contribution :)

would you mind signing your commit? :)

plwhite:
I haven't got authorisation to sign the CLA yet. I expect this to be just a formality, but there may be a slight delay until the right boxes are ticked here.

pkg/virt-controller/services/template.go
```diff
@@ -870,7 +870,7 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 			Resources: k8sv1.ResourceRequirements{
 				Limits: map[k8sv1.ResourceName]resource.Quantity{
 					k8sv1.ResourceCPU:    resource.MustParse("1m"),
-					k8sv1.ResourceMemory: resource.MustParse("12Mi"), // this is the minimum memory limit for cri-o!
+					k8sv1.ResourceMemory: resource.MustParse("40Mi"), // this is the minimum memory limit for cri-o!
```
slintes: Can you remove the comment please? 12Mi was a real minimum limit accepted by cri-o. If we go beyond that. it is not relevant anymore.

slintes:
> I haven't got authorisation to sign the CLA yet. I expect this to be just a formality, but there may be a slight delay until the right boxes are ticked here.

that DCO thing is about signing off commits with `git commit -s` only, see https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s

/ok-to-test

rmohr:
We should also consider moving forward with #2844. Totally fine for me to take this one in, but in general having to request 40Mi is a little bit extreme.

pkg/virt-controller/services/template.go
plwhite: Done.

plwhite:
Apologies for the force-push - I had to add the signed-off-by, and in the process I fixed the comment.

slintes:
formatting of the NONE release note seems to be broken, let's fix it with

/release-note-none

plwhite:
I don't see how I could have broken that test case, so going to retest as it suggests.
/retest

rmohr:
@plwhite yep all is good. The tests are just merge-gated, so once a PR gets merged all tests will run again, and sometimes a lane fails, but the retest bot will also take care of writing retest. Just needs some time.

PR #3036, opened 2020-01-29
Added a basic feature to rename a stopped vm

omeryahud:

Added a basic feature to rename a stopped vm with `virtctl` and `kubevirt.io/client-go`:
`virtctl rename [vm_name] [new_vm_name]`

**Release note**:

```release-note
It is now possible to rename a stopped vm with virtctl:
    virtctl rename vmname newvmname
```


pkg/virt-api/rest/subresource.go
```diff
@@ -517,6 +518,99 @@ func (app *SubresourceAPIApp) RestartVMRequestHandler(request *restful.Request,
...
+			Annotations:     vm.ObjectMeta.Annotations,
+			OwnerReferences: vm.ObjectMeta.OwnerReferences,
+			Finalizers:      vm.ObjectMeta.Finalizers, // Need to update the old VM and remove its finalizers and apply them here
+			//Initializers:    vm.ObjectMeta.Initializers, // Need to discover which initializers the VM had and apply them here
+		},
+		Spec: vm.Spec,
+	}
+
+	// Creating a new VM with the new name
+	newVM, err = app.virtCli.VirtualMachine(namespace).Create(newVM)
```
mhenriks: The problem I see with this approach is failure handling.  What if the apiserver crashes after creating the new vm but before deleting the old one?  Maybe just have the apiserver set a field/annotation in the VM that triggers a controller to do the rest in a resilient way?

omeryahud: That's a great point, I totally agree and will adapt your approach, thanks!

omeryahud: @mhenriks Hi Michael, I implemented your approach, you're welcome to take a look.
If you want to wait for tests to be implemented that's fine too.
Thanks!

pkg/virt-controller/watch/vm.go
```diff
@@ -271,7 +271,110 @@ func (c *VMController) execute(key string) error {
...
+				errMsg :=
+					fmt.Sprintf("can't determine existence of newVM: %s while processing rename of oldVM: %s",
+						renameTo, vm.Name)
+				logger.Reason(err).Error(errMsg)
+				return err
+			} else {
+				// New vm is not found, create it
+				newVM := &virtv1.VirtualMachine{
+					TypeMeta: vm.TypeMeta,
+					ObjectMeta: v1.ObjectMeta{
```
danielBelenky: I think that you should deep copy the given vm object and just change the name on it. Otherwise, if any new fields will be added we risk in missing them

pkg/virt-controller/watch/vm.go
```diff
@@ -271,7 +271,110 @@ func (c *VMController) execute(key string) error {
...
+				if err != nil {
+					errMsg := fmt.Sprintf("Could not create a new vm during the rename process of: %s", vm.Name)
+					logger.Reason(err).Error(errMsg)
+					return err
+				}
+			}
+		}
+
+		// New VM exists, remove renameTo annotation
+		patchString := fmt.Sprintf("{\"metadata\":{\"annotations\":{\"%s\":null}}}",
```
danielBelenky: Consider replacing with
```
fmt.Sprintf(`{"..."}`)
```

pkg/virt-controller/watch/vm.go
```diff
@@ -271,7 +271,110 @@ func (c *VMController) execute(key string) error {
...
+
+		if err != nil {
+			logger.Reason(err).Error(fmt.Sprintf("Failed removing 'renameTo' annotation from vm: %s", vm.Name))
+			return err
+		}
+
+		return nil
+	}
+
+	renameFrom, hasRenameFromAnnotation := vm.ObjectMeta.Annotations[virtv1.RenameFromAnnotation]
```
danielBelenky: I think that the flow here can go into a different function.. maybe `removeRenamedVM`?

pkg/virt-controller/watch/vm.go
```diff
@@ -271,7 +271,110 @@ func (c *VMController) execute(key string) error {
...
+
+	// Handle destination VM
+	if hasRenameFromAnnotation {
+		// Attempt to delete the old vm
+		err := c.clientset.VirtualMachine(vm.Namespace).Delete(renameFrom, &v1.DeleteOptions{})
+
+		if err != nil {
+			if !strings.Contains(err.Error(), "not found") {
+				errMsg :=
+					fmt.Sprintf("can't delete oldVM: %s while processing newVM: %s",
```
danielBelenky: I think that we should hide implementation details from the user... as far as the user knows, he asked us to rename a VM, not create new and remove the old one...
The error should at least tell the user  that the rename process failed...

pkg/virtctl/vm/vm.go
```diff
@@ -104,12 +105,33 @@ func NewMigrateCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
...
+}
+
 type Command struct {
 	clientConfig clientcmd.ClientConfig
 	command      string
 }
 
 func usage(cmd string) string {
+	if cmd == COMMAND_RENAME {
+		usage := fmt.Sprintf("	# %s a virtual machine called 'myvm' to 'notmyvm'\n", strings.Title(COMMAND_RENAME))
```
danielBelenky: Maybe just write `rename` here? If someone will change the command from `rename` to `rename-vm` the grammar here will be wrong...

pkg/virt-api/rest/subresource.go
```diff
@@ -517,6 +518,82 @@ func (app *SubresourceAPIApp) RestartVMRequestHandler(request *restful.Request,
...
+		writeError(errors.NewConflict(v1.Resource("virtualmachine"),
+			name, fmt.Errorf("Renaming a running vm is not allowed")), response)
+		return
+	}
+
+	// Check for renaming annotation on vm
+	if vm.Annotations != nil {
+		_, isScheduledForRenaming := vm.Annotations[v1.RenameToAnnotation]
+
+		if isScheduledForRenaming {
```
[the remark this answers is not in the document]
omeryahud: I don't think it's relevant, because if a `renameTo` annotation exists, regardless of its value, virt-controller has already started handling the rename process and will create a new VM and delete the old one.
Say I checked if `vm.Annotations["renameTo"] == newName`, what would you suggest I do then?

pkg/virt-api/rest/subresource.go
```diff
@@ -517,6 +518,82 @@ func (app *SubresourceAPIApp) RestartVMRequestHandler(request *restful.Request,
...
+			writeError(errors.NewConflict(v1.Resource("virtualmachine"),
+				name, fmt.Errorf("VM is already scheduled for renaming")), response)
+			return
+		}
+	}
+
+	patchString := fmt.Sprintf("{\"metadata\":{\"annotations\":{\"%s\":\"%s\"}}}",
```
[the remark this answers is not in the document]
omeryahud: Done, thanks!

pkg/virt-controller/watch/vm.go
omeryahud: Done

pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go
```diff
@@ -140,6 +140,12 @@ func (admitter *VMsAdmitter) authorizeVirtualMachineSpec(ar *v1beta1.AdmissionRe
 		}
 	}
 
+	if len(causes) > 0 {
+		return causes, nil
+	}
+
+	causes = validateVMStartAllowed(vm)
```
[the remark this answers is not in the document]
omeryahud: You mean like this right?
`if len(causes) > 0 {
		return causes, nil
	}`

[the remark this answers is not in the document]
omeryahud: I think it's redundant, if I understood correctly, that's how it would look like afterwards
```
causes = validateVMStartAllowed(vm)

	if len(causes) > 0 {
		return causes, nil
	}
	
	return causes, nil
```

pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go
```diff
@@ -222,5 +228,39 @@ func ValidateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
 		}
 	}
 
+	return causes
+}
+
+func validateVMStartAllowed(vm *v1.VirtualMachine) []metav1.StatusCause {
```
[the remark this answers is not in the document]
omeryahud: Not sure, maybe `preventStartIfRenameInProgress`?

pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go
```diff
@@ -222,5 +228,39 @@ func ValidateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
...
+	var causes []metav1.StatusCause
+
+	if vm.Annotations != nil {
```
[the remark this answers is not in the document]
omeryahud: I agree, done.

pkg/virt-controller/watch/vm.go
omeryahud: I agree, done

pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go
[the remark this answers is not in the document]
omeryahud: I think I should prevent ANY kind of change to a vm (not only starting it) if a renaming process is in progress, to prevent any updates or patches to the old vm that would not carry over to the new one, what do you think?

pkg/virt-controller/watch/vm.go
omeryahud: You are right about hiding implementation details, but the rename process did not fail at this point, only this attempt at the process has failed
The VM would be re-enqueued, and the controller would try again. Notice that this code is reached when an ambiguous error occurs (`err != 'not found'`), meaning I cannot be certain that the old VM was actually deleted, so I have to re-enqueque the VM untill I successfully delete it or receive `err == 'not found'`, only then I am certain it was actually deleted.
What's an error message you would recommend? I don't want to write vague error messages like "Rename process failed, will retry" because it doesn't hint to what the exact problem was in the process

pkg/virtctl/vm/vm.go
omeryahud: Fixed

pkg/virt-controller/watch/vm.go
omeryahud: I don't know, I think this function is pretty straight-forward in the way it handles both source/target VMs and it's not too big

pkg/virt-controller/watch/vm.go
omeryahud: I've modified the error messages

pkg/virt-controller/watch/vm.go
```diff
@@ -271,7 +276,133 @@ func (c *VMController) execute(key string) error {
...
+	// Handle source VM
+	if hasRenameToAnnotation {
+		// We don't need to check for the running status of the VM,
+		// it was already verified in the admission webhook
+
+		// If a VM with the new name exists, remove annotation
+		// TODO can I check for the existence of a vm using c.vmiVMInformer.GetStore().GetByKey()
+		// TODO and be 100% sure its consistent with the cluster?
+		// TODO If its possible, tests should be fixed
+		_, err := c.clientset.VirtualMachine(vm.Namespace).Get(renameTo, &v1.GetOptions{})
```
omeryahud: Looking for thoughts on the comments above this line

pkg/virt-controller/watch/vm.go
```diff
@@ -271,7 +276,133 @@ func (c *VMController) execute(key string) error {
...
+	// Check if renameFrom annotation exists
+	renameFrom, hasRenameFromAnnotation := vm.ObjectMeta.Annotations[virtv1.RenameFromAnnotation]
+
+	// Handle destination VM
+	if hasRenameFromAnnotation {
+		// Attempt to delete the old vm
+		// TODO can I delete the old vm using c.vmiVMInformer.GetStore().Delete()
+		// TODO and be 100% sure it will be removed from the cluster?
+		// TODO If its possible, tests should be fixed
+		err := c.clientset.VirtualMachine(vm.Namespace).Delete(renameFrom, &v1.DeleteOptions{})
```
omeryahud: Looking for thoughts on the comments above this line

pkg/virt-api/rest/subresource.go
```diff
@@ -517,6 +518,98 @@ func (app *SubresourceAPIApp) RestartVMRequestHandler(request *restful.Request,
...
+
+		if isScheduledForRenaming {
+			writeError(errors.NewConflict(v1.Resource("virtualmachine"),
+				name, fmt.Errorf("VM is already scheduled for renaming")), response)
+			return
+		}
+	}
+
+	// Make sure VM is stopped
+	if vm.Spec.Running == nil {
```
stu-gott: Why not just use vm.RunStrategy()? Then there's no need for nested if's here.

pkg/virt-api/rest/subresource_test.go
```diff
@@ -519,6 +532,212 @@ var _ = Describe("VirtualMachineInstance Subresources", func() {
...
+					server.AppendHandlers(
+						ghttp.CombineHandlers(
+							ghttp.VerifyRequest("GET", getVMPath("v1alpha3", vm.Namespace, vm.Name)),
+							ghttp.RespondWithJSONEncodedPtr(&vmGetStatus, vm),
+						),
+					)
+				})
+
+				Context("With no running/runStrategy applied", func() {
+					It("should fail if no running state or runStrategy are applied", func(done Done) {
```
stu-gott: Is this actually possible? I thought there was a validating webhook preventing a VM from being posted if neither Running nor RunStrategy were defined. If you've found a way to make this state actually happen, that's a bug.

pkg/virt-api/rest/subresource.go
```diff
@@ -517,6 +518,98 @@ func (app *SubresourceAPIApp) RestartVMRequestHandler(request *restful.Request,
...
+			name, err), response)
+		return
+	}
+
+	newName := string(newNameBuf)
+
+	request.Request.Body.Close()
+
+	if newName == "" {
+		writeError(errors.NewBadRequest("The new name is empty"), response)
```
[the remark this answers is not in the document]
omeryahud: Something like `"Please provide a new name for the VM"` sounds better?

pkg/virt-api/rest/subresource.go
```diff
@@ -517,6 +518,98 @@ func (app *SubresourceAPIApp) RestartVMRequestHandler(request *restful.Request,
...
+		return
+	}
+
+	if name == newName {
+		writeError(errors.NewBadRequest("The new name cannot be identical to the old name"), response)
```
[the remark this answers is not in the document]
omeryahud: Will do

pkg/virt-api/rest/subresource.go
omeryahud: I was not aware of the existence of this method, just read it and I totally agree.
Changing it right now

pkg/virt-api/rest/subresource_test.go
omeryahud: Yes, there is a validating webhook that checks for that, But because I was checking for running/runStrategy separately I wanted to make sure it works.
But now after you informed me about `vm.RunStrategy()`, this can be removed

omeryahud:
> A few minor concerns listed.
> 
> In general I'm not a fan of the idea of scribbling status messages in annotations. Is there a compelling reason for doing it this way instead of defining a field in the vm.Status stanza?

The reason I used annotations was that I didn't know of any other way.
I will take a look at `vm.Status` and see if it fits my needs.
Thanks!

stu-gott:
Be careful `VirtualMachineStateChangeRequest` -- I think it specifically applies to RunStrategy (at least that's the intent). There's all sorts of accounting logic tied up with that. We'll just need to carefully review the ramifications.

omeryahud:
> Be careful `VirtualMachineStateChangeRequest` -- I think it specifically applies to RunStrategy (at least that's the intent). There's all sorts of accounting logic tied up with that. We'll just need to carefully review the ramifications.

Yea I noticed, I just finished my proposal, will push soon for you to review

pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go
```diff
@@ -224,3 +234,65 @@ func ValidateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
 
 	return causes
 }
+
+// This function rejects VM patches/edits if the VM has a rename annotation (renaming process is in progress)
```
omeryahud: Redundant comment, function name is self explanatory

pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go
```diff
@@ -224,3 +234,65 @@ func ValidateVirtualMachineSpec(field *k8sfield.Path, spec *v1.VirtualMachineSpe
...
+				})
+			}
+		}
+	}
+
+	return causes
+}
+
+func hasRenameCreationRequest(vm *v1.VirtualMachine) bool {
+	for i := 0; i < len(vm.Status.StateChangeRequests); i++ {
```
omeryahud: Replace with iterator for loop

pkg/virt-api/rest/subresource.go
```diff
@@ -517,6 +525,97 @@ func (app *SubresourceAPIApp) RestartVMRequestHandler(request *restful.Request,
...
+
+	if err != nil {
+		writeError(errors.NewConflict(v1.Resource("virtualmachine"),
+			name, err), response)
+		return
+	}
+
+	newName := string(newNameBuf)
+
+	request.Request.Body.Close()
```
omeryahud: Defer this above `newNameBuff...`

pkg/virt-api/rest/subresource.go
```diff
@@ -517,6 +525,97 @@ func (app *SubresourceAPIApp) RestartVMRequestHandler(request *restful.Request,
...
+			"newName": newName,
+		},
+	})
+
+	if err != nil {
+		writeError(errors.NewInternalError(err), response)
+		return
+	}
+
+	_, err = app.virtCli.VirtualMachine(namespace).Patch(name, types.JSONPatchType, []byte(renameRequestJson))
```
omeryahud: Check if StrategicMergePatch is the better option

pkg/virt-controller/watch/vm.go
```diff
@@ -1150,6 +1241,45 @@ func (c *VMController) updateStatus(vmOrig *virtv1.VirtualMachine, vmi *virtv1.V
 				log.Log.Object(vm).Errorf("Start request shouldn't be honored for RunStrategyHalted.")
 				clearChangeRequest = true
 			}
+		case virtv1.RenameCreateRequest:
+			runningStatus, _ := vm.RunStrategy()
```
omeryahud: This can be removed, validated in webhooks

pkg/virt-controller/watch/vm.go
```diff
@@ -271,7 +276,94 @@ func (c *VMController) execute(key string) error {
...
+			// Attempt creation of the new VM
+			_, err = c.clientset.VirtualMachine(vm.Namespace).Create(newVM)
+
+			if err != nil {
+				return true, err
+			}
+		}
+	}
+
+	return false, nil
```
omeryahud: Specify that a VM with the new name was found

danielBelenky:
/lgtm
@stu-gott could you approve if you think that all the changes you requested were fulfilled?

omeryahud:
> Don't want to block this for the struct change, but that would be good practice to follow

Great idea, I'm on it

pkg/virt-api/rest/subresource.go
```diff
@@ -520,6 +528,96 @@ func (app *SubresourceAPIApp) RestartVMRequestHandler(request *restful.Request,
...
+	defer request.Request.Body.Close()
+	newNameBuf, err := ioutil.ReadAll(request.Request.Body)
+
+	if err != nil {
+		writeError(errors.NewConflict(v1.Resource("virtualmachine"),
+			name, err), response)
+		return
+	}
+
+	newName := string(newNameBuf)
```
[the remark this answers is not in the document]
omeryahud: Done, please take a look

pkg/virt-controller/watch/vm.go
```diff
@@ -271,7 +276,94 @@ func (c *VMController) execute(key string) error {
...
+			}
+		}
+	}
+
+	return false, fmt.Errorf("a VM with the new name was found, rejecting rename request")
+}
+
+// Handles VM rename delete requests
+// First return value is a boolean indicating if the controller should retry the request
+func (c *VMController) handleVMRenameDeleteRequest(vm *virtv1.VirtualMachine, oldName string) (bool, error) {
```
mhenriks: I think we need additional security here.  Imagine a user with RBAC permission limited to a single VM in a namespace. 
 Correct me if I'm wrong, but is not possible for that user to add a `RenameDeleteRequest` to the status of that vm allowing them to delete any arbitrary VM in the namespace?

omeryahud: The specific scenario you described is not possible because of
https://github.com/kubevirt/kubevirt/blob/8a51882bf5754c497579bddee9be236c693a9ba4/pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go#L253
Which prevents updating a VM with a `RenameDeleteRequest`.
The other issue that I found is that a user can **create** a new VM with a `RenameDeleteRequest` and delete an arbitrary VM in the namespace, which is still bad because it means that `create == delete` permission wise.
In order to make the entire process safe and failure proof, I might need to utilize a config map that will hold the current status of the renaming process and take that into consideration every time the controller notices the VM that has to be renamed.

pkg/virt-controller/watch/vm.go
```diff
@@ -271,7 +276,94 @@ func (c *VMController) execute(key string) error {
...
+// First return value is a boolean indicating if the controller should retry the request
+func (c *VMController) handleVMRenameCreateRequest(vm *virtv1.VirtualMachine, newName string) (bool, error) {
+	// Check if a VM named 'newName' exists
+	_, err := c.clientset.VirtualMachine(vm.Namespace).Get(newName, &v1.GetOptions{})
+
+	if err != nil {
+		// If the VM existence could not be determined, retry
+		if !strings.Contains(err.Error(), "not found") {
+			return true, err
+		} else {
```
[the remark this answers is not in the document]
omeryahud: Yes it will, fixed

pkg/virt-controller/watch/vm.go
```diff
@@ -271,7 +276,94 @@ func (c *VMController) execute(key string) error {
...
+}
+
+// Handles VM rename delete requests
+// First return value is a boolean indicating if the controller should retry the request
+func (c *VMController) handleVMRenameDeleteRequest(vm *virtv1.VirtualMachine, oldName string) (bool, error) {
+	// Attempt deletion of the VM with the old name
+	err := c.clientset.VirtualMachine(vm.Namespace).Delete(oldName, &v1.DeleteOptions{})
+
+	if err != nil && !strings.Contains(err.Error(), "not found") {
+		return true, err
```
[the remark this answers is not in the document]
omeryahud: Also fixed

pkg/virt-controller/watch/vm.go
omeryahud: I would like to know your opinion on this solution which I think would solve the issue:
https://github.com/kubevirt/kubevirt/pull/3036#discussion_r381870477

mhenriks: I think it may make sense verify a field in the vm before it is deleted.  However, don't you possibly have a chicken/egg situation or race condition?  Isn't the `RenameCreateRequest` removed from the source vm right after the new vm is created?

omeryahud: You are right, I think that was the original reason why I haven't checked for a field in the old vm before (I knew I had a reason for not doing that, but forgot what it was).
I'll think about it some more and try coming up with a proper solution

pkg/virt-api/rest/subresource.go
```diff
@@ -520,6 +527,97 @@ func (app *SubresourceAPIApp) RestartVMRequestHandler(request *restful.Request,
...
+			"newName": opts.NewName,
+		},
+	})
+
+	if err != nil {
+		writeError(errors.NewInternalError(err), response)
+		return
+	}
+
+	_, err = app.virtCli.VirtualMachine(namespace).Patch(name, types.JSONPatchType, []byte(renameRequestJson))
```
mhenriks: Per the resolved convo above:  I forget the details of how the different patch types work, but some (I think json patch) will replace lists rather than add elements to the list.  Have you confirmed (is there a test) where a rename request is added to a non empty list?

pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go
```diff
@@ -140,6 +140,16 @@ func (admitter *VMsAdmitter) authorizeVirtualMachineSpec(ar *v1beta1.AdmissionRe
 		}
 	}
 
+	if len(causes) > 0 {
+		return causes, nil
+	}
+
+	causes = validateNoModificationsDuringRename(ar, vm)
+
+	if len(causes) > 0 {
```
mhenriks: this statement seems unnecessary

omeryahud: I put it there so when another person adds code to this function, they can just add code above the return line, and won't have to deal with the error handling of my code. It feels more comfortable to me this way, we can discuss further if it still bothers you

pkg/virt-api/rest/subresource.go
omeryahud: I am using `getChangeRequestJson` (a few lines above), which prevents adding a `RenameRequest` to `Status.StateChangeRequests` if its not empty.
That is a valid scenario because I don't want to rename a VM that is scheduled to run because it might have an active VMI by the time renaming is done, and I don't want to rename a VM that is scheduled to stop for the same reason.
Because of that I think there is no need to add a `RenameRequest` to the list, only replace the list when its empty.

pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go
mhenriks: If that is your intention, it should probably be made explicit with a comment.  Obviously, this is not a huge deal but it looks kind of weird otherwise.

omeryahud: Sounds fair, I will add a comment to make it more clear, thanks!

stu-gott:
@omeryahud "make apidocs" is failing?

omeryahud:
> @omeryahud "make apidocs" is failing?

Travis fails on:
`Failed to get resource: GET. [HTTP HTTP/1.1 504 Gateway Time-out: https://jcenter.bintray.com/com/fasterxml/jackson/core/jackson-core/2.8.4/jackson-core-2.8.4.pom]`
I just checked the URL manually and it seems to work, I'll test it again...

omeryahud:
@stu-gott @mhenriks 
Could you guys take a final look please?

omeryahud:
Rebased on new master

PR #3037, opened 2020-01-30
Add dockerfile ci

dhiller:

Is _required_ to get openshift/release#6146 in. Checks fixes for @sallyom 's add-dockerfile-ci PR kubevirt/kubevirt#2991, as I don't want to mess with her branch.

As discussed with @rmohr we should clean up on this, and the cleanup should not be deferred that long. I.e. it looks like there is a couple of things we could remove if we extend from the builder image that has been updated recently. But as we need to get this in in order to be able to test this with the other PR openshift/release#6146 I'd be fine to merge this and create an issue for the cleanup.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/ci/Dockerfile.ci
+++ b/hack/ci/Dockerfile.ci
@@ -90,6 +90,7 @@ RUN yum remove -y git && \
     yum install -y git && \
     yum clean all && rm -rf /var/cache/yum/*
 
+# TODO: remove patching of external provider, after kubevirtci#199 has been merged and kubevirt updates
 RUN \
     export DOCKER_PREFIX='dhiller' && \
     export DOCKER_TAG="latest" && \
@@ -98,7 +99,7 @@ RUN \
     export GOPATH="/go" && \
     export GOBIN="/usr/bin" && \
     source /etc/profile.d/gimme.sh && \
-    curl -L -O -o cluster-up/cluster/external/provider.sh https://raw.githubusercontent.com/dhiller/kubevirtci/fix-external-provider/cluster-up/cluster/external/provider.sh && \
+    (cd cluster-up/cluster/external && curl -L -O -o provider.sh https://raw.githubusercontent.com/dhiller/kubevirtci/fix-external-provider/cluster-up/cluster/external/provider.sh) && \
     bash -x ./hack/build-manifests.sh && \
     bash -x ./hack/build-func-tests.sh && \
     curl -L -O -o kvm-ds.yaml https://raw.githubusercontent.com/kubevirt/kubernetes-device-plugins/master/manifests/kvm-ds.yml
--- a/hack/ci/Dockerfile.ci
+++ b/hack/ci/Dockerfile.ci
@@ -1,0 +1,129 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
+
+# Install packages
+# dnf copr enable -y @virtmaint-sig/virt-preview with yum?
+# python3-pip-9.0.3-5.el7.noarch \
+RUN yum install -y yum-plugin-copr && \
+    yum copr enable -y vbatts/bazel epel-7 && \
+    yum -y install \
+        libvirt-devel \
+        bazel \
+        cpio \
+        patch \
+        make \
+        git \
+        mercurial \
+        sudo \
+        gcc \
+        gcc-c++ \
+        glibc-devel \
+        findutils \
+        qemu-img \
+        protobuf-compiler \
+        python3-devel \
+        python3-pip \
+        redhat-rpm-config \
+        jq \
+        unzip && \
+    yum clean all && rm -rf /var/cache/yum/*
+
+ENV GIMME_GO_VERSION=1.12.8
+
+RUN mkdir -p /gimme && curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | HOME=/gimme bash >> /etc/profile.d/gimme.sh
+
+ENV GOPATH="/go" GOBIN="/usr/bin"
+
+# Install persistent go packages
+RUN \
+    mkdir -p /go && \
+    source /etc/profile.d/gimme.sh && \
+    # Install goimports
+    go get -u -d golang.org/x/tools/cmd/goimports && \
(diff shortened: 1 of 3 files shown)


dhiller:
@sallyom I meant this one :)

dhiller:
As discussed with @rmohr we should clean up on this, and the cleanup should not be deferred that long. I.e. it looks like there is a couple of things we could remove if we extend from the builder image that has been updated recently. But as we need to get this in in order to be able to test this with the other PR I'd be fine to merge this and create an issue for the cleanup.

hack/ci/test.sh
```diff
@@ -0,0 +1,61 @@
...
+export GIMME_GO_VERSION=1.12.8
+export GOPATH="/go"
+export GOBIN="/usr/bin"
+source /etc/profile.d/gimme.sh
+echo "checking configuration location"
+echo "KUBECONFIG: ${KUBECONFIG}"
+# enable nested-virt
+oc get machineset -n openshift-machine-api -o json >/tmp/machinesets.json
+MACHINE_IMAGE=$(jq .items[0].spec.template.spec.providerSpec.value.disks[0].image /tmp/machinesets.json)
+NESTED_VIRT_IMAGE="sotest-rhcos-nested-virt"
```
rmohr: No expert here, but I expected that we could in openshift-CI pre-select this and directly get the right cluster. Can't that be solved in https://github.com/openshift/release/pull/6932?

Btw. where is the image coming from?

hack/ci/wandisco-git.repo
```diff
@@ -0,0 +1,6 @@
+[wandisco-git]
```
rmohr: This is really weird.

hack/ci/Dockerfile.ci
```diff
@@ -0,0 +1,130 @@
...
+    cd $GOPATH/src/k8s.io/code-generator/cmd/deepcopy-gen && \
+    git checkout kubernetes-1.13.4 && \
+    go install && \
+    # Install defaulter-gen
+    cd $GOPATH/src/k8s.io/code-generator/cmd/defaulter-gen && \
+    git checkout kubernetes-1.13.4 && \
+    go install && \
+    # Install openapi-gen
+    cd $GOPATH/src/k8s.io/kube-openapi/cmd/openapi-gen && \
+    git checkout c59034cc13d587f5ef4e85ca0ade0c1866ae8e1d && \
```
qinqon: What about doing all this at a bash script and use COPY directive ? it's going to make a layer too, well maybe two one for the script and one for the run of the script.

hack/ci/Dockerfile.ci
```diff
@@ -0,0 +1,130 @@
...
+    go install && \
+    # Install protoc-gen-go
+    cd $GOPATH/src/github.com/golang/protobuf/protoc-gen-go && \
+    git checkout 1643683e1b54a9e88ad26d98f81400c8c9d9f4f9 && \
+    go install
+
+
+RUN \
+    source /etc/profile.d/gimme.sh && \
+    go get github.com/mattn/goveralls && \
```
qinqon: We don't update this one ?

hack/ci/test.sh
dhiller: I think I remember that the template @sallyom created was for another use case where only the `src` image type was required. So I figured she chose not to create another template and to enable nested virt directly instead.

hack/ci/wandisco-git.repo
dhiller: Is there a better approach (besides using HEREDOC) ?

hack/ci/Dockerfile.ci
dhiller: What is the advantage of doing this in a bash script over doing it here? I would find it easier to do the cleanup later on when comparing it to the [builder file](https://github.com/kubevirt/kubevirt/blob/master/hack/builder/Dockerfile) if we'd leave it like this.

hack/ci/wandisco-git.repo
rmohr: I just meant that we have to update git from somewhere ...

hack/ci/test.sh
rmohr: hm, that pulls in a lot of responsibility into the kubevirt/kubevirt codebase and probably increases the test time drastically. I guess we will need our own template then, or an additional parameter ...

But ok for now.

hack/ci/Dockerfile.ci
dhiller: Sorry, I don't understand?

hack/ci/test.sh
dhiller: I'll add this to the cleanup issue.

hack/ci/wandisco-git.repo
dhiller: Oh, yes, you are right.

hack/ci/test.sh
dhiller: The image currently used is [this one](https://github.com/openshift/release/blob/master/ci-operator/templates/openshift/installer/cluster-launch-installer-custom-test-image.yaml). In the cleanup I can move the setup code for enabling nested virt there.

EDIT: I meant I'll create a template from that and move the setup code to the copy.

rmohr:
/lgtm
/approve

/hold
let's ensure we have the cleanup issue. Can go in afterwards.

dhiller:
Created #3048 for tracking the cleanup. Please review if I touched all areas required @rmohr

hack/ci/test.sh
sallyom: right, I tested with a stock template, rather than create a new one.  We can add a new template that does the enabling of nested virt later, though.  The idea here I _think_ is to get the tests running, then optimize after.

hack/ci/wandisco-git.repo
sallyom: : ) git in centos7 is old :woman_shrugging: originally you were basing off of fedora and there git version was later.

sallyom:
also the `somalley-nested-virt` image used here will be replaced with a more official image.  I'll work on that asap.

sallyom:
this is proof-of-concept to be cleaned up in a follow-up : )

PR #3039, opened 2020-01-31
User filesystem list subresource

petrkotas:

Extends the feature introduced in pr #2827 by exposing the 
full user and filesystem list as additional VMI subresources.

The guest agent data is returned from the api endpoint
```
/apis/subresources.kubevirt.io/v1alpha3/namespaces/{namespace}/virtualmachineinstances/{name}/userlist
/apis/subresources.kubevirt.io/v1alpha3/namespaces/{namespace}/virtualmachineinstances/{name}/filesystemlist
```

Accessing through client-go is possible through new functions:

```
virtClient.VirtualMachineInstance(namespace).UserList(vmiName)
virtClient.VirtualMachineInstance(namespace).FilesystemList(vmiName)
```

The data for userlist looks like:
```json
{
    "items": [                                                                                            
        {                                                                                                 
            "loginTime": 1580467675.876078,                                                               
            "userName": "fedora"                                                                          
        }                                                                                                 
    ],                                                                                                    
    "metadata": {}
}
```

User LoginTime is is in fractional seconds since epoch time. It is left for consumer to convert to desired format.

The data for filesystem looks like:
```json
{
    "items": [
        {
            "diskName": "vda1",
            "fileSystemType": "ext4",
            "mountPoint": "/",
            "totalBytes": 3927900160,
            "usedBytes": 1029201920
        }
    ],
    "metadata": {}
}
```

**Release note**:

```release-note
Add ability to list all active users and all active filesystems on the guest machine.
```


petrkotas:
Hi Daniel and Daniel, would you please review this follow up PR for guest agent data? Thank you.

pkg/virt-launcher/virtwrap/cmd-server/server.go
```diff
@@ -341,6 +341,60 @@ func (l *Launcher) GetGuestInfo(ctx context.Context, request *cmdv1.EmptyRequest
 	return response, nil
 }
 
+// GetUsers returns the list of active users on the guest machine
+func (l *Launcher) GetUsers(ctx context.Context, request *cmdv1.EmptyRequest) (*cmdv1.GuestUserListResponse, error) {
```
dhiller: Why is the error never returned here?

pkg/virt-launcher/virtwrap/cmd-server/server.go
```diff
@@ -341,6 +341,60 @@ func (l *Launcher) GetGuestInfo(ctx context.Context, request *cmdv1.EmptyRequest
...
+		return response, nil
+	} else {
+		response.GuestUserListResponse = string(jUsers)
+	}
+
+	return response, nil
+}
+
+// GetFilesystems returns a full list of active filesystems on the guest machine
+func (l *Launcher) GetFilesystems(ctx context.Context, request *cmdv1.EmptyRequest) (*cmdv1.GuestFilesystemsResponse, error) {
```
dhiller: Why is the error never returned here?

pkg/virt-launcher/virtwrap/cmd-server/server_test.go
```diff
@@ -179,6 +179,38 @@ var _ = Describe("Virt remote commands", func() {
...
+				},
+			}
+
+			domainManager.EXPECT().GetUsers().Return(userList, nil)
+
+			fetchedList, err := client.GetUsers()
+			Expect(err).ToNot(HaveOccurred(), "should fetch users without any issue")
+			Expect(fetchedList.Items).To(Equal(userList), "fetched list should be the same")
+		})
+
```
dhiller: Maybe add unit tests for error cases? I.e. domainManager.getUsers() fails or json deserialization fails?

pkg/virt-launcher/virtwrap/cmd-server/server_test.go
```diff
@@ -179,6 +179,38 @@ var _ = Describe("Virt remote commands", func() {
...
+				},
+			}
+
+			domainManager.EXPECT().GetFilesystems().Return(fsList, nil)
+
+			fetchedList, err := client.GetFilesystems()
+			Expect(err).ToNot(HaveOccurred(), "should fetch filesystems without any issue")
+			Expect(fetchedList.Items).To(Equal(fsList), "fetched list should be the same")
+		})
 	})
```
dhiller: Maybe add unit tests for error cases? I.e. domainManager.getFilesystems() fails of json deserialization fails?

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -1455,3 +1457,37 @@ func (l *LibvirtDomainManager) GetGuestInfo() (v1.VirtualMachineInstanceGuestAge
 
 	return guestInfo, nil
 }
+
+// GetUsers return the full list of users on the guest machine
+func (l *LibvirtDomainManager) GetUsers() ([]v1.VirtualMachineInstanceGuestOSUser, error) {
```
dhiller: Error is never returned here?

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -1455,3 +1457,37 @@ func (l *LibvirtDomainManager) GetGuestInfo() (v1.VirtualMachineInstanceGuestAge
...
+			Domain:    user.Domain,
+			LoginTime: user.LoginTime,
+		})
+	}
+
+	return userList, nil
+}
+
+// GetFilesystems return the full list of filesystems on the guest machine
+func (l *LibvirtDomainManager) GetFilesystems() ([]v1.VirtualMachineInstanceFileSystem, error) {
```
dhiller: Error is never returned here?

staging/src/kubevirt.io/client-go/kubecli/vmi_test.go
```diff
@@ -256,6 +256,41 @@ var _ = Describe("Kubevirt VirtualMachineInstance Client", func() {
 		Expect(fetchedInfo).To(Equal(osInfo), "fetched info should be the same as passed in")
 	})
 
+	It("should fetch UserLIst from VirtualMachineInstance via subresource", func() {
```
dhiller: nit: spelling s/UserLIst/UserList/

tests/vmi_configuration_test.go
```diff
@@ -1020,9 +1020,36 @@ var _ = Describe("Configurations", func() {
 			})
 		})
 
-		Context("[rfe_id:140][crit:medium][vendor:cnv-qe@redhat.com][level:component]with guestAgent", func() {
+		FContext("[rfe_id:140][crit:medium][vendor:cnv-qe@redhat.com][level:component]with guestAgent", func() {
```
dhiller: Please remove focus here.

tests/vmi_configuration_test.go
```diff
@@ -1020,9 +1020,36 @@ var _ = Describe("Configurations", func() {
...
 			var agentVMI *v1.VirtualMachineInstance
 
+			prepareAgentVM := func() *v1.VirtualMachineInstance {
+				// TODO: actually review this once the VM image is present
```
dhiller: What's up with this?

tests/vmi_configuration_test.go
```diff
@@ -1196,6 +1184,47 @@ var _ = Describe("Configurations", func() {
...
+						return false
+					}
+					if len(userList.Items) > 0 && userList.Items[0].UserName == "fedora" {
+						return true
+					}
+					return false
+				}, 240*time.Second, 2).Should(BeTrue(), "Should have fedora users")
+			})
+
+			It("should return user list", func() {
```
dhiller: replace user with filesystem

pkg/virt-launcher/virtwrap/cmd-server/server.go
petrkotas: It follows the same pattern as the rest of the file. The error is returned as part of the message.

pkg/virt-launcher/virtwrap/cmd-server/server.go
petrkotas: It follows the same pattern as the rest of the file. The error is returned as part of the message.

pkg/virt-api/rest/subresource.go
```diff
@@ -816,3 +816,85 @@ func (app *SubresourceAPIApp) GuestOSInfo(request *restful.Request, response *re
...
+		return
+	}
+
+	response.WriteEntity(userList)
+}
+
+// FilesystemList handles the subresource for providingguest filesystem list
+func (app *SubresourceAPIApp) FilesystemList(request *restful.Request, response *restful.Response) {
+	validate := func(vmi *v1.VirtualMachineInstance) *errors.StatusError {
+		if vmi == nil || vmi.Status.Phase != v1.Running {
```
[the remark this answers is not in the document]
petrkotas: I do believe the code is more readable written this way.

pkg/virt-launcher/virtwrap/cmd-server/server.go
dhiller: OK, but what I don't understand is then if the error is never returned, why is it declared as a return value which effectively is always `nil`? Can you point me towards why this has to be like this please?

tests/vmi_configuration_test.go
petrkotas: This comment is not mine. My undrstanding is, that commented code will get deprecated once the VM image with GA installed is present.

pkg/virt-api/rest/subresource.go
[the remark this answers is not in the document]
petrkotas: If you agree I leave it like this.

pkg/virt-launcher/virtwrap/cmd-server/server_test.go
petrkotas: The tests cover the regression when it happens. I do not see the value of adding the negative test case. Would you agree to skip it?

pkg/virt-launcher/virtwrap/cmd-server/server_test.go
petrkotas: The tests cover the regression when it happens. I do not see the value of adding the negative test case. Would you agree to skip it?

pkg/virt-launcher/virtwrap/cmd-server/server.go
petrkotas: I do believe that the format of the command is dictated by the gRPC protocol used to communicate these functions to the virt-handler.

Therefore the code does not have another choice but to comply.

pkg/virt-launcher/virtwrap/manager.go
petrkotas: It was a design choice. There is no error to be returned. 
However, having the error returned, the usage is exactly the same as for the other functions, which makes reading the code easier.

I can remove it if this is something that have to be changed.

pkg/virt-launcher/virtwrap/manager.go
petrkotas: It was a design choice. There is no error to be returned. 
However, having the error returned, the usage is exactly the same as for the other functions, which makes reading the code easier.

I can remove it if this is something that have to be changed.

tests/vmi_configuration_test.go
dhiller: OK then. Although I'd love to rather see an issue referenced here.

pkg/virt-launcher/virtwrap/cmd-server/server.go
dhiller: I still don't understand why the code follows a pattern of returning an error embedded in a message even though there's a return value present that is much better suited to return it. Is the error message checked for the error somehow?

IMHO tthis should be cleaned up as this is a case of [stringly typed](https://wiki.c2.com/?StringlyTyped). But not in this PR then :-)

dhiller: Created #3127 for this.

dhiller:
Looks like a `make generate` is required.

pkg/virt-launcher/virtwrap/cmd-server/server.go
slintes: The grpc/protobuf implementation generates the func signature with err even though we don't declare it (see cmd.proto and cmd.pb.go). Also, it uses the error in case something goes wrong during the grpc call itself.
We, on the other hand, only use our own declared response type for reporting non-grpc related errors. It is stringly typed for sure, but IIRC it has always been this way, even in the pre-grpc implementation. If not, blame me, I introduced grpc a while ago ;) However, I don't say they is no room for improvement, just wanted give some more context :)

fangzhixiang:
filesystemlist api cannot get totalBytes and usedBytes, why? Is it any preconditions before use this api interface?
image

PR #3040, opened 2020-01-31
Refactor migration unit test

petrkotas:

Refactor the is migratable storage check in unit tests.

Fixes #

/cc @phoracek 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-handler/vm_test.go
+++ b/pkg/virt-handler/vm_test.go
@@ -835,7 +835,8 @@ var _ = Describe("VirtualMachineInstance", func() {
 		}, 3)
 	})
 
-	Context("check VMI disks for migration", func() {
+	Context("check if migratable", func() {
+
 		var testBlockPvc *k8sv1.PersistentVolumeClaim
 
 		BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1152,82 @@ var _ = Describe("VirtualMachineInstance", func() {
...
+							Masquerade: &v1.InterfaceMasquerade{},
+						},
+					},
+				}
+
+				err := controller.checkNetworkInterfacesForMigration(vmi)
+				Expect(err).ToNot(HaveOccurred(), "should not block migration for masquerade binding method")
+			})
+
+			It("should not block migration for bridge interface assigned to multus network", func() {
```
EdDev, line 1205: How about "should pass migration for bridge binding method on the secondary network"

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1152,82 @@ var _ = Describe("VirtualMachineInstance", func() {
...
+				vmi := v1.NewMinimalVMI("testvmi")
+				interface_name := "interface_name"
```
EdDev, line 1207: It will be nice to recognize that there are two interfaces, one primary and one secondary.
Something like this:
```
primaryInterface = "nic0"
secondaryInterface = "nic1"
```

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1152,82 @@ var _ = Describe("VirtualMachineInstance", func() {
...
+						Name:          "pod",
+						NetworkSource: v1.NetworkSource{Pod: &v1.PodNetwork{}},
+					},
+					v1.Network{
+						Name:          interface_name,
+						NetworkSource: v1.NetworkSource{Multus: &v1.MultusNetwork{}},
+					},
+				}
+				vmi.Spec.Domain.Devices.Interfaces = []v1.Interface{
+					v1.Interface{
```
EdDev, line 1220: I think the primary interface is missing for completion of the scenario.

pkg/virt-handler/vm_test.go
```diff
@@ -1151,6 +1152,82 @@ var _ = Describe("VirtualMachineInstance", func() {
...
+
+				err := controller.checkNetworkInterfacesForMigration(vmi)
+				Expect(err).ToNot(HaveOccurred(), "should not block migration for masquerade binding method")
+			})
+
+			It("should not block migration for bridge interface assigned to multus network", func() {
+				vmi := v1.NewMinimalVMI("testvmi")
+				interface_name := "interface_name"
+
+				vmi.Spec.Networks = []v1.Network{
```
EdDev, line 1209: The tests seem to have a similar pattern, could you please create two helpers to build the spec? One for the Networks and one for the Interfaces. It can be even one if you have a good name for it.

petrkotas:
I do believe this PR is no longer required. It was needed for a different purpose, but it was resolved long time ago.

Closing.

PR #3044, opened 2020-02-04
Add basic alert rules for virt-operator

danielBelenky:

In order to properly monitor the state of kubevirt deployment, we want to provide
alert rules that will tell the operator of the system that something it wrong.

This PR features a subset of the metrics that we want to collect.
Only metrics that we want to collect and can be collected w/o any
changes to virt-operator itself are included here.

-->
```release-note
Added alert rules to monitor that virt-operator application is up
```

### DIFF
--- a/pkg/virt-operator/creation/components/crds.go
+++ b/pkg/virt-operator/creation/components/crds.go
@@ -415,6 +415,68 @@ func NewPrometheusRuleSpec(ns string) *promv1.PrometheusRuleSpec {
 							"summary": "More than 80% of the rest calls failed in virt-controller for the last 5 minutes",
 						},
 					},
+					{
+						Record: "num_of_running_virt_operators",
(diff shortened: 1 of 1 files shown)


stu-gott:
follow up from offline conversation: the trouble with functional tests for this alert in particular is that we'd need to disrupt the cluster and wait for possibly up to 30 minutes. not really a great candidate for CI as that can be verified manually more easily.

danielBelenky:
> follow up from offline conversation: the trouble with functional tests for this alert in particular is that we'd need to disrupt the cluster and wait for possibly up to 30 minutes. not really a great candidate for CI as that can be verified manually more easily.

It doesn't have to be manual verification, but just not run as part of the CI gate to merge a new PR (just because it takes too long to verify it). But, with that said, I have opened a task to integrate unit-tests for the rules.

PR #2944, opened 2019-12-19
[RFC] Add build and deployment support for ppc64le arch

murilofv:
This commit adds support in Kubevirt for building, deploying and testing
VMs for IBM Power Systems little-endian architecture (ppc64le).

A .bazelrc file was created in order to move common and architecture
dependent options to it, this makes it easier to add new architectures
support in the future, the architecture dependent options are selected at
runtime by checking the output of uname -m and passing it to the --config
option added. The .bazelrc file generated at runtime by the CI was renamed
to ci.bazelrc and made and untracked file, the .bazelrc added attempts to
import the ci.bazelrc in case it exists.

The WORKSPACE file now contains pointers to the ppc64le specific
dependencies in order to properly build kubevirt and a select was added to
the BUILD.bazel files that use them to use the appropriate ones depending
on the architecture in which Kubevirt is build, the default dependencies
for an unmatched architecture is to use the amd64 ones that were already
defined by Kubevirt.

As part of the WORKSPACE a ppc64le binary puller was added given that
after the migration of rules_docker to a Go implementation the support for
container_pull was broken for architectures that are not amd64, due to the
binary the tool uses. So whenever pulling a container in ppc64le the
rebuilt puller is used by pointing it in the puller_linux attribute.

The .travis.yml file was modified to add ppc64le arch as well, but for the
deploy steps of it they are currently only done in amd64. Some of the tests
had their expected output adjusted to support this new arch. Two test cases
to note are the ones that check privileged mode, for now in Power we are
running in it so these two test cases are currently being skipped.

Signed-off-by: Murilo Fossa Vicentini <muvic@linux.ibm.com>

**What this PR does / why we need it**: This PR adds support to ppc64le architecture and does some modifications on the current upstream infrastructure in order to 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2144 

**Special notes for your reviewer**:
~There are still some pending concerns / feedback that I would like to hear from the community.
1- The patch changes from io_bazel_rules_docker from release 0.12.0 to 0.10.0 for amd64 as well, so there is the need to make sure it didn't break anything when doing that, using the latest release breaks the build in Power, I still need to get a better understanding as to why, I bisected to a commit in rules_docker where things broke, but it is still not clear to me the reason why and how to address it.~

[...]

PR #3045, opened 2020-02-04
Add support for building a multi-arch builder manifest

rmohr:

Build a ppc64le and amd64 multi-arch image with qemu-user on amd64
nodes.

The multi-arch image can be created for all architectures on amd64 machines. This works by injecting qemu-user binaries into the container which will be used by the kernel to run the binaries from foreign achritectures.

A prerequisite to bring in #2944

Fixes #

```release-note
NONE
```


hack/builder/build.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -e
```
danielBelenky: Maybe add `-x` too?

hack/builder/publish.sh
```diff
@@ -7,5 +8,12 @@ SCRIPT_DIR="$(
...
-docker push docker.io/kubevirt/builder:${VERSION}
+for ARCH in ${ARCHITECTURES}; do
+    docker push kubevirt/builder:${VERSION}-${ARCH}
+    TMP_IMAGES="${TMP_IMAGES} kubevirt/builder:${VERSION}-${ARCH}"
+done
+
+export DOCKER_CLI_EXPERIMENTAL=enabled
+docker manifest create --amend kubevirt/builder:${VERSION} ${TMP_IMAGES}
+docker manifest push kubevirt/builder:${VERSION}
+rm manifests/ -rf
```
danielBelenky: This will not be called if we we create the manifest and fail to push it

hack/builder/publish.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -e
```
danielBelenky: Same here

hack/builder/Dockerfile
```diff
@@ -1,4 +1,7 @@
-FROM fedora:30
+ARG ARCH
+FROM ${ARCH}/fedora:30
```
danielBelenky: is there a technical reason to use docker here? is it possible to use centos 8 instead so that we'll not have to worry about image being removed from dockerhub?

rmohr: are there plans from the fedora people to not push images to dockerhub anymore?

hack/builder/publish.sh
rmohr: I can move the cleanup into a trap ...

hack/builder/build.sh
rmohr: Can do that ...

hack/builder/Dockerfile
danielBelenky: no, but since a new fedora version comes out every 6 month, and they don't keep all the versions on dockerhub, we will have to update the image here every now and then... not too often, but with centos we won't have to update for several years...

rmohr: True. We can discuss that. But maybe not in this PR?

PR #3050, opened 2020-02-05
Removes superfluous calls to res.WriteHeader()

omeryahud:

Removes the unnecessary calls to `response.WriteHeader()` to make `virt-api` and `virt-handler` logs cleaner.
The `net/http` package returns `http.StatusOK` as a default if a different status code was not explicitly 
written, this means that every call to `res.WriteHeader(http.StatusOK)` writes an info log like the two below and floods the logs for `virt-api` and `virt-handler`.

`virt-api` and `virt-handler` logs are being flooded because of unnecessary calls to `response.WriteHeader()` 

**Notes for the reviewer**
The logs that are flooding `virt-api`:
```
`{"component":"virt-api","level":"info","msg":"http: superfluous response.WriteHeader call from kubevirt.io/kubevirt/pkg/virt-api/webhooks/mutating-webhook.serve (mutating-webhook.go:68)\n","pos":"server.go:3012","timestamp":"2020-02-05T12:43:31.454691Z"}
{"component":"virt-api","level":"info","msg":"http: superfluous response.WriteHeader call from kubevirt.io/kubevirt/pkg/util/webhooks/validating-webhooks.Serve (validating-webhook.go:61)\n","pos":"server.go:3012","timestamp":"2020-02-05T12:43:31.458961Z"}`
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-api/api.go
+++ b/pkg/virt-api/api.go
@@ -278,9 +278,7 @@ func (app *virtAPIApp) composeSubresources() {
 			Doc("Open a websocket connection to connect to VNC on the specified VirtualMachineInstance."))
 
 		subws.Route(subws.GET(rest.ResourcePath(subresourcesvmiGVR) + rest.SubResourcePath("test")).
-			To(func(request *restful.Request, response *restful.Response) {
-				response.WriteHeader(http.StatusOK)
-			}).
+			To(func(request *restful.Request, response *restful.Response) {}).
 			Param(rest.NamespaceParam(subws)).Param(rest.NameParam(subws)).
 			Operation("test").
 			Doc("Test endpoint verifying apiserver connectivity."))
--- a/pkg/virt-api/api.go
+++ b/pkg/virt-api/api.go
@@ -277,6 +277,7 @@ func (app *virtAPIApp) composeSubresources() {
 			Operation("vnc").
 			Doc("Open a websocket connection to connect to VNC on the specified VirtualMachineInstance."))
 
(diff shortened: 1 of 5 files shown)


omeryahud:
/hold
Will check if there are other superfluous calls to `resp.WriteHeader()`

pkg/virt-api/api.go
```diff
@@ -277,10 +277,9 @@ func (app *virtAPIApp) composeSubresources() {
 			Operation("vnc").
 			Doc("Open a websocket connection to connect to VNC on the specified VirtualMachineInstance."))
 
+		// An empty handler function would respond with HTTP OK by default
 		subws.Route(subws.GET(rest.ResourcePath(subresourcesvmiGVR) + rest.SubResourcePath("test")).
-			To(func(request *restful.Request, response *restful.Response) {
-				response.WriteHeader(http.StatusOK)
-			}).
+			To(func(request *restful.Request, response *restful.Response) {}).
```
slintes: At this place I find it better readable to explicitly set the OK status... but that might be a matter of taste

omeryahud: I agree, that's why I added the comment above, do you think it's worth putting it back, I don't mind either way after I read your explanation (which is great, thanks!).

slintes: ah I missed the comment. Works for me as it is then :)

rmohr:
@omeryahud thx!

PR #3051, opened 2020-02-05
Add bump-kubevirtci make target

oshoval:

Add bump-kubevirtci make target

```release-note
NONE
```


hack/bump-kubevirtci.sh
```diff
@@ -1,4 +1,4 @@
-#/bin/bash
+#!/bin/bash
```
oshoval: /cc @danielBelenky

danielBelenky: :)

PR #3053, opened 2020-02-06
Check new tests for flakes by running them repeatedly

dhiller:

In order to catch flaky tests **before** they get merged, we run new/changed tests a couple times to check whether they fail on any of these runs. This goes hand in hand with a new presubmit job
that will be created to call this.

In detail:
* `new_tests.sh` fetched tests added/changed since latest merge commit
* `test.sh` is modified to support parameterized focusing of tests
* `repeated_test.sh` puts these two together

Fixes #

/cc @danielBelenky @rmohr if you find the time...

**Release note**:

```release-note
NONE
```


dhiller:
IDK we had some discussion about not wanting to write bash scripts any more but as the existing script also was bash I decided to extend it. WDYT? Should I rather be writing go code and use the script?

automation/repeated_test.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+SCRIPT_PATH="$(
+    cd "$(dirname "$BASH_SOURCE[0]")/"
+    echo "$(pwd)/"
+)"
+
+NUM_TESTS=${NUM_TESTS:-3}
+if [[ -z ${NEW_TESTS} ]]; then
+    NEW_TESTS=$(${SCRIPT_PATH}/new_tests.sh)
```
danielBelenky: could you please double quote `SCRIPT_PATH`?

automation/repeated_test.sh
```diff
@@ -0,0 +1,29 @@
...
+fi
+
+if [[ -z ${NEW_TESTS} ]]; then
+    echo "Nothing to test"
+    exit 0
+fi
+
+declare -a TEST_LANES
```
danielBelenky: I think that the declare here is redundant

automation/repeated_test.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+SCRIPT_PATH="$(
+    cd "$(dirname "$BASH_SOURCE[0]")/"
+    echo "$(pwd)/"
+)"
+
+NUM_TESTS=${NUM_TESTS:-3}
+if [[ -z ${NEW_TESTS} ]]; then
```
danielBelenky: Personally, I usually prefer a script with a proper CLI instead of implicitly using env vars... I just think that it's less confusing... but, I know that this is just a personal preference so feel free to not address this comment.

danielBelenky:
> IDK we had some discussion about not wanting to write bash scripts any more but as the existing script also was bash I decided to extend it. WDYT? Should I rather be writing go code and use the script?

I don't think that we should write everything in Go... Both Bash and Golang are tools, each of them is good at something else so I don't think that we should stop writing bash scripts... while Go is great with concurrency, Bash is great with you need to do simple processing and and run other programs. But, I do think that we should give Bash scripts the honor that they deserve and treat a Bash script just as any other code we write in Go... Though we are far from that today, I hope that we'll be able to run ShellCheck on our scripts during CI.

dhiller:
> > IDK we had some discussion about not wanting to write bash scripts any more but as the existing script also was bash I decided to extend it. WDYT? Should I rather be writing go code and use the script?
> 
> I don't think that we should write everything in Go... Both Bash and Golang are tools, each of them is good at something else so I don't think that we should stop writing bash scripts... while Go is great with concurrency, Bash is great with you need to do simple processing and and run other programs. But, I do think that we should give Bash scripts the honor that they deserve and treat a Bash script just as any other code we write in Go... Though we are far from that today, I hope that we'll be able to run ShellCheck on our scripts during CI.
+1 for your thoughts about giving nah coffee the honor it deserves.

Just thought about that check myself. I have prepared a pre commit hook that runs shellcheck on all changed shell scripts before commit. Maybe we should just convert that into a check that we can then add to the set of checks we already run...

danielBelenky:
> 
> 
> > > IDK we had some discussion about not wanting to write bash scripts any more but as the existing script also was bash I decided to extend it. WDYT? Should I rather be writing go code and use the script?
> > 
> > 
> > I don't think that we should write everything in Go... Both Bash and Golang are tools, each of them is good at something else so I don't think that we should stop writing bash scripts... while Go is great with concurrency, Bash is great with you need to do simple processing and and run other programs. But, I do think that we should give Bash scripts the honor that they deserve and treat a Bash script just as any other code we write in Go... Though we are far from that today, I hope that we'll be able to run ShellCheck on our scripts during CI.
> 
> +1 for your thoughts about giving bash code the honor it deserves.
> 
> Just thought about that check myself. I have prepared a pre commit hook that runs shellcheck on all changed shell scripts before commit. Maybe we should just convert that into a check that we can then add to the set of checks we already run...

We can, but it means that we either have to refactor our scripts before this check is added or make this check not blocker for merges (and then it loses it's point). Maybe we should do it as part of a scripts refactoring task or at first stage run ShellCheck on a whitelisted subset of our scripts?

dhiller:
> > > > IDK we had some discussion about not wanting to write bash scripts any more but as the existing script also was bash I decided to extend it. WDYT? Should I rather be writing go code and use the script?
> > > 
> > > 
> > > I don't think that we should write everything in Go... Both Bash and Golang are tools, each of them is good at something else so I don't think that we should stop writing bash scripts... while Go is great with concurrency, Bash is great with you need to do simple processing and and run other programs. But, I do think that we should give Bash scripts the honor that they deserve and treat a Bash script just as any other code we write in Go... Though we are far from that today, I hope that we'll be able to run ShellCheck on our scripts during CI.
> > 
> > 
> > +1 for your thoughts about giving bash code the honor it deserves.
> > Just thought about that check myself. I have prepared a pre commit hook that runs shellcheck on all changed shell scripts before commit. Maybe we should just convert that into a check that we can then add to the set of checks we already run...
> 
> We can, but it means that we either have to refactor our scripts before this check is added or make this check not blocker for merges (and then it loses it's point). Maybe we should do it as part of a scripts refactoring task or at first stage run ShellCheck on a whitelisted subset of our scripts?

Maybe it's enough to just check changed files, then flag on warning level and make the check not required. I mean, a red x looks like there's something that should be fixed? In my experience those special purpose refactoring efforts normally never happen.

BTW, do I  need to create another PR or what else is wrong with the lane?

dhiller:
@danielBelenky @vatsalparekh I've updated the PR, can you please take another look?

automation/new_tests.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if [[ "$#" -eq 0 ]] || [[ -z "$1" ]]; then
+    TARGET_COMMIT=$(git log -1 --format=%H --merges)
+else
+    TARGET_COMMIT="$1"
+fi
+git diff --name-only "${TARGET_COMMIT}".. -- tests/ \
```
danielBelenky: would you mind adding a comment describing what this block does?
decoding the regex and all the pipes is not that straight forward... :)

automation/repeated_test.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+SCRIPT_PATH="$(
+    cd "$(dirname "${BASH_SOURCE[0]}")/" || exit
+    echo "$(pwd)/"
+)"
+
+NUM_TESTS=${NUM_TESTS:-3}
+if [[ -z ${NEW_TESTS} ]]; then
+    NEW_TESTS=$("${SCRIPT_PATH}"/new_tests.sh)
```
danielBelenky: could `new_tests.sh` become a function in this script? is there a reason to keep them separate?

automation/repeated_test.sh
```diff
@@ -0,0 +1,46 @@
...
+        make cluster-sync
+        ginko_params="--ginkgo.noColor --ginkgo.focus=${NEW_TESTS} --ginkgo.regexScansFilePath=true"
+        FUNC_TEST_ARGS="$ginko_params" make functest
+        if [[ $? -ne 0 ]]; then
+            echo "test lane: $lane, run: $i, tests failed!"
+            exit 1
+        fi
+    done
+
+    make cluster-down
```
danielBelenky: Since you have a trap for `EXIT`, this line here is redundant and will actually call `make cluster-down` twice

automation/repeated_test.sh
```diff
@@ -0,0 +1,46 @@
...
+    NEW_TESTS=$("${SCRIPT_PATH}"/new_tests.sh)
+    # skip certain tests for now, as we don't have a strategy currently
+    NEW_TESTS=$(echo "$NEW_TESTS" | sed -E 's/\|?[^\|]*(sriov|multus|windows|genie|gpu)[^\|]*//g')
+fi
+if [[ -z "${NEW_TESTS}" ]]; then
+    echo "Nothing to test"
+    exit 0
+fi
+
+TEST_LANES=( 'k8s-1.14.6' 'k8s-1.15.1' 'k8s-1.16.2' )
```
danielBelenky: Does it makes sense to have it configurable?

automation/repeated_test.sh
dhiller: Yes, but calling `cluster-down` twice should not be a problem, right? But if we don't call `cluster-down` here the loop would not work any more, as the next cluster-up (line 29) wouldn't work any more.

automation/repeated_test.sh
dhiller: I didn't make it configurable, thought about it at first, but don't see the need at the moment.

automation/new_tests.sh
dhiller: Makes sense.

automation/repeated_test.sh
dhiller: Yes, easier testing.

automation/repeated_test.sh
danielBelenky: you are right, I forgot the loop

automation/repeated_test.sh
dhiller: Made it into an argument.

automation/repeated_test.sh
dhiller: Done.

PR #3056, opened 2020-02-06
Allow setting of custom SELinux type for virt-launcher

stu-gott:
This PR introduces a ConfigMap entry (`selinuxLauncherType`) that overrides the built in SELinux type for KubeVirt.

This can be used to cause virt-launcher to run as `spc_t` or any other custom type the cluster admin might define.

**Release note**:
```release-note
Allow setting a custom SELinux type for virt-launcher.
```

### DIFF
--- a/cmd/virt-handler/virt-handler.go
+++ b/cmd/virt-handler/virt-handler.go
@@ -198,30 +198,6 @@ func (app *virtHandlerApp) Run() {
 		panic(err)
 	}
 
-	se, exists, err := selinux.NewSELinux()
-	if err == nil && exists {
-		for _, dir := range []string{app.VirtShareDir, app.VirtLibDir} {
-			if labeled, err := se.IsLabeled(dir); err != nil {
-				panic(err)
-			} else if !labeled {
-				err := se.Label("container_file_t", dir)
-				if err != nil {
-					panic(err)
-				}
-			}
-			err := se.Restore(dir)
-			if err != nil {
-				panic(err)
-			}
-		}
-		err = se.InstallPolicy("/var/run/kubevirt")
-		if err != nil {
-			panic(fmt.Errorf("failed to install virt-launcher selinux policy: %v", err))
-		}
-	} else if err != nil {
-		//an error occured
-		panic(fmt.Errorf("failed to detect the presence of selinux: %v", err))
-	}
 	// Create event recorder
 	app.virtCli, err = kubecli.GetKubevirtClient()
 	if err != nil {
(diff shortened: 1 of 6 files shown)


stu-gott:
/hold

I'd like to do more verification, and look into writing functional tests before merging this.

pkg/virt-config/config_test.go
```diff
@@ -288,4 +288,16 @@ var _ = Describe("ConfigMap", func() {
...
+
+	table.DescribeTable(" when SELinuxLauncherType", func(value string, result string) {
+		clusterConfig, _, _ := testutils.NewFakeClusterConfig(&kubev1.ConfigMap{
+			Data: map[string]string{virtconfig.SELinuxLauncherTypeKey: value},
+		})
+		selinuxLauncherType := clusterConfig.GetSELinuxLauncherType()
+		Expect(selinuxLauncherType).To(Equal(result))
+	},
+		table.Entry("when set, GetSELinuxLauncherType should return the value", "spc_t", "spc_t"),
+		table.Entry("when unset, GetSELinuxLauncherType should return the default", "", ""),
```
danielBelenky: I think that you are setting it explicitly to an empty value here instead of testing the default value... maybe pass the entire config in the table?

cmd/virt-handler/virt-handler.go
```diff
@@ -334,7 +311,37 @@ func (app *virtHandlerApp) Run() {
...
+					panic(err)
+				}
+			}
+			err := se.Restore(dir)
+			if err != nil {
+				panic(err)
+			}
+		}
+		// Only install KubeVirt's policy if not using a custom one
+		if selinuxLauncherType == "" {
```
danielBelenky: Maybe compare it to `se.DefaultSELinuxLauncherType`?

danielBelenky:
@stu-gott just 2 tiny nits.

rmohr:
Code looks apart from @danielBelenky's comments good, tests are missing.

stu-gott:
/hold cancel

functional test added

cmd/virt-handler/virt-handler.go
stu-gott: Thanks! It's always better to use a named value than a magic string--even an empty one.

rmohr: @stu-gott did you want to address something here?

tests/operator_test.go
```diff
@@ -1434,6 +1435,66 @@ spec:
...
+			By("Starting a New VMI")
+			vmi = tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskCirros))
+			vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+			Expect(err).ToNot(HaveOccurred())
+			tests.WaitForSuccessfulVMIStart(vmi)
+
+			By("Fetching virt-launcher Pod")
+			pod = tests.GetPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
+
+			By("Verifying SELinux context contains custom type")
```
rmohr: @stu-gott we should probably at least also check if we can log in to the machine?

PR #3058, opened 2020-02-09
Update test description to indicate its negative

danielBelenky:
Just a tiny addition to the description of a negative test case where we
check that a VMI with cloud-init secret is tarted when the secret itself
is missing.

**Release note**:

```release-note
NONE
```


danielBelenky:
@ILpinto fyi

Issue #3059, opened 2020-02-10
Invalid OpenAPI 2 specification


petrkotas:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:

When introduced a minimal change to the API(#3060) these tests started to fail:

* [test_id:836] Creating a VM with DataVolumeTemplates should succeed
* [test_id:837]deleting VM with cascade=true should automatically delete DataVolumes and VMI owned by VM
* [test_id:838]deleting VM with cascade=false should orphan DataVolumes and VMI owned by VM

with the following error:

```json
{
    "command": "/home/pkotas/Projects/redhat/upstream/kubevirt/_ci-configs/k8s-1.16.2/.kubectl -n kubevirt-test-default create -f /tmp/kubevirt-test-723165289/testvmifgpr866bxnvnr4r2k4dd7lpf6cqd4r7c857dlvv5rldp2dfx.json",
    "component": "tests",
    "level": "error",
    "msg": "command failed: cannot run command",
    "output": "",
    "pos": "utils.go:2957",
    "reason": "exit status 1",
    "stderr": "error: SchemaError(v1.ObjectMeta.creationTimestamp): definitions with multiple types aren't supported\n",
    "timestamp": "2020-02-10T13:39:55.050427Z"
}
```

The issue also happens for pure
```
kubectl create -f examples/vmi-fedora.yaml 

error: SchemaError(v1.VirtualMachineInstanceCondition.lastProbeTime): definitions with multiple types aren't supported
```

Both issues mean that the OpenAPI schema is invalid with types for field `v1.ObjectMeta.creationTimestamp`. The code indeed has multiple types present:
https://github.com/kubevirt/kubevirt/blob/9809e36b9f16c10d9068126bc4da3960f2712c59/pkg/util/openapi/openapi.go#L74

I understand this is a workaround for wrong CRDs validation as stated in the comment.

However, there is an invalid OpenAPI v2 present in the KubeVirt repository, which is causing these failures.

The full list of invalid points gathered from importing `api/openapi-spec/swagger.json` into the [online editor and validator](https://editor.swagger.io/) is here:

```
Structural error at info
should have required property 'version'
missingProperty: version
Jump to line 2

Semantic error at paths./apis/kubevirt.io/v1alpha3/namespaces/{namespace}/virtualmachineinstancemigrations
Declared path parameter "namespace" needs to be defined within every operation in the path (missing in "delete"), or moved to the path-level parameters object
[...]

PR #3060, opened 2020-02-11
Minimalist example of validation failure

petrkotas:
Minimalist example of API validation failure.

When adding new API subresource with return value containing Kubernetes metadata, validation fails due to multiple types in creationTimestamp field.

Without adding the object with problematic fields, the validation is correct.

Confirms #3059 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -3716,6 +3716,44 @@
      }
     }
    },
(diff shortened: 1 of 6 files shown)


rmohr:
@petrkotas isn't there some generated go code missing which should be used for validating the subresource? We don't use the swagger api json in any way. We just generate it for other uses.

PR #3062, opened 2020-02-12
Add autoattachSerialConsole flag

rmohr:
For completeness also allow opting out of the default serial console
device, like we allow for VNC.

For some usecases on some Operating Systems an unneeded device can have
unwanted side-effects. The most recent case we saw is Windows, where
every user in the Guest OS by default sees the serial console as an
unpluggable PCI device.

First this is confusing for the user and second the serial console does
not even serve any purpose on windows guests.

Fixes #

**Release note**:

```release-note
Add "autoattachSerialConsole" flag to the devices section to opt out of having the default serial console added to a VMI.
```

### DIFF
--- a/tests/console_test.go
+++ b/tests/console_test.go
@@ -178,7 +178,7 @@ var _ = Describe("[rfe_id:127][posneg:negative][crit:medium][vendor:cnv-qe@redha
 				_, err := virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
 				Expect(err).ToNot(HaveOccurred())
 
-				_, err = virtClient.VirtualMachineInstance(vmi.Namespace).SerialConsole(vmi.Name, 30*time.Second)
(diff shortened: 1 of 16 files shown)


tests/console_test.go
```diff
@@ -231,5 +231,37 @@ var _ = Describe("[rfe_id:127][posneg:negative][crit:medium][vendor:cnv-qe@redha
...
+				vmi = tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskAlpine))
+				f := false
+				vmi.Spec.Domain.Devices.AutoattachSerialConsole = &f
+			})
+
+			It("should create the vmi without any issue", func() {
+				tests.RunVMIAndExpectLaunch(vmi, 30)
+			})
+
+			It("should not have the  serial console in xml", func() {
```
stu-gott: whitespace

PR #3063, opened 2020-02-12
virt-handler should only update guest-agent details if interface doesn't have masquerade binding

vatsalparekh:

Currently `virt-handler` updates the VMI status with the interface data `guest-agent` provides when it is present. But, for VMI with `Masquerade` binding, this would update VMI with link-local IP address, which would not be a reachable address.

This PR changes this behavior so virt-handler would only update details from `guest-agent` on VMI status if interface does not have masquerade binding

This also unblocks https://github.com/kubevirt/kubevirt/pull/2963/, because both `virt-handler` and `virt-controller` would try to update VMI status when guest-agent is present without these patch.

**Fixes** # https://bugzilla.redhat.com/show_bug.cgi?id=1795889

**Release note**:

```release-note
virt-handler should only update guest-agent details if interface doesn't have masquerade binding
```

### DIFF
--- a/pkg/virt-handler/vm.go
+++ b/pkg/virt-handler/vm.go
@@ -344,10 +344,20 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 				domainInterfaceStatusByMac[domainInterfaceStatus.Mac] = domainInterfaceStatus
 			}
 
+			existingInterfacesSpecByName := map[string]v1.Interface{}
+			for _, existingInterfaceSpec := range vmi.Spec.Domain.Devices.Interfaces {
+				existingInterfacesSpecByName[existingInterfaceSpec.Name] = existingInterfaceSpec
+			}
+
 			// Iterate through all domain.Spec interfaces
 			for _, domainInterface := range domain.Spec.Devices.Interfaces {
 				interfaceMAC := domainInterface.MAC.MAC
 				var newInterface v1.VirtualMachineInstanceNetworkInterface
+				var isForwardingBindingInterface bool = false
+
+				if existingInterfacesSpecByName[domainInterface.Alias.Name].Masquerade != nil || existingInterfacesSpecByName[domainInterface.Alias.Name].Slirp != nil {
+					isForwardingBindingInterface = true
+				}
 
 				if existingInterface, exists := existingInterfaceStatusByName[domainInterface.Alias.Name]; exists {
 					// Reuse previously calculated interface from vmi.Status.Interfaces, updating the MAC from domain.Spec
@@ -365,9 +375,13 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 				// Update IP info based on information from domain.Status.Interfaces (Qemu guest)
 				// Remove the interface from domainInterfaceStatusByMac to mark it as handled
 				if interfaceStatus, exists := domainInterfaceStatusByMac[interfaceMAC]; exists {
-					newInterface.IP = interfaceStatus.Ip
-					newInterface.IPs = interfaceStatus.IPs
 					newInterface.InterfaceName = interfaceStatus.InterfaceName
+					// Do not update if interface has Masquerede binding
+					// virt-controller should update VMI status interface with Pod IP instead
+					if !isForwardingBindingInterface {
+						newInterface.IP = interfaceStatus.Ip
+						newInterface.IPs = interfaceStatus.IPs
+					}
 					delete(domainInterfaceStatusByMac, interfaceMAC)
 				}
 				newInterfaces = append(newInterfaces, newInterface)
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -2361,4 +2361,27 @@ var _ = Describe("Configurations", func() {
 			table.Entry("[test_id:3808] Should be rejected when using ide", "ide", "IDE bus is not supported"),
 		)
 	})
+
+	It("VMI with masquerade binding and guest agent should expose Pod IP as its public address", func() {
+		vmi := tests.NewRandomFedoraVMIWitGuestAgent()
+
+		By("Starting a VirtualMachineInstance")
+		vmi, err := virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+		Expect(err).ToNot(HaveOccurred(), "Should successfully create VMI")
+		tests.WaitForSuccessfulVMIStart(vmi)
+
+		vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(vmi.Name, &metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred(), "Should successfully get VMI")
+
+		vmiPod := tests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
+
+		tests.WaitAgentConnected(virtClient, vmi)
+
+		// Ensure that VMI IP Stays equal to PodIP even after Guest-Agent kicks in
+		Consistently(func() bool {
+			vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(vmi.Name, &metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred(), "Should successfully get VMI")
+			return vmi.Status.Interfaces[0].IP == vmiPod.Status.PodIP
+		}, 180*time.Second, 1*time.Second).Should(BeTrue(), "VMI status IP should match VMI Pod IP")
+	})
 })
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -2382,6 +2382,6 @@ var _ = Describe("Configurations", func() {
 			vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(vmi.Name, &metav1.GetOptions{})
 			Expect(err).ToNot(HaveOccurred(), "Should successfully get VMI")
 			return vmi.Status.Interfaces[0].IP == vmiPod.Status.PodIP
-		}, 180*time.Second, 1*time.Second).Should(BeTrue(), "VMI status IP should match VMI Pod IP")
+		}, 30*time.Second, 1*time.Second).Should(BeTrue(), "VMI status IP should match VMI Pod IP")
 	})
 })


tests/vmi_configuration_test.go
```diff
@@ -2110,4 +2110,19 @@ var _ = Describe("Configurations", func() {
...
 	})
+
+	FIt("Masquerede VMI with Guest agent should not update VMI IP", func() {
+		vmi := tests.NewRandomFedoraVMIWitGuestAgent()
+
+		By("Starting a VirtualMachineInstance")
+		vmi, err := virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+		Expect(err).ToNot(HaveOccurred())
+		tests.WaitForSuccessfulVMIStart(vmi)
+
```
enp0s3: @vatsalparekh Maybe its better to do another check using
```
tests.WaitAgentConnected(virtClient, vmi)
```
 thus making sure that GA is up.

tests/vmi_configuration_test.go
```diff
@@ -2110,4 +2110,19 @@ var _ = Describe("Configurations", func() {
...
+		vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(vmi.Name, &metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred())
+
+		vmiPod := tests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
+		Expect(vmi.Status.Interfaces[0].IP).To(Equal(vmiPod.Status.PodIP))
```
stu-gott: @vatsalparekh, @enp0s3 made me realize there's a race condition here. Ensuring that the Agent is running helps, but there's a still finite amount of time between when the agent reports as ready and when the IP gets updated. This Expect should probably be replaced with an Eventually clause.

vladikr: +1

vatsalparekh: Added `Eventually`

tests/vmi_configuration_test.go
```diff
@@ -2110,4 +2110,21 @@ var _ = Describe("Configurations", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 	})
+
+	It("Masquerede VMI with Guest agent should not update VMI IP", func() {
+		vmi := tests.NewRandomFedoraVMIWitGuestAgent()
```
oshoval: /Wit/With/ ?

vatsalparekh: That's the typo in the utils itself
https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L1913
Will do a separate PR to fix that

pkg/virt-handler/vm.go
```diff
@@ -312,9 +320,11 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 				// Update IP info based on information from domain.Status.Interfaces (Qemu guest)
 				// Remove the interface from domainInterfaceStatusByMac to mark it as handled
 				if interfaceStatus, exists := domainInterfaceStatusByMac[interfaceMAC]; exists {
-					newInterface.IP = interfaceStatus.Ip
-					newInterface.IPs = interfaceStatus.IPs
-					newInterface.InterfaceName = interfaceStatus.InterfaceName
+					if !IsMasqueredeInterface {
+						newInterface.IP = interfaceStatus.Ip
+						newInterface.IPs = interfaceStatus.IPs
+						newInterface.InterfaceName = interfaceStatus.InterfaceName
```
phoracek: Interface name can be reported even with Masquerade, the issue is only on IP.

pkg/virt-handler/vm.go
```diff
@@ -295,12 +295,20 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 			for _, domainInterface := range domain.Spec.Devices.Interfaces {
 				interfaceMAC := domainInterface.MAC.MAC
 				var newInterface v1.VirtualMachineInstanceNetworkInterface
+				var IsMasqueredeInterface bool
```
phoracek: lower case **i**

pkg/virt-handler/vm.go
```diff
@@ -312,9 +320,11 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 				// Update IP info based on information from domain.Status.Interfaces (Qemu guest)
 				// Remove the interface from domainInterfaceStatusByMac to mark it as handled
 				if interfaceStatus, exists := domainInterfaceStatusByMac[interfaceMAC]; exists {
-					newInterface.IP = interfaceStatus.Ip
-					newInterface.IPs = interfaceStatus.IPs
-					newInterface.InterfaceName = interfaceStatus.InterfaceName
+					if !IsMasqueredeInterface {
```
phoracek: If the previous if on line 300 is false, we never set this variable

enp0s3: @vatsalparekh @phoracek Maybe it worth to put a comment before the condition to explain the possible race condition?

vatsalparekh: Corrected.
Brought the check out of the if condition now.

tests/vmi_configuration_test.go
```diff
@@ -2110,4 +2110,21 @@ var _ = Describe("Configurations", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 	})
+
+	It("Masquerede VMI with Guest agent should not update VMI IP", func() {
```
phoracek: You are suggesting that there is an issue here. It should verify that the code performs as expected. So something along lines:

"VMI with masquerade binding and guest agent should expose Pod IP as its public address".

pkg/virt-handler/vm.go
```diff
@@ -295,12 +295,20 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
...
+					if iface.Name == domainInterface.Alias.Name && iface.Masquerade != nil {
+						isMasqueredeInterface = true
+					}
+				}
 				if existingInterface, exists := existingInterfaceStatusByName[domainInterface.Alias.Name]; exists {
 					// Reuse previously calculated interface from vmi.Status.Interfaces, updating the MAC from domain.Spec
 					// Only interfaces defined in domain.Spec are handled here
 					newInterface = existingInterface
-					newInterface.MAC = interfaceMAC
+					if !isMasqueredeInterface {
```
phoracek: Add an explanation similar to the MAC one.

tests/vmi_configuration_test.go
vatsalparekh: Corrected

pkg/virt-handler/vm.go
vatsalparekh: Added

tests/vmi_configuration_test.go
```diff
@@ -2110,4 +2110,21 @@ var _ = Describe("Configurations", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 	})
+
+	It("VMI with masquerade binding and guest agent should expose Pod IP as its public address", func() {
```
phoracek: We moved the MAC address reporting under a condition. Can you create a similar test for that?

vatsalparekh: Added within the same test

pkg/virt-handler/vm.go
```diff
@@ -295,12 +295,22 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 			for _, domainInterface := range domain.Spec.Devices.Interfaces {
 				interfaceMAC := domainInterface.MAC.MAC
 				var newInterface v1.VirtualMachineInstanceNetworkInterface
+				var isMasqueredeInterface bool = false
 
+				for _, iface := range vmi.Spec.Domain.Devices.Interfaces {
```
rmohr: How about iterating once through the interfaces and creating a map with the name? I would prefer if we avoid nested loops. What do you think? I know that most likely these arrays are not that big, but let's try to avoid n^2  runtime.

vatsalparekh: Good suggestion, changed it now

pkg/virt-handler/vm.go
```diff
@@ -291,16 +291,32 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 				domainInterfaceStatusByMac[domainInterfaceStatus.Mac] = domainInterfaceStatus
 			}
 
+			existingInterfacesSpecByName := map[string]v1.Interface{}
```
stu-gott: We named a type "Interface"? 😱

rmohr: `NetworkInterface` might have been a better name, but at least that matches the name in the domain xml. :D

tests/vmi_configuration_test.go
```diff
@@ -2110,4 +2110,31 @@ var _ = Describe("Configurations", func() {
...
+		oldMacOnVMI := vmi.Status.Interfaces[0].MAC
+
+		tests.WaitAgentConnected(virtClient, vmi)
+
+		vmiPod := tests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
+		Eventually(func() string {
+			vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(vmi.Name, &metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			return vmi.Status.Interfaces[0].IP
+		}, 180*time.Second, 1*time.Second).Should(Equal(vmiPod.Status.PodIP))
```
stu-gott: Can you add a brief explanation as to what was being tested for in case of failure?

`Error: <something> != <something>`

makes devs have to inspect the code to figure out what went wrong.

vatsalparekh: Added

pkg/virt-handler/vm.go
```diff
@@ -291,16 +291,32 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
...
 
 				if existingInterface, exists := existingInterfaceStatusByName[domainInterface.Alias.Name]; exists {
 					// Reuse previously calculated interface from vmi.Status.Interfaces, updating the MAC from domain.Spec
 					// Only interfaces defined in domain.Spec are handled here
 					newInterface = existingInterface
-					newInterface.MAC = interfaceMAC
+					// Do not update if interface has Masquerade binding
+					// virt-controller should update VMI status interface instead
+					if !isMasqueredeInterface {
+						newInterface.MAC = interfaceMAC
```
phoracek: It is clear that for IP we want to expose the one externally accessible. I'm not so sure what to do with MAC though. Should we report the MAC address that is used inside the guest or the one that is on the virt-launcher pod?

When a user explicitly specifies a MAC address for their VM, I suspect they would expect that MAC to appear on the VMI report.

Does that sound right @rmohr ?

enp0s3: @rmohr @phoracek @vatsalparekh If a custom MAC was configured on the VM, should it appear in the VMI spec? if so, we can try to locate it and if it exist we won't overwrite it but just compare it.

pkg/virt-handler/vm.go
```diff
@@ -291,16 +291,32 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
...
+
+				if existingInterfacesSpecByName[domainInterface.Alias.Name].Masquerade != nil {
+					isMasqueredeInterface = true
+				}
 
 				if existingInterface, exists := existingInterfaceStatusByName[domainInterface.Alias.Name]; exists {
 					// Reuse previously calculated interface from vmi.Status.Interfaces, updating the MAC from domain.Spec
 					// Only interfaces defined in domain.Spec are handled here
 					newInterface = existingInterface
-					newInterface.MAC = interfaceMAC
```
enp0s3: @vatsalparekh Where do you update the MAC address for masquerade-enabled interfaces?

vatsalparekh: @enp0s3 in this PR I'm not updating it yet at any place, just letting it be what it was previously
But I think it should be done at controller level, somewhere in this https://github.com/kubevirt/kubevirt/pull/2963

phoracek:
@vatsalparekh please use the PR description to explain what is the current state, why is it wrong and how do you fix it / how it works with this PR.

vatsalparekh:
@phoracek does it look good now?

phoracek:
Beautiful!

pkg/virt-handler/vm.go
phoracek: I believe so. As a user, I'd expect to see the internal MAC used in the VM.

I don't understand what you mean by "if so, we can try to locate it and if it exist we won't overwrite it but just compare it.".

fabiand: I see a few options:
1. Expose external facing details (public IP and public MAC)
2. Expose internal details (vNIC IP and vNIC MAC)
3. Expose external IP and vNIC MAC
4. Expose vNIC IP and external MAC
5. Expose external and internal (vNIC) details on the status

TO me 4 does not make much sense, but we shoul dunderstand which use-cases are supported by options 1-3. We should ask ourselves: When is the MAC needed and when i the IP needed.
Why do we expose these informations on the API, and thus which of the options above does answer those user questions.

/cc @vladikr 
xref #3090

vladikr: @fabiand 
In my view, the IPs in the status are there to reflect what IPs does the guest sees. This also serves as an indication to the user that the guest managed to get an IP(s).
For me, `status.IPs` is not about how the API user can connect to the VMI.
I don't think that all networks that the guest will be able to connect to will also be reachable by the API user.

Therefore, I'm not in favor of this PR. :/

Wrt #3090 , this is a completely different use case, from my POV.  This is about the guest OS user being about to identify which vNIC is connected to which network. In some cases, it is needed to configure the interface properly or to configure a vlan or anything else, but the guest user needs to see the vNIC information to be able to identify it.

phoracek: @vladikr I, as a user, would expect to publicly accessible IP on the VMI. Seeing guest agent report about the inside configuration may be interesting, but I find it way more useful to be able to get the public IP through the VMI. Then I can use it to create my own Endpoints/Services, connect to the running VM directly from another one etc.

Also, the default now is to report the Pod IP. I consider the report done by guest-agent something extra. If we want to make the report stable with or without guest-agent, using the default (pod IP) would be less destructive to the API.

phoracek: @vladikr 

We have two bindings, `bridge` and `masquerade`. We also have two modes of reporting IP, with and without guest-agent.

With `bridge`, it is simple. Since we move the Pod IP inside the VM and extend the Pod's L2 network to the VM, what guest-agent sees in the VM matches the Pod IP. No problem here.

Now with `masquerade`, we keep the original IP in the Pod. The VM is connected through NAT and keeps a link-local IP which gets traffic forwarded from the Pod. If we don't gave guest-agent running, we report the Pod IP. With that, other Pods/VMs can use this IP to access the VM. However, when guest-agent is present, we report the internal link-local IP. That has two issues: 1) Reported IP is not accessible from the outside and 2) The reported IP changes during the booting, first we report Pod IP, once guest-agent kicks in, we start reporting the link-local.

Mission of this PR is to adjust reporting of IP when using `masquerade` so we always report the Pod IP, even once guest-agent starts. That way we always report the publicly accessible IP and we keep the reporting consistent with or without guest-agent.

Please note that reporting of the link-local would have little to no use since users cannot change it after the VM starts (that would break forwarding).

stu-gott:
@vatsalparekh what's the state of this?

vatsalparekh:
@stu-gott as discussed in the community meeting, this is blocked on the various opinions about this PR, to have this change or not, and it's change in the API

pkg/virt-handler/vm.go
```diff
@@ -291,16 +291,32 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 				domainInterfaceStatusByMac[domainInterfaceStatus.Mac] = domainInterfaceStatus
 			}
 
+			existingInterfacesSpecByName := map[string]v1.Interface{}
+			for _, existingInterfaceSpec := range vmi.Spec.Domain.Devices.Interfaces {
+				if existingInterfaceSpec.Name != "" {
```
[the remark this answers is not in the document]
vatsalparekh: Nope, removed it now

phoracek:
@vatsalparekh could you please review https://github.com/nmstate/kubernetes-nmstate/pull/503 and make sure this PR aligns to it?

pkg/virt-handler/vm.go
```diff
@@ -344,16 +344,27 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
...
+			existingInterfacesSpecByName := map[string]v1.Interface{}
+			for _, existingInterfaceSpec := range vmi.Spec.Domain.Devices.Interfaces {
+				existingInterfacesSpecByName[existingInterfaceSpec.Name] = existingInterfaceSpec
+			}
+
 			// Iterate through all domain.Spec interfaces
 			for _, domainInterface := range domain.Spec.Devices.Interfaces {
 				interfaceMAC := domainInterface.MAC.MAC
 				var newInterface v1.VirtualMachineInstanceNetworkInterface
+				var isMasqueredeInterface bool = false
```
phoracek, line 356: Could we please hijack this PR to not update based on guest agent with both masqurade and slirp to implement  https://github.com/kubevirt/kubevirt/pull/3241.

You can rename it to `isForwardingBindingInterface`, so we keep that logic here and pick the reporting method based on a single variable. If you decide not to, fix the type in masquerade.

tests/vmi_configuration_test.go
```diff
@@ -2361,4 +2361,32 @@ var _ = Describe("Configurations", func() {
 			table.Entry("[test_id:3808] Should be rejected when using ide", "ide", "IDE bus is not supported"),
 		)
 	})
+
+	It("VMI with masquerade binding and guest agent should expose Pod IP as its public address", func() {
+		vmi := tests.NewRandomFedoraVMIWitGuestAgent()
+
+		By("Starting a VirtualMachineInstance")
+		vmi, err := virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+		Expect(err).ToNot(HaveOccurred())
```
phoracek, line 2370: To make debugging of failed tests easier, please add expectation strings to all Expects. Here it would be:
`Expect(err).ToNot(HaveOccurred(), "Should successfully create VMI")`

tests/vmi_configuration_test.go
```diff
@@ -2361,4 +2361,32 @@ var _ = Describe("Configurations", func() {
...
+		tests.WaitAgentConnected(virtClient, vmi)
+
+		vmiPod := tests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
+		Eventually(func() string {
+			vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(vmi.Name, &metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			return vmi.Status.Interfaces[0].IP
+		}, 180*time.Second, 1*time.Second).Should(Equal(vmiPod.Status.PodIP), "VMI status IP should match VMI Pod IP")
+
+		// Test that MAC on the interface has not changed
```
phoracek, line 2386: The MAC handling was dropped from this PR, do we need to keep testing it?

tests/vmi_configuration_test.go
vatsalparekh: Added

pkg/virt-handler/vm.go
phoracek: @vatsalparekh what about slirp?

tests/vmi_configuration_test.go
```diff
@@ -2361,4 +2361,26 @@ var _ = Describe("Configurations", func() {
...
+		Expect(err).ToNot(HaveOccurred(), "Should successfully create VMI")
+		tests.WaitForSuccessfulVMIStart(vmi)
+
+		vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(vmi.Name, &metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred(), "Should successfully get VMI")
+
+		tests.WaitAgentConnected(virtClient, vmi)
+
+		vmiPod := tests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
+		Eventually(func() string {
```
phoracek, line 2379: Why do we we need Eventually here? What are we waiting for exactly?

tests/vmi_configuration_test.go
```diff
@@ -2361,4 +2361,26 @@ var _ = Describe("Configurations", func() {
...
+			vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(vmi.Name, &metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
```
phoracek, line 2381: Missing explanation

pkg/virt-handler/vm.go
```diff
@@ -344,10 +344,20 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
...
+				existingInterfacesSpecByName[existingInterfaceSpec.Name] = existingInterfaceSpec
+			}
+
 			// Iterate through all domain.Spec interfaces
 			for _, domainInterface := range domain.Spec.Devices.Interfaces {
 				interfaceMAC := domainInterface.MAC.MAC
 				var newInterface v1.VirtualMachineInstanceNetworkInterface
+				var isForwardingBindingInterface bool = false
+
+				if existingInterfacesSpecByName[domainInterface.Alias.Name].Masquerade != nil {
```
phoracek, line 358: My comment about slirp handling is still pending

tests/vmi_configuration_test.go
vatsalparekh: on L#2376 ensuring that the Agent is running helps, but there's a still finite amount of time between when the agent reports as ready and when the IP gets updated, so this is needed to reduce flaky results

tests/vmi_configuration_test.go
vatsalparekh: Added now

pkg/virt-handler/vm.go
vatsalparekh: Missed that, added now

tests/vmi_configuration_test.go
phoracek: I still don't get the Eventually. The IP is reported there even before the guest agent has started since we use PodIP, right? So you don't need to wait at all to successfully pass the check.

Isn't Consistently what you are looking for? To make sure that the IP remains on PodIP even once the agent kicks in?

vatsalparekh: @phoracek Yes, just added `Consistently` block also

phoracek: I still don't get why Eventually is needed here.

Before the guest agent kicks in, PodIP is reported as VMs IP (that's the default). After guest agent kicks in, it is exactly the same, so there is nothing we can wait for with Eventually. The block would return right away on the original report.

Is there any need to keep the Eventually?

vatsalparekh: @petrkotas I'm not sure about `Before the guest agent kicks in, PodIP is reported as VMs IP`
is this something guaranteed? Because without `Eventually` we will insert a flaky test here

phoracek: Is it flaky without Eventually? Or do you just expect it to be?

It seems like we report the IP once the VM Pod is ready: https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-controller/watch/vmi.go#L369

Since we wait for successful start here on https://github.com/kubevirt/kubevirt/pull/3063/files#diff-d47bf036797965d742615e64adc3b777R2371, I would expect it to be ready. But I don't know the codebase by heart, please convince me otherwise.

vatsalparekh: @phoracek It was guess
I think we can remove `Eventually` because of this https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-controller/watch/vmi.go#L369

vatsalparekh: @phoracek Done!

tests/vmi_configuration_test.go
```diff
@@ -2361,4 +2361,27 @@ var _ = Describe("Configurations", func() {
...
+		vmiPod := tests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
+
+		tests.WaitAgentConnected(virtClient, vmi)
+
+		// Ensure that VMI IP Stays equal to PodIP even after Guest-Agent kicks in
+		Consistently(func() bool {
+			vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(vmi.Name, &metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred(), "Should successfully get VMI")
+			return vmi.Status.Interfaces[0].IP == vmiPod.Status.PodIP
+		}, 180*time.Second, 1*time.Second).Should(BeTrue(), "VMI status IP should match VMI Pod IP")
```
phoracek, line 2385: Do we need to wait whole 3 minutes to be really sure guest-agent has already started reporting?  We already wait with `tests.WaitAgentConnected(virtClient, vmi)`, I expected that we need to give it just a couple of extra seconds to avoid races.

How much time we really need here? Let's try to keep it low, not to make our e2e tests even worse.

phoracek, line 2385: Do we need to wait whole 3 minutes to be really sure guest-agent has already started reporting?  We already wait with `tests.WaitAgentConnected(virtClient, vmi)`, I expected that we need to give it just a couple of extra seconds to avoid races.

How much time we really need here? Let's try to keep it low, not to make our e2e tests even worse.

vatsalparekh: @phoracek makes sense, reduced it to 30s now, just a couple of seconds might be too short :smile:

phoracek: Fair enough!

PR #3073, opened 2020-02-14
Missing pod anti-affinity rule in virt-operator fixed

alosadagrande:
Signed-off-by: Alberto Losada <alosadag@redhat.com>

Currently reviewing some advancing scheduling options in KubeVirt (affinity and antiaffinity rules basically). I've found that the deployment object of virt-operator, which is configured as replica 2 by default does not have any antiaffinity rules. `virt-controller` and `virt-api` already do have, which basically says that if it is possible (preferredDuringSchedulingIgnoredDuringExecution:) do not run both replicas together (same node). 

This PR fixes this missing pod anti-affinity rule in `virt-operator` 

Fixes #

Thanks to @slintes for showing me the right place to write the fix.

**Release note**:

```release-note
Fixed missing pod anti-affinity rule in virt-operator
```

### DIFF
--- a/pkg/virt-operator/creation/components/deployments.go
+++ b/pkg/virt-operator/creation/components/deployments.go
@@ -437,6 +437,7 @@ func NewOperatorDeployment(namespace string, repository string, imagePrefix stri
 	kubeVirtVersionEnv string, virtApiShaEnv string, virtControllerShaEnv string,
 	virtHandlerShaEnv string, virtLauncherShaEnv string) (*appsv1.Deployment, error) {
 
+	podAntiAffinity := newPodAntiAffinity("kubevirt.io", "kubernetes.io/hostname", metav1.LabelSelectorOpIn, []string{"virt-operator"})
 	name := "virt-operator"
 	version = AddVersionSeparatorPrefix(version)
 	image := fmt.Sprintf("%s/%s%s%s", repository, imagePrefix, name, version)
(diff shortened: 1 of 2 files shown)


alosadagrande:
See this error in travisCI: "The command "if [[ -n "$(git status --porcelain)" ]] ; then echo "It seems like you need to run `make`. Please run it and commit the changes"; git status --porcelain; false; fi" exited with 1."

I just modified the `deployments.go`, should I run the make generate and commit all changes or just committing the modified file is enough?

pkg/virt-operator/creation/components/deployments.go
```diff
@@ -481,6 +481,27 @@ func NewOperatorDeployment(namespace string, repository string, imagePrefix stri
...
+							PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
+								{
+									Weight: 1,
+									PodAffinityTerm: corev1.PodAffinityTerm{
+										LabelSelector: &metav1.LabelSelector{
+											MatchExpressions: []metav1.LabelSelectorRequirement{
+												{
+													Key:      "kubevirt.io",
+													Operator: metav1.LabelSelectorOpIn,
+													Values:   []string{"virt-operator"},
```
danielBelenky: Maybe it's better to use the `name` var here?

alosadagrande: I am not sure since they are different types. Name is a string since Values requires a []string.

I can easily use split:

```                                                                                                       Values:   strings.Split(name, " "),
```

However, I do not think it makes things clearer. And on the other, what will happen if the name contains a space...

danielBelenky: What I mean is: `[]string{name}`
The thing is, that you specify affinity with the label hard-coded here while it's controlled by the `name` parameter. So is the `kubevirt.io` label which you used as a key... you probably should use `virtv1.AppLabel` instead.

alosadagrande: Ok, understood. Yes, agree, make sense to use the virtv1.AppLabel.

Thanks for your input @danielBelenky.

danielBelenky:
/lgtm

Thank you

alosadagrande:
@slintes, correct, looks clearer using the podAntiAffinity function.

slintes:
thanks!

/lgtm

PR #3075, opened 2020-02-16
Fix "Multiple VM connectivity" tests to allow any mtu > 1000

oshoval:
Fix "Multiple VM connectivity" tests to allow any mtu > 1000 

```release-note
None
```

### DIFF
--- a/tests/vmi_networking_test.go
+++ b/tests/vmi_networking_test.go
@@ -189,29 +189,19 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 				virtClient,
 				vmiPod,
 				"compute",
-				[]string{"ip", "address", "show", "k6t-eth0"},
+				[]string{"cat", "/sys/class/net/k6t-eth0/mtu"},
 			)
-			log.Log.Infof("%v", output)
+			log.Log.Infof("k6t-eth0 mtu is %v", output)
 			Expect(err).ToNot(HaveOccurred())
 
-			// The MTU of the Pod network varies, depending on the environment in use. We want to
-			// verify that the actual MTU is from the possible range, which is {1500, 1450}, minus
-			//  50 bytes for vxlan overhead.
-			possibleMtus := []int{1400, 1450}
-			mtuMatch := false
-			expectedMtuString := ""
+			output = strings.TrimSuffix(output, "\n")
+			mtu, err := strconv.Atoi(output)
+			Expect(err).ToNot(HaveOccurred())
 
-			for _, expectedMtu := range possibleMtus {
-				payloadSize = expectedMtu - ipHeaderSize
+			Expect(mtu > 1000).To(BeTrue())
 
-				// the following substring is part of 'ip address show' output
-				expectedMtuString = fmt.Sprintf("mtu %d", expectedMtu)
-				mtuMatch = strings.Contains(output, expectedMtuString)
-				if mtuMatch == true {
-					break
-				}
-			}
-			Expect(mtuMatch).To(BeTrue())
+			payloadSize = mtu - ipHeaderSize
+			expectedMtuString := fmt.Sprintf("mtu %d", mtu)
 
 			By("checking eth0 MTU inside the VirtualMachineInstance")
 			expecter, err := tests.LoggedInCirrosExpecter(outboundVMI)


oshoval:
according https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2020-02-15-168h.html

looks like pull-kubevirt-e2e-kind-k8s-sriov-1.14.2 is skipped ?

oshoval:
according one of the runs it looks like those 4 tests that i work on are working

https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3075/pull-kubevirt-e2e-k8s-1.17.0/1229016525357191168

tests 1539-1542
but i will refactor the code to use mtu > 1000 as an acceptable prerequisites as suggested by Dan

if needed i will split it to 2 PRs since they the 2nd one is  dependent on the 1st one

oshoval:
/hold
we need to groom before we merge it
this will be an helper for fixing tests of k8s-1.17.0 as groomed in the sprint

tests/vmi_networking_test.go
```diff
@@ -189,29 +189,19 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
 
-			// The MTU of the Pod network varies, depending on the environment in use. We want to
-			// verify that the actual MTU is from the possible range, which is {1500, 1450}, minus
-			//  50 bytes for vxlan overhead.
-			possibleMtus := []int{1400, 1450}
-			mtuMatch := false
-			expectedMtuString := ""
+			output = strings.TrimSuffix(output, "\n")
+			mtu, err := strconv.Atoi(output)
+			Expect(err).ToNot(HaveOccurred())
```
yossisegev: I'm not sure that an error of strconv.Atoi is something we should trace with gomega's (the test framework) Expect, and hence fail the test. What is the meaning of Atoi returning an error?
I might be wrong here, but it's worth checking.

oshoval: for example i had to use TrimSuffix, because there was \n and then the atoi failed
we need a valid value for mtu, so it can hint good if there isnt imo, what do you think?

yossisegev: OK, makes sense.
So the changes in vmi_networking_test.go LGTM.

oshoval:
cleaned, fixed the test , checked with bump, 
removed now the bump since it will be part of Daniel PR

tests/vmi_networking_test.go
```diff
@@ -189,29 +189,19 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 				virtClient,
 				vmiPod,
 				"compute",
-				[]string{"ip", "address", "show", "k6t-eth0"},
+				[]string{"cat", "/sys/class/net/k6t-eth0/mtu"},
```
phoracek: Why do we need this change? `ip address show` is not working anymore?

oshoval: it does, but we just want the mtu, so no need to parse the whole line
it was Suggested by Dan

oshoval:
travis fails @ make apidocs
its just because 

Failed to get resource: GET. [HTTP HTTP/1.1 504 Gateway Time-out: https://jcenter.bintray.com/io/swagger/swagger-models/1.5.12/swagger-models-1.5.12.pom]

how can we rerun it please ?

oshoval:
@dhiller 
Travis passed, thanks

dankenigsberg:
@oshoval the title of this PR is now out of date, I believe.

oshoval:
> @oshoval the title of this PR is now out of date, I believe.

hmm by this title i mean a range ( > 1000) , instead of a scalar list, 
what would you suggest please to write ?
also the description has some more info

dankenigsberg:
> > @oshoval the title of this PR is now out of date, I believe.
> 
> hmm by this title i mean a range ( > 1000) , instead of a scalar list,
> what would you suggest please to write ?
> also the description has some more info

ah, got it. I think that 
```
Fix "Multiple VM connectivity" tests to allow any mtu>1000
```
would be shorter and simpler.

phoracek:
@oshoval when you fix PR title+description, you should reflect that in the commit title+description.

oshoval:
> @oshoval when you fix PR title+description, you should reflect that in the commit title+description.

commit say what went bad and how we fixed it actually will little more words, 
or you prefer it to be changed as well ? 

mtu was 1440 and wasnt part of the supported mtus
added a supported range of mtu instead of a discrete list
(mtu > 1000), because there isnt really a need for a discrete list

tests/vmi_networking_test.go
```diff
@@ -189,29 +189,19 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
-			possibleMtus := []int{1400, 1450}
-			mtuMatch := false
-			expectedMtuString := ""
+			output = strings.TrimSuffix(output, "\n")
+			mtu, err := strconv.Atoi(output)
+			Expect(err).ToNot(HaveOccurred())
 
-			for _, expectedMtu := range possibleMtus {
-				payloadSize = expectedMtu - ipHeaderSize
+			Expect(mtu > 1000).To(BeTrue())
```
dhiller: nit: for a more fluent reading I'd recommend using

`Expect(mtu > 1000).To(BeNumerically(">", 1000))`

next time :)

PR #3076, opened 2020-02-16
Dummy to check status of master

oshoval:
```release-note
None
```


oshoval:
@rmohr 
isnt it bad that the bot adds "approved" ?
for example if due to a bug, a later push to this PR wont remove the approve, and people can add malicious code in this case ?

anyway this is just a dummy PR to check the state of the vanilla lanes 
thanks

rmohr:
@oshoval I habmve no idea why this is approved. That should not have happened.

rmohr:
> for example if due to a bug, a later push to this PR wont remove the approve, and people can add malicious code in this case ?

Ok, see https://github.com/kubernetes/test-infra/issues/13200.

Prow approves it because no file was changed, which requires that no approver for changes files is needed (since none are changed). It is a not-addressed edge case where people could merge strangte commit messages if someone can lgtm a PR, but it would remove the approve if you change files in a follow-up.

oshoval:
thanks @rmohr

rmohr:
> thanks @rmohr

Thanks for pinging.

PR #3071, opened 2020-02-14
Add automation helper binaries and reduce builder image size

rmohr:
Installing pr-creator, issue-creator and issue-labeler into the builder
image while at the same time also cleaning all docker caches to reduce
the builder image size.

Especially `pr-creator` will help to automate basic tasks like pushing kubevirtci, updating vendor dependencies and so on from prow jobs.

Signed-off-by: Roman Mohr <rmohr@redhat.com>

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Related to https://github.com/kubevirt/project-infra/issues/344

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #3077, opened 2020-02-17
Automate things

rmohr:

Add scripts which can be used to do the following automatic code bumps:

 * Bump to latest kubevirtci version
 * Update patch or minor versions of vendor dependencies

Further a KUBEVIRT_RUN_UNNUESTED environment variable is added, which
allows to execute all Makefile targets without nested docker. That is
useful when already running inside a builder image in e.g. a prow job.

Follow up of #3071 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/autobump-kubevirtci.sh
+++ b/hack/autobump-kubevirtci.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+# Copyright 2018 The Kubernetes Authors.
+# Copyright 2020 The KubeVirt Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Taken from https://github.com/kubernetes/test-infra/blob/4d7f26e59a5e186eef3a7de55486b7a40bbd79d7/hack/autodeps.sh
+# and modified for kubevirt.
+
+set -o errexit
+set -o pipefail
+
+source $(dirname "$0")/common.sh
+source $(dirname "$0")/config.sh
+
+cd $(git rev-parse --show-toplevel)
+
+if [[ $# -lt 2 ]]; then
+    echo "Usage: $(basename "$0") <github-login> </path/to/github/token> [git-name] [git-email]" >&2
+    exit 1
+fi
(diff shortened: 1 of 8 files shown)


rmohr:
/hold

needs #3071 first.

rmohr:
@dhiller would be ready for a review.

hack/dockerized
```diff
@@ -3,6 +3,11 @@ set -e
 
 source $(dirname "$0")/common.sh
 
+if [ "${KUBEVIRT_RUN_UNNESTED}" == "true" ]; then
```
dhiller: I tried this, for me it didn't work. What I did was:

```
> export KUBEVIRT_RUN_UNNESTED=true
> make generate
hack/dockerized "DOCKER_PREFIX= DOCKER_TAG= IMAGE_PULL_POLICY= VERBOSITY= ./hack/generate.sh"
hack/dockerized: line 7: DOCKER_PREFIX= DOCKER_TAG= IMAGE_PULL_POLICY= VERBOSITY= ./hack/generate.sh: No such file or directory
make: *** [Makefile:29: generate] Error 127
```

hack/autobump-kubevirtci.sh
```diff
@@ -0,0 +1,65 @@
...
+hack/bump-kubevirtci.sh
+
+git add -A
+if git diff --name-only --exit-code HEAD; then
+    echo "Nothing changed" >&2
+    exit 0
+fi
+
+title="Run hack/bump-kubevirtci.sh, updating to ${kubevirtci_git_hash:0:8}..."
+git commit -m "${title}"
```
dhiller: Same here for signoff?

hack/dockerized
rmohr: Thanks. It seems like within our builder container, the `$@` is executed with int `sh` mode. Changed it to `/bin/bash -c "$@"` and it started to work.

This works now:

```bash
hack/dockerized /bin/bash
export KUBEVIRT_RUN_UNNESTED=true
make generate
```

hack/autobump-kubevirtci.sh
rmohr: kubevirt-bot does not need it. Might be nice to sign it for convenience if someone runs the scripts locally ...

hack/dockerized
dhiller: For me this worked:

```
hack/dockerized "export KUBEVIRT_RUN_UNNESTED=true && make generate"
```

I think we should document this somehow as the usual approach people would try to use this would be the above one. And then they'll start to complain, no?

qinqon:
@rmohr Can you create separate PR for KUBEVIRT_RUN_UNNUESTED ? I will use it as you propose moving travis to prow.

rmohr:
> @rmohr Can you create separate PR for KUBEVIRT_RUN_UNNUESTED ? I will use it as you propose moving travis to prow.

The PR is already approved and I don't think that having another PR would make things faster :/

PR #3079, opened 2020-02-17
Add secureBoot EFI flag

jean-edouard:
This PR enables the use of secure boot by adding a boolean option under EFI.
If set, alternative (secboot) OVMF roms will be loaded and secure boot will be enabled.
If not set or set to false, the usual OVMF roms will be loaded and secure boot will not be enabled.

Fixes #

The SMM domain feature must be enabled along with EFI secure boot. The dependency is enforced by libvirt.
The state of secure boot should be reflected in the VM's dmesg (`dmesg | grep secureboot`)

**Release note**:

```release-note
Add "secureBoot" EFI option to enable secure boot in OVMF.
```


jean-edouard:
Force-pushed to add missing update of generated files.

staging/src/kubevirt.io/client-go/api/v1/schema.go
```diff
@@ -209,6 +209,12 @@ type BIOS struct {
 // ---
 // +k8s:openapi-gen=true
 type EFI struct {
+	// If set, SecureBoot will be enabled and the OVMF roms will be swapped for
+	// SecureBoot-enabled ones.
+	// Requires SMM to be enabled.
```
rmohr: Since this requires SMM, which one can explicitly enable or disable, could we do the following:

In the defaulter:
 * If secureBoot is set to true and SMM is unset (neither explicit true or false), enable SMM too.

In the validator:
 * If secureBoot is set to true and SSM is explicitlly disabled, reject the VM/VMI/VMIRS?

jean-edouard: The validator part is definitely missing, I'll add it.
However I don't like the idea of defaulting SMM to true without the user explicitly enabling it. SMM is quite a ... dangerous feature.
Thanks for the review!

rmohr: > SMM is quite a ... dangerous feature.

Could you elaborate on that? I can understand that applications inside the guest may attack SMM RAM and succeed, but I don't understand the difference to not having protected RAM at all. Would be happy to learn about the dangers.

jean-edouard: I believe that if the application in question manages to write malicious code into SMM RAM and then trigger an SMI, there could now be an invisible rootkit controlling the VM.
Multiple security features are in place to prevent all that, but it still seems to me that enabling SMM slightly increases the attack surface of the VM.
Most likely not a problem here, but it's still worth making sure the user knows it's there IMO.

rmohr: > I believe that if the application in question manages to write malicious code into SMM RAM and then trigger an SMI, there could now be an invisible rootkit controlling the VM.
> Multiple security features are in place to prevent all that, but it still seems to me that enabling SMM slightly increases the attack surface of the VM.
> Most likely not a problem here, but it's still worth making sure the user knows it's there IMO.

Thanks, so let's continue without ssm auto-enable for now. I don't know how many signed firmwares requires SMM too. I think that if most do, then it would still make sense to enable it automatically, but let's wait for user-feedback. If no one complains it is definitely good enough.

jean-edouard:
Force-pushed to add a validator

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -55,6 +55,8 @@ const (
 	defaultIOThread        = uint(1)
 	EFIPath                = "/usr/share/OVMF/OVMF_CODE.fd"
 	EFIVarsPath            = "/usr/share/OVMF/OVMF_VARS.fd"
+	EFIPathSecureBoot      = "/usr/share/OVMF/OVMF_CODE.secboot.fd"
```
rmohr: I wonder if we need to make this configurable ... I think that I have also read somewhere that libvirt may know these locations on each OS it is running on, but that may be wrong.

jean-edouard: About the functional test, I started writing one but, as you predicted, `NewRandomVMIWithEFIBootloader()` starts an unsigned binary and fails.
Locally I've been testing with the Fedora 31 live CD / installed, that boots just fine.

About the paths, they can be configured in /etc/libvirt/qemu.conf, but I don't think we can count on it. We probably want to be able to switch between regular and secboot anyway.
However, there is indeed a recent libvirt option that automatically sets paths and stuff (including SMM! yikes). I tried adding support for it but miserably failed... This is where I'm at:
https://github.com/jean-edouard/kubevirt/commit/92a9abec6b541188448495d1b42912de909abeeb

rmohr: Do we need a way to set the paths? I guess we should at least be able to successfully boot with the right pathes on centos/fedora/rhel.

jean-edouard: I'd say it's fine to hardcode the paths since libvirt does too.
Before v.6.0.0, libvirt had a (compile-time) exception for Fedora. Seems like before v31, Fedora used different paths than the rest of the world.
The exception was removed there:
https://github.com/libvirt/libvirt/commit/5a5e92000d12a671f491c5fb90677f63b1ae7e75#diff-c49a1d135157a3c9aac84639bc92a285

Maybe we should work with libvirt to improve their logic for choosing the right firmware files, ultimately allowing us to remove those hardcoded paths from our codebase.

jean-edouard: I'd say hardcoded paths are fine as a temporary measure, until libvirt gets upgraded to a better version.

From v6.0.0 forward, the paths are already fixed, but it seems that they'll always use the secboot images if present, which I don't think is what we want.

(
Before v.6.0.0, libvirt had a (compile-time) exception for Fedora. Seems like before v31, Fedora used different paths than the rest of the world.
The exception was removed there:
https://github.com/libvirt/libvirt/commit/5a5e92000d12a671f491c5fb90677f63b1ae7e75#diff-c49a1d135157a3c9aac84639bc92a285
)

Maybe we should work with them to implement a switch that picks the secboot files only if secureboot was requested, ultimately lifting the need for hardcoded paths in our codebase.

stu-gott:
@rmohr Is there something further missing from this PR that you'd like to see before we merge?

jean-edouard:
Rebased and added a commit to make the OVMF path configurable.

jean-edouard:
@rmohr the path is now configurable, a functional test is being worked on as a separate effort.
This should be good to go.

stu-gott:
@jean-edouard this PR has merge conflicts. Once you resolve that I'll approve.

PR #3088, opened 2020-02-19
added error messages for pause cmd

### LINKED ISSUE #3061: Better error message for pausing a stopped VM

/kind enhancement

**What happened**:
When trying to pause a stopped VM, the error message is `Error pausing VirtualMachineInstance test-vm: virtualmachineinstance.kubevirt.io "test-vm" not found`.

Error message should suggest to check if the VM is running.

Create a not running VM, try to pause it.

**Environment**:
- KubeVirt version (use `virtctl version`): v0.26.0

xpivarc:
PR is adding better error messages for user which is trying to pause stopped VM.

Fixes #3061

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virtctl/pause/pause.go
+++ b/pkg/virtctl/pause/pause.go
@@ -130,7 +130,7 @@ func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
 						return fmt.Errorf("Error pausing VirutalMachineInstance %s: %v", vmiName, err)
 					}
 					if runningStrategy == kubevirtV1.RunStrategyHalted {
-						return fmt.Errorf("Error pausing VirtualMachineInstance %s. VirtualMachine %s is not set to run", vmiName, vmiName)
+						return fmt.Errorf("Error pausing VirtualMachineInstance %s. VirtualMachine %s is not set to run", vmiName, vm.Name)
 					}
 					return fmt.Errorf("Error pausing VirtualMachineInstance %s, it was not found", vmiName)
 
--- a/pkg/virtctl/pause/pause.go
+++ b/pkg/virtctl/pause/pause.go
@@ -95,7 +95,6 @@ type VirtCommand struct {
 }
 
 func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
-
 	resourceType := strings.ToLower(args[0])
 	resourceName := args[1]
 	namespace, _, err := vc.clientConfig.Namespace()
@@ -122,7 +121,13 @@ func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
 			}
 			err = virtClient.VirtualMachineInstance(namespace).Pause(vmiName)
 			if err != nil {
-				return fmt.Errorf("Error pausing VirtualMachineInstance %s: %v", vmiName, err)
+				if vm.Spec.Running != nil && !*vm.Spec.Running {
+					return fmt.Errorf("Error pausing VirtualMachineInstance %s: %v. VirtualMachine %s is not set to running", vmiName, err, vmiName)
+				} else if vm.Spec.Running != nil && *vm.Spec.Running {
+					return fmt.Errorf("Error pausing VirutalMachineInstance %s: %v", vmiName, err)
+				}
+				return fmt.Errorf("Error pausing VirtualMachineInstance %s: %v, check if VirtualMachine is running", vmiName, err)
+
 			}
 			fmt.Printf("VMI %s was scheduled to %s\n", vmiName, vc.command)
 		case ARG_VMI_LONG, ARG_VMI_SHORT:
--- a/pkg/virtctl/pause/pause.go
+++ b/pkg/virtctl/pause/pause.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
+
 	kubevirtV1 "kubevirt.io/client-go/api/v1"
 
 	"kubevirt.io/client-go/kubecli"
--- a/pkg/virtctl/pause/pause.go
+++ b/pkg/virtctl/pause/pause.go
@@ -125,14 +125,14 @@ func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
 			err = virtClient.VirtualMachineInstance(namespace).Pause(vmiName)
 			if err != nil {
 				if errors.IsNotFound(err) {
-					runningStrategy, invalid := vm.RunStrategy()
-					if invalid != nil {
-						return invalid
+					runningStrategy, err := vm.RunStrategy()
+					if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/virtctl/pause/pause.go
```diff
@@ -122,7 +121,13 @@ func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
 			}
 			err = virtClient.VirtualMachineInstance(namespace).Pause(vmiName)
 			if err != nil {
-				return fmt.Errorf("Error pausing VirtualMachineInstance %s: %v", vmiName, err)
+				if vm.Spec.Running != nil && !*vm.Spec.Running {
```
slintes: I would start with a check if `err` is a NotFound error. Only then it makes sense to check if the VM is running IMHO.
For the running check then you also need to check the `vm.Spec.RunStrategy`.

pkg/virtctl/pause/pause.go
```diff
@@ -122,7 +121,13 @@ func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
...
+					return fmt.Errorf("Error pausing VirtualMachineInstance %s: %v. VirtualMachine %s is not set to running", vmiName, err, vmiName)
```
slintes: When the VM is not started, no need to return the error itself.

pkg/virtctl/pause/pause.go
```diff
@@ -122,7 +121,13 @@ func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
...
+				} else if vm.Spec.Running != nil && *vm.Spec.Running {
+					return fmt.Errorf("Error pausing VirutalMachineInstance %s: %v", vmiName, err)
```
slintes: When the VM is started (and this is within the NotFound check), you can just return that the VirutalMachineInstance was not found (yet?), without err in the message.

pkg/virtctl/pause/pause.go
```diff
@@ -122,7 +121,13 @@ func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
...
+				}
+				return fmt.Errorf("Error pausing VirtualMachineInstance %s: %v, check if VirtualMachine is running", vmiName, err)
```
slintes: Last but not least, when err != NotFound, probably something worse happened, and we can just keep the generic error message as before. The "check if VirtualMachine is running" advice probably does not help here.

pkg/virtctl/pause/pause.go
xpivarc: Makes sense and is more readable IMO. Thanks

pkg/virtctl/pause/pause.go
```diff
@@ -122,7 +124,18 @@ func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
 			}
 			err = virtClient.VirtualMachineInstance(namespace).Pause(vmiName)
 			if err != nil {
-				return fmt.Errorf("Error pausing VirtualMachineInstance %s: %v", vmiName, err)
+				if errors.IsNotFound(err) {
+					runningStrategy, invalid := vm.RunStrategy()
```
slintes: please use `err` instead of `invalid`, it's best practice

pkg/virtctl/pause/pause.go
```diff
@@ -122,7 +124,18 @@ func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
...
+					if invalid != nil {
+						return invalid
```
slintes: Very unlikely that we ever get here, but I think we should return a better error here. I'd suggest just the generic `return fmt.Errorf("Error pausing VirutalMachineInstance %s: %v", vmiName, err)`. Wdyt?

pkg/virtctl/pause/pause.go
```diff
@@ -122,7 +124,18 @@ func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
...
+					}
+					if runningStrategy == kubevirtV1.RunStrategyHalted {
+						return fmt.Errorf("Error pausing VirtualMachineInstance %s. VirtualMachine %s is not set to run", vmiName, vmiName)
```
slintes: nit: no need to repeat the VM name. Maybe just `Error pausing VirtualMachineInstance %s, it is not running`?

pkg/virtctl/pause/pause.go
```diff
@@ -122,7 +124,18 @@ func (vc *VirtCommand) Run(cmd *cobra.Command, args []string) error {
...
+					}
+					return fmt.Errorf("Error pausing VirtualMachineInstance %s was not found", vmiName)
```
slintes: nit `..., it was not found`

pkg/virtctl/pause/pause.go
xpivarc: Sure very unlikely, maybe we can also remove it in RunStrategy....

pkg/virtctl/pause/pause.go
xpivarc: Just for consistency ... as we always return it with name and to not confuse users.

pkg/virtctl/pause/pause.go
xpivarc: Isn`t it redundant ? VM test was not found vs VM test, it was not found  ?

pkg/virtctl/pause/pause.go
xpivarc: thought I was using both... my bad

pkg/virtctl/pause/pause.go
slintes: sure, but you do it twice here ;)

pkg/virtctl/pause/pause.go
slintes: maybe just add a comma after `VirtualMachineInstance`, `Error pausing VirtualMachineInstance test was not found` does not sound right?

pkg/virtctl/pause/pause.go
xpivarc: Because of VMI and VM, not sure if they can be different ...

slintes: ah, I see, fair point. However, you need to use `vm.Name` instead of `vmiName` for the 2nd parameter then

xpivarc: https://github.com/kubevirt/kubevirt/blob/f77ae94e563d36424c65352826b848a31c58c9fa/pkg/virtctl/pause/pause.go#L121
Seems like they are same.... should I change it ?

pkg/virtctl/pause/pause.go
xpivarc: ah I see 2 info in one long sentence

pkg/virtctl/pause/pause.go
slintes: vmiName is potentially updated right after that line

slintes: so this is the last one, everything else looks to me on the latest commit :+1:
Travis failed downloading a dependency, I restarted the job, I guess it will fine.

slintes:
thanks! :+1: 

/lgtm
/approve

PR #3095, opened 2020-02-23
Fix cluster-sync

oshoval:
Need to retry in case of errors such as
Unable to connect to the server, request timed out
which are temporary

```release-note
None
```

### DIFF
--- a/hack/cluster-deploy.sh
+++ b/hack/cluster-deploy.sh
@@ -85,7 +85,12 @@ until _kubectl -n kubevirt get kv kubevirt; do
     sleep 1
 done
 
-# wait until KubeVirt is ready
-_kubectl wait -n kubevirt kv kubevirt --for condition=Available --timeout 6m || (echo "KubeVirt not ready in time" && dump_kubevirt && exit 1)
+# Wait until KubeVirt is ready
+count=0
+until _kubectl wait -n kubevirt kv kubevirt --for condition=Available --timeout 5m; do
+    ((count++)) && ((count == 5)) && echo "KubeVirt not ready in time" && dump_kubevirt && exit 1
+    echo "Error waiting for KubeVirt to be Available, sleeping 1m and retrying"
+    sleep 1m
+done
 
 echo "Done"


oshoval:
general question out of this PR scope
are we ok with the cluster giving EOF when installing kubevirt ?
or it is an hint to something about cluster management that doesn't get enough priority or so?
(or just a local unimportant error)

oshoval:
failures here arent due to this PR
they are few specific tests

hack/cluster-deploy.sh
```diff
@@ -85,7 +85,12 @@ until _kubectl -n kubevirt get kv kubevirt; do
     sleep 1
 done
 
-# wait until KubeVirt is ready
-_kubectl wait -n kubevirt kv kubevirt --for condition=Available --timeout 6m || (echo "KubeVirt not ready in time" && dump_kubevirt && exit 1)
+# Wait until KubeVirt is ready
+count=0
+until _kubectl wait -n kubevirt kv kubevirt --for condition=Available --timeout 5m; do
```
dhiller: Why don't we just either increase the timeout value here or use an env variable?

oshoval: The problem is that increasing timeout wont help, because there are real hiccups such as (all those i saw in this case)

Unable to connect to the server: EOF
Unable to connect to the server: net/http: TLS handshake timeout
request timed out

they are temporary in our case, so just had to sleep, let the cluster relax and retry the command and it worked

dhiller: Or asked the other way round, did you observe that kubevirt got into ready state after more that 5 minutes? When and where did this happen?

oshoval: what we see is the following:
the original code gets one of the 3 errors above
and the make cluster-sync "fail"

now if you continue to try the wait manually or just oc get pods -A
if will give errors those errors for around 1 minutes, after 1 minute, 
all commands will back to normal and triggering the wait command again will success

dhiller: OK then. Thanks for your answers!

PR #3080, opened 2020-02-17
virt-launcher: lookup qemu gid

jfehlig:
When setting ownership of /dev/kvm, avoid hardcoding a gid of 107
and lookup qemu gid instead.

Signed-off-by: Jim Fehlig <jfehlig@suse.com>

**What this PR does / why we need it**:
Allows dynamically setting group ownership of /dev/kvm

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

NONE
```release-note

```

PR #3100, opened 2020-02-24
Ensure virt-launcher runs with qemu as a supplemental group

stu-gott:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1805627 by specifying fsGroup

**Release note**:
```release-note
None
```

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -95,6 +95,7 @@ type templateService struct {
 	persistentVolumeClaimStore cache.Store
 	virtClient                 kubecli.KubevirtClient
 	clusterConfig              *virtconfig.ClusterConfig
+	launcherSubGid             int64
 }
 
 type PvcNotFoundError error
@@ -317,7 +318,6 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 	var volumeDevices []k8sv1.VolumeDevice
 	var userId int64 = 0
 	var privileged bool = false
-	var qemuUser int64 = 107
 	var volumeMounts []k8sv1.VolumeMount
 	var imagePullSecrets []k8sv1.LocalObjectReference
 
@@ -978,7 +978,7 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 				SELinuxOptions: &k8sv1.SELinuxOptions{
 					Type: selinuxLauncherType,
 				},
-				FSGroup: &qemuUser,
+				FSGroup: &t.launcherSubGid,
 			},
 			TerminationGracePeriodSeconds: &gracePeriodKillAfter,
 			RestartPolicy:                 k8sv1.RestartPolicyNever,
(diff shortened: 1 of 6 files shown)


tests/vmi_lifecycle_test.go
```diff
@@ -1542,6 +1542,33 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
 			Expect(event).To(BeNil(), "virt-handler tried to sync on a VirtualMachineInstance in final state")
 		})
 	})
+
+	Describe("Defaults", func() {
+		Context("FSGroup", func() {
+			FIt("Should run with qemu as supplemental group", func() {
```
awels: I think you forgot to remove the F here

pkg/virt-controller/services/template.go
```diff
@@ -317,6 +317,7 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 	var volumeDevices []k8sv1.VolumeDevice
 	var userId int64 = 0
 	var privileged bool = false
+	var qemuUser int64 = 107
```
slintes: Can we (re-)use something like in https://github.com/kubevirt/kubevirt/pull/3080 and avoid hardcoded group id?

stu-gott: Fair point, thanks! Fixed.

tests/vmi_lifecycle_test.go
stu-gott: Indeed I did. fixed.

pkg/virt-controller/services/template.go
stu-gott: Oops. I fixed the tests.

Fixing this is a little odd--this code is executed in the virt-controller container. The group id in question is in a different container. Retrieving the GID might require a fair bit of engineering. Is that really worth it?

dhiller: Maybe a compromise could be sth like leveraging getent? I.e. sth like this:

```bash
getent group qemu | cut -d ':' -f 3
```

dhiller: I guess there's a package in go for that, at least we could get rid of that magic number here. Praying the GID will stay stable ;-)

stu-gott: The qemu user isn't present in virt-controller's container, so there's nothing to look up. It would require looking up the user info from inside virt-launcher. that would probably require firing up a dummy container.

The GID *is* stable, fwiw. On every redhat-derived system I'm aware of, it's always 107. iirc libvirt-qemu on debian is 64055, which is the closest analogy afaik.

It's really easy to over-engineer this.

slintes: oh, that's bad...
IIRC the context of the other PR was that he was rebuilding KubeVirt on Suse Linux. So having this hardcoded here will break his build again...
Getting the actual GID from virt-launcher might be overengineering for this edge usecase indeed, but maybe it's easy to make it at least configurable here?

stu-gott: Added a config option.

slintes: thanks! :+1:

slintes:
@stu-gott can you please add a release note, so that people get aware that they potentially need to configure the gid? Can also be done after merge. Thanks!

/approve

PR #3103, opened 2020-02-26
Pin the IO and the Emulator threads on the same pCPU when iothreads policy is Automatic

vladikr:

Currently, IOThreads are being placed on the same physical CPUs as the guest's virtual CPUs.
However, this may affect the expected latency of a vCPU
When `isolateEmulatorThread` is requested along with with `dedicatedCpuPlacement`, KubeVirt will allocate an additional pCPU and pin the QEMU emulator thread to it. In order to not affect the vcpu latency, this PR will pin the IOThread and the emulator thread on the same pCPU, if the iothread policy was set to the default "automatic". 

oVirt follows a similar pattern of placing io and the emulation threads on the same pCPU.
https://www.ovirt.org/develop/release-management/features/virt/high-performance-vm.html

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/vmi_iothreads_test.go
+++ b/tests/vmi_iothreads_test.go
@@ -218,6 +218,68 @@ var _ = Describe("IOThreads", func() {
 			// though the limit would have supported 8.
 			table.Entry("[test_id:3096]for four CPUs", 4, 6),
 		)
+
+		// IOThread with Emulator Thread
+
+		It("Should place io and emulator threads on the same pcpu with auto ioThreadsPolicy", func() {
+			policy := v1.IOThreadsPolicyAuto
+			vmi.Spec.Domain.IOThreadsPolicy = &policy
+			vmi.Spec.Domain.CPU = &v1.CPU{
+				Cores:                 1,
+				DedicatedCPUPlacement: true,
+				IsolateEmulatorThread: true,
+			}
+			vmi.Spec.Domain.Resources = v1.ResourceRequirements{
+				Requests: k8sv1.ResourceList{
+					k8sv1.ResourceMemory: resource.MustParse("64M"),
+				},
+			}
+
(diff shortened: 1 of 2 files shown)


tests/vmi_iothreads_test.go
```diff
@@ -218,6 +218,60 @@ var _ = Describe("IOThreads", func() {
...
+				DedicatedCPUPlacement: true,
+				IsolateEmulatorThread: true,
+			}
+			vmi.Spec.Domain.Resources = v1.ResourceRequirements{
+				Requests: k8sv1.ResourceList{
+					k8sv1.ResourceMemory: resource.MustParse("64M"),
+				},
+			}
+
+			tests.AddEphemeralDisk(vmi, "ded2", "virtio", tests.ContainerDiskFor(tests.ContainerDiskCirros))
```
stu-gott: Please rename references to "ded2" to something like "disk1" ... in this test case, the disk isn't flagged as dedicated, so it's confusing me.

stu-gott: actually, it might be helpful to actually create a disk requesting dedicated IOthreads--just to prove they're also still isolated.

vladikr: I've added another disk with a dedicated iothread, to show that it will have a separate thread on the isolated pCPU.
Thanks @stu-gott

tests/vmi_iothreads_test.go
```diff
@@ -218,6 +218,68 @@ var _ = Describe("IOThreads", func() {
 			// though the limit would have supported 8.
 			table.Entry("[test_id:3096]for four CPUs", 4, 6),
 		)
+
+		// IOThread with Emulator Thread
+
+		FIt("Should place io and emulator threads on the same pcpu with auto ioThreadsPolicy", func() {
```
stu-gott: FIt

PR #3107, opened 2020-02-26
Reject CD-ROM devices that use the virtio bus

stu-gott:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1802001

**Release note**:

```release-note
Default bus for CD-ROM devices is now sata.
```

### DIFF
--- a/pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
+++ b/pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
@@ -517,6 +517,15 @@ func ValidateVirtualMachineInstanceSpec(field *k8sfield.Path, spec *v1.VirtualMa
 			}
 			bootOrderMap[order] = true
 		}
+
+		// verify that CD-ROM devices use only SATA
+		if (disk.CDRom != nil) && (disk.CDRom.Bus != "sata") {
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeFieldValueInvalid,
+				Message: fmt.Sprintf("Bus type %s is invalid for CD-ROM device.", disk.CDRom.Bus),
+				Field:   field.Child("domain", "devices", "disks").Index(idx).Child("cdrom", "bus").String(),
+			})
+		}
 	}
 
 	multusDefaultCount := 0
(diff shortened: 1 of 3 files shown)


stu-gott:
/hold

Let's think through the side effects of forcing/not-forcing a default bus before merging.

On the one hand, creating a default bus other than virtio might be surprising. On the other, existing manifests that didn't specify a bus (that can happen, right?) will be invalid on accident.

vladikr:
afaik, cdroms are not allowed on he virtio bus anymore: https://www.redhat.com/archives/libvir-list/2019-January/msg01104.html

stu-gott:
/hold cancel

Ok, then this PR basically does the right thing.

vladikr:
@stu-gott perhaps we should just reject `virtio`, I think cd-roms are allowed on `SCSI` or `ide`(not q35) ?

stu-gott:
I don't understand. We do reject virtio requests for CD-ROM with this PR.

stu-gott:
After testing I've confirmed that you're correct @vladikr. scsi bus should also be valid.

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter_test.go
```diff
@@ -2126,6 +2126,29 @@ var _ = Describe("Validating VMICreate Admitter", func() {
...
+			vmi.Spec.Volumes = append(vmi.Spec.Volumes, v1.Volume{
+				Name: "testcdrom",
+				VolumeSource: v1.VolumeSource{
+					ContainerDisk: &v1.ContainerDiskSource{Image: "fake"},
+				},
+			})
+
+			causes := ValidateVirtualMachineInstanceSpec(k8sfield.NewPath("fake"), &vmi.Spec, config)
+			Expect(len(causes)).To(Equal(1))
+			Expect(causes[0].Field).To(Equal("fake.domain.devices.disks[0].cdrom.bus"))
```
vladikr: @stu-gott this test fails. It expects: 
`fake.domain.devices.disks.domain.devices.disks[0].cdrom.bus`

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
```diff
@@ -1583,6 +1583,16 @@ func validateDisks(field *k8sfield.Path, disks []v1.Disk) []metav1.StatusCause {
 						Field:   field.Index(idx).Child(diskType, "bus").String(),
 					})
 				}
+
+				// special case. virtio is incompatible with CD-ROM for q35 machine types
+				if diskType == "cdrom" && bus == "virtio" {
+					causes = append(causes, metav1.StatusCause{
+						Type:    metav1.CauseTypeFieldValueInvalid,
+						Message: fmt.Sprintf("Bus type %s is invalid for CD-ROM device", bus),
+						Field:   field.Child("domain", "devices", "disks").Index(idx).Child("cdrom", "bus").String(),
```
vladikr: maybe you just need `field.Index(idx).Child("cdrom", "bus").String()` here? 
or `field.Child("domain", "devices", "disks", "disk").Index(idx).Child("cdrom", "bus").String()`

stu-gott: This actually caused the other error. Thanks for catching that!

Issue #2942, opened 2019-12-19
VMI created with the pod/masquerade network on Calico cannot be accessed


luckyfengyong:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
/kind bug
> /kind enhancement

**What happened**:
VMI created with the default network cannot be accessed

**What you expected to happen**:
User could access the VMI by allocated container ip address

**How to reproduce it (as minimally and precisely as possible)**:
In a k8s cluster created by kubeadm v1.15.1, install kubevirt v0.24.0. The container network is provided by calico v3.8.0. 

Create vmi with the following spec

```
apiVersion: kubevirt.io/v1alpha3
kind: VirtualMachineInstance
metadata:
  name: testvmi
spec:
  terminationGracePeriodSeconds: 30
  domain:
    resources:
      requests:
        memory: 1024M
    devices:
      disks:
      - name: containerdisk
        disk:
          bus: virtio
      - name: emptydisk
        disk:
          bus: virtio
      - disk:
          bus: virtio
        name: cloudinitdisk
      interfaces:
        - name: default
          masquerade: {}
          ports:
            - port: 22 # allow incoming traffic on port 80 to get into the virtual machine
  networks:
  - name: default
    pod: {}
  volumes:
  - name: containerdisk
    containerDisk:
      image: kubevirt/fedora-cloud-container-disk-demo:latest
  - name: emptydisk
    emptyDisk:
      capacity: "2Gi"
  - name: cloudinitdisk
    cloudInitNoCloud:
      userData: |-
        #cloud-config
        password: fedora
        chpasswd: { expire: False }
```

```
# kubectl describe vmi testvmi
Name:         testvmi
Namespace:    default
Labels:       kubevirt.io/nodeName=host03
Annotations:  kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"kubevirt.io/v1alpha3","kind":"VirtualMachineInstance","metadata":{"annotations":{},"name":"testvmi","namespace":"def...
              kubevirt.io/latest-observed-api-version: v1alpha3
              kubevirt.io/storage-observed-api-version: v1alpha3
API Version:  kubevirt.io/v1alpha3
Kind:         VirtualMachineInstance
Metadata:
  Creation Timestamp:  2019-12-19T06:20:27Z
  Generation:          9
  Resource Version:    23769874
[...]

PR #3112, opened 2020-02-27
Support ipv6 masquarade

AlonaKaplan:

- Adding masquarade and dnat rules to ip6tables.
- Adding masquarade and dnat rules to ip6 nat nftable.
- Giving ipv6 address to k6t-eth0 bridge.
- Turning on 'net.ipv6.conf.all.forwarding' on the virt-launcher pod
(done via the virt-handler).

**The PR doesn't contain**:
- ipv6 support to bind methods other than masquarade.

**Further changes needed**:
- DHCP server support for ipv6 or deciding on another method to set
the ip + default gateway on the vm.
- Adding vmIpv6NetworkCIDR to podInterface schema.
- Support migration -  PR https://github.com/kubevirt/kubevirt/pull/3221
- Remove the skip of the PVC tests once PR https://github.com/kubevirt/kubevirt/pull/3171 is merged.
- Remove the skip of [ref_id:1182]Probes for readiness should succeed [test_id:1200][posneg:positive]with working HTTP probe and http server - once https://issues.redhat.com/browse/CNV-4454 is done.

The PR was tested by specifying manually on the vm-
$ sudo ip -6 addr add fd2e:f1fe:9490:a8ff::2/120 dev eth0
$ sudo ip -6 route add default via fd2e:f1fe:9490:a8ff::1 src fd2e:f1fe:9490:a8ff::2

To configure dns on the vm
echo "nameserver <dnsClusterIp>" >> /etc/resolv.conf
dnsClusterIp - the output of `kubectl get services -n kube-system` -> ClusterIP

**Release note**:
```release-note
Add IPv6 support for masquerade. IPv6 support is still highly experimental and not supported.
```


qinqon:
@AlonaKaplan to fix Phrometeus at ipv6 lane 
https://github.com/kubevirt/kubevirt/pull/3142/commits/7e09eda9cd08bd6f4cfe8c80cdb408c74c4fda9d

I have take a look at expose ports ones, maybe we don't have ipv6 support at netcat either at the server inside the cirros VM or at the client at the netcat pod.

qinqon:
/test pull-kubevirt-e2e-k8s-1.16 

Infra issue related to docker daemon down.

qinqon:
/test pull-kubevirt-e2e-k8s-1.15-ceph

Infra issue related to docker daemon down.

qinqon:
/test pull-kubevirt-e2e-k8s-multus-1.13.3

Infra issue related to docker daemon down.

qinqon:
Running locally the windows job got the following

```
STEP: Creating winrm-cli pod for the future use                                                                                                                                                                                              
STEP: Starting the windows VirtualMachineInstance                                                                                                                                                                                            
{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"Event(v1.ObjectReference{Kind:\"VirtualMachineInstance\", Namespace:\"kubevirt-test-default\", Name:\"testvminsfz5ckgxdvjk8tw2vkdlmgf7pqbzdsh7nwrs62vtzw4xbr4\", U
ID:\"4cc9fe02-6ea6-4621-832a-c64501ce18ed\", APIVersion:\"kubevirt.io/v1alpha3\", ResourceVersion:\"2766\", FieldPath:\"\"}): type: 'Normal' reason: 'SuccessfulCreate' Created virtual machine pod virt-launcher-testvminsfz5ckgxdvjk8tw2vkd
lmgf7pqbzdsh7nwrsvfnwr","name":"testvminsfz5ckgxdvjk8tw2vkdlmgf7pqbzdsh7nwrs62vtzw4xbr4","namespace":"kubevirt-test-default","pos":"utils.go:351","timestamp":"2020-03-12T11:07:34.705277Z","uid":"4cc9fe02-6ea6-4621-832a-c64501ce18ed"}    
{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"Event(v1.ObjectReference{Kind:\"VirtualMachineInstance\", Namespace:\"kubevirt-test-default\", Name:\"testvminsfz5ckgxdvjk8tw2vkdlmgf7pqbzdsh7nwrs62vtzw4xbr4\", U
ID:\"4cc9fe02-6ea6-4621-832a-c64501ce18ed\", APIVersion:\"kubevirt.io/v1alpha3\", ResourceVersion:\"2808\", FieldPath:\"\"}): type: 'Normal' reason: 'NodeUnresponsive' virt-handler on node node01 is not responsive, marking VMI as failed","name":"testvminsfz5ckgxdvjk8tw2vkdlmgf7pqbzdsh7nwrs62vtzw4xbr4","namespace":"kubevirt-test-default","pos":"utils.go:351","timestamp":"2020-03-12T11:12:48.266157Z","uid":"4cc9fe02-6ea6-4621-832a-c64501ce18ed"}

```

Looks like there is something fishy with virt-handler.

qinqon:
virt-launcher errors 
```
{"component":"virt-launcher","level":"info","msg":"Sorted all collected sidecar sockets per hook point based on their priority and name: map[]","pos":"manager.go:71","timestamp":"2020-03-12T09:13:52.654206Z"}
{"component":"virt-launcher","level":"info","msg":"Watchdog file created at /var/run/kubevirt/watchdog-files/kubevirt-test-default_testvmipdt4ff5zs89jg2bqmh59tqrn22g5tmvhkfkqx7pbh97hfj5z","pos":"virt-launcher.go:155","timestamp":"2020-03-12T09:13:52.659617Z"}
{"component":"virt-launcher","level":"info","msg":"Connecting to libvirt daemon: qemu:///system","pos":"libvirt.go:355","timestamp":"2020-03-12T09:13:52.666457Z"}
{"component":"virt-launcher","level":"info","msg":"Connecting to libvirt daemon failed: virError(Code=38, Domain=7, Message='Failed to connect socket to '/var/run/libvirt/libvirt-sock': No such file or directory')","pos":"libvirt.go:363","timestamp":"2020-03-12T09:13:52.667827Z"}
{"component":"virt-launcher","level":"info","msg":"libvirt version: 5.6.0, package: 7.fc31 (Unknown, 2019-11-22-21:16:59, )","subcomponent":"libvirt","thread":"46","timestamp":"2020-03-12T09:13:52.745000Z"}

```

qinqon:
The intermintent error at 1.14 and 1.15 are related to some VMI interfaces not deleted after the VMI is removed 
```
{"component":"virt-handler","level":"info","msg":"Domain does not exist","pos":"vm.go:869","timestamp":"2020-03-12T09:15:20.034723Z"}
{"component":"virt-handler","level":"error","msg":"failed to create an interface: k6t-eth0-nic","pos":"podinterface.go:601","reason":"file exists","timestamp":"2020-03-12T09:15:20.058252Z"}
{"component":"virt-handler","level":"fatal","msg":"failed to prepare pod networking","pos":"podinterface.go:111","reason":"file exists","timestamp":"2020-03-12T09:15:20.058431Z"}
panic: file exists

goroutine 370 [running, locked to thread]:
kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/network.(*PodInterface).PlugPhase1(0x2ea0768, 0xc000cd8000, 0xc000b9ab40, 0xc0008a03c0, 0x1a0f767, 0x4, 0x78a5, 0xc000d36d08, 0xc000b9ab40)
	pkg/virt-launcher/virtwrap/network/podinterface.go:112 +0x915
```

Could this be related with ipv6 changes ? feels like some interfaces are not deleted now.

go.mod
```diff
@@ -111,3 +111,5 @@ replace (
 
 	kubevirt.io/client-go => ./staging/src/kubevirt.io/client-go
 )
+
+go 1.13
```
qinqon: Is this needed going to 1.13 have to be done carefully.

qinqon:
/retest

Not related 
```
kubevirt   0s    
+ count=0
+ _kubectl wait -n kubevirt kv kubevirt --for condition=Available --timeout 5m
+ export KUBECONFIG=/home/prow/go/src/github.com/kubevirt/kubevirt/_ci-configs/k8s-1.14/.kubeconfig
+ KUBECONFIG=/home/prow/go/src/github.com/kubevirt/kubevirt/_ci-configs/k8s-1.14/.kubeconfig
+ /home/prow/go/src/github.com/kubevirt/kubevirt/_ci-configs/k8s-1.14/.kubectl wait -n kubevirt kv kubevirt --for condition=Available --timeout 5m
kubevirt.kubevirt.io/kubevirt condition met
+ echo Done
Done
+ hack/dockerized bazel shutdown
go version go1.12.8 linux/amd64

docker: Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:297: getting the final child's pid from pipe caused \"EOF\"": unknown.
+ make cluster-down
```

qinqon:
@AlonaKaplan can you rebase the PR ? just in case.

go.mod
AlonaKaplan: It was auto-generated. Do you think it should be removed?

AlonaKaplan:
> @AlonaKaplan can you rebase the PR ? just in case.

sure, done

qinqon:
@AlonaKaplan bisecting at https://github.com/kubevirt/kubevirt/pull/3161 to discover ipv4 issue.

pkg/virt-launcher/virtwrap/api/defaults.go
```diff
@@ -8,6 +8,7 @@ const (
 	resolvConf        = "/etc/resolv.conf"
 	DefaultProtocol   = "TCP"
 	DefaultVMCIDR     = "10.0.2.0/24"
+	DefaultVMIpv6CIDR = "fd2e:f1fe:9490:a8ff::/120"
```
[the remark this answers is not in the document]
AlonaKaplan: Random

go.mod
qinqon: Was auto generated by `make deps-update` ? well we should put 1.12 here but it's not related to the PR.

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -142,47 +149,97 @@ func (h *NetworkUtilsHandler) UseIptables() bool {
 
 	return true
 }
-func (h *NetworkUtilsHandler) IptablesNewChain(table, chain string) error {
-	iptablesObject, err := iptables.New()
+
+func (h *NetworkUtilsHandler) ConfigureIpv6Forwarding() error {
+	_, err := exec.Command("sysctl", "net.ipv6.conf.all.forwarding=1").CombinedOutput()
```
qinqon: Are we adding the proper policy to run sysctl at container ? 
https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#podsecuritypolicy

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -142,47 +149,97 @@ func (h *NetworkUtilsHandler) UseIptables() bool {
...
+	return err
+}
+
+func (h *NetworkUtilsHandler) Ipv4NatEnabled() bool {
+	lsmod, _ := exec.Command("lsmod").CombinedOutput()
+	lsmodString := string(lsmod)
+	if !strings.Contains(lsmodString, "iptable_nat") && !strings.Contains(lsmodString, "nf_nat_ipv4") {
+		log.Log.Errorf("no ipv4 support")
```
qinqon: "ipv4 nat unsupported"

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -142,47 +149,97 @@ func (h *NetworkUtilsHandler) UseIptables() bool {
...
+		return false
+	}
+	return true
+}
+
+func (h *NetworkUtilsHandler) Ipv6NatEnabled() bool {
+	lsmod, _ := exec.Command("lsmod").CombinedOutput()
+	lsmodString := string(lsmod)
+	if !strings.Contains(lsmodString, "iptable_nat") && !strings.Contains(lsmodString, "nf_nat_ipv6") {
+		log.Log.Errorf("no ipv6 support")
```
qinqon: "ipv6 nat unsupported"

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -142,47 +149,97 @@ func (h *NetworkUtilsHandler) UseIptables() bool {
...
-		return fmt.Errorf("failed to create new nftable error %s", string(output))
-	}
 
-	return nil
+func (h *NetworkUtilsHandler) GetNftIpString(proto iptables.Protocol) string {
+	ipString := "ip"
+	if proto == iptables.ProtocolIPv6 {
+		ipString = "ip6"
+	}
+	return ipString
```
qinqon: We can simplify it.
```golang
func (h *NetworkUtilsHandler) GetNftIpString(proto iptables.Protocol) string {
	if proto == iptables.ProtocolIPv6 {
		return "ip6"
	}
	return "ip"
}
```

pkg/virt-launcher/virtwrap/network/generated_mock_common.go
```diff
@@ -6,6 +6,7 @@ package network
 import (
```
qinqon: This has being generated right ?

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -532,7 +552,32 @@ func (p *MasqueradePodInterface) discoverPodNetworkInterface() error {
...
+}
 
+func configureVifV6Addresses(p *MasqueradePodInterface, err error) error {
+	if p.vmIpv6NetworkCIDR == "" {
+		p.vmIpv6NetworkCIDR = api.DefaultVMIpv6CIDR
+	}
+
+	defaultGatewayIpv6, vmIpv6, err := Handler.GetHostAndGwAddressesFromCIDR(p.vmIpv6NetworkCIDR)
+	if err != nil {
+		log.Log.Errorf("failed to get gw and vm available ipv6 addresses from CIDR %s", p.vmIpv6NetworkCIDR)
```
qinqon: We can use log.Log.Reason(err).Errorf...

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -532,7 +552,32 @@ func (p *MasqueradePodInterface) discoverPodNetworkInterface() error {
...
+		return err
+	}
+
+	gatewayIpv6Addr, err := Handler.ParseAddr(defaultGatewayIpv6)
+	if err != nil {
+		return fmt.Errorf("failed to parse gateway ipv6 address %s", gatewayIpv6Addr)
```
qinqon: Print the error too either as "%v" or by using github.com/pkg/errors and errors.Wrapf

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -532,7 +552,32 @@ func (p *MasqueradePodInterface) discoverPodNetworkInterface() error {
...
+	}
+	p.vif.GatewayIpv6 = gatewayIpv6Addr.IP.To16()
+	p.gatewayIpv6Addr = gatewayIpv6Addr
+
+	vmAddr, err := Handler.ParseAddr(vmIpv6)
+	if err != nil {
+		return fmt.Errorf("failed to parse vm ipv6 address %s", vmIpv6)
```
qinqon: ditto.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -577,10 +622,25 @@ func (p *MasqueradePodInterface) preparePodNetworkInterfaces() error {
...
 	}
 
-	err = p.createNatRules()
-	if err != nil {
-		log.Log.Errorf("failed to create nat rules for vm error: %v", err)
-		return err
+	if Handler.Ipv4NatEnabled() {
+		err = p.createNatRules(iptables.ProtocolIPv4)
+		if err != nil {
+			log.Log.Errorf("failed to create ipv4 nat rules for vm error: %v", err)
```
qinqon: use Reason(err)

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -577,10 +622,25 @@ func (p *MasqueradePodInterface) preparePodNetworkInterfaces() error {
...
+			return err
+		}
+	}
+	if Handler.IsIpv6Enabled(p.podNicLink) && Handler.Ipv6NatEnabled() {
+		err = Handler.ConfigureIpv6Forwarding()
+		if err != nil {
+			log.Log.Errorf("failed to turn on net.ipv6.conf.all.forwarding")
```
qinqon: Reason(err)

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -577,10 +622,25 @@ func (p *MasqueradePodInterface) preparePodNetworkInterfaces() error {
...
+			return err
+		}
+
+		err = p.createNatRules(iptables.ProtocolIPv6)
+		if err != nil {
+			log.Log.Errorf("failed to create ipv6 nat rules for vm error: %v", err)
```
qinqon: Reason(err)

tests/infra_test.go
```diff
@@ -172,7 +172,7 @@ var _ = Describe("Infrastructure", func() {
 			By("Finding the prometheus endpoint")
 			pod, err = kubecli.NewVirtHandlerClient(virtClient).Namespace(tests.KubeVirtInstallNamespace).ForNode(nodeName).Pod()
 			Expect(err).ToNot(HaveOccurred(), "Should find the virt-handler pod")
-			metricsURL = fmt.Sprintf("https://%s:%d/metrics", pod.Status.PodIP, 8443)
+			metricsURL = fmt.Sprintf("https://%s:%d/metrics", tests.FormatIpForCurl(pod.Status.PodIP), 8443)
```
qinqon: Can't we use formatIpForUri ?

pkg/virt-launcher/virtwrap/network/common.go
AlonaKaplan: Would it work if don't?

AlonaKaplan:
Roman, thanks for the review. The pr message was outdated.

> A first superficial review.
> 
> One more question regarding to the scope of the PR:
> 
> > The PR doesn't contain:
> > 
> > * Adding masquarade and dnat rule to nftable.
> 
> Wouldn't we need to support things with `iptables` and `nftables`?
> 

The pr contains the nftables as well.

> > * ipv6 support to bind methods other than masquarade.
> 
> That sounds like it can be done in an extra PR.
> 

Sure, once we will finish with this one:)

> > * Fixing virtctl console to work with ipv6 vm.
> 
> That sounds like it should also work in the context of this PR.

Yes, the pr include the fix.

Updated the pr message.

pkg/virt-launcher/virtwrap/network/common.go
AlonaKaplan: virt-launcher has ipv4 enabled by default. But the ipv6 is not. Enabling ipv6 forwarding on the node doesn't help. Do you have another idea how we can make the ipv6 enabled on the virt-launcher?

rmohr: It is also not always the case for all ipv4 providers ( https://github.com/kubevirt/kubevirt/issues/2942). You could consider moving this part to `virt-handler` and enable it for both ipv4 and ipv6 always. virt-handler is fully privileged and can enter the right network namespace.

pkg/virt-launcher/virtwrap/network/common.go
AlonaKaplan: done

pkg/virt-launcher/virtwrap/network/common.go
AlonaKaplan: done

pkg/virt-launcher/virtwrap/network/common.go
AlonaKaplan: done

pkg/virt-launcher/virtwrap/network/podinterface.go
AlonaKaplan: done

pkg/virt-launcher/virtwrap/network/podinterface.go
AlonaKaplan: done

pkg/virt-launcher/virtwrap/network/podinterface.go
AlonaKaplan: done

pkg/virt-launcher/virtwrap/network/podinterface.go
AlonaKaplan: done

pkg/virt-launcher/virtwrap/network/podinterface.go
AlonaKaplan: done

pkg/virt-launcher/virtwrap/network/podinterface.go
AlonaKaplan: done

tests/infra_test.go
AlonaKaplan: Do you mean changing the name of the method?

rmohr: Best would be `formatIPForURL` ;)

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -142,47 +149,97 @@ func (h *NetworkUtilsHandler) UseIptables() bool {
...
 	return nil
 }
-func (h *NetworkUtilsHandler) NftablesNewTable(table string) error {
-	output, err := exec.Command("nft", "add", "table", table).CombinedOutput()
-	if err != nil {
-		return fmt.Errorf("failed to create new nftable error %s", string(output))
-	}
 
-	return nil
+func (h *NetworkUtilsHandler) GetNftIpString(proto iptables.Protocol) string {
```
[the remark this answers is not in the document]
AlonaKaplan: done

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -508,8 +513,23 @@ func (p *MasqueradePodInterface) discoverPodNetworkInterface() error {
 	}
 
 	// Get interface MTU
-	p.vif.Mtu = uint16(p.podNicLink.Attrs().MTU)
+	p.vif.Mtu = uint16(p.podNicLink.Attrs().MTU) // TODO - why do we override the mtu that was specified in the yaml - iface.MacAddress
```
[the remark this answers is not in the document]
AlonaKaplan: It was a comment to myself that I forgot to remove. Now I see the yaml has the mac and here it is the mtu. Removed the comment:)

automation/test.sh
```diff
@@ -280,9 +280,11 @@ elif [[ $TARGET =~ genie.* ]]; then
 elif [[ $TARGET =~ sriov.* ]]; then
   ginko_params="$ginko_params --ginkgo.focus=SRIOV"
 elif [[ $TARGET =~ gpu.* ]]; then
-  ginko_params="$ginko_params --ginkgo.focus=GPU" 
+  ginko_params="$ginko_params --ginkgo.focus=GPU"
 elif [[ $TARGET =~ (okd|ocp).* ]]; then
   ginko_params="$ginko_params --ginkgo.skip=Genie|SRIOV|GPU"
+elif [[ $TARGET =~ ipv6.* ]]; then
+  ginko_params="$ginko_params --ginkgo.skip=Multus|Genie|SRIOV|GPU|.*slirp.*|.*bridge.*|.*PVC.*"
```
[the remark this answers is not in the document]
AlonaKaplan: PVC is skipped until https://github.com/kubevirt/kubevirt/pull/3171 is merged.
I think I can remove the Multus, SRIOV and GPU skip.
@qinqon is there a spacial reason you skipped it? Are you ok with removing it?

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -142,47 +149,97 @@ func (h *NetworkUtilsHandler) UseIptables() bool {
...
 }
-func (h *NetworkUtilsHandler) IptablesNewChain(table, chain string) error {
-	iptablesObject, err := iptables.New()
+
+func (h *NetworkUtilsHandler) ConfigureIpv6Forwarding() error {
+	_, err := exec.Command("sysctl", "net.ipv6.conf.all.forwarding=1").CombinedOutput()
+	return err
+}
+
+func (h *NetworkUtilsHandler) Ipv4NatEnabled() bool {
```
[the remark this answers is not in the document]
AlonaKaplan: done.
For iptables - `iptables -t nat -L` and `ip6tables -t nat -L`
For nftables- `nft -f /etc/nftables/ipv4-nat.nft` and `nft -f /etc/nftables/ipv6-nat.nft` for nftables

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -142,47 +149,97 @@ func (h *NetworkUtilsHandler) UseIptables() bool {
...
+	lsmodString := string(lsmod)
+	if !strings.Contains(lsmodString, "iptable_nat") && !strings.Contains(lsmodString, "nf_nat_ipv4") {
+		log.Log.Errorf("no ipv4 support")
+		return false
+	}
+	return true
+}
+
+func (h *NetworkUtilsHandler) Ipv6NatEnabled() bool {
+	lsmod, _ := exec.Command("lsmod").CombinedOutput()
```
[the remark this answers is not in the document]
AlonaKaplan: Done. And all `phase1` including the iptables/nftables configuration is done in the virt-handler anyway since [Configure pod network for launcher from virt-handler](https://github.com/kubevirt/kubevirt/pull/2837) was merged.

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -142,47 +149,97 @@ func (h *NetworkUtilsHandler) UseIptables() bool {
...
+	addrs, err := h.AddrList(link, netlink.FAMILY_V6)
+	if err != nil {
+		linkLog.Reason(err).Errorf("ipv6 disabled, failed retrieving ipv6 addresses")
+		return false
+	}
+	if len(addrs) == 0 {
+		linkLog.Infof("ipv6 disabled, there is no ipv6 addresses")
+	}
+	for _, addr := range addrs {
+		if !addr.IP.IsLinkLocalUnicast() {
```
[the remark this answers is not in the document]
AlonaKaplan: We had to check whether eth0 has only link local address configured because of a mysteries bug that is reproduced in prow only. On some of the ipv4 lanes (not all of them) configuring ipv6 address on the k6t-eth0 bridge causes virt-handler to crash.

This is a workaround till we understand the actual root cause of this crash.

@qinqon can you please elaborate?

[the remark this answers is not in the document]
rmohr: I think it is sane to not try to configure anything ipv6 related if you are on an ipv4 cluster. Therefore I would suggest to keep a check, but check via the downwards api if you are on an ipv6 enabled cluster.

automation/test.sh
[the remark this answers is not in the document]
phoracek: I'm afraid we cannot test this with Multus, nor SR-IOV. kindnet (which provides us with IPv6 support) does not play well with Multus.

While we were developing kind-sriov lane, we were not able to use kindnet+multus, so we switched to flannel and later calico.

So since we need kindnet for IPv6 and kindnet does not work with Multus, we cannot enabled none of these two lanes.

We may try again enabling Multus on kind to see if anything changed since, but we cannot use the current provider.

I don't see this as that big of a risk since we connect Multus interfaces only through `bridge` binding and SR-IOV through `sriov`.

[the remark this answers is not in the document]
AlonaKaplan: @rmohr before I'm adding it back. What is GPU? Does it supported on kind provider?

pkg/virt-launcher/virtwrap/network/common.go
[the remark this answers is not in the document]
qinqon: Agree with Roman, is better to rely on PODs interface, at the end is what CNI provides with non link local address or other mechanisms, will add a commit for that.

automation/test.sh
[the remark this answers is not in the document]
qinqon: What I did was just copy the default lane and skip slirp and bridge, about multus, we can try to hack calico into kind provider but noone has do it, let's wait to have a proper provider the k8s-1.17 has already calico running and the k8s-1.17 is passing most of the times here, also I think Roman is preparing the dnsmasq container to assign ipv6 addresses too.

pkg/virt-launcher/virtwrap/network/common.go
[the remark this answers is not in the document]
phoracek: Yeah, link-local addresses can be used for connectivity with the same network.

But if we assume that some CNI relies on link-local, they may as well behave as IPv4 externally fo addressing between pods and use IPv6 link-local only to traverse between host and the pod (extreme example, I know). If we want to allow all CNIs to function, maybe we should forward even link-local through the masquerade, no matter if the network CNI is considered IPv4 or IPv6 externally.

Would that be possible? Can we just setup the binding for all found IPs on the pod interface?

automation/test.sh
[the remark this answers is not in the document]
qinqon: btw there are more PVC related test without PVC in the test name

cluster-up/cluster/kind/common.sh
```diff
@@ -95,7 +98,37 @@ function _configure_network() {
 function prepare_workers() {
     # appending eventual workers to the yaml
     for ((n=0;n<$(($KUBEVIRT_NUM_NODES-1));n++)); do
-        echo "- role: worker" >> ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
+        cat << EOF >> ${KUBEVIRTCI_CONFIG_PATH}/$KUBEVIRT_PROVIDER/kind.yaml
+- role: worker
+  kubeadmConfigPatches:
+  - |
+    kind: JoinConfiguration
```
phoracek: This should be in Cluster config, not per node: https://github.com/kubernetes-sigs/kind/blob/master/pkg/apis/config/v1alpha3/types.go#L41

phoracek: Not a change for this PR, but we need to resolve it on kubevirtci.

qinqon: Cannot be cluster config, kubevirt e2e tests expect this to not be activated are master, at qemu provides is only activated at workers, nodes.sh

phoracek: Yeah, I think I have to switch API version of that type. Was doing some more debugging. Will post all the changes so we can review it on Monday. On my test PR, sriov lane passes again

tests/utils.go
```diff
@@ -1940,16 +1940,29 @@ func NewRandomFedoraVMIWithDmidecode() *v1.VirtualMachineInstance {
...
+	virtClient, err := kubecli.GetKubevirtClient()
+	PanicOnError(err)
+
+	dnsServerIP, err := getClusterDnsServiceIP(virtClient)
+	PanicOnError(err)
+
+	ipv6UserDataString := ""
+	if netutils.IsIPv6String(dnsServerIP) {
+		ipv6UserDataString = fmt.Sprintf(`sudo ip -6 addr add fd2e:f1fe:9490:a8ff::2/120 dev eth0
+                             sudo ip -6 route add default via fd2e:f1fe:9490:a8ff::1 src fd2e:f1fe:9490:a8ff::2
```
maiqueb: I see that we're hard-coding the IP (also route) in different places in the tests (at least in the cloud init & in the expector). 

Furthermore, this IP has to match what is being assigned by the https://github.com/kubevirt/kubevirt/pull/3112/files#diff-d3ed8eb8cca020f37cab207c079f2627R563 algorithm. 

That's not good, and should be addressed either in this, or on a follow up PR.

AlonaKaplan: This PR doesn't include adding vmIpv6NetworkCIDR to podInterface schema.
It supports only the default one.

It would be addressed in a follow up pr, please refer to the PR message.

pkg/virt-launcher/virtwrap/network/common.go
AlonaKaplan: This code is currently in virt-handler. I would prefer to enable the ipv4 in a separate PR.

pkg/virt-launcher/virtwrap/network/common.go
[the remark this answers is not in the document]
AlonaKaplan: The iptables masquerade mechanism is automatically taking the ip address of the outgoing interface. I'm not sure it will take a link local address. If it is not working maybe we can change the rules to use snat instead of masquerade.
Is it ok to finish with this PR and do those experiments in a different one?

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -681,47 +742,53 @@ func (p *MasqueradePodInterface) createBridge() error {
 		return err
 	}
 
+	if Handler.IsIpv6Enabled(p.podNicLink) {
```
rmohr, line 745: `IsIPv6Enabled` seems to be called multiple times. Consider running the check when constructing the `MasqueratePodInterface` object and store the result in the struct.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -772,40 +839,59 @@ func (p *MasqueradePodInterface) createNatRulesUsingIptables() error {
 	return nil
 }
 
-func (p *MasqueradePodInterface) createNatRulesUsingNftables() error {
-	err := Handler.NftablesLoad("ipv4-nat")
-	if err != nil {
-		return err
+func getGatewayByProtocol(p *MasqueradePodInterface, proto iptables.Protocol) string {
```
rmohr, line 842: can you store the `proto` in the struct and just add a `getGateway()` method to `MasqueradePodInterface`?

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -772,40 +839,59 @@ func (p *MasqueradePodInterface) createNatRulesUsingIptables() error {
...
+	if proto == iptables.ProtocolIPv4 {
+		return p.gatewayAddr.IP.String()
+	} else {
+		return p.gatewayIpv6Addr.IP.String()
 	}
+}
+
+func getVifIpByProtocol(p *MasqueradePodInterface, proto iptables.Protocol) string {
```
rmohr, line 850: Also here, why not a method?

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -772,40 +839,59 @@ func (p *MasqueradePodInterface) createNatRulesUsingIptables() error {
...
+	if proto == iptables.ProtocolIPv4 {
+		return p.vif.IP.IP.String()
+	} else {
+		return p.vif.IPv6.IP.String()
+	}
+}
+
+func getLoopbackAdrress(proto iptables.Protocol) string {
```
rmohr, line 858: Also here, why not a method?

pkg/virt-launcher/virtwrap/network/podinterface_test.go
```diff
@@ -98,21 +108,53 @@ var _ = Describe("Pod Network", func() {
...
-			IP:      *masqueradeVmAddr,
-			MAC:     fakeMac,
-			Mtu:     1410,
-			Gateway: masqueradeGwAddr.IP.To4()}
+			IP:          *masqueradeVmAddr,
+			IPv6:        *masqueradeIpv6VmAddr,
+			MAC:         fakeMac,
+			Mtu:         1410,
+			Gateway:     masqueradeGwAddr.IP.To4(),
+			GatewayIpv6: masqueradeIpv6GwAddr.IP.To16()}
```
rmohr, line 129: We are targeting ipv6 only right now. Is this separation already done to have a way in the future to express dual-stack requirements?

tests/probes_test.go
```diff
@@ -47,11 +47,15 @@ var _ = Describe("[ref_id:1182]Probes", func() {
 			},
 		}
 		table.DescribeTable("should succeed", func(readinessProbe *v12.Probe, serverStarter func(vmi *v12.VirtualMachineInstance, port int)) {
+			if tests.IsRunningOnKindInfra() && readinessProbe == httpProbe {
+				Skip("Skip http probe test till card https://issues.redhat.com/browse/CNV-4454 is fixed")
```
rmohr, line 51: is that public accessable?

tests/storage_test.go
```diff
@@ -195,22 +199,39 @@ var _ = Describe("Storage", func() {
...
+
 			It("[test_id:3136]should be successfully started", func() {
+				tests.SkipPVCTestIfRunnigOnKindInfra()
+
 				// Start the VirtualMachineInstance with the PVC attached
 				vmi := tests.NewRandomVMIWithEphemeralPVC(tests.DiskAlpineHostPath)
-				tests.RunVMIAndExpectLaunch(vmi, 90)
-
+				if isRunOnKindInfra {
+					tests.RunVMIAndExpectLaunchIgnoreWarnings(vmi, 90)
```
rmohr, line 213: What is the exact warning you are ignoring?

maiqueb: All info about this topic - along with the specific warning you're asking about - can be found in [this mailing list question](https://groups.google.com/forum/#!topic/kubevirt-dev/u8DL-8-KFxY).

Unfortunately, I had no responses (yet?).

maiqueb: Thanks for replying on [the thread](https://groups.google.com/forum/#!topic/kubevirt-dev/u8DL-8-KFxY)!

I've pushed https://github.com/kubevirt/kubevirt/pull/3233 to address that, but it doesn't work for one of the tests - I'll need someone storage savy to help me out.

Could we 'get away with' opening an issue on kubevirt, linking against it on the test, and fix it afterwards ?

pkg/virt-launcher/virtwrap/network/podinterface.go
AlonaKaplan: Done, I hope I understood your intention correctly.

pkg/virt-launcher/virtwrap/network/podinterface_test.go
AlonaKaplan: yes

tests/probes_test.go
AlonaKaplan: no, will open a bugzilla, will change it to a PR link. Thanks for noticing it.

AlonaKaplan:
> I checked the iptables rules to a degree, but assume that they are double checked by someone from networking too.
> 

Yes, I verified the rules.
I think the 'KUBEVIRT_POSTINBOUND' chain is redundant. It changes the source address of all the VM incoming packets to the address of the gateway bridge.
All the scenarios that I verified are perfectly working without this chain.

@SchSeba thinks this chain was added for istio. I didn't verify it.

Anyway, this PR just replicates the ipv4 rules for ipv6. If we want to add improvements to the rules, it can be done in a separate PR.

> I thing that the MasqueradeInterface IPv6/IPv4 switches and detection is right now extremely hard to follow. It should be improved. I made some suggestions but any other way to make that easier to follow is fine too.

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -142,47 +146,77 @@ func (h *NetworkUtilsHandler) UseIptables() bool {
...
+		log.Log.Infof("ipv6 disabled, there is no ipv6 addresses from %s", linkName)
+		return false
+	}
+	for _, addr := range addrs {
+		if !addr.IP.IsLinkLocalUnicast() {
+			log.Log.Infof("ipv6 enabled, there is a non link local ipv6 address from %s", linkName)
+			return true
+		}
+	}
+	log.Log.Infof("ipv6 disabled, all the ipv6 addresses are link local from %s", linkName)
```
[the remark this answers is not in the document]
AlonaKaplan: I changed the detection way to check whether the 'MY_POD_IP' env var is ipv6 address as you suggested previously.  In the future when we will have dual stack I'm not sure how the status.podIP will be chosen. I guess we will have to change the code to use status.podIPs.
Also changed the log verbosity level to V(5).

tests/utils.go
```diff
@@ -2640,6 +2662,64 @@ func CheckForTextExpecter(vmi *v1.VirtualMachineInstance, expected []expect.Batc
 	return err
 }
 
+func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter, virtClient kubecli.KubevirtClient) error {
```
maiqueb, line 2665: This function is only working on cirros images, afaiu.

The FedoraExpecter - which also calls this function - has a [different prompt](https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L2721). As can be seen below, this function 'expects' the prompt to be '\$' which matches the [cirrOS prompt](https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L2672).

I think this slipped through because the fedora expecter is used (only?) on VMIs that were configured via cloud init - and as such don't rely on this function.

I guess we need to enable this function to handle multiple prompt types - depending on the image used - or, assume this function is only to be used on cirrOS images - leaving cloudInit as the only way to configure the IPs on fedora VMs.

If we choose the latter approach, I think the function should be renamed, to reflect that it must only be used on cirrOS VMIs.

tests/probes_test.go
```diff
@@ -47,11 +47,15 @@ var _ = Describe("[ref_id:1182]Probes", func() {
 			},
 		}
 		table.DescribeTable("should succeed", func(readinessProbe *v12.Probe, serverStarter func(vmi *v12.VirtualMachineInstance, port int)) {
+			if tests.IsRunningOnKindInfra() && readinessProbe == httpProbe {
+				Skip("Skip http probe test till PR https://github.com/kubevirt/kubevirt/pull/3224 is merged")
```
maiqueb, line 51: You could probably refer this [issue](https://github.com/kubevirt/kubevirt/issues/3247) I've opened about it instead; I'll abandon that PR soon, in favor of another solution.

examples/vmi-masquerade.yaml
```diff
@@ -40,6 +40,9 @@ spec:
       userData: |-
         #!/bin/bash
         echo "fedora" |passwd fedora --stdin
+        ip -6 addr add fd2e:f1fe:9490:a8ff::2/120 dev eth0
```
davidvossel, line 43: why is this necessary?

pkg/virt-launcher/virtwrap/api/defaults.go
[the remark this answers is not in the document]
davidvossel: It's a random address in the [ula](https://en.wikipedia.org/wiki/Unique_local_address) range, which is exactly what we want here.

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -142,47 +148,65 @@ func (h *NetworkUtilsHandler) UseIptables() bool {
...
 }
-func (h *NetworkUtilsHandler) IptablesNewChain(table, chain string) error {
-	iptablesObject, err := iptables.New()
+
+func (h *NetworkUtilsHandler) ConfigureIpv6Forwarding() error {
+	_, err := exec.Command("sysctl", "net.ipv6.conf.all.forwarding=1").CombinedOutput()
+	return err
+}
+
+func (h *NetworkUtilsHandler) IsIpv6Enabled(link netlink.Link) bool {
```
davidvossel, line 157: this function isn't using the `link` argument at all. It could be a bit deceiving

tests/utils.go
```diff
@@ -1937,16 +1946,29 @@ func NewRandomFedoraVMIWithDmidecode() *v1.VirtualMachineInstance {
...
 func GetGuestAgentUserData() string {
+	virtClient, err := kubecli.GetKubevirtClient()
+	PanicOnError(err)
+
+	dnsServerIP, err := getClusterDnsServiceIP(virtClient)
+	PanicOnError(err)
+
+	ipv6UserDataString := ""
+	if netutils.IsIPv6String(dnsServerIP) {
+		ipv6UserDataString = fmt.Sprintf(`sudo ip -6 addr add fd2e:f1fe:9490:a8ff::2/120 dev eth0
```
davidvossel, line 1957: why do we have to add the ip in cloud init? is the ip not automatically provided to the VMI via dhcp?

tests/probes_test.go
AlonaKaplan: done

tests/utils.go
AlonaKaplan: done - enabled the function to handle multiple prompt types

pkg/virt-launcher/virtwrap/network/common.go
AlonaKaplan: You're right! Done.

tests/utils.go
AlonaKaplan: We don't have dhcp server for ipv6. Vm ipv6 addresses should be configured manually (via cloud-init, console, etc.) Please refer to the PR description for more details.

examples/vmi-masquerade.yaml
AlonaKaplan: We don't have dhcp server for ipv6. Vm ipv6 addresses should be configured manually (via cloud-init, console, etc.) Please refer to the PR description for more details.

tests/probes_test.go
```diff
@@ -47,11 +47,15 @@ var _ = Describe("[ref_id:1182]Probes", func() {
...
+			}
 			By("Specifying a VMI with a readiness probe")
 			vmi := tests.NewRandomVMIWithEphemeralDiskAndUserdata(tests.ContainerDiskFor(tests.ContainerDiskCirros), "#!/bin/bash\necho 'hello'\n")
 			vmi.Spec.ReadinessProbe = readinessProbe
 			vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
 			Expect(err).ToNot(HaveOccurred())
+			tests.WaitUntilVMIReady(vmi, tests.LoggedInCirrosExpecter)
```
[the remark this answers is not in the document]
AlonaKaplan: Done, it indeed was added to configure the ipv6. But the tests are passing without it so removed it.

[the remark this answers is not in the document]
maiqueb: I'm puzzled as to how 0_o

I actually meant the `tests.WaitUntilVMIReady(vmi, tests.LoggedInCirrosExpecter)` clause to be kept, and the `tests.WaitForSuccessfulVMIStartIgnoreWarnings(vmi)` to be dropped (since we need to configure the ip address via console, afaiu).

I'll try to understand that a bit better, but so far, I'm almost sure configuring the ipv6 address via console **is needed** for the http probe test (the one that is currently skipped).

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -772,40 +839,59 @@ func (p *MasqueradePodInterface) createNatRulesUsingIptables() error {
...
+
+func (p *MasqueradePodInterface) getVifIpByProtocol(proto iptables.Protocol) string {
+	if proto == iptables.ProtocolIPv4 {
+		return p.vif.IP.IP.String()
+	} else {
+		return p.vif.IPv6.IP.String()
+	}
+}
+
+func getLoopbackAdrress(proto iptables.Protocol) string {
```
rmohr, line 858: `getLoopbackAddress`

PR #3113, opened 2020-02-27
Allow time for config changes to propagate before running tests

stu-gott:
Fixes a potential race condition in functional tests where kubevirt-config is changed and then immediately exercised.

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/security_features_test.go
+++ b/tests/security_features_test.go
@@ -22,6 +22,7 @@ package tests_test
 import (
 	"encoding/json"
 	"fmt"
+	"time"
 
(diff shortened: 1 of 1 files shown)


rmohr:
did we see issues with that in practice? The changes should be picked up very fast and I would try to avoid re-running tests. I would even prefer a short sleep.

stu-gott:
We actually did observe these tests becoming flakey in CI immediately in the wake of the previous change. If the group at large feels a sleep is better practice, I'll follow suit.

PR #3125, opened 2020-03-04
Ensure labels on our metrics are not being overwritten

danielBelenky:
Since OCP 4.3, the monitoring stack is featuring a Prometheus proxy
component (Thanos). One of the proxy components (prom-query-proxy) is modifying
the metrics on the fly: it changes our namespace label to exported_namespace.
In result, while virt-handler sets the namespace label to be the
namespace in which the VMI resides, it's being overwritten to the
namespace of virt-handler. Since virt-handler's namespace is privileged,
users without access to this namespace will not be able to view this
metric.

** Related BZ **:
https://bugzilla.redhat.com/show_bug.cgi?id=1805044

** Related PR on openshift-console **:
https://github.com/openshift/console/pull/4623

**Release note**:

```release-note
Ensure Prometheus operator honors labels on kubevirt exported metrics
```

### DIFF
--- a/pkg/virt-operator/creation/components/crds.go
+++ b/pkg/virt-operator/creation/components/crds.go
@@ -273,6 +273,12 @@ func NewServiceMonitorCR(namespace string, monitorNamespace string, insecureSkip
 					TLSConfig: &promv1.TLSConfig{
 						InsecureSkipVerify: insecureSkipVerify,
 					},
+					RelabelConfigs: []*promv1.RelabelConfig{
+						{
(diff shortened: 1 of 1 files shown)


danielBelenky:
/hold
Apparently, `honorLabels` is discouraged by the monitoring team because it introduces some security risks. The alternative is to use `relabelings` and ensure that Prometheus doesn't even attach the `namespace` label to our metrics. I'm checking this path now and will update the PR accordingly.

PR #3150, opened 2020-03-13
relabel the hostImages directory to container_file_t

vladikr:
This PR re-labels the `/tmp/hostImages` directory on nodes to  `container_file_t`
to support testing the virt-launcher on SELinux enabled environment.

The code uses `svirt_sandbox_file_t` instead of exclicitly using `container_file_t`.
`container_file_t` is aliased to `svirt_sandbox_file_t`, however, it is missing from some platforms such as RHEL 7.4

```release-note
NONE
```


jean-edouard:
/lgtm
This is great! Not only does this PR move container-disks away from tmp_t, but the disk image files also show up in virt-launcher with the correct categories!

PR #3162, opened 2020-03-17
Allow a manual adjustment of guest memory overhead

vladikr:
In some cases, manual adjustment of the calculated guest memory overhead is required.
However, until now this option was blocked.

This PR re-enables setting the guest's memory to a different value than the container requested memory. This will effectively allow the container to allocate more memory than the VMI and manually control the guest's memory overhead.

```
spec:
  domain:
    memory:
      guest: "2G"
    resources:
      requests:
        memory: "3G"
```

Fixes #

This PR also adjusts the memory overhead when huge pages are being requested.
Since until now, we didn't allow setting different values for `memory.guest` and `resources.requests.memory` we were requesting huge pages based on the `resources.requests.memory`. This PR requests huge pages based on guest memory (if present and different from requests.memory) and adds the difference to the containers `resources.requests.memory` in addition to the guest's memory overhead.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -583,8 +583,18 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 	// Consider hugepages resource for pod scheduling
 	if vmi.Spec.Domain.Memory != nil && vmi.Spec.Domain.Memory.Hugepages != nil {
 		hugepageType := k8sv1.ResourceName(k8sv1.ResourceHugePagesPrefix + vmi.Spec.Domain.Memory.Hugepages.PageSize)
-		resources.Requests[hugepageType] = resources.Requests[k8sv1.ResourceMemory]
-		resources.Limits[hugepageType] = resources.Requests[k8sv1.ResourceMemory]
+		hugepagesMemReq := vmi.Spec.Domain.Resources.Requests.Memory()
+
+		// If requested, use the guest memory to allocate hugepages
+		if vmi.Spec.Domain.Memory != nil && vmi.Spec.Domain.Memory.Guest != nil {
+			requests := vmi.Spec.Domain.Resources.Requests.Memory().Value()
+			guest := vmi.Spec.Domain.Memory.Guest.Value()
+			if requests > guest {
+				hugepagesMemReq = vmi.Spec.Domain.Memory.Guest
+			}
+		}
+		resources.Requests[hugepageType] = *hugepagesMemReq
+		resources.Limits[hugepageType] = *hugepagesMemReq
 
 		// Configure hugepages mount on a pod
 		volumeMounts = append(volumeMounts, k8sv1.VolumeMount{
@@ -600,10 +610,29 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 			},
 		})
 
+		reqMemDiff := resource.NewScaledQuantity(0, resource.Kilo)
+		limMemDiff := resource.NewScaledQuantity(0, resource.Kilo)
+		// In case the guest memory and the requested memeory are diffrent, add the difference
+		// to the to the overhead
+		if vmi.Spec.Domain.Memory != nil && vmi.Spec.Domain.Memory.Guest != nil {
+			requests := vmi.Spec.Domain.Resources.Requests.Memory().Value()
+			limits := vmi.Spec.Domain.Resources.Limits.Memory().Value()
+			guest := vmi.Spec.Domain.Memory.Guest.Value()
+			if requests > guest {
+				reqMemDiff.Add(*vmi.Spec.Domain.Resources.Requests.Memory())
+				reqMemDiff.Sub(*vmi.Spec.Domain.Memory.Guest)
+			}
+			if limits > guest {
+				limMemDiff.Add(*vmi.Spec.Domain.Resources.Limits.Memory())
+				limMemDiff.Sub(*vmi.Spec.Domain.Memory.Guest)
+			}
+		}
 		// Set requested memory equals to overhead memory
-		resources.Requests[k8sv1.ResourceMemory] = *memoryOverhead
+		reqMemDiff.Add(*memoryOverhead)
+		resources.Requests[k8sv1.ResourceMemory] = *reqMemDiff
 		if _, ok := resources.Limits[k8sv1.ResourceMemory]; ok {
-			resources.Limits[k8sv1.ResourceMemory] = *memoryOverhead
+			limMemDiff.Add(*memoryOverhead)
+			resources.Limits[k8sv1.ResourceMemory] = *limMemDiff
 		}
 	} else {
 		// Add overhead memory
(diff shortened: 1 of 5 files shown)


vladikr:
> We have some functional tests which verify that pods have `guaranteed` QoS if Requests and Limit are the same on the VMI, can we add one for a situation where the guest memory is less than the requested memory, to be sure?

@rmohr added, thanks

danielBelenky:
>  In some cases, manual adjustment of the calculated guest memory overhead is required.
Could you please elaborate on the case please? (asking just to expand my knowledge...)

vladikr:
> Could you please elaborate on the case please? (asking just to expand my knowledge...)

sure @danielBelenky ,  Our memory overhead calculation[1] is really a our best guess. It's hard to account for every possible memory increase by QEMU or other components that are running in the container.

One know issue is that we don't currently account for vfio related overhead. However, even if we will introduce a new factor to account for we will still need a way to manually adjust the overhead.

[1] https://github.com/kubevirt/kubevirt/blob/a53c612b3b4ad287c7c88f942973bf15dfbbd6f6/pkg/virt-controller/services/template.go#L1069

pkg/virt-controller/services/template.go
```diff
@@ -600,10 +610,29 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
+				reqMemDiff.Sub(*vmi.Spec.Domain.Memory.Guest)
+			}
+			if limits > guest {
+				limMemDiff.Add(*vmi.Spec.Domain.Resources.Limits.Memory())
+				limMemDiff.Sub(*vmi.Spec.Domain.Memory.Guest)
+			}
+		}
 		// Set requested memory equals to overhead memory
-		resources.Requests[k8sv1.ResourceMemory] = *memoryOverhead
+		reqMemDiff.Add(*memoryOverhead)
```
phoracek: Then it is not a diff anymore. Could you use a better name for the variable to make it clear that it is total memory consisting of the request of VM+overhead?

pkg/virt-controller/services/template.go
```diff
@@ -600,10 +610,29 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 			},
 		})
 
+		reqMemDiff := resource.NewScaledQuantity(0, resource.Kilo)
```
phoracek: What does this cover. What does this diff between?

pkg/virt-controller/services/template.go
```diff
@@ -600,10 +610,29 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
+		limMemDiff := resource.NewScaledQuantity(0, resource.Kilo)
+		// In case the guest memory and the requested memeory are diffrent, add the difference
```
phoracek: typo: memory

pkg/virt-controller/services/template.go
vladikr: Well, that's just a variable initialization.
you should look at lines 621-624 to see that the difference is between the guest memory and the requested memory for the compute container

pkg/virt-controller/services/template.go
vladikr: @phoracek - That's not a guest memory + overhead.

In general, when we start a VMI we need to request more memory for the compute container than the guest needs.
We refer to the difference between the container memory and the guest memory as a memory overhead[1].
The overhead calculation includes the memory needed for qemu threads, graphic card memory, os/code running in the compute container and more... see [1]

When it comes to huge pages, we need to request k8s to allocate some of the memory with specific page size (`resources.Requests[hugepageType]`) and the rest of the memory (overhead + difference between guest memory and the memory requested for the container) with a default page size (`resources.Requests[k8sv1.ResourceMemory]`)
The memory allocated for huge pages will be provided to the guest for it to use as it's memory.

In this PR, we allow the guest memory `vmi.Spec.Domain.Memory.Guest` to be different from `vmi.Spec.Domain.Resources.Requests` therefore, we need to allocate huge pages according to the guest's memory (`vmi.Spec.Domain.Memory.Guest`) and the difference (requested mem - guest memory + overhead) request to allocate as a memory with a default page size (`resources.Requests[k8sv1.ResourceMemory]`).  

[1] https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-controller/services/template.go#L1069

pkg/virt-controller/services/template.go
```diff
@@ -600,10 +610,29 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
+		// to the to the overhead
+		if vmi.Spec.Domain.Memory != nil && vmi.Spec.Domain.Memory.Guest != nil {
+			requests := vmi.Spec.Domain.Resources.Requests.Memory().Value()
+			limits := vmi.Spec.Domain.Resources.Limits.Memory().Value()
+			guest := vmi.Spec.Domain.Memory.Guest.Value()
+			if requests > guest {
+				reqMemDiff.Add(*vmi.Spec.Domain.Resources.Requests.Memory())
+				reqMemDiff.Sub(*vmi.Spec.Domain.Memory.Guest)
+			}
+			if limits > guest {
```
rmohr: Would here be a more appropriate check, if `limits` is set at all?

pkg/virt-controller/services/template.go
phoracek: Thanks.

pkg/virt-controller/services/template.go
phoracek: Later in the code, the value of the variable does not match its name. Could you initialize a new variable which name would reflect that it has the total memory, not only the diff?

pkg/virt-controller/services/template.go
phoracek: @vladikr

pkg/virt-controller/services/template.go
phoracek: I get it now.

Maybe it is just my ignorance and lack of knowledge about kubevirt, but the fact that there is no design document for this (is there) anyone new to kubevirt can read on is frightening. It makes is really hard for somebody new to start contributing to kubevirt even with reviews.

pkg/virt-controller/services/template.go
vladikr: I can check, but `.Memory().Value()` returns `0` if it's not set. I don't think it's necessary.

pkg/virt-controller/services/template.go
```diff
@@ -600,10 +610,29 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
+				limMemDiff.Add(*vmi.Spec.Domain.Resources.Limits.Memory())
+				limMemDiff.Sub(*vmi.Spec.Domain.Memory.Guest)
+			}
+		}
 		// Set requested memory equals to overhead memory
-		resources.Requests[k8sv1.ResourceMemory] = *memoryOverhead
+		reqMemDiff.Add(*memoryOverhead)
+		resources.Requests[k8sv1.ResourceMemory] = *reqMemDiff
```
phoracek: This is what I don't understand, assigning diff to the request of memory. But again, maybe I don't fully understand what `resources.Requests[k8sv1.ResourceMemory]` stands for. I would expect it gets `max(vmMemory, vmReqMemory) + overhead`.

stu-gott:
@vladikr looks like we need to run "make generate"?

vladikr:
> @vladikr looks like we need to run "make generate"?

@stu-gott I don't think so, it's something with the ppc lane :(

phoracek:
I saw it failing on other PRs. e.g. https://github.com/kubevirt/kubevirt/pull/3173

rmohr:
> I saw it failing on other PRs. e.g. #3173

I hope that #3184 fixes that.

PR #3165, opened 2020-03-17
Move fetching new tests into function, correct providers

dhiller:

Current script does not work as it returns an exit code of 1 in case of "no tests changed". This modification disables option `e` for the function in order to return 0 in case of no tests changed.

Furthermore it corrects the default providers that changed until recently.

Fixes #

/cc @danielBelenky 

**Release note**:

```release-note
NONE
```


dhiller:
If this works, the flake test job should be successful: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3165/pull-kubevirt-check-tests-for-flakes/1239903439069122560

PR #3166, opened 2020-03-17
Add guestagent virtctl commands

petrkotas:
Easies the access for guestagent installed in the VM.

Add
 - guestosinfo
 - userlist
 - fslist
as endpoints for gathering data from guestagent in VM, when present.

**Release note**:

```release-note
Introduce new virtctl commands:
- guestosinfo
- userlist
- fslist
```


pkg/virtctl/vm/vm.go
```diff
@@ -193,6 +242,45 @@ func (o *Command) Run(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return fmt.Errorf("Error renaming VirtualMachine %v", err)
 		}
+	case COMMAND_GUESTOSINFO:
+		guestosinfo, err := virtClient.VirtualMachineInstance(namespace).GuestOsInfo(vmiName)
+		if err != nil {
+			return fmt.Errorf("Error renaming VirtualMachine %v", err)
```
EdDev, line 248: The message is wrong here.
This seems to be repeated in the rest of the changes below.

pkg/virtctl/vm/vm.go
```diff
@@ -193,6 +242,45 @@ func (o *Command) Run(cmd *cobra.Command, args []string) error {
...
+		}
+
+		data, err := json.MarshalIndent(guestosinfo, "", "  ")
+		if err != nil {
+			return fmt.Errorf("Cannot unmarshal guestosinfo %v", err)
```
EdDev, line 253: The above "marshal" the data. Why did you used "unmarshal"?
Also, if the error is reported, it will be unclear what "marshal" is it all about (unless the error gives enough info, then this message is not really useful). If the err does not include the value of `guestosinfo`, it may be useful to add it.

Also note that it is not common to use capital letter in such messages because it is usually appended to other text.
(true for all other messages)

stu-gott:
/hold cancel

No need to request changes and hold at the same time. The former removes the LGTM label. @EdDev just so it's clear, @xpivarc is not the original author of this PR, so might not have insight into some of your questions/concerns--but your points are valid. Would you be able to add a patch to this PR?

PR #3174, opened 2020-03-18
Avoid timeout with travis on ppc64 on "make generate"

rmohr:

Ensure that we also print periodically a keep alive message already when
we do "make generate".

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,9 @@ install:
 - git reset --hard
 
 script:
+- |
+  while sleep 9m; do echo "Long running job - keep alive"; done &
+  LOOP_PID=$!
 - make generate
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -25,9 +28,6 @@ script:
   `make`. Please run it and commit the changes"; git status --porcelain; false; fi
 - make build-verify # verify that we set version on the packages built by bazel
 # The make bazel-test might take longer then the current timeout for a command in Travis-CI of 10 min, so adding a keep alive loop while it runs
```
dhiller: Maybe you should move the comment and change it to something like `several commands might take longer than ten minutes...`?

rmohr:
/close

this does not work as expected.

PR #3176, opened 2020-03-18
Copy 'Ready' condition from VMI to VM

### LINKED ISSUE #2864: Provide Ready condition for VM
/kind enhancement

**What happened**:
$ virtctl start testvm
$ kubectl wait --for=condition=Ready vm testvm
No such condition

The condition is just not available today, only a phsae.

$ virtctl start testvm
$ kubectl wait --for=condition=Ready vm testvm
# vm running
$ virtctl vnc testvm
# vnc can connect

Just run the lines above

**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

akrejcir:
When 'Ready' condition is added the VMI resource,
the virt-controller copies it to the VM resource.

Signed-off-by: Andrej Krejcir <akrejcir@redhat.com>

Users will be able to wait for the VM start using `kubectl wait --for=condition=Ready vm ${NAME}`

Fixes #2864

**Release note**:
```release-note
Added a new 'Ready' condition to the VirtualMachine resource.
The condition is copied from the associated VMI resource.

This makes it possible to wait while the VM starts on the VM resource:
$ kubectl wait --for=condition=Ready vm ${NAME}
```

### DIFF
--- a/pkg/controller/conditions.go
+++ b/pkg/controller/conditions.go
@@ -28,7 +28,41 @@ import (
 type VirtualMachineConditionManager struct {
 }
 
-func (d *VirtualMachineConditionManager) CheckFailure(vmi *v1.VirtualMachineInstance, syncErr error, reason string) (changed bool) {
+func NewVirtualMachineConditionManager() *VirtualMachineConditionManager {
+	return &VirtualMachineConditionManager{}
+}
+
+func (d *VirtualMachineConditionManager) GetCondition(vm *v1.VirtualMachine, cond v1.VirtualMachineConditionType) *v1.VirtualMachineCondition {
+	if vm == nil {
+		return nil
+	}
+	for _, c := range vm.Status.Conditions {
+		if c.Type == cond {
+			return &c
+		}
+	}
+	return nil
+}
+
+func (d *VirtualMachineConditionManager) HasCondition(vm *v1.VirtualMachine, cond v1.VirtualMachineConditionType) bool {
+	return d.GetCondition(vm, cond) != nil
+}
+
+func (d *VirtualMachineConditionManager) RemoveCondition(vm *v1.VirtualMachine, cond v1.VirtualMachineConditionType) {
+	var conds []v1.VirtualMachineCondition
+	for _, c := range vm.Status.Conditions {
+		if c.Type == cond {
+			continue
+		}
+		conds = append(conds, c)
+	}
+	vm.Status.Conditions = conds
+}
+
+type VirtualMachineInstanceConditionManager struct {
+}
+
+func (d *VirtualMachineInstanceConditionManager) CheckFailure(vmi *v1.VirtualMachineInstance, syncErr error, reason string) (changed bool) {
 	if syncErr != nil && !d.HasCondition(vmi, v1.VirtualMachineInstanceSynchronized) {
 		vmi.Status.Conditions = append(vmi.Status.Conditions, v1.VirtualMachineInstanceCondition{
 			Type:               v1.VirtualMachineInstanceSynchronized,
@@ -45,31 +79,28 @@ func (d *VirtualMachineConditionManager) CheckFailure(vmi *v1.VirtualMachineInst
 	return false
 }
 
-func (d *VirtualMachineConditionManager) HasCondition(vmi *v1.VirtualMachineInstance, cond v1.VirtualMachineInstanceConditionType) bool {
+func (d *VirtualMachineInstanceConditionManager) GetCondition(vmi *v1.VirtualMachineInstance, cond v1.VirtualMachineInstanceConditionType) *v1.VirtualMachineInstanceCondition {
 	if vmi == nil {
-		return false
+		return nil
 	}
 	for _, c := range vmi.Status.Conditions {
 		if c.Type == cond {
-			return true
+			return &c
 		}
 	}
-	return false
+	return nil
 }
 
-func (d *VirtualMachineConditionManager) HasConditionWithStatus(vmi *v1.VirtualMachineInstance, cond v1.VirtualMachineInstanceConditionType, status k8sv1.ConditionStatus) bool {
-	for _, c := range vmi.Status.Conditions {
-		if c.Type == cond {
-			if c.Status == status {
-				return true
-			}
-			return false
-		}
-	}
-	return false
+func (d *VirtualMachineInstanceConditionManager) HasCondition(vmi *v1.VirtualMachineInstance, cond v1.VirtualMachineInstanceConditionType) bool {
+	return d.GetCondition(vmi, cond) != nil
 }
 
-func (d *VirtualMachineConditionManager) RemoveCondition(vmi *v1.VirtualMachineInstance, cond v1.VirtualMachineInstanceConditionType) {
+func (d *VirtualMachineInstanceConditionManager) HasConditionWithStatus(vmi *v1.VirtualMachineInstance, cond v1.VirtualMachineInstanceConditionType, status k8sv1.ConditionStatus) bool {
+	c := d.GetCondition(vmi, cond)
+	return c != nil && c.Status == status
+}
+
+func (d *VirtualMachineInstanceConditionManager) RemoveCondition(vmi *v1.VirtualMachineInstance, cond v1.VirtualMachineInstanceConditionType) {
 	var conds []v1.VirtualMachineInstanceCondition
 	for _, c := range vmi.Status.Conditions {
 		if c.Type == cond {
@@ -81,7 +112,7 @@ func (d *VirtualMachineConditionManager) RemoveCondition(vmi *v1.VirtualMachineI
 }
 
 // AddPodCondition add pod condition to the VM.
-func (d *VirtualMachineConditionManager) AddPodCondition(vmi *v1.VirtualMachineInstance, cond *k8sv1.PodCondition) {
+func (d *VirtualMachineInstanceConditionManager) AddPodCondition(vmi *v1.VirtualMachineInstance, cond *k8sv1.PodCondition) {
 	if !d.HasCondition(vmi, v1.VirtualMachineInstanceConditionType(cond.Type)) {
 		vmi.Status.Conditions = append(vmi.Status.Conditions, v1.VirtualMachineInstanceCondition{
 			LastProbeTime:      cond.LastProbeTime,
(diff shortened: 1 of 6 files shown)


pkg/virt-controller/watch/vm.go
```diff
@@ -1285,6 +1294,38 @@ func (c *VMController) updateStatus(vmOrig *virtv1.VirtualMachine, vmi *virtv1.V
 	return err
 }
 
+func (c *VMController) copyReadyConditionFromVMI(vm *virtv1.VirtualMachine, vmi *virtv1.VirtualMachineInstance) {
```
dhiller: nit: rather call it `syncReadyConditionFromVMI`?

akrejcir: Done

pkg/virt-controller/watch/vm.go
```diff
@@ -1285,6 +1294,38 @@ func (c *VMController) updateStatus(vmOrig *virtv1.VirtualMachine, vmi *virtv1.V
...
+			vm.Status.Conditions = append(vm.Status.Conditions, newCond)
+		} else if vmiReadyCond.Status != vmReadyCond.Status {
+			log.Log.Object(vm).V(4).Info("Updating ready condition")
+			vmReadyCond.Status = vmiReadyCond.Status
+			vmReadyCond.LastTransitionTime = now
+			vmReadyCond.Reason = vmiReadyCond.Reason
+			vmReadyCond.Message = vmiReadyCond.Message
+
+		}
+	} else if vmReadyCond != nil {
```
danielBelenky: This edge case is redundant. I think that it will also improve the readability of this function:

```
if vmiReadyCond == nil {
    c.removeCondition(vm, virtv1.VirtualMachineReady)
    return
}

// now just copy the condition
```

pkg/virt-controller/watch/vm.go
```diff
@@ -1285,6 +1294,38 @@ func (c *VMController) updateStatus(vmOrig *virtv1.VirtualMachine, vmi *virtv1.V
...
+				Type:               virtv1.VirtualMachineReady,
+				Status:             vmiReadyCond.Status,
+				LastProbeTime:      now,
+				LastTransitionTime: now,
+				Reason:             vmiReadyCond.Reason,
+				Message:            vmiReadyCond.Message,
+			}
+			vm.Status.Conditions = append(vm.Status.Conditions, newCond)
+		} else if vmiReadyCond.Status != vmReadyCond.Status {
+			log.Log.Object(vm).V(4).Info("Updating ready condition")
```
danielBelenky: Why not always update the VM ready condition? regardless of it's value...
If we don't update it, the timestamps (at least the `LastProbeTime` will become stale and it usually indicates of an issue to the user).

pkg/virt-controller/watch/vm_test.go
```diff
@@ -818,6 +818,40 @@ var _ = Describe("VirtualMachine", func() {
...
+				objVM := obj.(*v1.VirtualMachine)
+				cond := getVmCondition(objVM, v1.VirtualMachineReady)
+				Expect(cond).ToNot(BeNil())
+				Expect(cond.Status).To(Equal(k8sv1.ConditionTrue))
+			}).Return(vm, nil)
+
+			controller.Execute()
+		})
+
+		It("should add ready condition False", func() {
```
danielBelenky: Can those tests become a table? since the flow is similar and they differ only in the value of the condition...

pkg/virt-controller/watch/vm_test.go
```diff
@@ -1011,3 +1045,15 @@ func DefaultVirtualMachineWithNames(started bool, vmName string, vmiName string)
 func DefaultVirtualMachine(started bool) (*v1.VirtualMachine, *v1.VirtualMachineInstance) {
 	return DefaultVirtualMachineWithNames(started, "testvmi", "testvmi")
 }
+
+func getVmCondition(vm *virtv1.VirtualMachine, condType virtv1.VirtualMachineConditionType) *virtv1.VirtualMachineCondition {
```
danielBelenky: why is it needed? can't we use `getCondition`? https://github.com/kubevirt/kubevirt/pull/3176/files#diff-3d28bf4913d346c1909faf1a6411f176R48

tests/vm_test.go
```diff
@@ -634,6 +634,31 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
 			}
 		})
 
+		It("should set the Ready condition on VM", func() {
+			vm := newVirtualMachine(false)
+
+			getVmReadyConditionStatus := func() *k8sv1.ConditionStatus {
```
danielBelenky: Same here... why can't we use `getCondition` and then check the Status?

pkg/virt-controller/watch/vm.go
akrejcir: I wanted to minimize update requests, but this makes sense. Should the `LastProbeTime` be copied from VMI or set to current time?

danielBelenky: I think that copying `LastProbeTime` makes sense...

pkg/virt-controller/watch/vm.go
akrejcir: That would call the `c.removeCondition` function on every update even when it is not needed. I would prefer to leave the condition there, but I have made it more readable.

pkg/virt-controller/watch/vm.go
akrejcir: Done

pkg/virt-controller/watch/vm_test.go
akrejcir: Done

pkg/virt-controller/watch/vm_test.go
akrejcir: Moved the function to a common place `pkg/controller/conditions.go` and it is called when needed.

tests/vm_test.go
akrejcir: Done

pkg/controller/conditions.go
```diff
@@ -25,10 +25,44 @@ import (
 	v1 "kubevirt.io/client-go/api/v1"
 )
 
-type VirtualMachineConditionManager struct {
+type virtualMachineConditionManager struct {
 }
 
-func (d *VirtualMachineConditionManager) CheckFailure(vmi *v1.VirtualMachineInstance, syncErr error, reason string) (changed bool) {
```
danielBelenky: Why you changed this? it used to be exported and now it is not...

pkg/controller/conditions.go
```diff
@@ -45,31 +79,28 @@ func (d *VirtualMachineConditionManager) CheckFailure(vmi *v1.VirtualMachineInst
 	return false
 }
 
-func (d *VirtualMachineConditionManager) HasCondition(vmi *v1.VirtualMachineInstance, cond v1.VirtualMachineInstanceConditionType) bool {
+func (d virtualMachineInstanceConditionManager) GetCondition(vmi *v1.VirtualMachineInstance, cond v1.VirtualMachineInstanceConditionType) *v1.VirtualMachineInstanceCondition {
```
danielBelenky: same here...

pkg/controller/conditions.go
```diff
@@ -129,14 +160,14 @@ func (d *VirtualMachineConditionManager) GetPodCondition(pod *k8sv1.Pod, conditi
 	return nil
 }
 
-func NewVirtualMachineInstanceConditionManager() *VirtualMachineConditionManager {
-	return &VirtualMachineConditionManager{}
+func VmiConditionManager() virtualMachineInstanceConditionManager {
```
danielBelenky: Is this change required for this PR?

pkg/virt-controller/watch/vm.go
danielBelenky: I don't see a real harm in calling `c.removeCondition` every time (it will do nothing if the condition is nil)...
But it's your code so I don't mind that much.
Thanks for improving the readability :+1:

pkg/controller/conditions.go
akrejcir: No, not required. I thought the names were too long. It is in a separate commit, should it be a separate PR too?

pkg/controller/conditions.go
akrejcir: Does the struct need to be exported? It has no use other than calling it's method, and it can be accessed through the exported method.

pkg/controller/conditions.go
danielBelenky: I prefer it to go to a separate PR if possible since its unrelated.

pkg/controller/conditions.go
danielBelenky: If this struct is not used anywhere else, we can un-export it. But, since this is not related to this PR, maybe we can do it in a separate PR where we can discuss if it's required or not?

(I'm sorry for being picky on this...)

pkg/controller/conditions.go
akrejcir: Ok, the other commit is removed.

akrejcir:
`/retest`

danielBelenky:
Hey @akrejcir, thanks for addressing the comments.
I see that there are some test failures so once those are resolved, we can get this PR merged.

akrejcir:
The travis CI is failing, but the failure does not seem related to this patch. Is there something I can do to fix it?

rmohr:
I see that your PR is still based on a branch were  we run the PPC build which we have now temporarily disabled. Could you just rebase?

pkg/controller/conditions.go
```diff
@@ -28,7 +28,41 @@ import (
...
+	for _, c := range vm.Status.Conditions {
+		if c.Type == cond {
+			continue
+		}
+		conds = append(conds, c)
+	}
+	vm.Status.Conditions = conds
+}
+
+type VirtualMachineInstanceConditionManager struct {
```
xpivarc, line 62: Is there a need to have this as struct? Could be this only simple functions?

akrejcir: I used the same design as the other structs in this file use, to be consistent. I agree that the structs are useless, we could remove them and rename all the functions in a new PR. Function overloading would be ideal for this use case, but Go does not support it.

akrejcir:
The tests have passed, can this be merged?

PR #3178, opened 2020-03-19
Increase memory overhead for VFIO devices

vladikr:
VMIs with assigned VFIO devices require an additional memory overhead of 1G to be added to the virt-launcher's compute container.

An assigned device potentially pins the entire guest's RAM and MMIO memory regions.  This is necessary for any IOMMU backed device where any guest RAM is potentially a DMA target.

Libvirt also calculates the locked memory with the 1GB overhead[1]. Based on this virt-handler also sets the memory lock limit to a similar value[2], for VFIO devices, however, it doesn't adjust the compute container resources to accommodate this overhead.

Since we don't take enough VFIO related memory overhead into account VMIs with SRIOV will get OOM. This is much often visible when the VMI has dedicated CPU since in this case, the vmipod will have a Guaranteed QOS (hard limits). PODs with Burstable QoS will keep running until the node has resources.

This PR adjusts the memory overhead to 1G and adds a functional test to make sure VMIs with SRIOV can start properly.

[1] https://github.com/libvirt/libvirt/blob/master/src/qemu/qemu_domain.c#L13437
[2] https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-handler/isolation/isolation.go#L199

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/virt-controller/services/template_test.go
+++ b/pkg/virt-controller/services/template_test.go
@@ -1312,7 +1312,10 @@ var _ = Describe("Template", func() {
 
 				pod, err := svc.RenderLaunchManifest(&vmi)
 				Expect(err).ToNot(HaveOccurred())
-				Expect(pod.Spec.Containers[0].Resources.Requests.Memory().String()).To(Equal("2163507557"))
+				expectedMemory := resource.NewScaledQuantity(0, resource.Kilo)
+				expectedMemory.Add(*getMemoryOverhead(&vmi))
+				expectedMemory.Add(*domain.Resources.Requests.Memory())
+				Expect(pod.Spec.Containers[0].Resources.Requests.Memory().Value()).To(Equal(expectedMemory.Value()))
 			})
 		})
 		Context("with slirp interface", func() {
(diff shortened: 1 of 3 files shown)


pkg/virt-controller/services/template_test.go
```diff
@@ -1293,6 +1293,27 @@ var _ = Describe("Template", func() {
...
+				vmi := v1.VirtualMachineInstance{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "testvmi", Namespace: "default", UID: "1234",
+					},
+					Spec: v1.VirtualMachineInstanceSpec{Domain: domain},
+				}
+
+				pod, err := svc.RenderLaunchManifest(&vmi)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(pod.Spec.Containers[0].Resources.Requests.Memory().String()).To(Equal("2163507557"))
```
phoracek: Would it be possible to get something human readable here? So it really is clear that it added the memory we expected.

vladikr: Done

phoracek:
I verified the new SR-IOV test in https://github.com/kubevirt/kubevirt/pull/3189. It passes!

PR #3182, opened 2020-03-20
fix mount in case docker rootfs is not /

xuxinkun:
If docker make use of other graph driver, for example, devicemapper, the rootfs will be /rootfs, not /.
```
cat /proc/{container-disk-container pid}/mountinfo
1889 1747 253:24 /rootfs / rw,relatime master:240 - ext4 /dev/mapper/docker-8:3-4456449-8ff323dff89027a4cafa614fccf06cea26f7519f07b33f1f44eb4dd62bb49b62 rw,stripe=16,data=ordered
```
This PR try to solve this case.


rmohr:
@xuxinkun thanks, looks like a duplicate for #3124

xuxinkun:
@rmohr  Yes. I will close this one.

PR #3191, opened 2020-03-21
Test sriov ipv6 ping

phoracek:

We should validate that connectivity over SR-IOV works with both IPv4 and IPv6.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/vmi_multus_test.go
+++ b/tests/vmi_multus_test.go
@@ -22,6 +22,7 @@ package tests_test
 import (
 	"flag"
 	"fmt"
+	"net"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


tests/vmi_multus_test.go
```diff
@@ -848,16 +853,33 @@ var _ = Describe("SRIOV", func() {
...
+		It("should connect to another machine with sriov interface over IPv6", func() {
+			pingThroughSriov("fc00::1/64", "fc00::2/64")
 		})
 	})
 })
 
+func cidrToIP(cidr string) string {
+	ip, _, err := net.ParseCIDR(cidr)
+	if err != nil {
+		panic(err)
```
vladikr: Instead of panic, better to use: Expect(err).ToNot(HaveOccurred(), "some error")

phoracek: Done

PR #3197, opened 2020-03-24
Allow generation of manifests with configurable Kubevirt namespace.

enp0s3:
In the current situation the `make manifests` action will create manifests with constant "kubevirt" namespace. This suggested change will allow us to control the namespace value via the `KUBEVIRT_INSTALLED_NAMESPACE` environment variable before running `make manifests`
 
In some situations Kubevirt is being deployed to a namespace other then the traditional "kubevirt".
In these situations we still want to use the generated manifests but we need to control the namespace value. 

Fixes #
Test automation

```release-note
Allow generation of manifests with configurable Kubevirt namespace.
```


hack/config-default.sh
```diff
@@ -4,7 +4,7 @@ docker_tag=${DOCKER_TAG:-latest}
 docker_tag_alt=${DOCKER_TAG_ALT}
 image_prefix=${IMAGE_PREFIX}
 image_prefix_alt=${IMAGE_PREFIX_ALT}
-namespace=kubevirt
+namespace=${KUBEVIRT_INSTALLED_NAMESPACE:-kubevirt}
 deploy_testing_infra=${DEPLOY_TESTING_INFRA:-false}
 csv_namespace=placeholder
 cdi_namespace=cdi
```
danielBelenky: should we also customize `CDI_NAMESPACE`?

enp0s3: @danielBelenky Currently there is no need to modify CDI installation namespace here because the consume of CDI generated manifests is done from the CDI project. Probably should be added there.

PR #3199, opened 2020-03-25
Adjust testing code to take proper URL of the files located at the HTTP import server.

enp0s3:
Introduce a new test utility function called `GetUrl` which will be used in order to take proper URL of testing files located at the HTTP import server. That is because the HTTP server can be deployed to a configurable namespace and it's FQDN will be affected by that.

Functional tests are using testing infrastructure pods. These pods can be deployed to a namespace other then the traditional "kubevirt" namespace. Therefore The FQDN of the HTTP server cannot be determined in compile time using constants. The suggested PR introduces the changes required to invoke the correct URL according to run-time parameters.

Fixes #
Test automation

**Release note**:

```release-note
Improvement of testing infrastructure code. Calculate FQDN of HTTP import server in run-time.
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -155,7 +155,6 @@ const (
 
 const (
 	AlpineHttpUrl = iota
-	FedoraHttpUrl
 	GuestAgentHttpUrl
 	StressHttpUrl
 	DmidecodeHttpUrl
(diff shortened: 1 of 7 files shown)


tests/utils.go
```diff
@@ -4197,3 +4196,21 @@ func GenerateRandomMac() (net.HardwareAddr, error) {
 	}
 	return net.HardwareAddr(append(prefix, suffix...)), nil
 }
+func GetUrl(urlIndex int) string {
+	var str string
+
+	switch urlIndex {
+	case AlpineHttpUrl:
+		str = fmt.Sprintf("http://cdi-http-import-server.%s/images/alpine.iso", KubeVirtInstallNamespace)
```
danielBelenky: Maybe instead of having a hard-coded URL we can get the name of the endpoint from the environment (and fallback to some default maybe) and extract the IP of the endpoint once at the beginning of the suite? This way we are not making any assumptions on the form URL itself...

PR #3201, opened 2020-03-25
[release-0.26] Increase memory overhead for VFIO devices

vladikr:
This is a manual cherry-pick of #3178

/assign stu-gott

```release-note
NONE
```

### DIFF
--- a/pkg/virt-controller/services/template.go
+++ b/pkg/virt-controller/services/template.go
@@ -561,7 +561,7 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 	gracePeriodKillAfter := gracePeriodSeconds + int64(15)
 
 	// Get memory overhead
-	memoryOverhead := getMemoryOverhead(vmi.Spec.Domain)
+	memoryOverhead := getMemoryOverhead(vmi)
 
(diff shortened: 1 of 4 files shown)


phoracek:
The SR-IOV e2e test won't pass. CPU manager is not enabled on this SR-IOV provider lane (we fixed that only recently on master).

rmohr:
Will this PR break the sriov lane now also for other PRs?

rmohr:
/hold

If that is the case, at least skip failing tests instead of breaking the lane.

rmohr:
Do we use dedicated CPUs because only that triggers the issue, or do we use dedicated CPUs to get a VM with QoS Guaranteed? If the second one I would suggest to change the test to simply use a VM whis has the Guaranteed class assigned.

vladikr:
> Do we use dedicated CPUs because only that triggers the issue, or do we use dedicated CPUs to get a VM with QoS Guaranteed? If the second one I would suggest to change the test to simply use a VM whis has the Guaranteed class assigned.

Well, just because we need a test that verifies this scenario. But the root cause of the issue is QoS Guaranteed. 
I've updated the PR.

vladikr:
> Will this PR break the sriov lane now also for other PRs?

Good point, thanks!

PR #3205, opened 2020-03-25
Open and close /dev/net/tun on virt-handler startup

jean-edouard:
Trying to access /dev/net/tun on a host which doesn't have the tun module loaded triggers a modprobe.
libvirtd in virt-launcher uses /dev/net/tun, which will trigger a modprobe if the module is not already loaded.
That won't work if/when virt-launcher runs under the SELinux type container_t.

In virt-handler, which runs as spc_t, access (open/close) /dev/net/tun and (potentially) trigger the modprobe.
That way the tun module will be loaded for sure when (libvirtd in) virt-launcher tries to access /dev/net/tun.

Fixes #

Hard to review with virt-launcher not running as container_t.
I will PR that next.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/virt-handler/virt-handler.go
+++ b/cmd/virt-handler/virt-handler.go
@@ -341,6 +341,13 @@ func (app *virtHandlerApp) Run() {
 		panic(fmt.Errorf("failed to detect the presence of selinux: %v", err))
 	}
 
+	// Make sure the tun module is loaded on the node
+	// Just opening and closing /dev/net/tun triggers a modprobe on the host if necessary
+	devnettun, err := os.Open("/dev/net/tun")
+	if err == nil {
+		devnettun.Close()
+	}
+
 	cache.WaitForCacheSync(stop, factory.ConfigMap().HasSynced, vmiInformer.HasSynced, factory.CRD().HasSynced)
 
 	go vmController.Run(10, stop)


pkg/virt-handler/vm.go
```diff
@@ -598,6 +598,13 @@ func (c *VirtualMachineController) Run(threadiness int, stopCh chan struct{}) {
 	defer c.Queue.ShutDown()
 	log.Log.Info("Starting virt-handler controller.")
 
+	// Make sure the tun module is loaded on the node
```
enp0s3: @jean-edouard I think a better place for this in-direct probing of TUN kernel module will be in the file cmd/virt-handler.go because looks like most of the initialization is done there. Another thing is that it may require a functional test where you unload the TUN module on a specific node and then re-create virt-handler pod and see if the TUN kernel module appears again.

vladikr:
> I think that this change worth a release note.

My take on this is this change should be transparent to users.
We have a device plugin in virt-handler that counts the available /dev/net/tun devices. A loaded module is required for this.

pkg/virt-handler/vm.go
vladikr: Regarding a functional test. We've discussed this in the past. On a system without selinux, libvirt (that's running in the virt-launcher pod) will do the work. It probes the module by default.
However, on a system with enabled selinux, virt-launcher will run as `conatiner_t` and will not be allowed to probe the kernel module, that's why this change was posted.

I think the best test is simply a running VMI on selinux enabled lane.

stu-gott: I'm uncomfortable with the notion of unloading kernel modules in automated tests. Can we guarantee this won't disrupt the system itself? other tests caught in the crossfire?

jean-edouard: @enp0s3 good points, I'll look at cmd/virt-handler.go. I didn't know what kind of test to write but that's a good idea.
@vladikr +1
@stu-gott is this a question about the test suggested by @enp0s3? This PR only triggers loads, no unloads.

jean-edouard:
Force-pushed to move the code to cmd/virt-handler/virt-handler.go.
`tun` will now get loaded on virt-handler startup rather than on VM creation.

pkg/virt-handler/vm.go
stu-gott: @jean-edouard yes that was my concern. I don't want to increase the scope of this PR to cover that.

PR #3210, opened 2020-03-26
Small fix of rename condition on VM

akrejcir:
The condition status should be `ConditionTrue`, instead of "Success"
The type should probably not contain space character.

**Release note**:

```release-note
The VM rename condition has now type 'RenameOperation'.
```


danielBelenky:

overriding known issue with ppc build

PR #3211, opened 2020-03-26
Remove the custom SELinux policies

jean-edouard:
This removes the custom SELinux policy applied to virt-launcher.
virt-launcher will default to container_t, which is mcs-constrained.

Fixes #

Start a VM and run `ps auxZ | grep virt-launcher` on the node.
virt-launcher should be running as container_t.

**Release note**:

```release-note
NONE
```


pkg/virt-controller/services/template.go
```diff
@@ -1004,9 +998,6 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 			Subdomain: vmi.Spec.Subdomain,
 			SecurityContext: &k8sv1.PodSecurityContext{
 				RunAsUser: &userId,
-				SELinuxOptions: &k8sv1.SELinuxOptions{
-					Type: selinuxLauncherType,
-				},
```
vladikr: I would still keep this. Just would set the default to `container_t`. Just in case some systems will need to use a different label. I've seen that there was a new `container_kvm_t` introduced recently.

vladikr:
@jean-edouard the installation of the policies should be removed as well:
https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-handler/selinux/labels.go#L161

pkg/virt-controller/services/template.go
jean-edouard: Well, I removed it on purpose.
We don't manage container_t. The reason we want virt-launcher to run as container_t is because that's what k8s wants to run it as, to be able to isolate it with SELinux.
If tomorrow k8s decides to use another type to label and secure containers, virt-launcher should follow.

vladikr: I'm not sure if that applies to the virt-launcher case. It's still special since we run virtualization components in it and should have room to set a different label.

jean-edouard: So this ended up being more than a philosophical question, since I just confirmed that forcing container_t leads to different behaviors than not forcing a type and letting it default to container_t.
I understand the need to leave room to set a different label, and I'll rework the code accordingly.

jean-edouard:
Closing, since the reworked fix is completely different. Will open a new PR.

PR #3216, opened 2020-03-27
Do not set an SELinux type on virt-launcher by default

jean-edouard:
virt-launcher will default to container_t, which is different than forcing it to container_t.
For example, container disks don't get relabelled when a type is forced.

Fixes #

With this fix, virt-launcher pods should have an SELinux type of container_t by default.
VMs that use a container disk should not trigger SELinux denials.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-controller/services/template_test.go
+++ b/pkg/virt-controller/services/template_test.go
@@ -173,6 +173,37 @@ var _ = Describe("Template", func() {
 				Expect(pod.Spec.Subdomain).To(BeEmpty())
 			})
 		})
+		Context("with SELinux types", func() {
+			It("should run under the SELinux type container_t if none specified", func() {
+				vmi := v1.VirtualMachineInstance{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "testvmi", Namespace: "default", UID: "1234",
+					},
+					Spec: v1.VirtualMachineInstanceSpec{Volumes: []v1.Volume{}, Domain: v1.DomainSpec{}},
+				}
+				pod, err := svc.RenderLaunchManifest(&vmi)
+				Expect(err).ToNot(HaveOccurred())
+				if pod.Spec.SecurityContext != nil && pod.Spec.SecurityContext.SELinuxOptions != nil {
+					Expect(pod.Spec.SecurityContext.SELinuxOptions.Type).To(Equal(""))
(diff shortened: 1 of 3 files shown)


jean-edouard:
@vladikr added a couple unit tests

pkg/virt-controller/services/template_test.go
```diff
@@ -173,6 +173,37 @@ var _ = Describe("Template", func() {
...
+				vmi := v1.VirtualMachineInstance{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "testvmi", Namespace: "default", UID: "1234",
+					},
+					Spec: v1.VirtualMachineInstanceSpec{Volumes: []v1.Volume{}, Domain: v1.DomainSpec{}},
+				}
+				pod, err := svc.RenderLaunchManifest(&vmi)
+				Expect(err).ToNot(HaveOccurred())
+				if pod.Spec.SecurityContext != nil && pod.Spec.SecurityContext.SELinuxOptions != nil {
+					Expect(pod.Spec.SecurityContext.SELinuxOptions.Type).To(Equal(""))
```
vladikr: maybe, `.To(BeEmpty())` instead?

pkg/virt-controller/services/template_test.go
```diff
@@ -173,6 +173,37 @@ var _ = Describe("Template", func() {
...
+				}
+			})
+			It("should run under the corresponding SELinux type if specified", func() {
+				testutils.UpdateFakeClusterConfig(configMapInformer, &kubev1.ConfigMap{
+					Data: map[string]string{virtconfig.SELinuxLauncherTypeKey: "virt_launcher.process"},
```
vladikr: I think we should use `spc_t` instead of `virt-launcher.process` as we are going to get rid of the later.

pkg/virt-controller/services/template_test.go
jean-edouard: I wasn't sure `.To(BeEmpty())` would work on a string. The rest of the file seems to use that only on arrays, and `""` on strings.

pkg/virt-controller/services/template_test.go
jean-edouard: Yes, good point, will do.

PR #3217, opened 2020-03-27
Allow bazel to have custom CA certs

### LINKED ISSUE #3194: Bazel can't push to private HTTPS registry
FEATURE REQUEST ?

/kind enhancement

**What happened**:
Trying to 'make push' to an HTTPS registry results in:

 unable to push image to <image name> Get https://registry/v2/: x509: certificate signed by unknown authority

Push succeeded

1. Create an HTTPS docker registry signed by a private or local CA
2. export DOCKER_PREFIX=<custom registry>
3. run 'make push'

I'm willing to fix this if I can get guidance and buy-in from the community as to how. Just as a test I did this and it fixed the problem, but I think we want to have a variable for developers to set and I'm not certain where this would best fit in the developer scripts. This is certainly not the real fix.

```
diff --git a/hack/bazel-push-images.sh b/hack/bazel-push-images.sh
index fcef26144..ad42def37 100755
--- a/hack/bazel-push-images.sh
+++ b/hack/bazel-push-images.sh
@@ -22,6 +22,9 @@ set -e
 source hack/common.sh
 source hack/config.sh


+cp /root/.docker/certs.d/custom-ca.crt /etc/pki/ca-trust/source/anchors/
+/usr/bin/update-ca-trust
+
 for tag in ${docker_tag} ${docker_tag_alt}; do
     for target in other-images virt-operator virt-api virt-controller virt-handler virt-launcher; do

```

It may be as simple as checking if a variable is set, e.g. "BAZEL_CUSTOM_CA_PATH" and mounting that into the build environment, checking if that path exists, copying it and running update-ca-trust.  Would this only apply to bazel-push-images.sh, or would it be in common.sh/config.sh?

mlsorensen:

This allows developers and CI systems to pull from and publish to private HTTPS docker registries by passing in a custom CA PEM file in the DOCKER_CA_CERT_FILE environment variable.  This then gets mounted into the build container and integrated into the container's system certs.

It seems necessary to do it this way since Bazel uses go-containerregistry package, and doesn't support the /etc/docker/certs.d functionality.

Fixes #3194 

**Reviewer Notes**:

There seem to be a lot of configurations/adjustable variables for the development process that aren't explicitly documented, I'm not certain if this needs further documentation or not. Please let me know if you want to see something added in the way of documenting this feature, and where.

```release-note
NONE
```

### DIFF
--- a/hack/bazel-build.sh
+++ b/hack/bazel-build.sh
@@ -26,6 +26,8 @@ rm -rf ${CMD_OUT_DIR}
 mkdir -p ${CMD_OUT_DIR}/virtctl
 mkdir -p ${CMD_OUT_DIR}/dump
 
+check_update_trust
+
 # Build all binaries for amd64
 bazel build \
     --config=${ARCHITECTURE} \
--- a/hack/bazel-build.sh
+++ b/hack/bazel-build.sh
@@ -26,8 +26,6 @@ rm -rf ${CMD_OUT_DIR}
 mkdir -p ${CMD_OUT_DIR}/virtctl
 mkdir -p ${CMD_OUT_DIR}/dump
 
-check_update_trust
-
 # Build all binaries for amd64
 bazel build \
     --config=${ARCHITECTURE} \
--- a/hack/dockerized
+++ b/hack/dockerized
@@ -93,6 +93,11 @@ volumes="-v ${BUILDER}:/root:rw,z"
 mkdir -p "${HOME}/.docker"
 volumes="$volumes -v ${HOME}/.docker:/root/.docker:ro,z"
 
+# add custom docker certs, if needed
+if [ -n "$DOCKER_CA_CERT_FILE" ] && [ -f "$DOCKER_CA_CERT_FILE" ]; then
+    volumes="$volumes -v ${DOCKER_CA_CERT_FILE}:${DOCKERIZED_CUSTOM_CA_PATH}:ro,z"
+fi
+
 # Ensure that a bazel server which is running is the correct one
 if [ -n "$(docker ps --format '{{.Names}}' | grep ${BUILDER}-bazel-server)" ]; then
     # check if the image is correct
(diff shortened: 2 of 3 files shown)


hack/bazel-build.sh
```diff
@@ -26,6 +26,8 @@ rm -rf ${CMD_OUT_DIR}
 mkdir -p ${CMD_OUT_DIR}/virtctl
 mkdir -p ${CMD_OUT_DIR}/dump
 
+check_update_trust
```
rmohr: I am not sure we want to run this every time. It is also possible to run the scripts directly without docker.

Can we instead call it in hack/dockerized via `exec` at the right moment?

rmohr: And there also only if a custom ca is provided.

mlsorensen: There is already a check for this to only run if custom CA is provided. I will look though to see how to move this to an exec.

hack/dockerized
```diff
@@ -109,6 +114,11 @@ if [ -z "$(docker ps --format '{{.Names}}' | grep ${BUILDER}-bazel-server)" ]; t
     docker run --network host -d ${volumes} --security-opt label:disable --name ${BUILDER}-bazel-server -w "/root/go/src/kubevirt.io/kubevirt" --rm ${KUBEVIRT_BUILDER_IMAGE} hack/bazel-server.sh
 fi
 
+# Update cert trust, if custom is provided
+if [ -n "$DOCKER_CA_CERT_FILE" ] && [ -f "$DOCKER_CA_CERT_FILE" ]; then
+    docker exec ${BUILDER}-bazel-server /entrypoint.sh "/usr/bin/update-ca-trust"
```
rmohr: I guess this does not take too long, to be an issue, if it is executed over and over?

mlsorensen: It doesn't take any noticeable time, not compared to the rsyncs and other things. I'm not sure it is worth adding change detection to, in order to determine if trust store needs an update per custom CA contents.

mlsorensen: Also might be worth pointing out that the user can set the variable only when they want to push a CA change to the image, and it won't run otherwise without the variable.

PR #3221, opened 2020-03-31
Fix Ipv6 Migration

oshoval:
Use IPv6 bind address and local address unless IPv6 is disabled,
Its done in order to support IPv6 only when will be available,
since at the moment it was using IPv4 0.0.0.0, and 127.0.0.1 as bind and localhost address.

Fix migration server to support IPv6, by using
well formatted IPv6 address in case IPv6 is enabled.

Move IsIpv6Enabled from common to network, in order to reuse it,
and make it independent on the handler at common (which isn't needed anyhow).

To remove this part once #3374 is merged.
Create a PodPreset in order to supply each virt-launcher with unique UUID,
this is mandatory for migration to pass, and since we are using Kind provider,
all the containers inherit the same UUID from DMI which is shared between the containers,
that act as cluster nodes.

Possible needed follow up changes:
1. Allow migration via the cli in kind providers (including ipv6) - for this we need to add for kubevirt, The script of Fabian, if its acceptable as part of kubevirt repository,
(The script creates the PodPreset which is created as part of the e2e tests, in the default name space, as the namespace of the vm in examples folder).

```release-note
NONE
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -32,25 +32,17 @@ func IsGPUVMI(vmi *v1.VirtualMachineInstance) bool {
 	return false
 }
 
-func isIpv6Disabled() bool {
-	res, err := exec.Command("cat", "/proc/sys/net/ipv6/conf/default/disable_ipv6").Output()
-	return err != nil || string(res) == "1"
+// IsIpv6Disabled returns if IPv6 is disabled according sysctl
+func IsIpv6Disabled() bool {
+	ipv6Disabled, err := exec.Command("cat", "/proc/sys/net/ipv6/conf/default/disable_ipv6").Output()
+	return err != nil || string(ipv6Disabled) == "1"
 }
 
 // GetIPBindAddress returns IP bind address (either 0.0.0.0 or [::] according sysctl disable_ipv6)
 func GetIPBindAddress() string {
-	if isIpv6Disabled() {
+	if IsIpv6Disabled() {
 		return "0.0.0.0"
 	}
 
 	return "[::]"
-}
-
-// GetLoopbackAddress returns the loopback IP address (either 127.0.0.1 or [::1] according sysctl disable_ipv6)
-func GetLoopbackAddress() string {
-	if isIpv6Disabled() {
-		return "127.0.0.1"
-	}
-
-	return "[::1]"
 }
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -1,6 +1,8 @@
 package util
 
 import (
+	"os/exec"
+
 	v1 "kubevirt.io/client-go/api/v1"
 )
 
@@ -28,4 +30,27 @@ func IsGPUVMI(vmi *v1.VirtualMachineInstance) bool {
 		return true
 	}
 	return false
+}
+
+func isIpv6Disabled() bool {
+	res, err := exec.Command("cat", "/proc/sys/net/ipv6/conf/default/disable_ipv6").Output()
+	return err != nil || string(res) == "1"
+}
+
+// GetIPBindAddress returns IP bind address (either 0.0.0.0 or [::] according sysctl disable_ipv6)
+func GetIPBindAddress() string {
+	if isIpv6Disabled() {
+		return "0.0.0.0"
+	}
+
+	return "[::]"
+}
+
+// GetLoopbackAddress returns the loopback IP address (either 127.0.0.1 or [::1] according sysctl disable_ipv6)
+func GetLoopbackAddress() string {
+	if isIpv6Disabled() {
+		return "127.0.0.1"
+	}
+
+	return "[::1]"
 }
--- a/pkg/virt-handler/migration-proxy/migration-proxy.go
+++ b/pkg/virt-handler/migration-proxy/migration-proxy.go
@@ -32,7 +32,6 @@ import (
 	netutils "k8s.io/utils/net"
 
 	"kubevirt.io/client-go/log"
-	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/network"
 )
 
 const (
@@ -97,7 +96,8 @@ func SourceUnixFile(baseDir string, key string) string {
 }
 
 func ipBindAddress() string {
-	if network.IsIpv6Enabled() {
+	podIP := os.Getenv("MY_POD_IP")
+	if netutils.IsIPv6String(podIP) {
 		return "[::]"
 	}
 
--- a/pkg/virt-handler/migration-proxy/migration-proxy.go
+++ b/pkg/virt-handler/migration-proxy/migration-proxy.go
@@ -29,7 +29,10 @@ import (
 	"strings"
 	"sync"
 
+	netutils "k8s.io/utils/net"
+
 	"kubevirt.io/client-go/log"
+	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/network"
 )
 
 const (
@@ -93,6 +96,14 @@ func SourceUnixFile(baseDir string, key string) string {
 	return filepath.Join(baseDir, "migrationproxy", key+"-source.sock")
 }
 
+func ipBindAddress() string {
+	if network.IsIpv6Enabled() {
+		return "[::]"
+	}
+
+	return "0.0.0.0"
+}
+
 func (m *migrationProxyManager) StartTargetListener(key string, targetUnixFiles []string) error {
 	m.managerLock.Lock()
 	defer m.managerLock.Unlock()
@@ -129,7 +140,7 @@ func (m *migrationProxyManager) StartTargetListener(key string, targetUnixFiles 
 	proxiesList := []*migrationProxy{}
 	for _, targetUnixFile := range targetUnixFiles {
 		// 0 means random port is used
-		proxy := NewTargetProxy("0.0.0.0", 0, m.serverTLSConfig, m.clientTLSConfig, targetUnixFile)
+		proxy := NewTargetProxy(ipBindAddress(), 0, m.serverTLSConfig, m.clientTLSConfig, targetUnixFile)
 
 		err := proxy.StartListening()
 		if err != nil {
@@ -211,6 +222,10 @@ func (m *migrationProxyManager) StopTargetListener(key string) {
 func (m *migrationProxyManager) StartSourceListener(key string, targetAddress string, destSrcPortMap map[int]int, baseDir string) error {
 	m.managerLock.Lock()
 	defer m.managerLock.Unlock()
+
+	if netutils.IsIPv6String(targetAddress) {
+		targetAddress = "[" + targetAddress + "]"
+	}
 
 	isExistingProxy := func(curProxies []*migrationProxy, targetAddress string, destSrcPortMap map[int]int) bool {
 		if len(curProxies) != len(destSrcPortMap) {
@@ -315,12 +330,16 @@ func NewTargetProxy(tcpBindAddress string, tcpBindPort int, serverTLSConfig *tls
 
 }
 
+func isLoopbackAddress(ipAddress string) bool {
+	return (strings.Contains(ipAddress, "127.0.0.1") || strings.Contains(ipAddress, "[::1]"))
+}
+
 func (m *migrationProxy) createTcpListener() error {
 	var listener net.Listener
 	var err error
 	if m.serverTLSConfig != nil {
 		listener, err = tls.Listen("tcp", fmt.Sprintf("%s:%d", m.tcpBindAddress, m.tcpBindPort), m.serverTLSConfig)
-	} else if strings.Contains(m.tcpBindAddress, "127.0.0.1") {
+	} else if isLoopbackAddress(m.tcpBindAddress) {
 		listener, err = net.Listen("tcp", fmt.Sprintf("%s:%d", m.tcpBindAddress, m.tcpBindPort))
 	} else {
 		return fmt.Errorf("Unsecured tcp migration proxy listeners are not permitted")
--- a/pkg/virt-handler/migration-proxy/migration-proxy.go
+++ b/pkg/virt-handler/migration-proxy/migration-proxy.go
@@ -32,6 +32,7 @@ import (
 	netutils "k8s.io/utils/net"
 
 	"kubevirt.io/client-go/log"
+	"kubevirt.io/kubevirt/pkg/util"
 )
 
 const (
@@ -95,15 +96,6 @@ func SourceUnixFile(baseDir string, key string) string {
 	return filepath.Join(baseDir, "migrationproxy", key+"-source.sock")
 }
 
-func ipBindAddress() string {
-	podIP := os.Getenv("MY_POD_IP")
-	if netutils.IsIPv6String(podIP) {
-		return "[::]"
-	}
-
-	return "0.0.0.0"
-}
-
 func (m *migrationProxyManager) StartTargetListener(key string, targetUnixFiles []string) error {
 	m.managerLock.Lock()
 	defer m.managerLock.Unlock()
@@ -140,7 +132,7 @@ func (m *migrationProxyManager) StartTargetListener(key string, targetUnixFiles 
 	proxiesList := []*migrationProxy{}
 	for _, targetUnixFile := range targetUnixFiles {
 		// 0 means random port is used
-		proxy := NewTargetProxy(ipBindAddress(), 0, m.serverTLSConfig, m.clientTLSConfig, targetUnixFile)
+		proxy := NewTargetProxy(util.GetIPBindAddress(), 0, m.serverTLSConfig, m.clientTLSConfig, targetUnixFile)
 
 		err := proxy.StartListening()
 		if err != nil {
--- a/pkg/virt-launcher/virtwrap/manager.go
+++ b/pkg/virt-launcher/virtwrap/manager.go
@@ -59,7 +59,7 @@ import (
 	"kubevirt.io/kubevirt/pkg/hooks"
 	hostdisk "kubevirt.io/kubevirt/pkg/host-disk"
 	"kubevirt.io/kubevirt/pkg/ignition"
-	generalutil "kubevirt.io/kubevirt/pkg/util"
+	utils "kubevirt.io/kubevirt/pkg/util"
 	migrationproxy "kubevirt.io/kubevirt/pkg/virt-handler/migration-proxy"
 	agentpoller "kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/agent-poller"
 	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
@@ -96,6 +96,7 @@ type DomainManager interface {
 	GetUsers() ([]v1.VirtualMachineInstanceGuestOSUser, error)
 	GetFilesystems() ([]v1.VirtualMachineInstanceFileSystem, error)
 	SetGuestTime(*v1.VirtualMachineInstance) error
+	GetLoopbackAddress() string
 }
 
 type LibvirtDomainManager struct {
@@ -409,7 +410,7 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
 		isBlockMigration := (vmi.Status.MigrationMethod == v1.BlockMigration)
 		migrationPortsRange := migrationproxy.GetMigrationPortsList(isBlockMigration)
 
-		loopbackAddress := generalutil.GetLoopbackAddress()
+		loopbackAddress := l.GetLoopbackAddress()
 
 		// Create a tcp server for each direct connection proxy
 		for _, port := range migrationPortsRange {
@@ -498,6 +499,14 @@ func (l *LibvirtDomainManager) SetGuestTime(vmi *v1.VirtualMachineInstance) erro
 	// try setting the guest time. Fail early if agent is not configured.
 	l.setGuestTime(vmi, dom)
 	return nil
+}
+
+func (l *LibvirtDomainManager) GetLoopbackAddress() string {
+	if utils.IsIpv6Disabled() {
+		return "127.0.0.1"
+	}
+
+	return "[::1]"
 }
 
 func (l *LibvirtDomainManager) getGuestTimeContext() context.Context {
@@ -749,7 +758,7 @@ func (l *LibvirtDomainManager) MigrateVMI(vmi *v1.VirtualMachineInstance, option
 		return nil
 	}
 
-	if err := updateHostsFile(fmt.Sprintf("%s %s\n", generalutil.GetLoopbackAddress(), vmi.Status.MigrationState.TargetPod)); err != nil {
+	if err := updateHostsFile(fmt.Sprintf("%s %s\n", l.GetLoopbackAddress(), vmi.Status.MigrationState.TargetPod)); err != nil {
 		return fmt.Errorf("failed to update the hosts file: %v", err)
 	}
 	l.asyncMigrate(vmi, options)
@@ -857,8 +866,7 @@ func (l *LibvirtDomainManager) PrepareMigrationTarget(vmi *v1.VirtualMachineInst
 		return fmt.Errorf("executing custom preStart hooks failed: %v", err)
 	}
 
-	loopbackAddress := generalutil.GetLoopbackAddress()
-	if err := updateHostsFile(fmt.Sprintf("%s %s\n", loopbackAddress, vmi.Status.MigrationState.TargetPod)); err != nil {
+	if err := updateHostsFile(fmt.Sprintf("%s %s\n", l.GetLoopbackAddress(), vmi.Status.MigrationState.TargetPod)); err != nil {
 		return fmt.Errorf("failed to update the hosts file: %v", err)
 	}
 
@@ -867,7 +875,7 @@ func (l *LibvirtDomainManager) PrepareMigrationTarget(vmi *v1.VirtualMachineInst
 	for _, port := range migrationPortsRange {
 		// Prepare the direct migration proxy
 		key := migrationproxy.ConstructProxyKey(string(vmi.UID), port)
-		curDirectAddress := fmt.Sprintf("%s:%d", loopbackAddress, port)
+		curDirectAddress := fmt.Sprintf("%s:%d", l.GetLoopbackAddress(), port)
 		unixSocketPath := migrationproxy.SourceUnixFile(l.virtShareDir, key)
 		migrationProxy := migrationproxy.NewSourceProxy(unixSocketPath, curDirectAddress, nil, nil)
 
--- a/pkg/virt-launcher/virtwrap/manager.go
+++ b/pkg/virt-launcher/virtwrap/manager.go
@@ -407,11 +407,12 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
 
 		isBlockMigration := (vmi.Status.MigrationMethod == v1.BlockMigration)
 		migrationPortsRange := migrationproxy.GetMigrationPortsList(isBlockMigration)
+		loopbackAddress := network.GetLoopbackAddress()
 
 		// Create a tcp server for each direct connection proxy
 		for _, port := range migrationPortsRange {
 			key := migrationproxy.ConstructProxyKey(string(vmi.UID), port)
-			migrationProxy := migrationproxy.NewTargetProxy("127.0.0.1", port, nil, nil, migrationproxy.SourceUnixFile(l.virtShareDir, key))
+			migrationProxy := migrationproxy.NewTargetProxy(loopbackAddress, port, nil, nil, migrationproxy.SourceUnixFile(l.virtShareDir, key))
 			defer migrationProxy.StopListening()
 			err := migrationProxy.StartListening()
 			if err != nil {
@@ -421,7 +422,7 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
 		}
 
 		//  proxy incoming migration requests on port 22222 to the vmi's existing libvirt connection
-		libvirtConnectionProxy := migrationproxy.NewTargetProxy("127.0.0.1", LibvirtLocalConnectionPort, nil, nil, migrationproxy.SourceUnixFile(l.virtShareDir, string(vmi.UID)))
+		libvirtConnectionProxy := migrationproxy.NewTargetProxy(loopbackAddress, LibvirtLocalConnectionPort, nil, nil, migrationproxy.SourceUnixFile(l.virtShareDir, string(vmi.UID)))
 		defer libvirtConnectionProxy.StopListening()
 		err := libvirtConnectionProxy.StartListening()
 		if err != nil {
@@ -430,8 +431,8 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
 		}
 
 		// For a tunnelled migration, this is always the uri
-		dstUri := fmt.Sprintf("qemu+tcp://127.0.0.1:%d/system", LibvirtLocalConnectionPort)
-		migrUri := "tcp://127.0.0.1"
+		dstUri := fmt.Sprintf("qemu+tcp://%s:%d/system", loopbackAddress, LibvirtLocalConnectionPort)
+		migrUri := "tcp://" + loopbackAddress
 
 		domName := api.VMINamespaceKeyFunc(vmi)
 		dom, err := l.virConn.LookupDomainByName(domName)
@@ -746,7 +747,7 @@ func (l *LibvirtDomainManager) MigrateVMI(vmi *v1.VirtualMachineInstance, option
 		return nil
 	}
 
-	if err := updateHostsFile(fmt.Sprintf("%s %s\n", "127.0.0.1", vmi.Status.MigrationState.TargetPod)); err != nil {
+	if err := updateHostsFile(fmt.Sprintf("%s %s\n", network.GetLoopbackAddress(), vmi.Status.MigrationState.TargetPod)); err != nil {
 		return fmt.Errorf("failed to update the hosts file: %v", err)
 	}
 	l.asyncMigrate(vmi, options)
@@ -854,7 +855,8 @@ func (l *LibvirtDomainManager) PrepareMigrationTarget(vmi *v1.VirtualMachineInst
 		return fmt.Errorf("executing custom preStart hooks failed: %v", err)
 	}
 
-	if err := updateHostsFile(fmt.Sprintf("%s %s\n", "127.0.0.1", vmi.Status.MigrationState.TargetPod)); err != nil {
+	loopbackAddress := network.GetLoopbackAddress()
+	if err := updateHostsFile(fmt.Sprintf("%s %s\n", loopbackAddress, vmi.Status.MigrationState.TargetPod)); err != nil {
 		return fmt.Errorf("failed to update the hosts file: %v", err)
 	}
 
@@ -863,7 +865,7 @@ func (l *LibvirtDomainManager) PrepareMigrationTarget(vmi *v1.VirtualMachineInst
 	for _, port := range migrationPortsRange {
 		// Prepare the direct migration proxy
 		key := migrationproxy.ConstructProxyKey(string(vmi.UID), port)
-		curDirectAddress := fmt.Sprintf("%s:%d", "127.0.0.1", port)
+		curDirectAddress := fmt.Sprintf("%s:%d", loopbackAddress, port)
 		unixSocketPath := migrationproxy.SourceUnixFile(l.virtShareDir, key)
 		migrationProxy := migrationproxy.NewSourceProxy(unixSocketPath, curDirectAddress, nil, nil)
 
--- a/pkg/virt-launcher/virtwrap/manager.go
+++ b/pkg/virt-launcher/virtwrap/manager.go
@@ -407,12 +407,11 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
 
 		isBlockMigration := (vmi.Status.MigrationMethod == v1.BlockMigration)
 		migrationPortsRange := migrationproxy.GetMigrationPortsList(isBlockMigration)
-		loopbackAddress := network.GetLoopbackAddress()
 
 		// Create a tcp server for each direct connection proxy
 		for _, port := range migrationPortsRange {
 			key := migrationproxy.ConstructProxyKey(string(vmi.UID), port)
-			migrationProxy := migrationproxy.NewTargetProxy(loopbackAddress, port, nil, nil, migrationproxy.SourceUnixFile(l.virtShareDir, key))
+			migrationProxy := migrationproxy.NewTargetProxy("127.0.0.1", port, nil, nil, migrationproxy.SourceUnixFile(l.virtShareDir, key))
 			defer migrationProxy.StopListening()
 			err := migrationProxy.StartListening()
 			if err != nil {
@@ -422,7 +421,7 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
 		}
 
 		//  proxy incoming migration requests on port 22222 to the vmi's existing libvirt connection
-		libvirtConnectionProxy := migrationproxy.NewTargetProxy(loopbackAddress, LibvirtLocalConnectionPort, nil, nil, migrationproxy.SourceUnixFile(l.virtShareDir, string(vmi.UID)))
+		libvirtConnectionProxy := migrationproxy.NewTargetProxy("127.0.0.1", LibvirtLocalConnectionPort, nil, nil, migrationproxy.SourceUnixFile(l.virtShareDir, string(vmi.UID)))
 		defer libvirtConnectionProxy.StopListening()
 		err := libvirtConnectionProxy.StartListening()
 		if err != nil {
@@ -431,8 +430,8 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
 		}
 
 		// For a tunnelled migration, this is always the uri
-		dstUri := fmt.Sprintf("qemu+tcp://%s:%d/system", loopbackAddress, LibvirtLocalConnectionPort)
-		migrUri := "tcp://" + loopbackAddress
+		dstUri := fmt.Sprintf("qemu+tcp://127.0.0.1:%d/system", LibvirtLocalConnectionPort)
+		migrUri := "tcp://127.0.0.1"
 
 		domName := api.VMINamespaceKeyFunc(vmi)
 		dom, err := l.virConn.LookupDomainByName(domName)
@@ -747,7 +746,7 @@ func (l *LibvirtDomainManager) MigrateVMI(vmi *v1.VirtualMachineInstance, option
 		return nil
 	}
 
-	if err := updateHostsFile(fmt.Sprintf("%s %s\n", network.GetLoopbackAddress(), vmi.Status.MigrationState.TargetPod)); err != nil {
+	if err := updateHostsFile(fmt.Sprintf("%s %s\n", "127.0.0.1", vmi.Status.MigrationState.TargetPod)); err != nil {
 		return fmt.Errorf("failed to update the hosts file: %v", err)
 	}
 	l.asyncMigrate(vmi, options)
@@ -855,8 +854,7 @@ func (l *LibvirtDomainManager) PrepareMigrationTarget(vmi *v1.VirtualMachineInst
 		return fmt.Errorf("executing custom preStart hooks failed: %v", err)
 	}
 
-	loopbackAddress := network.GetLoopbackAddress()
-	if err := updateHostsFile(fmt.Sprintf("%s %s\n", loopbackAddress, vmi.Status.MigrationState.TargetPod)); err != nil {
+	if err := updateHostsFile(fmt.Sprintf("%s %s\n", "127.0.0.1", vmi.Status.MigrationState.TargetPod)); err != nil {
 		return fmt.Errorf("failed to update the hosts file: %v", err)
 	}
 
@@ -865,7 +863,7 @@ func (l *LibvirtDomainManager) PrepareMigrationTarget(vmi *v1.VirtualMachineInst
 	for _, port := range migrationPortsRange {
 		// Prepare the direct migration proxy
 		key := migrationproxy.ConstructProxyKey(string(vmi.UID), port)
-		curDirectAddress := fmt.Sprintf("%s:%d", loopbackAddress, port)
+		curDirectAddress := fmt.Sprintf("%s:%d", "127.0.0.1", port)
 		unixSocketPath := migrationproxy.SourceUnixFile(l.virtShareDir, key)
 		migrationProxy := migrationproxy.NewSourceProxy(unixSocketPath, curDirectAddress, nil, nil)
 
--- a/pkg/virt-launcher/virtwrap/manager.go
+++ b/pkg/virt-launcher/virtwrap/manager.go
@@ -59,6 +59,7 @@ import (
 	"kubevirt.io/kubevirt/pkg/hooks"
 	hostdisk "kubevirt.io/kubevirt/pkg/host-disk"
 	"kubevirt.io/kubevirt/pkg/ignition"
+	generalutil "kubevirt.io/kubevirt/pkg/util"
 	migrationproxy "kubevirt.io/kubevirt/pkg/virt-handler/migration-proxy"
 	agentpoller "kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/agent-poller"
 	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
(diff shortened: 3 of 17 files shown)


maiqueb:
> I'm suggesting you create a function like `isIPv6Address(address string) bool {...}` ; it returns true only if it's argument is an ipv6 address - a unix socket is not an ipv6 address.
> 
> I think this would both simplify the code(remove duplicated code) and improve readability.

No need to create anything; I would just use https://github.com/kubernetes/utils/blob/master/net/net.go#L122

oshoval:
> > I'm suggesting you create a function like `isIPv6Address(address string) bool {...}` ; it returns true only if it's argument is an ipv6 address - a unix socket is not an ipv6 address.
> > I think this would both simplify the code(remove duplicated code) and improve readability.
> 
> No need to create anything; I would just use https://github.com/kubernetes/utils/blob/master/net/net.go#L122

it comes ipv6:port sometimes so that function wont work unless it support also port

AlonaKaplan:
> isn't there an utility function to know if an ip address is v4 or v6 ? if not, this is a good time to create it - and use it here.

You have k8s.io/utils/net - IsIPv6String

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -382,6 +383,16 @@ func getDiskTargetsForMigration(dom cli.VirDomain, vmi *v1.VirtualMachineInstanc
 	return copyDisks
 }
 
+func loopbackAddress() string {
+	ipString := os.Getenv("KUBERNETES_SERVICE_HOST")
+	if netutils.IsIPv6String(ipString) {
+		return "[::1]"
+	}
+
+	// TODO in case env is "" do we want to return error or to fallback to IPv4 ?
```
oshoval, line 392: @AlonaKaplan 
what to do in case of an error ? just fall back to ipv4 ? return err all the way ?
just print and faill back to ipv4 ?

same question on the 2nd function

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -92,6 +94,16 @@ func SourceUnixFile(virtShareDir string, key string) string {
 	return filepath.Join(virtShareDir, "migrationproxy", key+"-source.sock")
 }
 
+func ipAnyAddress() string {
```
oshoval, line 97: see question below on the function like this please

tests/migration_test.go
```diff
@@ -386,9 +387,8 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 			})
 		})
 		Context("with a Cirros disk", func() {
-			It("should be successfully migrate with cloud-init disk with devices on the root bus", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
-
+			It("[test_id:3227]should be successfully migrate with cloud-init disk with devices on the root bus", func() {
+				//tests.SkipMigrationTestIfRunnigOnKindInfra()
```
oshoval, line 391: still wip, when i finish i will remove all those ,thx

tests/migration_test.go
```diff
@@ -386,9 +387,8 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 			})
 		})
 		Context("with a Cirros disk", func() {
-			It("should be successfully migrate with cloud-init disk with devices on the root bus", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
-
+			It("[test_id:3227]should be successfully migrate with cloud-init disk with devices on the root bus", func() {
```
oshoval, line 390: added a new number , hope its fine

pkg/virt-controller/services/template.go
```diff
@@ -370,6 +370,25 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 			},
 		})
 	}
+
+        // Fix for kind "same host" migration, how to do it right ?
+        // keep in mind that the creation of the proudct_uuid i still ran manually
```
oshoval, line 375: for WORKER in kind-1.17.0-control-plane kind-1.17.0-worker; do
  docker exec $WORKER sh -c 'cat /proc/sys/kernel/random/uuid > /kind/product_uuid'
done

at the moment its not fully automatic, and the lines above are needed
if the file doesnt exists it would fail, again its WIP and known

pkg/virt-controller/services/template.go
```diff
@@ -370,6 +370,25 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
...
+	volumeMounts = append(volumeMounts, k8sv1.VolumeMount{
```
AlonaKaplan, line 376: I think @maiqueb already wrote it and I totally agree. I wouldn't touch the code base of all the virt-launchers only for fixing a kind bug. This should be done as part of the make cluster-sync, only for kind provider.

AlonaKaplan: I would add it to the tests. Would create a migratable vm that all the migration tests can use (in case it is a kind cluster).

oshoval: but at the end we understood that it need to be here, but later on now there is maybe a way that we can research

on the other hand , there were couple people that do thought its good on kubevirt
anyway i am focusing a moment for a research

the bug is not Kind ,but non robustness of kubevirt for containers, this is a side note

oshoval: > I would add it to the tests. Would create a migratable vm that all the migration tests can use (in case it is a kind cluster).

i need it in the virt-launchers, i will investigate first why virtctl migrate works
and from tests it doesnt, and we will see after

cluster-up/cluster/kind-k8s-1.17.0-ipv6/provider.sh
```diff
@@ -24,10 +24,22 @@ function up() {
             mount_disk $node $i
         done
         $DOCKER exec $node bash -c "chmod -R 777 /var/local/kubevirt-storage/local-volume"
+
+        # Create a unique UUID file reference file for each node, that will be mounted in order to support
+        # Migration in Kind providers.
+        $DOCKER exec $node bash -c 'cat /proc/sys/kernel/random/uuid > /kind/product_uuid'
```
oshoval, line 30: this is already merged thanks to Daniel in kubevirtci
there is a PR that will bump it to kubevirt and then it will gone from this PR

pkg/virt-handler/migration-proxy/migration-proxy.go
oshoval: saw that Alona updated is ipv6 enable func, and saw there are some remarks to save state and such,
once it will be done i will align the 2 uses of this func

pkg/virt-launcher/virtwrap/manager.go
oshoval: obsolete , see please previous comment

tests/migration_test.go
```diff
@@ -820,7 +811,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				By("Waiting for NFS pod to disappear")
 				tests.WaitForPodToDisappearWithTimeout(tests.NFSTargetName, 120)
 			})
-			It("[test_id:2653]  should be migrated successfully, using guest agent on VM", func() {
+			PIt("[test_id:2653]  should be migrated successfully, using guest agent on VM", func() {
```
oshoval, line 814: will need to adjust to a real skip

tests/utils.go
```diff
@@ -1175,6 +1180,48 @@ func cleanupSubresourceServiceAccount() {
 	}
 }
 
+func createPodPreset(ppName string) {
```
oshoval, line 1183: guess we need to template some vars in it ?

tests/migration_test.go
oshoval: can i usetests.SkipPVCTestIfRunnigOnKindInfra() ?
since it runs NewRandomVMIWithPVC

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -319,7 +331,7 @@ func (m *migrationProxy) createTcpListener() error {
 	var err error
 	if m.tlsConfig != nil {
 		listener, err = tls.Listen("tcp", fmt.Sprintf("%s:%d", m.tcpBindAddress, m.tcpBindPort), m.tlsConfig)
```
AlonaKaplan, line 333: Don't you need here [] around the tcpBindAddress in case of ipv6?

pkg/virt-controller/services/template.go
```diff
@@ -370,6 +370,7 @@ func (t *templateService) RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (
 			},
 		})
 	}
+
```
oshoval, line 373: will do cosmetics soon

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -366,6 +378,29 @@ func (m *migrationProxy) StopListening() {
...
+	if strings.Contains(targetAddress, "[") {
+		return targetAddress
+	}
+
+	// Unix socket doesnt need to be formatted
+	if strings.Count(targetAddress, ":") == 0 {
+		return targetAddress
+	}
+
+	stringSlice := strings.Split(targetAddress, ":")
```
AlonaKaplan, line 392: If you get an ipv6 socket with no brackets it means the sender of this socket has a bug. IMO, you shouldn't fix a non-valid ipv6 socket here, the sender should send the correct format. If you get a wrong format from the sender you should fail.

pkg/virt-handler/migration-proxy/migration-proxy.go
oshoval: ips when they were specific ip and not ::1 , already had '['
i can add print to make sure we pass here and see what the ip is

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -382,6 +383,16 @@ func getDiskTargetsForMigration(dom cli.VirDomain, vmi *v1.VirtualMachineInstanc
 	return copyDisks
 }
 
+func loopbackAddress() string {
+	ipString := os.Getenv("KUBERNETES_SERVICE_HOST")
```
AlonaKaplan, line 387: I'm not sure we can use "KUBERNETES_SERVICE_HOST" for this purpose. Maybe we can add "MY_POD_IP" env to the virt-launcher. Please consult with @rmohr about it.

pkg/virt-handler/migration-proxy/migration-proxy.go
oshoval: it means we need to find all the situations of the callers, and for each one of them wrap the ip,
also need to take into account that if there are cases where the migration that i will run doesnt pass across , i wont be able to find them so fast

tests/migration_test.go
```diff
@@ -230,7 +230,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 			migration, err = virtClient.VirtualMachineInstanceMigration(migration.Namespace).Create(migration)
 			return err
 		}, timeout, 1*time.Second).ShouldNot(HaveOccurred())
-		By("Waiting until the Migration Completes")
+		By("Waiting until the Migration Completes zzz")
```
AlonaKaplan, line 233: zzz?

pkg/virt-launcher/virtwrap/manager.go
oshoval: i think that someone the env already have MY_POD_IP, 
i saw you are using it in launcher, just waiting because saw as well Roman asked some refactor there

tests/migration_test.go
oshoval: i will remove it of course, its still WIP, 
its a way for me to find fast the place where i want to add sleep

tests/migration_test.go
AlonaKaplan: yes, but we need to make sure Adam's PR is indeed fixing it. If not, we should keep investigating.

oshoval: adam PR is still WIP, and is needed for it , so we better wait with it and focus on the other first ?

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -765,7 +780,7 @@ func (l *LibvirtDomainManager) PrepareMigrationTarget(vmi *v1.VirtualMachineInst
 	for _, port := range migrationPortsRange {
 		// Prepare the direct migration proxy
 		key := migrationproxy.ConstructProxyKey(string(vmi.UID), port)
-		curDirectAddress := fmt.Sprintf("%s:%d", "127.0.0.1", port)
+		curDirectAddress := fmt.Sprintf("%s:%d", loopbackAddress(), port)
```
AlonaKaplan, line 783: The ipv6 pr also introduced getLoopbackAddress(..) method. Maybe it can be moved to Common and reused.

oshoval: cant see it now on 3112  ? once it there i will look on it if its ok

pkg/virt-handler/migration-proxy/migration-proxy.go
oshoval: • [SLOW TEST:137.866 seconds]
[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system] VM Live Migration
/root/go/src/kubevirt.io/kubevirt/tests/migration_test.go:60
  Starting a VirtualMachineInstance 
  /root/go/src/kubevirt.io/kubevirt/tests/migration_test.go:345
    migration security
    /root/go/src/kubevirt.io/kubevirt/tests/migration_test.go:869
      [test_id:2303][posneg:negative] should secure migrations with TLS
      /root/go/src/kubevirt.io/kubevirt/tests/migration_test.go:886


a TLS test passes, it means the ip already have it (and i did saw ips that have [ ] already)
else the listen wont success

pkg/virt-handler/migration-proxy/migration-proxy.go
rmohr: @oshoval could you rephrase? I can't follow the reply.

rmohr: I don't understand why we expect situations where we don't get either a valid ipv4 or ipv6 address.

oshoval: Hi, sure

Generally saying that if we want that the caller will have the responsibility to format its ip:port as [ipv6]:port, in case needed (ipv6 and not ipv4 / unix socket),
it means that i will need to find all the places where it call the function, and add there the logic that checks if wrap is needed / wrap according. 

just wonder what is the right approach you both prefer, since Miguel and i liked more the approach that already exists, at least from what i understood on previous comments, but you both decide.

Thanks

oshoval: > I don't understand why we expect situations where we don't get either a valid ipv4 or ipv6 address.

callers doesnt format the ipv6:port as [ipv6]:port
so according this comment, i believe you prefer the way Alona said as well,
to make the caller format the TargetAddress

rmohr: Not sure I can follow. If something creates invalid sockets, then that should be fixed. Does that answer the question? We should not create wrong things on one end, only to fix it then somewhere else.

rmohr: You can always guard against invalid input and fail with noise, but trying to fix input somehow is not good.

oshoval: it doesn't create wrong sockets since this trap and format it well, but i understand that you prefer that its will be the caller responsibility to format it well and not the endpoint, as Alona asked, thanks on it.

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -93,6 +95,16 @@ func SourceUnixFile(baseDir string, key string) string {
 	return filepath.Join(baseDir, "migrationproxy", key+"-source.sock")
 }
 
+func ipAnyAddress() string {
+	ipString := os.Getenv("MY_POD_IP")
+	if netutils.IsIPv6String(ipString) {
+		return "[::]"
+	}
+
+	// TODO in case env is "" do we want to return error or to fallback to IPv4 ?
```
AlonaKaplan: I think fallback to ipv4 to avoid regression.

pkg/virt-launcher/virtwrap/manager.go
AlonaKaplan: I'm using it in virt-handler, not launcher (although the code is in virt-launcher it is running in virt-handler. Confusing, I know).

tests/migration_test.go
AlonaKaplan: Adam's Pr is ready, just has some ci issues.
You can meanwhile take the commit from Adam's PR and add it to your PR to test if it fixes this test.

If fixes, you can use the tests.SkipPVCTestIfRunnigOnKindInfra(). Otherwise, you will need to introduce a new skip.

pkg/virt-launcher/virtwrap/network/podinterface_test.go
```diff
@@ -378,7 +384,7 @@ var _ = Describe("Pod Network", func() {
 				for _, proto := range ipProtocols() {
 					mockNetwork.EXPECT().HasNatIptables(proto).Return(true).Times(2)
 				}
-				mockNetwork.EXPECT().IsIpv6Enabled().Return(true).Times(3)
+				Expect(CheckIsIpv6Enabled()).To(BeTrue())
```
AlonaKaplan, line 387: Why do you need to check the value of a stub method? If it possible I would keep the 'Times' check.

tests/migration_test.go
oshoval: Storage  Starting a VirtualMachineInstance  [rfe_id:2298][crit:medium][vendor:cnv-qe@redhat.com][level:component]  With HostDisk and PVC initialization With smaller than requested PVCs  [test_id:3109]Should initialize an empty PVC with a disk.img when disk  is too small but within toleration expand_less | 38s
-- | --
/root/go/src/kubevirt.io/kubevirt/tests/storage_test.go:577 Unexpected Warning event received: testvmisj68cp8p66cf9cm2vdzz5wq585hd72lxwprqz6w4xlmq2cfn,1db27be8-9653-41bc-b23c-d06fad462249: server error. command SyncVMI failed: "preparing host-disks failed: unable to create /tmp/kubevirt/wzz5b89nh7/mount/disk.img, not enough space, demanded size 227781854400 B is bigger than available space 134950559744 B, also after taking 30 % toleration into account"


this test started to fail , maybe lets wait for the PR of Adam to be merged and this storage issues to be fixed and then we can try at once?

oshoval: as we spoke, it doesnt fix it yet, we will skip and back to it in following PR once its merged

oshoval:
@rmohr 
Hi Roman , can you please review this PR ?
thanks

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -93,6 +96,14 @@ func SourceUnixFile(baseDir string, key string) string {
 	return filepath.Join(baseDir, "migrationproxy", key+"-source.sock")
 }
 
+func ipAnyAddress() string {
```
maiqueb, line 99: I'd rename this to `ipBindAddress` . "any" is misleading.

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -320,7 +335,7 @@ func (m *migrationProxy) createTcpListener() error {
 	var err error
 	if m.serverTLSConfig != nil {
 		listener, err = tls.Listen("tcp", fmt.Sprintf("%s:%d", m.tcpBindAddress, m.tcpBindPort), m.serverTLSConfig)
-	} else if strings.Contains(m.tcpBindAddress, "127.0.0.1") {
+	} else if strings.Contains(m.tcpBindAddress, "127.0.0.1") || strings.Contains(m.tcpBindAddress, "[::1]") {
```
maiqueb, line 338: I'd encapsulate "if strings.Contains(m.tcpBindAddress, "127.0.0.1") || strings.Contains(m.tcpBindAddress, "[::1]")" in a method named `isLocallyBound` or `isBoundToLocalhost` or something like that.

pkg/virt-handler/vm.go
```diff
@@ -1740,7 +1740,6 @@ func (d *VirtualMachineController) processVmUpdate(origVMI *v1.VirtualMachineIns
 
 	if d.isPreMigrationTarget(vmi) {
 		if !isMigrating(vmi) {
-
```
maiqueb, line 1743: unrelated, isn't it ?

pkg/virt-handler/vm.go
```diff
@@ -1787,7 +1786,6 @@ func (d *VirtualMachineController) processVmUpdate(origVMI *v1.VirtualMachineIns
 			d.recorder.Event(vmi, k8sv1.EventTypeNormal, v1.Migrating.String(), "VirtualMachineInstance is migrating.")
 		}
 	} else {
-
```
maiqueb, line 1790: ditto

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -742,7 +743,7 @@ func (l *LibvirtDomainManager) MigrateVMI(vmi *v1.VirtualMachineInstance, option
 		return nil
 	}
 
-	if err := updateHostsFile(fmt.Sprintf("%s %s\n", "127.0.0.1", vmi.Status.MigrationState.TargetPod)); err != nil {
+	if err := updateHostsFile(fmt.Sprintf("%s %s\n", network.GetLoopbackAddress(), vmi.Status.MigrationState.TargetPod)); err != nil {
```
maiqueb, line 746: I'm confused here; why aren't you re-using the `loopbackAddress` you've created previously ? 

Did I miss something ?

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -850,7 +851,7 @@ func (l *LibvirtDomainManager) PrepareMigrationTarget(vmi *v1.VirtualMachineInst
 		return fmt.Errorf("executing custom preStart hooks failed: %v", err)
 	}
 
-	if err := updateHostsFile(fmt.Sprintf("%s %s\n", "127.0.0.1", vmi.Status.MigrationState.TargetPod)); err != nil {
+	if err := updateHostsFile(fmt.Sprintf("%s %s\n", network.GetLoopbackAddress(), vmi.Status.MigrationState.TargetPod)); err != nil {
```
maiqueb, line 854: ditto

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -859,7 +860,7 @@ func (l *LibvirtDomainManager) PrepareMigrationTarget(vmi *v1.VirtualMachineInst
 	for _, port := range migrationPortsRange {
 		// Prepare the direct migration proxy
 		key := migrationproxy.ConstructProxyKey(string(vmi.UID), port)
-		curDirectAddress := fmt.Sprintf("%s:%d", "127.0.0.1", port)
+		curDirectAddress := fmt.Sprintf("%s:%d", network.GetLoopbackAddress(), port)
```
maiqueb, line 863: ditto

pkg/virt-launcher/virtwrap/manager.go
oshoval: here its 127.0.0.1  / ::1
there its 0.0.0.0 / ::

pkg/virt-launcher/virtwrap/network/common.go
```diff
@@ -154,15 +151,6 @@ func (h *NetworkUtilsHandler) ConfigureIpv6Forwarding() error {
...
 
-func (h *NetworkUtilsHandler) IsIpv6Enabled() bool {
-	podIp := os.Getenv("MY_POD_IP")
-	if !netutils.IsIPv6String(podIp) {
-		log.Log.V(5).Info("Since the pod ip is non IPv6, IPv6 is disabled")
-		return false
-	}
-	return true
-}
-
```
maiqueb, line 165: why move this function from `common` to `network` ? 

If it is really needed (not sure), I think it deserves to be done in an isolated commit, where the motivation for it is explained.

oshoval: its needed in order use IsIpv6Enabled in several places - but have one function
it was groomed with Alona , and i need it in this PR please

tests/utils.go
```diff
@@ -1221,6 +1226,48 @@ func cleanupSubresourceServiceAccount() {
 	}
 }
 
+func createPodPreset(ppName, volName, podName, srcPath, dstPath string) {
```
maiqueb, line 1229: Please use the full name of the variables - at least for ppName => podPresetName .

I also find it more readable to be explicit about the types, but it could be just me; feel free to ignore me on this.

Finally, I think your 'podName' is really not a pod name, but a pod *label* instead, isn't it ? e.g. `virt-launcher-<vmi name>` is the name of a virt-launcher pod. At least, I that's what I understood from the [docs](https://kubernetes.io/docs/concepts/workloads/pods/podpreset/#how-it-works).

tests/utils.go
```diff
@@ -4551,15 +4598,14 @@ func IsRunningOnKindInfra() bool {
 	return strings.HasPrefix(provider, "kind")
 }
 
-func SkipPVCTestIfRunnigOnKindInfra() {
-	if IsRunningOnKindInfra() {
-		Skip("Skip PVC tests till PR https://github.com/kubevirt/kubevirt/pull/3171 is merged")
-	}
+func IsRunningOnKindInfraIPv6() bool {
+	provider := os.Getenv("KUBEVIRT_PROVIDER")
+	return strings.HasPrefix(provider, "kind-k8s-1.17.0-ipv6")
```
maiqueb, line 4603: I don't understand this; isn't the issue that kind (the only infra where we can currently run ipv6 tests on) is docker on docker, and as such we get the same machine id on all k8s nodes ? Why add a new function about ipv6 on kind, instead of re-using https://github.com/kubevirt/kubevirt/pull/3221/files#diff-a5a178b4c66835110ceff63c61d28c96R4596 ?

tests/utils.go
```diff
@@ -757,6 +758,10 @@ func BeforeTestSuitSetup() {
 	CreatePVC(osWindows, defaultWindowsDiskSize, Config.StorageClassWindows)
 	CreatePVC(osRhel, defaultRhelDiskSize, Config.StorageClassRhel)
 
+	if IsRunningOnKindInfraIPv6() {
```
maiqueb, line 761: Why not use `IsRunningOnKindInfra` ? Isn't the issue here related to kind itself ?

pkg/virt-launcher/virtwrap/network/common.go
maiqueb: > its needed in order use IsIpv6Enabled in several places - but have one function
> it was groomed with Alona , and i need it in this PR please
> since its an integral part
> h isnt used and not needed (was groomed with Petr that its logical to create a standalone function), also the mock changes arent needed (once removed it from here, the mock generate removed them from the autogen files)

OK, makes sense - the function doesn't require any real HW behind, and as such doesn't need to be on the handler. I'm OK with it.

But anyway you look at this, it deserves to be on a separate commit, with that reasoning written in the commit msg. That commit would, of course, be part of this PR.

pkg/virt-launcher/virtwrap/manager.go
maiqueb: I have no idea what you meant here, but I'll clarify; I mean why aren't you re-using the variable you've created in https://github.com/kubevirt/kubevirt/pull/3221/files#diff-9b1d64950edb57271a1a2fc36192d464R410 ?

AFAIU, they both refer to `network.GetLoopbackAddress()`. Don't they ?

tests/utils.go
oshoval: i added to kubevirtci a need part that generates the files only for ipv6, this is the reason
in the future if we add it to all the other kind providers (i.e sriov)
this can be changed
mounting when file doesnt exists will break virt launcher

tests/utils.go
oshoval: since its not in the same func

tests/utils.go
oshoval: answered in a sep comment, since github had problems, changed parts according my understanding

pkg/virt-launcher/virtwrap/manager.go
oshoval: not same func

pkg/virt-launcher/virtwrap/manager.go
oshoval: not same func

pkg/virt-launcher/virtwrap/manager.go
oshoval: not same func, so var aint there

pkg/virt-launcher/virtwrap/network/common.go
oshoval: i think otherwise, and also as part of this task wont fix sriov migration which wasnted tested, as i understood, 
i did the changes i agreed with, every other change must be accepted by other reviewers, and be coordinated with Petr / Alona if they want me to extend the task scope
thanks

oshoval:
> machine-id

machine-id will need change of libvirt conf in people machines, agreed with Petr is not a good decision

oshoval:
will update the commit message to include all changes
and will make one var reuse twice.
but not atm , letting the tests finish first, since they started

tests/migration_test.go
```diff
@@ -856,7 +848,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 			var pvName string
 			var vmi *v1.VirtualMachineInstance
 			BeforeEach(func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
+				tests.SkipPVCTestIfRunnigOnKindInfra()
```
AlonaKaplan, line 851: AFAIU, this test is not solved by https://github.com/kubevirt/kubevirt/pull/3171. If so, please introduce a new skip message.

oshoval: thats right , but maybe 3171 is mandatory as well or all that will be needed once its done,
(now it stuck on VM not getting scheduled), so what you suggest please to say in the skip message ?

tests/migration_test.go
```diff
@@ -1063,6 +1055,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
 			It("[test_id:2227]should abort a vmi migration without progress", func() {
+				tests.IsRunningOnKindInfraIPv6()
```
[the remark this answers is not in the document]
oshoval: plese view the commit above it , this code is outdated
the reason is written there, this is the test where the stress isnt enough to kill the ipv6 migration, so skipping as what you suggested to do yesterday

thanks

tests/migration_test.go
AlonaKaplan: I suggest to create a new kubevirt issue telling the test fails and explaing what you know about it. And add a link to this issue in the skip. Similar to https://github.com/kubevirt/kubevirt/blob/master/tests/probes_test.go#L51

tests/migration_test.go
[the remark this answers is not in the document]
AlonaKaplan: Ignore, I see it was removed in the following commit.

tests/migration_test.go
```diff
@@ -1055,7 +1055,10 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
 			It("[test_id:2227]should abort a vmi migration without progress", func() {
-				tests.IsRunningOnKindInfraIPv6()
+				// stress isnt enough in kind provider to cause the migration to fail,
```
AlonaKaplan, line 1058: We don't know what is the root cause for the problem, kind/ipv6 or other. Please create a kubevirt issue (telling the test is not working and all the findings you have about it) and link the skip to the issue (you can ask @rmohr what's better - bugzilla or issue, I'm not sure).

oshoval:
i think that all the non passing tests are also flaky without this PR according 
what i saw in flake finder

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -29,7 +29,10 @@ import (
 	"strings"
 	"sync"
 
+	netutils "k8s.io/utils/net"
+
 	"kubevirt.io/client-go/log"
+	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/network"
```
SchSeba, line 35: It can be nice to have `network.IsIpv6Enabled()` in another place.
This import connect two completely different binaries

pkg/virt-launcher/virtwrap/network/network.go
```diff
@@ -133,3 +136,18 @@ func getNetworkClass(network *v1.Network) (NetworkInterface, error) {
 	}
 	return nil, fmt.Errorf("Network not implemented")
 }
+
+// IsIpv6Enabled returns if IPv6 is enabled
+func IsIpv6Enabled() bool {
+	podIP := os.Getenv("MY_POD_IP")
```
SchSeba, line 142: This is used by both the virt-handler and the virt-launcher so the question is they both have this environment variable?

tests/utils.go
```diff
@@ -4551,15 +4598,26 @@ func IsRunningOnKindInfra() bool {
 	return strings.HasPrefix(provider, "kind")
 }
 
+func IsRunningOnKindInfraIPv6() bool {
+	provider := os.Getenv("KUBEVIRT_PROVIDER")
+	return strings.HasPrefix(provider, "kind-k8s-1.17.0-ipv6")
```
SchSeba, line 4603: should this be more generic something like `kind` and `ipv6`?

pkg/virt-handler/migration-proxy/migration-proxy.go
vladikr: This `IsIpv6Enabled` is so small that you can easily reimplement it here (or in a common library perhaps pkg/utils) and you already have `netutils "k8s.io/utils/net"` imported here.

pkg/virt-launcher/virtwrap/network/network.go
vladikr: I'm also confused. What sets this variable?

tests/utils.go
oshoval: atm we added to kubevirtci the needed part that enable PodPreset and create the unique uuid file
that will be mounted, only for kind ipv6, for example not for sriov, according scope of this task.

tests/utils.go
oshoval: moreover the change need to be both on kci and here, as groomed before, because it was decided by Petr / Alona and even you at the first that kubevirt shouldnt be changed, but the change is part of the test
(kci cant do it alone, because it depends on the namespace name, which is decided, created and deleted in the tests themself)

pkg/virt-handler/migration-proxy/migration-proxy.go
oshoval: Hi, Thanks for your review,
so i took the time to check deeper and here are the results:
1. indeed pkg/virt-launcher/virtwrap/manager.go doesnt have it , i was mistaken because
pkg/virt-launcher/virtwrap/network/podinterface.go does have the var (not part of this PR),
(from what i was explained ,since Ihar change
makes the virt-handler runs code of virt-launcher in its network namespace, this is how it works, right ?)

2. pkg/virt-handler/migration-proxy/migration-proxy.go does have the VAR
3. eliminating both of this places which actually mux between [::] and 0.0.0.0 , the migration would still work,
so action items

1. i will remove the wrong place (1 above)
2. should i remove also the place which is right but isnt needed ? 
is there a case where the kernel will enforce ipv6 only and then 0.0.0.0 wont be usable ?

thanks

pkg/virt-launcher/virtwrap/network/network.go
oshoval: please view the comment in the other thread which also refer to this question

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -430,7 +430,7 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
 		}
 
 		// For a tunnelled migration, this is always the uri
-		dstUri := fmt.Sprintf("qemu+tcp://127.0.0.1:%d/system", LibvirtLocalConnectionPort)
+		dstUri := fmt.Sprintf("qemu+tcp://%s:%d/system", "127.0.0.1", LibvirtLocalConnectionPort)
```
oshoval, line 433: i can revert this if needed

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -407,12 +407,11 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
 
 		isBlockMigration := (vmi.Status.MigrationMethod == v1.BlockMigration)
 		migrationPortsRange := migrationproxy.GetMigrationPortsList(isBlockMigration)
-		loopbackAddress := network.GetLoopbackAddress()
 
 		// Create a tcp server for each direct connection proxy
 		for _, port := range migrationPortsRange {
 			key := migrationproxy.ConstructProxyKey(string(vmi.UID), port)
-			migrationProxy := migrationproxy.NewTargetProxy(loopbackAddress, port, nil, nil, migrationproxy.SourceUnixFile(l.virtShareDir, key))
+			migrationProxy := migrationproxy.NewTargetProxy("127.0.0.1", port, nil, nil, migrationproxy.SourceUnixFile(l.virtShareDir, key))
```
AlonaKaplan: @oshoval can you please explain how it works for ipv6?

pkg/virt-handler/migration-proxy/migration-proxy.go
AlonaKaplan: > This `IsIpv6Enabled` is so small that you can easily reimplement it here (or in a common library perhaps pkg/utils) and you already have `netutils "k8s.io/utils/net"` imported here.

virt-handler has two places using 'IsIpv6Enabled' -
pkg/virt-launcher/virtwrap/network/podinterface.go &
pkg/virt-handler/migration-proxy/migration-proxy.go

IMO, we should reuse the same method in both of the places and not re-implement exactly the same code twice.
Since currently (unfortunately), the code under virt-launcher directory is a mixture of a code that is running on virt-launcher and code that is running on virt-handler I'm not sure the import is connecting completely different binaries.

@SchSeba @vladikr in what common place do you suggest to put it? Since the code is relevant to virt-handler only, I'm not sure pkg/utils is the best place.

oshoval: please consider the following when deciding:
1. even the addon at migration proxy, isnt mandatory , i mean that as i wrote before,
removing it , wont cause migration to fail under current circumstances.

2. having it in a common place would be prune to errors, of usage of it in places where the MY_POD_IP doesnt exists
since its a one liner i would suggest to leave it reimplemented or remove it totally since without it, migration works as well

pkg/virt-launcher/virtwrap/manager.go
oshoval: i believe that since ipv4 isnt disabled, the 127.0.0.1 and 0.0.0.0 can act as local listening
for all ports, both for ipv6 and ipv4

oshoval: https://serverfault.com/questions/256811/how-to-completely-dynamically-disable-ipv4-stack-from-linux
see last remark  for example please

oshoval:
some new flakes appear (flake also without this PR of course i mean)
this test doesnt appear in flakefinder, but 
the reason is as other flaky test (1783) say
"failed to invoke qemu-img: open /dev/null: operation not permitted"
will ask Daniel if his PR of chmod solve it (not sure it does because its for OpenShift IIUC)

[rfe_id:151][crit:high][vendor:cnv-qe@redhat.com][level:component]CloudInit  UserData  [rfe_id:151][crit:medium][vendor:cnv-qe@redhat.com][level:component]A  new VirtualMachineInstance with cloudInitConfigDrive userDataBase64  source with injected ssh-key [test_id:3178]should have ssh-key under  authorized keys expand_less | 53s
-- | --
/root/go/src/kubevirt.io/kubevirt/tests/vmi_cloudinit_test.go:185 Unexpected Warning event received: testvmitj5n2wcj9rwzvcnzxx6k6s7rvjgvmxm2tbr6gx7gkchpwrwt,6d6892c1-4a97-450d-a5d6-cf1ce3d22e2b: server error. command SyncVMI failed: "generating local cloud-init data failed: open /dev/null: operation not permitted" Expected     <string>: Warning not to equal     <string>: Warning /root/go/src/kubevirt.io/kubevirt/tests/utils.go:391

rerunning lane to see if its consistent
/test pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6

pkg/virt-launcher/virtwrap/manager.go
oshoval: https://github.com/nodejs/node/issues/9390
0.0.0.0 works as long as bindv6only is false

127.0.0.1 works if /etc/hosts have it pointing to 127.0.0.1
https://eklitzke.org/localhost

pkg/virt-handler/migration-proxy/migration-proxy.go
oshoval: btw both virt-launcher and virt-handler pods have
KUBERNETES_SERVICE_HOST
according 
https://github.com/kubernetes/kubernetes/issues/40973
they are mandatory
their value is an IPv6 address of the cluster service
can we then just use in all places this env var in order to determine the address to be used ?
i mean both in podInterface (which was before me)
and also for virt-launcher to determine between 127.0.0.1 and ::1 
(in order to support cases where just one of them would be in the /etc/hosts and not both)
and for virt-handler to support cases where bindv6only will be true ?

oshoval:
@vladikr 
@AlonaKaplan 
@SchSeba 
@maiqueb 

if you can review please the answers, according your leads i researched better thanks
and there are some decision that should be made so we can continue and wrap it please.
wrote all the info in the comments, in case its not ordered enough would summarize.

thanks

oshoval:
will move the PodPreset to a separated PR.
in case its desired that we fix it for all providers, please let me know, it will extend a bit the current task.

oshoval:
Thanks Alona and Vladik
to summarize the changes that were groomed and pushed today
use 
	res, err := exec.Command("cat", "/proc/sys/net/ipv6/conf/default/disable_ipv6").Output()
to determine if ipv6 is not disabled and then use ::1 / :: which works for both ipv6 and ipv4
else use 127.0.0.1 / 0.0.0.0

Running all lanes to see it doesnt hurt anything

@AlonaKaplan 
@maiqueb 
can you please review it ?

this is the lane of ipv6 (still running)
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3221/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1257699018016821248

pkg/virt-launcher/virtwrap/manager.go
maiqueb: I'm very confused by this answer.

> [nodejs/node#9390](https://github.com/nodejs/node/issues/9390)
> 0.0.0.0 works as long as bindv6only is false
> cat /proc/sys/net/ipv6/bindv6only
> 
> 127.0.0.1 works if /etc/hosts have it pointing to localhost
> https://eklitzke.org/localhost

Correct me if I'm wrong, but this means that for it (migration) to work, we depend on a certain `/etc/hosts` configuration. That is not good, imho. What are the advantages of this approach ? 

> 
> both virt-launcher and virt-handler have it 0 and with localhost pointed by 127.0.0.1
> in kind ipv6.
> 
> in virt-launcher we would always have KUBERNETES_SERVICE_HOST
> because its a mandatory var for k8s,
> so we can use this instead of MY_POD_IP in case we want to mux between 127.0.0.1 and ::1 for cases where localhost wont have this entry, or enforce that it will have it
> wdyt ? if it can be done on different PR it would be awesome, since i really prefer to wrap this one. flakiness happens a lot and make it longer each time, another reason is that for example for the new ipv6 provider we are dual stack, we dont directly support bindv6only. sometimes coming fresher to a task make it goes smoother, and we learn from each task more.
> (for now i removed the code that looked on MY_POD_IP in the virt-launcher because it wasnt right)

pkg/util/util.go
```diff
@@ -29,3 +31,26 @@ func IsGPUVMI(vmi *v1.VirtualMachineInstance) bool {
 	}
 	return false
 }
+
+func isIpv6Disabled() bool {
```
maiqueb, line 35: I think it's more readable when positive logic is used - i.e. I would have a `isIpv6Enabled` function instead.

This is subjective though.

pkg/util/util.go
```diff
@@ -29,3 +31,26 @@ func IsGPUVMI(vmi *v1.VirtualMachineInstance) bool {
...
+	res, err := exec.Command("cat", "/proc/sys/net/ipv6/conf/default/disable_ipv6").Output()
```
maiqueb, line 36: Please use a meaningful variable name - `res` can be anything. `ipv6Enabled` / `ipv6Disabled` clearly indicate what they are.

pkg/virt-launcher/virtwrap/manager.go
oshoval: i was wrong here
the summary that was been groomed with Alona and Vladik was that as ovirt was doing, 
we are going to use ::1 / :: unless ipv6 is disabled

please see the 3rd commit, as you already did from what you told me

tests/migration_test.go
```diff
@@ -1079,6 +1071,8 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
 			It("[test_id:2227]should abort a vmi migration without progress", func() {
+				tests.SkipStressTestIfRunnigOnKindInfraIPv6()
```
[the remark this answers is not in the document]
oshoval: for kind the stress isnt enough to make the migration fail 
the stress tries to dirty ram pages, in order to fail the migration.
it was approved by Alona and Vladik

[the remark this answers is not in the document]
maiqueb: Then, if the issue happens for all kind providers, why skip it for kind ipv6 ? It is misleading; readers will think it has to do with ipv6 on kind, while it does not. 

I think a comment above the skip with that reasoning (or better yet, have this in an isolated commit with that reasoning in the commit message) is necessary. That is not obvious. At all.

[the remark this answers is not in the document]
oshoval: even that we have ? (and issue with all the info, this is what Alona asked before)
https://github.com/kubevirt/kubevirt/pull/3221/files#diff-a5a178b4c66835110ceff63c61d28c96R4639

there arent other kind providers that run it / fail on it
so its right to skip for all ? i am not sure
i dont want to close gates in future or past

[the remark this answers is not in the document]
maiqueb: > even that we have ? (and issue with all the info, this is what Alona asked before)
https://github.com/kubevirt/kubevirt/pull/3221/files#diff-a5a178b4c66835110ceff63c61d28c96R4639

Oh, didn't follow through. Since there's an opened issue detailing it, no need to add comment / split commit. Sorry for misleading, should have looked more carefully.

> there arent other kind providers that run it / fail on it
so its right to skip for all ? i am not sure
i dont want to close gates in future or past

I honestly think if the issue happens in all kind providers, it should be skipped in all kind providers. The fact that the test is only executed on the kind ipv6 provider is besides the point, and something people reading the code will not know - unless they go dig the automation script. Which, by the way, is not mandatory to execute tests - I don't use it locally, for instance.

Test code also serves as documentation; whoever reads the test code (and fails to see your issue & read it) will get the feeling the issue is related to ipv6. That's what I'm trying to avoid.

What's the advantage of skipping the test only for kind ipv6, when, according to the [issue](https://github.com/kubevirt/kubevirt/issues/3323), it is orthogonal to all kind providers ? I've stated the disadvantages above.

pkg/util/util.go
oshoval: just because this is the name of the var, and can be mistaken to think that ipv6 is used if 
the word enabled appear, i prefer this method if it ok (as you said subjective),
else i would change to positive logic

tests/migration_test.go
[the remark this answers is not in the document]
oshoval: done at 
https://github.com/kubevirt/kubevirt/pull/3379
which will be either pushed here, or we move there
just made an ordered one, and not directly pushed to here,
to see i dont have mistakes

pkg/util/util.go
oshoval: done at https://github.com/kubevirt/kubevirt/pull/3379, see reason before as well please

oshoval:
closed in favour of splitting to separated PRs

PR #3222, opened 2020-03-31
Allow developer to 'make push' subset of images

mlsorensen:

Some developer workflows (specifically those that target non-local registries/development environments) can benefit from the efficiency of not pushing "other-images" every time. This PR allows developers to set "PUSH_TARGETS" environment variable to override the list of push targets used during "make push".

It defaults to the original list of targets, so general behavior should not change.

**Release note**:

```release-note
NONE
```


oshoval:
Thanks , its great and can save time
for example when syncing an external remote cluster
for Travis need to rebase please i think,
the 1.15-ceph failing test is known to be flaky

/lgtm

rmohr:
Thanks!

dhiller:
Sound like a really useful feature, so maybe it would be wortwhile to add this to the developer documentation somewhere?

PR #3223, opened 2020-03-31
Explicitly test packages in case of error

dhiller:

We found out that `go test -cover` seems to swallow the error output (in this case it was 
`FAIL    kubevirt.io/kubevirt/pkg/virt-api/webhooks/validating-webhook/admitters [build failed]`
). In order to keep the output still helpful we test each package again in case of error to have at least a hint for the user what to do about this.

Fixes #

/cc @rmohr @phoracek mind taking a look whether you find is helpful? Alternative ideas welcome :)

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/coverage.sh
+++ b/hack/coverage.sh
@@ -5,5 +5,23 @@ set -e
 path=${1:-./pkg/...}
 profile=.coverprofile
 
-go test -cover -v -coverprofile=$profile $(go list ${path})
+packages=$(go list ${path})
+
+# in rare cases go test with coverage seems to swallow test
+# failure output, so we keep the exit code
+set +e
+go test -cover -v -coverprofile=$profile ${packages}
+return_value=$?
+
+# in case of exit code we explicitly test
+# each package again to have the output of all
+# failing tests
+if [[ $return_value -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


dhiller:
FYI @danielBelenky @jean-edouard

rmohr:
@dhiller so far I always saw my failing tests. Did you locally run the tests also with `-v`? Otherwise the output is swallowed. I agree that the failing tests (especially if they don't compile) are hard to see in travis, but they should be there.

dhiller:
> @dhiller so far I always saw my failing tests. Did you locally run the tests also with `-v`? Otherwise the output is swallowed. I agree that the failing tests (especially if they don't compile) are hard to see in travis, but they should be there.

Actually I just overlloked the error in the build log: https://travis-ci.org/github/kubevirt/kubevirt/jobs/669180603#L1623

So it's all there. I don't think it makes sense to change anything. Although I'd have this line more prominent. Do you know if there's a way to customize travis build log with error patterns or the like?

rmohr:
I don't think so. We could probably do something like `tee` the output, store interesting patterns in an extra file and then `cat` the file after the test run.

PR #3225, opened 2020-04-01
Fix travis ppc64le lane

maya-r:
```release-note
NONE
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ services:
 
 arch:
 - amd64
-# - ppc64le # temporary disable until we have it under control again
+- ppc64le
 
 os: linux
 
--- a/.bazelrc
+++ b/.bazelrc
@@ -11,6 +11,6 @@ build:x86_64 --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64_cgo
 run:x86_64 --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64_cgo
 test:x86_64 --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64_cgo
 
-build:ppc64le --platforms=@io_bazel_rules_go//go/toolchain:linux_ppc64le_cgo
-run:ppc64le --platforms=@io_bazel_rules_go//go/toolchain:linux_ppc64le_cgo
-test:ppc64le --platforms=@io_bazel_rules_go//go/toolchain:linux_ppc64le_cgo --host_javabase=@local_jdk//:jdk
+build:ppc64le --host_javabase="@local_jdk//:jdk" --platforms=@io_bazel_rules_go//go/toolchain:linux_ppc64le_cgo
+run:ppc64le --host_javabase="@local_jdk//:jdk" --platforms=@io_bazel_rules_go//go/toolchain:linux_ppc64le_cgo
+test:ppc64le --host_javabase="@local_jdk//:jdk" --platforms=@io_bazel_rules_go//go/toolchain:linux_ppc64le_cgo


maya-r:
Seems to pass! :tada:

danielBelenky:
@maya-r very nice!
Shouldn't we install a JDK on the host? how it event exists there now? :thinking:

hack/bazel-build.sh
```diff
@@ -28,6 +28,7 @@ mkdir -p ${CMD_OUT_DIR}/dump
 
 # Build all binaries for amd64
 bazel build \
+    --host_javabase="@local_jdk//:jdk" \
```
rmohr, line 31: This seems to improve things drastically. The choice is also good, because it uses the embedded jdk from bazel (again).

Can you however instead of putting it here, add to `.bazelrc`? So that we pick this setting transparently up?

maya-r: oh, that's a much nicer place!

.travis.yml
```diff
@@ -16,12 +16,12 @@ install:
 
 script:
 - make generate
-- if [[ -n "$(git status --porcelain)" ]] ; then echo "It seems like you need to run
+- if [[ -n "$(git status --porcelain)" && $TRAVIS_CPU_ARCH == "amd64" ]] ; then echo "It seems like you need to run
```
rmohr, line 19: I see that you wrote that this causes differences. That is explicitly not what we want. What are the differences? They should result in predictable results.

oshoval: looks like it creates files in cluster-up that it doesnt need to
its not the current problem maybe, but generally can cause headache i think ?
cluster-up/cluster/kind-k8s-sriov-1.17.0/csrcreator/BUILD.bazel

rmohr: This is suspicious, yes.

maya-r:
I was having a hard time parsing why I had that `make generate` failure (I misread the output before).  it looks like it's passing without that hack now.

maya-r:
The travis build is "pending", but clicking on the details link, the build passed.

rmohr:
let me re-trigger travis

rmohr:
@ murilofv FYI that seems to fix everything.

rmohr:
@murilofv ^^

murilofv:
I still think the keep alive is needed to run for the whole build, looking at the re-triggered build it seems to be the case, it failed with the same error: `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`

The --host_javabase is needed because commit 7b4deb2 ("Add a basic unit test to verify the
container-binary") ended up breaking the build in Power, without it we now start to see the following error: 
```
Configurable attribute "actual" doesn't match this configuration: Could
not find a JDK for host execution environment, please explicitly provide
one using `--host_javabase.` 
```

So we need to at least add --host_javabase to the `build:ppc64le` .bazelrc , I don't think the run so far is necessary, but I don't see any harm in adding it as well.

murilofv:
With the keep alive in the whole build, like in this patch ( https://github.com/murilofv/kubevirt/commit/aff88fbae68f6deecdda377be3ee8258d13aaf7a ) I don't see the issue anymore of it dying due to travis CI not getting any output, I have never gotten any actual stalls as well in over 100 runs.

I sent you a note in Slack, but let me add it here as well, in case someone has any insight of what might be going on. The one issue I've been occasionally getting is that `git status --porcelain` step fails and I can’t understand why would that be the case if the previous step completed successfully, there are runs where I don’t see it complain and ones that I do see, for example:
https://travis-ci.com/github/murilofv/kubevirt/builds/158851093#L1090 -> Working
https://travis-ci.com/github/murilofv/kubevirt/builds/158712840#L1153 -> Not working

The only odd thing I noticed when comparing both is in the not working in line 447 there was the error: `gazelle: finding module path for import github.com/cloudflare/cfssl/csr: exit status 1: go: finding github.com/cloudflare/cfssl/csr latest`. So if anyone has any insight on what might be going on I'd appreciate it. 

I can send out the patch for the keep alive and re enable the CI in Power, but I am scared this will be the issue that folks will start hitting occasionally.

maya-r:
/hold
let the bot take a break, we will probably want #3273 merged instead of this PR, but not closing the pull request until it is merged.

murilofv:
> /hold
> let the bot take a break, we will probably want #3273 merged instead of this PR, but not closing the pull request until it is merged.

Do you want to make this one about fixing the build in ppc64le? So ditch the reenabling of the CI for now and just address the ppc64le build issue of the needed --host_javabase added in the .bazelrc and we can work on the CI part on the other PR?

maya-r:
Please keep going with your PR as-is, keeping this PR not-closed is more of a reminder to myself to have a look that things are fixed in a few days.

Right now the bot is constantly re-invoking the tests every few hours which is probably unnecessary :-)

maya-r:
Closing this on further thought. Keeping a pull request opened is probably creating weird effects for CI.

PR #3238, opened 2020-04-02
Increase the amount of memory used by stress to delay migrations

vladikr:
Some migrations related functional rests are failing randomly on the 1.17 lane. This is probably because stress doesn't have enough time to pollute the guest's memory. This change increases the amount of memory stress uses instead of increasing the time.

```release-note
NONE
```

### DIFF
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -328,7 +328,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 	runStressTest := func(expecter expect.Expecter) {
 		By("Run a stress test to dirty some pages and slow down the migration")
 		_, err = expecter.ExpectBatch([]expect.Batcher{
-			&expect.BSnd{S: "stress --vm 1 --vm-bytes 600M --vm-keep --timeout 1600s&\n"},
+			&expect.BSnd{S: "stress --vm 1 --vm-bytes 800M --vm-keep --timeout 1600s&\n"},
 		}, 15*time.Second)
 		Expect(err).ToNot(HaveOccurred(), "should run a stress test")
 		// give stress tool some time to trash more memory pages before returning control to next steps


rmohr:
Unrelated to your PR, but apparently there was no `/dev/kvm` available just for one test:

```
/root/go/src/kubevirt.io/kubevirt/tests/migration_test.go:1195
Unexpected Warning event received: testvmiwr8gwwqbcckdfs4vk4ddwh28hrl4j2r2gndpd5b5jgskvkxb,3edf3a02-a19d-473d-aafc-06ac96cf251e: server error. command SyncVMI failed: "LibvirtError(Code=67, Domain=10, Message='unsupported configuration: Domain requires KVM, but it is not available. Check that virtualization is enabled in the host BIOS, and host configuration is setup to load the kvm modules.')"
Expected
    <string>: Warning
not to equal
    <string>: Warning
/root/go/src/kubevirt.io/kubevirt/tests/utils.go:380
```

rmohr:
/test pull-kubevirt-e2e-k8s-1.17

Let's create a little bit more data.

tests/migration_test.go
```diff
@@ -1155,7 +1155,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				table.Entry("[test_id:2226]with ContainerDisk", newVirtualMachineInstanceWithFedoraContainerDisk),
 				table.Entry("[test_id:2731]with OCS Disk", newVirtualMachineInstanceWithFedoraOCSDisk),
 			)
-			It("[test_id:3241]should be able to cancel a migration right after posting it", func() {
+			FIt("[test_id:3241]should be able to cancel a migration right after posting it", func() {
```
rmohr, line 1158: `FIt` does not work if  a `--skip` or `--focus` is added to the test execution and that is what we do in `automation/test.sh`.

vladikr:
@rmohr 1.17 passes with this change.

PR #3241, opened 2020-04-02
docs: clarify what will be reported in VMI status IPs

phoracek:

It was not clear which IPs should be reported in the status. This doc change
should make it clear and we should be able to adjust the existing code to follow
this.

Fixes #

This should unblock @vatsalparekh on #3063

**Release note**:

```release-note
NONE
```

### DIFF
--- a/staging/src/kubevirt.io/client-go/api/v1/types.go
+++ b/staging/src/kubevirt.io/client-go/api/v1/types.go
@@ -318,15 +318,21 @@ func (m *VirtualMachineInstanceMigration) TargetIsHandedOff() bool {
 //
 // +k8s:openapi-gen=true
 type VirtualMachineInstanceNetworkInterface struct {
-	// IP address of a Virtual Machine interface. It is always the first item of
-	// IPs
+	// (deprecated) IP address of a Virtual Machine interface. This will show
+	// the first IP from IPs list.
 	IP string `json:"ipAddress,omitempty"`
 	// Hardware address of a Virtual Machine interface
 	MAC string `json:"mac,omitempty"`
 	// Name of the interface, corresponds to name of the network assigned to the interface
 	// TODO: remove omitempty, when api breaking changes are allowed
 	Name string `json:"name,omitempty"`
-	// List of all IP addresses of a Virtual Machine interface
+	// List of all IP addresses of a Virtual Machine interface accessible on
+	// given network.
+	// For L2 binding mechanisms (bridge, sriov), this would be addresses
+	// assigned by CNI IPAM. In case IPAM was not set, this would be addresses
+	// reported by guest agent (if the guest agent is available in the guest).
+	// For forwarding binding mechanisms (masquerade, slirp), this would be
+	// always addresses assigned by CNI IPAM.
 	IPs []string `json:"ipAddresses,omitempty"`
 	// The interface name inside the Virtual Machine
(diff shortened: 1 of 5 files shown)


staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -302,14 +302,19 @@ func (m *VirtualMachineInstanceMigration) TargetIsHandedOff() bool {
...
 	// Hardware address of a Virtual Machine interface
 	MAC string `json:"mac,omitempty"`
 	// Name of the interface, corresponds to name of the network assigned to the interface
 	// TODO: remove omitempty, when api breaking changes are allowed
 	Name string `json:"name,omitempty"`
-	// List of all IP addresses of a Virtual Machine interface
+	// List of all IP addresses of a Virtual Machine interface accessible on
+	// given network. For NAT-based binding mechanisms (masquerade), this will
+	// be the PodIP assigned by CNI. For L2 binding mechanisms (sriov, bridge,
+	// slirp), this would be either IP addresses observed on the interface via
```
rmohr, line 315: A few things:

I don't think that it makes sense to report IP addresses from slirp. Only the CNI IP makes sense.

I think that whenever we get an IP from CNI,  CNI must be the only point of truth and the guest agent needs to be ignored. At the end we mostly care about the IP from a k8s reachability perspective (so that people can e.g. build custom services on top of VMIs) and not about the real IP inside the VM.

When we finally add readiness Probe support for the guest agent we may even take IP mismatches into account.

phoracek: > I don't think that it makes sense to report IP addresses from slirp. Only the CNI IP makes sense.

Slirp is doing forwarding on L2, isn't it? So it can be used without any IPAM and IPs can be later obtained by a dhcp client running inside the guest. Correct me if I'm wrong.

> I think that whenever we get an IP from CNI, CNI must be the only point of truth and the guest agent needs to be ignored.

This would be indeed a good security mechanism. But that would require us to have the probe in place connected with IP spoofing filter (we cannot trust guest agent).

> At the end we mostly care about the IP from a k8s reachability perspective (so that people can e.g. build custom services on top of VMIs) and not about the real IP inside the VM.

Since the secondary networks are not exposed in PodIP, but only in an annotation set by Multus, users that need to expose a service on the secondary networks have to create Endpoints manually. And for that, it does not make a difference if they take IPs which are based on guest-agent or CNI IPAM, doest it?

phoracek: Update: We discussed it online with Roman and came to conclusion that if CNI IPAM is used, we should always respect that. The administrator explicitly gave some IPs to the user and those should be used, nothing else. In case there is no IPAM, the user is responsible of configuring the network and we help them by exposing guest agent IPs in the status.

In a follow-up, we should add an IP spoofing filter that would enforce that the guest is only communicating through IPs assigned by IPAM.

rmohr: > In a follow-up, we should add an IP spoofing filter that would enforce that the guest is only communicating through IPs assigned by IPAM.

Perfect. Lets' ensure that we create an issue for that.

phoracek: https://github.com/kubevirt/kubevirt/issues/3242

PR #3248, opened 2020-04-03
Open and close /dev/vhost-net on virt-handler startup

jean-edouard:
Similarly to /dev/net/tun, opening vhost-net for the first time triggers a modprobe of the vhost_net module.
virt-launcher sometimes uses vhost-net, but is not allowed to modprobe since it runs under the SELinux type container_t.
In virt-handler, which runs as spc_t, access (open/close) /dev/vhost-net and (potentially) trigger the modprobe.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-handler/device-manager/device_controller.go
+++ b/pkg/virt-handler/device-manager/device_controller.go
@@ -46,9 +46,9 @@ type DeviceController struct {
 func NewDeviceController(host string, maxDevices int) *DeviceController {
 	return &DeviceController{
 		devicePlugins: []GenericDevice{
-			NewGenericDevicePlugin(KVMName, KVMPath, maxDevices),
-			NewGenericDevicePlugin(TunName, TunPath, maxDevices),
-			NewGenericDevicePlugin(VhostNetName, VhostNetPath, maxDevices),
+			NewGenericDevicePlugin(KVMName, KVMPath, maxDevices, false),
+			NewGenericDevicePlugin(TunName, TunPath, maxDevices, true),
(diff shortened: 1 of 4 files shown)


rmohr:
Since you are touching that area again, would you mind moving that startup code to our device manager package?

jean-edouard:
@rmohr done, please review

pkg/virt-handler/device-manager/device_controller.go
```diff
@@ -65,8 +65,18 @@ func (c *DeviceController) nodeHasDevice(devicePath string) bool {
 func (c *DeviceController) startDevicePlugin(dev GenericDevice, stop chan struct{}) {
 	logger := log.DefaultLogger()
 	deviceName := dev.GetDeviceName()
+	devicePath := dev.GetDevicePath()
 	retries := 0
 
+	// Making sure the tun and vhost-net modules are loaded on the host
+	// Opening and closing the corresponding dev nodes will trigger modprobes if necessary
+	if deviceName == TunName || deviceName == VhostNetName {
```
rmohr, line 73: Would it make sense to add a parameter to `NewGenericDevicePlugin` which indicates that it should pre-open the device? Right now that we have to pre-open is even the "default" case. It might be worth reflecting the need when creating the plugin for the specific device instead of implicit additional checks.

jean-edouard: Pushed, how does that look?

rmohr: Perfect.

PR #3252, opened 2020-04-05
Migration tests - use proper node drain key

enp0s3:
In order to test a migration in case of node drain, the migration tests were using a hardcoded taint key "kubevirt.io/drain" for tainting the node and trigger VMI eviction. This taint key is configurable and is read from the KubeVirt configmap by the virt-handler evacuation controller. Thus, the tests also should read this key from the configmap and not assume a constant value.

In some cases the node drain key is being customized (One good example is the HCO which builds the KV configmap and uses drain key "node.kubernetes.io/unschedulable"). This PR will fix potential issues with running functional tests.

Fixes #
Live migration tests

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -48,6 +48,7 @@ import (
 	"kubevirt.io/kubevirt/pkg/certificates/triple"
 	"kubevirt.io/kubevirt/pkg/certificates/triple/cert"
 	migrations "kubevirt.io/kubevirt/pkg/util/migrations"
+	virtconfig "kubevirt.io/kubevirt/pkg/virt-config"
 	"kubevirt.io/kubevirt/tests"
 )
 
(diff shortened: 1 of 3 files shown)


tests/migration_test.go
```diff
@@ -1356,7 +1353,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 					// Taint Node.
 					By("Tainting node with kubevirt.io/drain=NoSchedule")
```
omeryahud, line 1354: `kubevirt.io/drain` might not be the taint key anymore, maybe you should change that for better readability in the test log

danielBelenky:
@enp0s3 nice fix.
Can you please run `make generate` and commit the changes? This is the reason why Travis fails.

rmohr:
Has a merge conflict

rmohr:
/lgtm

@enp0s3 we tend to not create relese notes when we refactor tests.

PR #3255, opened 2020-04-06
Configurable increase for test timeouts

stu-gott:
This introduces a configurable timeout for the test suite. This is useful in slow environments--our tests have hard coded timeouts, and there's always a counterexample.
**Release note**:
```release-note
NONE
```


rmohr:
@stu-gott I am not sure if we want that. At a certain point one just has to admit that the cluster does not have enough resources to work properly. We should have right now already very very generous timeouts on almost all locations.

stu-gott:
Perhaps. I would note that the vmi_lifecycle tests identified in this PR are particularly unforgiving

stu-gott:
This doesn't appear to be a universally supported idea. Closing.

Issue #3041, opened 2020-01-31
Support changing scheduler name + allowing live migrations on PVCs with multiple node replicas.


azaiter:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>

/kind enhancement

**What happened**:
Can't set `schedulerName` for any `virt-launcher` pods from either VM or VMI resource types. Adding the option to do so will allow the usage of some storage-aware schedulers such as stork to only schedule to nodes that has the block replicas of the PVCs, and that gives kubevirt node failover advantage when using portworx as storage provider. Going that route will allow live migrations without needing to have shared PVCs (it's the current condition that triggers live migration flag) as shared PVCs have higher IO overhead than the PVCs node replicas.

**What you expected to happen**:
**How to reproduce it (as minimally and precisely as possible)**:
**Anything else we need to know?**:

**Environment**:
- KubeVirt version (use `virtctl version`): v0.25.0
- Kubernetes version (use `kubectl version`): v1.16.2
- VM or VMI specifications: both
- Cloud provider or hardware configuration: on prem, portworx as storage provider, stork as scheduler.
- OS (e.g. from /etc/os-release): RancherOS and CentOS
- Kernel (e.g. `uname -a`): 4.12
- Install tools: n/a
- Others: n/a

rmohr:
Allowing to set the scheduler name should not be an issue. We can just add the scheduler field to our VMI spec and pass it through, to the pod.

Allowing live migrations with replicas sounds nice. Is it somehow visible on the PVC that it has node replicas? Could you point us to some documentation?

azaiter:
Great.

As for your replicas question, setting up the replicas would be done in the [storage class that the PVC is created from (`parameters.repl`)](https://kubernetes.io/docs/concepts/storage/storage-classes/#portworx-volume). The PVC has a `spec.storageClassName` key that points to the storage class that has the replicas count.

rmohr:
@azaiter #3258 should add the first step. Regarding to the PVCs. We would end up with PVCs which are readwriteonce, but where the scheduler can schedule pods to different hosts based on the replica count?

azaiter:
@rmohr Let's take a simple 3 storage/compute nodes topology (A, B, and C):
if the PVC that is tied to a storage class has repl value of 2, two of the above nodes will have the volume replicated to them and synced at all times, say A and C.

[...]

PR #3258, opened 2020-04-07
Support setting custom schedulers on VMIs

rmohr:

Some people have custom schedulers. We can support them by adding a `schedulerName` field in the VMI and passing it to the launcher pod.

Related to #3041

**Release note**:
```release-note
Allow setting custom schedulers on VMI specs
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -6948,6 +6948,10 @@
       "description": "Periodic probe of VirtualMachineInstance service readiness.\nVirtualmachineInstances will be removed from service endpoints if the probe fails.\nCannot be updated.\nMore info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes\n+optional",
       "$ref": "#/definitions/v1.Probe"
      },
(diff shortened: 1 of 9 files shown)


staging/src/kubevirt.io/client-go/api/v1/defaults.go
```diff
@@ -116,6 +117,10 @@ func SetDefaults_VirtualMachineInstance(obj *VirtualMachineInstance) {
 		obj.Spec.Domain.Features = &Features{}
 	}
 
+	if obj.Spec.SchedulerName == "" {
+		obj.Spec.SchedulerName = v1.DefaultSchedulerName
```
vladikr, line 121: So far we didn't set the schedulerName on the pod spec, perhaps we should continue and set it only if a specific scheduler name was provided? Can't there be an admission controller or something else that would set a non-default scheduler name if nothing is set?

rmohr: Good point. Fixed.

rmohr:
@vladikr would you mind doing another review?

PR #3263, opened 2020-04-07
Use a properly SELinux-labelled location for the storage tests

jean-edouard:
/tmp is labelled tmp_t, which containers (labelled container_t) can't write to.
/var/run/kubevirt is also a tmpfs location, but it has the label we want (container_file_t).
Switching to that.

Fixes #

Should fix at least test 3107 on the 1.17 test lane.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/storage_test.go
+++ b/tests/storage_test.go
@@ -324,7 +324,7 @@ var _ = Describe("Storage", func() {
 
 			Context("With a HostDisk defined", func() {
 
-				const hostDiskDir = "/tmp/kubevirt-hostdisks"
+				const hostDiskDir = "/var/run/kubevirt/kubevirt-hostdisks"
 				var nodeName string
 
 				BeforeEach(func() {


tests/storage_test.go
```diff
@@ -324,7 +324,7 @@ var _ = Describe("Storage", func() {
 
 			Context("With a HostDisk defined", func() {
 
-				const hostDiskDir = "/tmp/kubevirt-hostdisks"
+				const hostDiskDir = "/var/run/kubevirt/kubevirt-hostdisks"
```
rmohr, line 327: This one may be properly labeled but it has a special purpose and should not be used to place arbitrary files inside it.

jean-edouard:
Replaced by https://github.com/kubevirt/kubevirtci/pull/340 and https://github.com/kubevirt/kubevirt/pull/3276

PR #3268, opened 2020-04-07
Bump libvirt to 6.0.0

phoracek:

In order to get latest bug fixes and features of libvirt, we should bump from
5.0 to 6.0. It would allow us to offload some of the tasks that are done in
virt-launcher to virt-handler and improve the overall security.

Fixes #

* This is WIP testing changes made in https://github.com/kubevirt/libvirt/pull/48

**Release note**:

```release-note
Upgrade to libvirt 6.0.0.
```

### DIFF
--- a/tests/vmi_lifecycle_test.go
+++ b/tests/vmi_lifecycle_test.go
@@ -165,7 +165,7 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
 			Eventually(logs,
 				2*time.Second,
 				500*time.Millisecond).
-				Should(And(ContainSubstring("internal error: Unable to get DBus system bus connection"), ContainSubstring(`"subcomponent":"libvirt"`)))
+				Should(And(ContainSubstring("At least one cgroup controller is required: No such device or address"), ContainSubstring(`"subcomponent":"libvirt"`)))
 		})
 
 		It("[test_id:3197]should log libvirtd debug logs when enabled", func() {
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -323,15 +323,16 @@ container_pull(
 # Pull base image libvirt
 container_pull(
     name = "libvirt",
-    digest = "sha256:cfff1e8729a7466581ba71d953b132beaef28f61b330f52caef6d0184e5f9af0",
+    digest = "sha256:584c7b49f1615e3f229668d530a79945ebf7f3afbab4ec554187e9ba2b754008",
     registry = "index.docker.io",
     repository = "kubevirt/libvirt",
-    #tag = "tmp.20191126.748bd4d.x86_64",
+    #tag = "av-8.2-20200514-cc2bc34",
 )
 
+# TODO: Update this once we have PPC builds of the base image available
 container_pull(
     name = "libvirt_ppc64le",
-    digest = "sha256:9a24e6e9269c395e226b1dc949adce2076aaf5f20f882047116b3614b740294b",
+    digest = "sha256:NOT_AVAILABLE",  # Make sure we don't use outdated image by mistake
     puller_linux = "@go_puller_linux_ppc64le//file:downloaded",
     registry = "index.docker.io",
     repository = "kubevirt/libvirt",


phoracek:
virt-handler seems to be stuck on:

```
virt-handler-qv2jg   0/1     Init:CrashLoopBackOff   5          4m23s
```

I don't get how is it related to the change of the base image. @vladikr, is virt-handler using libvirt for its base image (sorry, but I have troubles figuring that out from bazel config).

phoracek:
I have one concern. We both pull the new base-image and download libvirt libs and sources (to support libvirt Go package AFAIK). Those RPMs are now outdated. Is this mismatch putting us in risk?

RPMs that are not updated: https://github.com/kubevirt/kubevirt/blob/master/WORKSPACE#L69

@vladikr

phoracek:
node-labeler init container in virt-handler fails with:

```
error: failed to get emulator capabilities
error: invalid argument: unable to find any emulator to serve 'x86_64' architecture
```

phoracek:
@andreabolognani would you please take a look at the failure describe in the code? Not sure if we should use a different libvirt version, ignore this or just open a bug.

andreabolognani:
> @andreabolognani would you please take a look at the failure describe in the code? Not sure if we should use a different libvirt version, ignore this or just open a bug.

I'm quite lost to be honest. Where did you get that error message from? Can you provide a link to the relevant logs? Sorry, I can't navigate KubeVirt's CI confidently quite yet :)

Anyway, I think the issue is caused by the fact that we're missing https://github.com/libvirt/libvirt/commit/6d786f95a366600e7bbae68c1b324a8131f5e2c5, which is part of libvirt 6.1.0. We need to either backport that patch to the 6.0.0 build in the COPR, or consider skipping 6.0.0 and moving to 6.2.0 directly.

andreabolognani:
Filed the issue as [RHBZ#1836351](https://bugzilla.redhat.com/show_bug.cgi?id=1836351).

phoracek:
@andreabolognani The log was obtained from virt-handlers init container. The init container (called virt-launcher) calls [/bin/node-labeller.sh](https://github.com/kubevirt/kubevirt/blob/master/cmd/virt-launcher/node-labeller/node-labeller.sh) from the virt-launcher image. The two lines were all that was reported. I added the `virtctl capabilities` call to the script to avoid it.

Thanks for the bug! I added link to it to the workaround in the script. Could we consider merging this (if it passes tests), so we can move forward with fixes requiring libvirt 6.0.0 and get rid of it once the bug is solved?

andreabolognani:
> Could we consider merging this (if it passes tests), so we can move forward with fixes requiring libvirt 6.0.0 and get rid of it once the bug is solved?

I think it's fine to have the workaround there temporarily, we just have to remember to drop it once we pick up a fixed libvirt build.

tests/vmi_lifecycle_test.go
```diff
@@ -165,7 +165,7 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
 			Eventually(logs,
 				2*time.Second,
 				500*time.Millisecond).
-				Should(And(ContainSubstring("internal error: Unable to get DBus system bus connection"), ContainSubstring(`"subcomponent":"libvirt"`)))
+				Should(And(ContainSubstring("At least one cgroup controller is required: No such device or address"), ContainSubstring(`"subcomponent":"libvirt"`)))
```
andreabolognani, line 168: @phoracek how does this work?

If I'm reading the code correctly, you're starting a VM and ensuring that succeeds, but afterwards you look at libvirt's logs and specifically look for an error message? Wouldn't that error have prevented libvirt (and thus the VM) from starting in the first place?

The change in the specifics of the error might be connected to internal changes in the order libvirt performs its checks, but I'm just uneasy that errors are raised in the first place.

phoracek: @andreabolognani  I was asking myself the same.

I think it just wants to see that libvirt has started, is running and printing logs. This was not the only error one I saw there...

This follows the patch that fixed this check for 5.6: https://github.com/kubevirt/kubevirt/commit/67e1d8b629737be30e254bac0aba68dcbfef00aa

andreabolognani: @phoracek looking at dc6b75cf31517f34c8c7b1aedb43139010a9a470, it seems like all three errors have been around for quite a long time, so I guess it's fine to look for any one of them.

I still wonder why some of them have stopped showing up over time? Order of initialization shouldn't affect this, since AFAICT you're just making sure that they eventually show up, so it's as if libvirt is emitting fewer errors over time. Not necessarily a bad thing :) but we should make sure that it's intentional and not masking some potential issue.

Another thing: `pkg/virt-launcher/virtwrap/util/libvirt_helper_test.go` contains what appears to be a fairly old libvirt log snippet, would it make sense to update it with a more recent one? Could be done with a separate PR of course.

phoracek: They seem terribly outdated. I found the formatted output here https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3268/pull-kubevirt-e2e-k8s-cnao-1.17/1261314930372513792, but I don't know how to get the basic one. This PR is blocking us from multiple tasks. The unit test is long outdated. I vote for a different PR.

vladikr:
pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6 lane is completely not stable. overriding this here to unblock other PRs that depend on this one.

phoracek:
Rebased after node-labeler was reverted/dropped from KubeVirt.

PR #3273, opened 2020-04-09
Address travis CI stalls in ppc64le

murilofv:
This PR addresses the build failures seen in Travis for ppc64le due to no output being generated, the following error is seen in these cases:
"No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."

It does so by moving the keep alive mechanism to run through the whole job. The patch also increases the number of jobs in bazel to 2 given it provides some performance benefit and so far it has not caused the issue of bazel server being abruptly ended.

We can ignore the first patch ("Add --host_javabase to fix ppc64le build") in the series and pick it up from PR #3225, I am just sending it along given it hasn't been incorporated in the upstream code yet and without it the build process in ppc64le is broken. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ services:
 
 arch:
 - amd64
-# - ppc64le # temporary disable until we have it under control again
+- ppc64le
 
 os: linux
 
@@ -15,6 +15,10 @@ install:
 - git reset --hard
 
 script:
+- |
+  if [[ $TRAVIS_CPU_ARCH == "ppc64le" ]]; then
+  while sleep 9m; do echo "Long running job - keep alive"; done & LOOP_PID=$!
+  fi
 - make generate
 - if [[ -n "$(git status --porcelain)" ]] ; then echo "It seems like you need to run
   `make generate`. Please run it and commit the changes"; git status --porcelain; false; fi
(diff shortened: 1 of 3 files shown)


murilofv:
Just copying a note that I've mentioned in the PR #3225 and that I still haven't figure out how to address it:
The one issue I've been occasionally getting is that git status --porcelain step fails and I can’t understand why would that be the case if the previous step completed successfully, there are runs where I don’t see it complain and ones that I do see, for example:
https://travis-ci.com/github/murilofv/kubevirt/builds/158851093#L1090 -> Working
https://travis-ci.com/github/murilofv/kubevirt/builds/158712840#L1153 -> Not working

The only odd thing I noticed when comparing both is in the not working in line 447 there was the `error: gazelle: finding module path for import github.com/cloudflare/cfssl/csr: exit status 1: go: finding github.com/cloudflare/cfssl/csr latest.`

.travis.yml
```diff
@@ -7,14 +7,18 @@ services:
...
+- ppc64le
 
 os: linux
 
 install:
 - git reset --hard
 
 script:
+- |
+  if [[ $TRAVIS_CPU_ARCH == "ppc64le" ]]; then
```
rmohr, line 19: I tried to use this too, but that does not help. When playing with it, it turned out that every `-` is executed in its own subshell and all backgroupd processes are terminated. This means that we start the background process and it quits immediately.

murilofv: From what I've been testing it did the trick for me, I used to see the issue of the no output pretty often and with this I didn't see it once in a lot of runs.

For example in the current run on the CI: https://travis-ci.org/github/kubevirt/kubevirt/jobs/672952277
You can see the keep alive message in Lines 1133 and 1213. The keep alive process only gets killed at the end when I use the kill command to its PID, as seen in line 2737:
```
1827 Terminated              while sleep 9m; do
    echo "Long running job - keep alive";
done
```

I think I found the run in which you ran into an issue when trying a similar approach: https://travis-ci.org/github/kubevirt/kubevirt/jobs/665757942#L99
I was taking a look at it, and what I found interesting is if you look at the line 1072 it seems like the output got cropped mid print: ” 	versi”
I think you somehow got unlucky and an actual stall seemed to have happened. I’ve been attempting to recreate it with your patch, but the process keeps running in background and it actually prints the keep alive: https://travis-ci.com/github/murilofv/kubevirt/builds/160393191
The process only gets killed at the end of the build

rmohr: I am convinced now. Thanks for the detailed analysis.

maya-r:
ping. if this looks good, please approve :-)
/test pull-kubevirt-e2e-k8s-1.15-ceph

maya-r:
ping. if this looks good, please approve :-)
/test pull-kubevirt-e2e-k8s-1.15-ceph

murilofv:
Not much of an update, but just wanted to mention I am still working out with the Travis CI folks to address some issues I've been having with Travis on Power architecture. So let's hold on a bit further on enabling it again.

rmohr:
/close

Since we had  to move out of travis, closing this. I am sure that at some point we can re-enable ppc64le but we have no hardware for it at the moment.

PR #3289, opened 2020-04-13
Add option to run only VNC Proxy in virtctl 

vatsalparekh:

Add option to run only VNC Proxy with new flag `--proxy-only`, letting users connect to VNC Proxy localhost port manually

Fixes #

Writing a test is tricky here, because we cannot have good test with external viewers verifying this works.
Existing VNC command also has no tests with external viewer, it only checks directly communicating with streaming api

**Release note**:

```release-note
Add option to run only VNC Proxy in virtctl
```

### DIFF
--- a/pkg/virtctl/vnc/vnc.go
+++ b/pkg/virtctl/vnc/vnc.go
@@ -40,7 +40,7 @@ import (
 )
 
 const (
-	LISTEN_TIMEOUT = 180 * time.Second
+	LISTEN_TIMEOUT = 60 * time.Second
 	FLAG           = "vnc"
 
 	//#### Tiger VNC ####
@@ -75,8 +75,8 @@ func NewCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
 		},
 	}
 	cmd.Flags().BoolVar(&proxyOnly, "proxy-only", proxyOnly, "--proxy-only=false: Setting this true will run only the virtctl vnc proxy and show the localhost port where VNC viewers can connect")
-	cmd.Flags().IntVar(&customPort, "custom-port", customPort,
-		"--custom-port=0: Assigning a port value to this will try to run the proxy on the given port if the port is accessible; If unassigned, the proxy will run on a random port")
+	cmd.Flags().IntVar(&customPort, "port", customPort,
+		"--port=0: Assigning a port value to this will try to run the proxy on the given port if the port is accessible; If unassigned, the proxy will run on a random port")
 	cmd.SetUsageTemplate(templates.UsageTemplate())
 	return cmd
 }
@@ -143,9 +143,11 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
 	go func() {
 		start := time.Now()
 		glog.Infof("connection timeout: %v", LISTEN_TIMEOUT)
-		// exit early if spawning remote-viewer fails
-		ln.SetDeadline(time.Now().Add(LISTEN_TIMEOUT))
-
+		// Don't set deadline if only proxy is running and VNC is to be connected manually
+		if !proxyOnly {
+			// exit early if spawning remote-viewer fails
+			ln.SetDeadline(time.Now().Add(LISTEN_TIMEOUT))
+		}
 		fd, err := ln.Accept()
 		if err != nil {
 			glog.V(2).Infof("Failed to accept unix sock connection. %s", err.Error())
--- a/pkg/virtctl/vnc/vnc.go
+++ b/pkg/virtctl/vnc/vnc.go
@@ -20,6 +20,7 @@
 package vnc
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"net"
@@ -39,7 +40,7 @@ import (
 )
 
 const (
-	LISTEN_TIMEOUT = 60 * time.Second
+	LISTEN_TIMEOUT = 180 * time.Second
 	FLAG           = "vnc"
 
 	//#### Tiger VNC ####
@@ -59,6 +60,8 @@ const (
 	TIGER_VNC     = "vncviewer"
 )
 
+var proxyOnly bool
+
 func NewCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "vnc (VMI)",
@@ -70,6 +73,7 @@ func NewCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
 			return c.Run(cmd, args)
 		},
 	}
+	cmd.Flags().BoolVar(&proxyOnly, "proxy-only", proxyOnly, "--proxy-only=false: Setting this true will run only the virtctl vnc proxy and show the localhost port where VNC viewers can connect")
 	cmd.SetUsageTemplate(templates.UsageTemplate())
 	return cmd
 }
@@ -165,78 +169,21 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
 		listenResChan <- err
 	}()
 
-	// execute VNC
-	go func() {
+	port := ln.Addr().(*net.TCPAddr).Port
+
+	if proxyOnly {
 		defer close(doneChan)
-		port := ln.Addr().(*net.TCPAddr).Port
-		args := []string{}
-
-		vncBin := ""
-		osType := runtime.GOOS
-		switch osType {
-		case "darwin":
-			if matches, err := filepath.Glob(MACOS_TIGER_VNC_PATTERN); err == nil && len(matches) > 0 {
-				// Always use the latest version
-				vncBin = matches[len(matches)-1]
-				args = tigerVncArgs(port)
-			} else if err == filepath.ErrBadPattern {
-				viewResChan <- err
-				return
-			} else if _, err := os.Stat(MACOS_CHICKEN_VNC); err == nil {
-				vncBin = MACOS_CHICKEN_VNC
-				args = chickenVncArgs(port)
-			} else if !os.IsNotExist(err) {
-				viewResChan <- err
-				return
-			} else if _, err := os.Stat(MACOS_REAL_VNC); err == nil {
-				vncBin = MACOS_REAL_VNC
-				args = realVncArgs(port)
-			} else if !os.IsNotExist(err) {
-				viewResChan <- err
-				return
-			} else if _, err := exec.LookPath(REMOTE_VIEWER); err == nil {
-				// fall back to user supplied script/binary in path
-				vncBin = REMOTE_VIEWER
-				args = remoteViewerArgs(port)
-			} else if !os.IsNotExist(err) {
-				viewResChan <- err
-				return
-			}
-		case "linux", "windows":
-			if _, err := exec.LookPath(REMOTE_VIEWER); err == nil {
-				vncBin = REMOTE_VIEWER
-				args = remoteViewerArgs(port)
-			} else if _, err := exec.LookPath(TIGER_VNC); err == nil {
-				vncBin = TIGER_VNC
-				args = tigerVncArgs(port)
-			} else {
-				viewResChan <- fmt.Errorf("could not find %s or %s binary in $PATH",
-					REMOTE_VIEWER, TIGER_VNC)
-				viewResChan <- err
-				return
-			}
-		default:
-			viewResChan <- fmt.Errorf("virtctl does not support VNC on %v", osType)
-			return
-		}
-
-		if vncBin == "" {
-			glog.Errorf("No supported VNC app found in %s", osType)
-			err = fmt.Errorf("No supported VNC app found in %s", osType)
+		optionString, err := json.Marshal(struct {
+			Port int `json:"port"`
+		}{port})
+		if err != nil {
+			return fmt.Errorf("Error encountered: %s", err.Error())
+			fmt.Println(string(optionString))
 		} else {
-			if glog.V(4) {
-				glog.Infof("Executing commandline: '%s %v'", vncBin, args)
-			}
-			cmnd := exec.Command(vncBin, args...)
-			output, err := cmnd.CombinedOutput()
-			if err != nil {
-				glog.Errorf("%s execution failed: %v, output: %v", vncBin, err, string(output))
-			} else {
-				glog.V(2).Infof("remote-viewer output: %v", string(output))
-			}
+			// execute VNC Viewer
+			go checkAndRunVNCViewer(doneChan, viewResChan, port)
 		}
-		viewResChan <- err
-	}()
+	}
 
 	go func() {
 		defer close(stopChan)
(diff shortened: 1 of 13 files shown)


pkg/virtctl/vnc/vnc.go
```diff
@@ -164,76 +167,84 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
+
+	if onlyProxy {
+		fmt.Printf("VNC Proxy localhost port: %d\n", port)
+	}
+
+	// If onlyProxy flag is passed, do not execute VNC
+	if !onlyProxy {
+		// execute VNC
+		go func() {
+			defer close(doneChan)
```
akrejcir, line 180: In case `onlyProxy == true`, the `doneChan` will not be closed and the above goroutine would not terminate.

It may be what we want, that the proxy does not terminate when the client disconnects. But in that case, could it support reconnect?

pkg/virtctl/vnc/vnc.go
```diff
@@ -164,76 +167,84 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
+	port := ln.Addr().(*net.TCPAddr).Port
+
+	if onlyProxy {
+		fmt.Printf("VNC Proxy localhost port: %d\n", port)
+	}
+
+	// If onlyProxy flag is passed, do not execute VNC
+	if !onlyProxy {
+		// execute VNC
+		go func() {
```
akrejcir, line 179: Could you extract this goroutine to a new function? It would improve readability.

pkg/virtctl/vnc/vnc.go
```diff
@@ -164,76 +167,84 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
-				args = realVncArgs(port)
-			} else if !os.IsNotExist(err) {
-				viewResChan <- err
-				return
-			} else if _, err := exec.LookPath(REMOTE_VIEWER); err == nil {
-				// fall back to user supplied script/binary in path
+	port := ln.Addr().(*net.TCPAddr).Port
+
+	if onlyProxy {
+		fmt.Printf("VNC Proxy localhost port: %d\n", port)
```
fabiand, line 173: Maybe worth to output this in a machine readable format, aka json?

staging/src/kubevirt.io/client-go/kubecli/kubevirt.go
```diff
@@ -114,6 +114,10 @@ type StreamOptions struct {
 	Out io.Writer
 }
 
+type VNCOptions struct {
```
akrejcir, line 117: Why is this structure exported in this file? It seems it is only used to format JSON output in the `vnc.go`. In that case it may be easier to use an anonymous struct.

pkg/virtctl/vnc/vnc.go
```diff
@@ -164,76 +168,25 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
-			} else if err != nil {
-				viewResChan <- err
-				return
-			}
-			vncBin = REMOTE_VIEWER
-			args = remoteViewerArgs(port)
-		default:
-			viewResChan <- fmt.Errorf("virtctl does not support VNC on %v", osType)
-			return
+		optionString, err := json.Marshal(vncoptions)
```
akrejcir, line 178: If the `vncoptions` variable is only used for JSON formatting, it could be an anonymous struct:
```
json.Marshal(struct {Port int `json:"port"`}{port})
```

pkg/virtctl/vnc/vnc.go
```diff
@@ -164,76 +168,25 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
-			if err != nil {
-				glog.Errorf("%s execution failed: %v, output: %v", vncBin, err, string(output))
-			} else {
-				glog.V(2).Infof("remote-viewer output: %v", string(output))
-			}
-		}
-		viewResChan <- err
-	}()
+	// If onlyProxy flag is passed, do not execute VNC
+	if !onlyProxy {
```
akrejcir, line 186: Can you merge this `if` and the previous one, to a single `if {} else {}` ?

pkg/virtctl/vnc/vnc.go
```diff
@@ -164,76 +168,25 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
-				return
-			}
-			vncBin = REMOTE_VIEWER
-			args = remoteViewerArgs(port)
-		default:
-			viewResChan <- fmt.Errorf("virtctl does not support VNC on %v", osType)
-			return
+		optionString, err := json.Marshal(vncoptions)
+		if err != nil {
+			return fmt.Errorf("Error encountered: %s", err.Error())
```
akrejcir, line 180: This error string does not contain any useful information, so it is enough to just return the `err`.

staging/src/kubevirt.io/client-go/kubecli/kubevirt.go
vatsalparekh: ack, corrected

pkg/virtctl/vnc/vnc.go
vatsalparekh: ack, corrected

pkg/virtctl/vnc/vnc.go
vatsalparekh: ack, corrected

pkg/virtctl/vnc/vnc.go
vatsalparekh: It's similar to other error msg in the file, not changing anything here in this PR

vatsalparekh:
@akrejcir @jean-edouard This should be good to review now
Added description, release-note and functional test now

tests/vnc_test.go
```diff
@@ -155,6 +161,54 @@ var _ = Describe("[rfe_id:127][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 			_, err = wrappedRoundTripper.RoundTrip(req)
 			Expect(err).ToNot(HaveOccurred())
 		})
+
+		It("should connect to vnc with --proxy-only flag", func() {
+			By("Invoking virtctl vnc with --proxy-only")
+			proxyOnlyCommand := tests.NewVirtctlCommand("vnc", "--proxy-only", "--namespace", vmi.Namespace, vmi.Name)
+
+			r, w, _ := os.Pipe()
```
akrejcir, line 169: This creates a pipe on the OS level. It may be simpler to specify an `io.Writer` as the output of the `proxyOnlyCommand`. Something like:

```go
pipeReader, pipeWriter := io.Pipe()
proxyOnlyCommand.SetOut(pipeWriter)
go func() {
	defer GinkgoRecover()
	Expect(proxyOnlyCommand.Execute()).ToNot(HaveOccurred())
}()

var result map[string]interface{}
err := json.NewDecoder(pipeReader).Decode(&result)

// Discards the rest of the output from the pipe, so the command does not block
go io.Copy(ioutil.Discard, pipeReader)
```

Also this way there is no need to call `time.Sleep()`, the `Decode()` method should block until some data is available.

akrejcir: Ignore my previous comment. I incorrectly thought that `proxyOnlyCommand` would run `virtctl` as a new process.
The current approach with redirecting `os.Stdout` is ok.

tests/vnc_test.go
```diff
@@ -155,6 +161,54 @@ var _ = Describe("[rfe_id:127][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+			os.Stdout = w
+
+			outC := make(chan string)
+			go func() {
+				var buf bytes.Buffer
+				io.Copy(&buf, r)
+				outC <- buf.String()
+			}()
```
akrejcir, line 177: This block (lines 172 - 177) is not necessary, because the `proxyOnlyCommand` will be executed in a goroutine, so this method can simply read from `r`.

tests/vnc_test.go
```diff
@@ -155,6 +161,54 @@ var _ = Describe("[rfe_id:127][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+			// No simple way to check if proxy has started, so arbitrary delay
+			time.Sleep(10 * time.Second)
+
+			w.Close()
+			out := <-outC
+
+			Expect(out).ToNot(BeNil())
+
+			var result map[string]interface{}
+			err := json.Unmarshal([]byte(out), &result)
```
akrejcir, line 191: Json can be directly read from the pipe: `err := json.NewDecoder(r).Decode(&result)`.

It may be useful to add a timeout for the read, in case the pipe is not written to because of some error. To do it, the read can be moved to a new goroutine, and then use a select statement to wait on the result or error.

tests/vnc_test.go
```diff
@@ -155,6 +161,54 @@ var _ = Describe("[rfe_id:127][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 			_, err = wrappedRoundTripper.RoundTrip(req)
 			Expect(err).ToNot(HaveOccurred())
 		})
+
+		It("should connect to vnc with --proxy-only flag", func() {
+			By("Invoking virtctl vnc with --proxy-only")
+			proxyOnlyCommand := tests.NewVirtctlCommand("vnc", "--proxy-only", "--namespace", vmi.Namespace, vmi.Name)
+
+			r, w, _ := os.Pipe()
+			os.Stdout = w
```
akrejcir, line 170: After the json is read from the pipe, the `os.Stdout` should be returned to it's original value. The same in case of a test failure:
```go
oldStdout := os.Stdout
os.Stdout = w
defer func() { os.Stdout = oldStdout }()
```

Please also add a short comment explaining why is the redirection needed.

tests/vnc_test.go
```diff
@@ -155,6 +161,54 @@ var _ = Describe("[rfe_id:127][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+				outC <- buf.String()
+			}()
+
+			// Run this as go routine because to keep proxy open in the background
+			go proxyOnlyCommand.Execute()
+
+			// No simple way to check if proxy has started, so arbitrary delay
+			time.Sleep(10 * time.Second)
+
+			w.Close()
```
akrejcir, line 185: The pipe should be closed in a `defer` statement, so that it will be closed also in case of test failure.

tests/vnc_test.go
```diff
@@ -155,6 +161,54 @@ var _ = Describe("[rfe_id:127][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+
+			outC := make(chan string)
+			go func() {
+				var buf bytes.Buffer
+				io.Copy(&buf, r)
+				outC <- buf.String()
+			}()
+
+			// Run this as go routine because to keep proxy open in the background
+			go proxyOnlyCommand.Execute()
```
akrejcir, line 180: The goroutone should pass the returned error to the main test method, or handle it using `Expect()` and `GinkgoRecover()`.

tests/vnc_test.go
vatsalparekh: Added

tests/vnc_test.go
vatsalparekh: Corrected

tests/vnc_test.go
vatsalparekh: Added an `Eventually` block for this

tests/vnc_test.go
```diff
@@ -155,6 +160,49 @@ var _ = Describe("[rfe_id:127][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+			// Run this as go routine to keep proxy open in the background
+			go func() {
+				defer GinkgoRecover()
+				Expect(proxyOnlyCommand.Execute()).ToNot(HaveOccurred())
+			}()
+
+			var result map[string]interface{}
+			Eventually(func() error {
+				return json.NewDecoder(r).Decode(&result)
+			}, 60*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
```
akrejcir, line 186: The timeout and the polling interval should be the same, it does not make sense to call the method more than once.

vatsalparekh: Corrected, in this case we only need the timeout

akrejcir:
/lgtm

The PR looks good. 

Just one small thing, I have noticed that the first commit introduces a bug and the second commit fixes it.
In the future, please change directly the commit that introduces the bug, or squash them together.

pkg/virtctl/vnc/vnc.go
```diff
@@ -165,78 +169,21 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
+		defer close(doneChan)
+		optionString, err := json.Marshal(struct {
+			Port int `json:"port"`
+		}{port})
+		if err != nil {
+			return fmt.Errorf("Error encountered: %s", err.Error())
 		}
-		viewResChan <- err
-	}()
+		fmt.Println(string(optionString))
```
rmohr, line 182: Can you replace this line with `fmt.Fprintln(cmd.OutOrStdout(), string(optionString))`?

tests/vnc_test.go
```diff
@@ -155,6 +160,49 @@ var _ = Describe("[rfe_id:127][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+		It("should connect to vnc with --proxy-only flag", func() {
+
+			By("Invoking virtctl vnc with --proxy-only")
+			proxyOnlyCommand := tests.NewVirtctlCommand("vnc", "--proxy-only", "--namespace", vmi.Namespace, vmi.Name)
+
+			r, w, _ := os.Pipe()
+
+			// Stdout should be set back to previous value after the test
+			oldStdout := os.Stdout
+			os.Stdout = w
```
rmohr, line 173: Instaed of this, can you please do `cmd.SetOut(w)` and remove the Stdout mangling of the process?

pkg/virtctl/vnc/vnc.go
vatsalparekh: Done

tests/vnc_test.go
vatsalparekh: Done

vatsalparekh: I tried this initially with `io.Pipe()` which didn't seem to be working when passing writer to `cmd.SetOut`; but with `os.Pipe()` seems working fine currently

pkg/virtctl/vnc/vnc.go
```diff
@@ -39,7 +40,7 @@ import (
 )
 
 const (
-	LISTEN_TIMEOUT = 60 * time.Second
+	LISTEN_TIMEOUT = 180 * time.Second
```
rmohr, line 43: Why is the timeout increased?

pkg/virtctl/vnc/vnc.go
```diff
@@ -59,6 +60,8 @@ const (
 	TIGER_VNC     = "vncviewer"
 )
 
+var proxyOnly bool
```
rmohr, line 63: We should not have started creating gloabls for these flags. They could just have been checked in the actual command. Can be changed for all commands in a follow-up. No need to change it here.

pkg/virtctl/vnc/vnc.go
```diff
@@ -70,6 +73,7 @@ func NewCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
 			return c.Run(cmd, args)
 		},
 	}
+	cmd.Flags().BoolVar(&proxyOnly, "proxy-only", proxyOnly, "--proxy-only=false: Setting this true will run only the virtctl vnc proxy and show the localhost port where VNC viewers can connect")
```
rmohr, line 76: What I would immediately look for is for a way to set the listen port.

vatsalparekh: Added an option for it now

pkg/virtctl/vnc/vnc.go
rmohr: Ah ok. I think we have to change that. If the proxy option is set, we should probably not timeout.

pkg/virtctl/vnc/vnc.go
rmohr: :+1

pkg/virtctl/vnc/vnc.go
```diff
@@ -70,6 +74,9 @@ func NewCommand(clientConfig clientcmd.ClientConfig) *cobra.Command {
...
+	cmd.Flags().IntVar(&customPort, "custom-port", customPort,
+		"--custom-port=0: Assigning a port value to this will try to run the proxy on the given port if the port is accessible; If unassigned, the proxy will run on a random port")
```
rmohr, line 79: would `--port ` and `-p` make sense?

pkg/virtctl/vnc/vnc.go
```diff
@@ -165,78 +172,21 @@ func (o *VNC) Run(cmd *cobra.Command, args []string) error {
...
-				viewResChan <- fmt.Errorf("could not find %s or %s binary in $PATH",
-					REMOTE_VIEWER, TIGER_VNC)
-				viewResChan <- err
-				return
-			}
-		default:
-			viewResChan <- fmt.Errorf("virtctl does not support VNC on %v", osType)
-			return
-		}
+	port := ln.Addr().(*net.TCPAddr).Port
```
rmohr, line 175: Do we validate if we don't select a proxy, but set a port and fail then?

rmohr (changes requested):
Thanks for the updates. I have some questions.

cluster-up/cluster/local/provider.sh
```diff
@@ -32,8 +32,7 @@ function up() {
 }
 
 function prepare_config() {
-    PROVIDER_CONFIG_FILE_PATH="${BASE_PATH}/$KUBEVIRT_PROVIDER/config-provider-$KUBEVIRT_PROVIDER.sh"
-    cat > "$PROVIDER_CONFIG_FILE_PATH" <<EOF
+    cat >hack/config-provider-local.sh <<EOF
```
oshoval, line 35: This reverts a needed fix, see please 
https://github.com/kubevirt/kubevirt/issues/4041
i suggest please to not include this file in this commit (`make generate` did this change for you)

vatsalparekh: Yes `make generate` made this change for me, removed now

pkg/virtctl/vnc/vnc.go
vatsalparekh: @rmohr Yes, changed it to just `--port` now

davidvossel (approved):
/lgtm
/approve

PR #3292, opened 2020-04-13
VMI configuration tests - Clear SMBIOS test data from configmap when the test ends

enp0s3:
When testing custom SMBIOS settings we use KV configmap in order to propagate these settings. The issue was that the test does not clear these settings from the CM at the end. This PR fixes this issue.

Not cleaning the custom SMBIOS settings can lead to a failure of the default SMBIOS settings test. As it already happened on cluster that is frequently used to run functional tests on a nightly basis.

Fixes #
Test 2752 - Custom SMBIOS settings

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -4576,3 +4576,18 @@ func GetKubeVirtConfigMap() (*k8sv1.ConfigMap, error) {
 
 	return cfgMap, err
 }
+
+func ClearKubeVirtConfigMap(key string) error {
+	virtClient, err := kubecli.GetKubevirtClient()
+	PanicOnError(err)
+
+	cfgMap, err := GetKubeVirtConfigMap()
+	if err == nil {
+		if _, ok := cfgMap.Data[key]; ok {
+			data := fmt.Sprintf(`[{ "op": "remove", "path": "/data/%s"}]`, key)
+			_, err = virtClient.CoreV1().ConfigMaps(KubeVirtInstallNamespace).Patch(virtconfig.ConfigMapName, types.JSONPatchType, []byte(data))
+		}
+	}
+
+	return err
+}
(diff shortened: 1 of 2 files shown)


tests/utils.go
```diff
@@ -4576,3 +4576,18 @@ func GetKubeVirtConfigMap() (*k8sv1.ConfigMap, error) {
...
 }
+
+func ClearKubeVirtConfigMap(key string) error {
+	virtClient, err := kubecli.GetKubevirtClient()
+	PanicOnError(err)
+
+	cfgMap, err := GetKubeVirtConfigMap()
+	if err == nil {
+		if _, ok := cfgMap.Data[key]; ok {
+			data := fmt.Sprintf(`[{ "op": "remove", "path": "/data/%s"}]`, key)
```
stu-gott, line 4587: This is likely good enough, but because we're messing about with KubeVirt's ConfigMap here, perhaps we should be slightly paranoid:
```
		if val, ok := cfgMap.Data[key]; ok {
                        testJson := fmt.Sprintf(`[{ "op": "test", "path": "/data/%s", "value": %s}]`, key, val)
			removeJson := fmt.Sprintf(`[{ "op": "remove", "path": "/data/%s"}]`, key)
                        data := fmt.Sprintf("[%s, %s]", testJson, removeJson)
			_, err = virtClient.CoreV1().ConfigMaps(KubeVirtInstallNamespace).Patch(virtconfig.ConfigMapName, types.JSONPatchType, []byte(data))
```

Note: I just scribbled that snippet in this comment box. it's untested. Basically the gist here is that we should send a patch that compares the configMap state to what we expect it to be before removing the key. I can't think of a scenario where this could actually happen to trip us, but coding it this way will trigger an error if the cluster's in a state we didn't expect.

enp0s3: @stu-gott I did exactly what you mentioned but the JSON test failed although the key was there. I could not understand how to do JSON test for existence of keys, not values.

enp0s3: Looks like JSON test not works of intermediate nodes but only for leaves, if thinking of JSON as of an analogy to a tree.

PR #3303, opened 2020-04-16
Test fixes - dmidecode file was missing on NGINX DL path of the HTTP CDI import server 

enp0s3:
The dmidecode utility is being downloaded from the CDI HTTP import server at the cloud-init stage.
We noticed that instead of a binary, 404 was returned. The issue was with incorrect source copy path in the entrypoint.sh script of the CDI HTTP container. Path is fixed using the which command.

Dmidecode is required is some tests in order to test various VMI configurations

Fixes #
dmidecode related tests

**Release note**:
```release-note
None
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -561,6 +561,14 @@ http_file(
     ],
 )
 
+http_file(
+    name = "which",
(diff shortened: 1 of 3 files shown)


enp0s3:
@rmohr Do we need to upload `which` RPM to Google as backup?

PR #3315, opened 2020-04-21
Fix tests relying on bearer token behavior

stu-gott:
A handful of functional tests relied on using "--token" to indicate preference as to which service account the test should run as. These tests might have worked at some point in the past, but don't behave correctly with current k8s or openshift clients.

Since the intention of the test is to simply verify VM creation behavior/permissions, these tests have been changed to instead use the "auth can-i" subcommand--which unambiguously checks permissions for the explicitly named account.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/vm_test.go
+++ b/tests/vm_test.go
@@ -1483,15 +1483,15 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
 
 		Context("as ordinary OCP user trough test service account", func() {
 			var testUser string
-			var token string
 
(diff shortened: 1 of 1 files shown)


stu-gott:
/cc @danielBelenky 
No code changes in forced commit -- added the PR description to the changeset as requested.

rmohr:
> @stu-gott could you please add the description to the commit message? it will be lost when we merge it otherwise...

Why would it be lost otherwise?

rmohr:
Tests look good. Merging without tide right now.

PR #3319, opened 2020-04-22
Flakey test 3468 fix

omeryahud:
No longer sleeping for a static duration after starting the watch process, instead waiting for the VM to be successfully created and then starting the watch process

Fixes testid 3468

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/vm_watch_test.go
+++ b/tests/vm_watch_test.go
@@ -1,10 +1,14 @@
 package tests_test
 
 import (
+	"fmt"
 	"io"
+	"net"
 	"os/exec"
 	"strings"
 	"time"
+
+	v12 "kubevirt.io/client-go/api/v1"
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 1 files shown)


tests/vm_watch_test.go
```diff
@@ -21,338 +24,380 @@ const (
...
-		ip,
-		node string
+	return stat
+}
+
+func newVMIStatus(fields []string) *vmiStatus {
+	flen := len(fields)
+	stat := &vmiStatus{}
+
+	switch {
```
rmohr, line 73: @omeryahud I really don't get this test. If I start until a VMI is waiting, I should get a pretty well defined response. Why is there all this extra parsing going on. Do you know?

omeryahud: The point of this test is to make sure that the end user receives all VM or VMI updates when using kubectl get vm/vmi -w.
Parsing the response to vmiStatus and vmStatus objects is mainly for readability instead of asserting on slices

rmohr: Hm, regarding to readability. I think I would have preferred if the test just compares it to a string which uses `fmt.Spprintf()` to insert the expected node and the expected IP. I think that is much simpler and makes it much easier to follow.

omeryahud: String comparison won't work since the number of white space characters between columns is not constant, so I will have to use `strings.Fields()` (which I already do right now) and mess around with slices, that's why I added the helper structs to give it some more context.

rmohr: how about something like this:

```go
expected := []string{"vmi-nocloud"   2m   Failed   10.244.1.25   node02}
fields := strings.Fields(line)
Expect(fields).Should(ConsistOf("vmi-nocloud",MatchRegex("[0-9]+[ms]"),"Failed", ip, node))
```

tests/vm_watch_test.go
```diff
@@ -21,338 +24,337 @@ const (
 	relevantk8sVer = "1.16.2"
 
 	// Define a timeout for read opeartions in order to prevent test hanging
-	readTimeout     = 1 * time.Minute
-	processWaitTime = 2 * time.Minute
+	readTimeout       = 30 * time.Second
+	vmCreationTimeout = 1 * time.Minute
+	vmAgeRegex        = "^[0-9]+[sm]$"
+	ipRegex           = "^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$"
```
jean-edouard, line 30: This only works with IPv4.
Maybe `net.ParseIP()` could be used instead? (https://golang.org/pkg/net/#ParseIP)

tests/vm_watch_test.go
```diff
@@ -21,338 +24,337 @@ const (
...
+	}
 
-			Fail("Timeout reached on read operation")
+	newStatus := strings.Fields(statusLine)
 
-			return ""
+	// Skip status line with similar running state for VM or phase state for VMI
+	// newStatus[2] and oldStatus[2] point to the VM running state or VMI phase
+	if oldStatus != nil {
+		if len(oldStatus) == 2 && len(oldStatus) == len(newStatus) {
```
jean-edouard, line 108: `len(oldStatus) == len(newStatus)` can be refactored to the parent `if`

omeryahud: You are right, thanks!

tests/vm_watch_test.go
omeryahud: Do you know how would I go using `net.ParseIP()` in this assertion:
```
Expect(vmiStatus).To(ConsistOf(vmi.Name, MatchRegexp(vmAgeRegex), string(v12.Scheduled), MatchRegexp(ipRegex), vmi.Status.NodeName),
			"VMI should be in the Scheduled phase")
```
Maybe I should just use a regex that matches both IPv4 and IPv6?

jean-edouard: Right, to use `ParseIP()` you'd still need to parse the string into fields... That might still be worth it, at least for tests which check IPs.
With that I would ParseIP() on fields[3] to `ip`, Expect() it to be non-nil, and in the ConsistOf, use ip.String() (or just fields[3]) to match the IP.
Expending that regex with a big "or" block for ipv6 would make the string pretty huge!

omeryahud:
@jean-edouard @rmohr 
Can you guys take a final look please?

tests/vm_watch_test.go
omeryahud: I just now saw your reply, sorry about that!
I like your idea, will work on it right now

PR #3334, opened 2020-04-24
Avoid unusable taints in test suite

stu-gott:
Some tests in our test suite use taints to trigger a migration. This is normally fine. However, in deployments where the node maintenance operator is installed, the `node.kubernetes.io/unschedulable` built-in taint is used. This taint cannot be applied manually. If this is the case, we can simply use another key.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -103,6 +103,22 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 		if len(nodes.Items) < 2 {
 			Skip("Migration tests require at least 2 nodes")
 		}
+
+		// Taints defined by k8s are special and can't be applied manually.
+		// Temporarily configure KubeVirt to use something else for the duration of these tests.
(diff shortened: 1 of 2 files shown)


tests/migration_test.go
```diff
@@ -103,6 +103,22 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 		if len(nodes.Items) < 2 {
 			Skip("Migration tests require at least 2 nodes")
 		}
+
+		// Taints defined by k8s are special and can't be applied manually.
+		// Temporarily configure KubeVirt to use something else for the duration of these tests.
+		if tests.IsUsingBuiltinNodeDrainKey() {
```
vladikr, line 109: Do we need to clean this after the tests?

stu-gott: Nope! there's a general "revert any changes to the configmap" step in the AfterEach already.

vladikr: Ah, that's right. Great.
/lgtm

PR #3336, opened 2020-04-24
Fix file access in docker build

dhiller:

When running the tests the cluster-up writes configuration into _ci-configs/external, which it can't currently as somehow during the docker build the directory seems to be put in readonly mode for the user. To avoid this we reset access rights during image build.

See openshift/release#8545 for reference

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/ci/build.sh
+++ b/hack/ci/build.sh
@@ -16,6 +16,11 @@ bash -x ./hack/build-manifests.sh
 # build dump
 CMD_OUT_DIR="$(pwd)/_out/cmd"
 export CMD_OUT_DIR
-mkdir -p _out/cmd/dump/
+mkdir -p "$CMD_OUT_DIR/dump/"
 GOPROXY=off GOFLAGS=-mod=vendor go build -o "$CMD_OUT_DIR/dump/dump" ./cmd/dump
 bash -x ./hack/build-func-tests.sh
+
+rm -rf _ci-configs/
+
+# to avoid any permission problems we reset access rights recursively
+chmod -R 777 .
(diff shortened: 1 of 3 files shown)


dhiller:
The image build runs fine, but [the build itself falls apart](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/kubevirt_kubevirt/3299/pull-ci-kubevirt-kubevirt-master-e2e-gcp-nested-virt/73) due to we are running on oepnshift/release master branch here.

hack/ci/build.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# to avoid any permission problems we reset access rights recursively
+chmod -R 777 .
```
oshoval, line 4: maybe worth make it more subtle ?
644 or whatever needed?

rmohr, line 4: How long does this take on an average system with a few gigabyte in place?

dhiller: So you're suggesting that we only use this on selected subdirectories in the kubevirt repository directory? Like i.e. _out and _ci-configs?

rmohr: In general yes.

dhiller: I mean, at that time we are within the kubevirt repository directory so there should just be the contents of that repo there.

rmohr: ah wait, it is the CI build file only. So no interaction with the normal developer flow.

PR #3345, opened 2020-04-27
CONTRIBUTING: Add a procedure for joining the organization

EdDev:
Introduce a simple procedure for contributors to join the kubevirt organization.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,6 +92,17 @@ Maintainers are:
  * @stu-gott
  * @vladikr
 
+### Becoming a memeber
+
+Contributors that frequently contribute to the project may ask to join the
+kubevirt organization.
+Please send a request to join by sending an email to our
+[Google Group](https://groups.google.com/forum/#!forum/kubevirt-dev).
+The request should include a self-introduction, and an explanation why you
+should become a member.
+
+Responses will be provided privately by one of the organization owners.
+
 ## Projects & Communities
 
 ### [KubeVirt](https://github.com/kubevirt/)


CONTRIBUTING.md
```diff
@@ -92,6 +92,15 @@ Maintainers are:
  * @stu-gott
  * @vladikr
 
+### Becoming a memeber
+
+Contributors that frequently contribute to the project may ask to join the
+kubevirt organization.
+Please send a request to join by sending an email to our
```
dankenigsberg, line 99: the request should include a self-introduction, and an explanation why you should be a member.

CONTRIBUTING.md
```diff
@@ -92,6 +92,15 @@ Maintainers are:
  * @stu-gott
  * @vladikr
 
+### Becoming a memeber
+
+Contributors that frequently contribute to the project may ask to join the
+kubevirt organization.
```
dankenigsberg, line 98: Can you add a word on the benefits of being a member? Can you get CI running automatically? Any other permission granted for members?

EdDev: It may be in our best interest to leave vague.
This may change with time and not committing to specifics is not always bad.

CONTRIBUTING.md
EdDev: Done

CONTRIBUTING.md
dankenigsberg: ok, don't update the doc, but would you remind me: why would one currently want to be a member.

EdDev: I just wanted to be added as a reviewer or to be mentioned in comments.

PR #3358, opened 2020-04-28
VM destruction should always result in Failed phase

stu-gott:
Lifecycle rules for VMI phases in virt-controller are written around the
assumption that graceful shutdown of a VMI should result in the
Succeeded phase and every form of involuntary shutdown should be
considered a Failure. If ACPI is not enabled, the VMI cannot have
succeeded by definition.

**Release note**:
```release-note
Fix a bug where VMI's without ACPI support incorrectly entered the "Succeeded" lifecycle phase.
```

### DIFF
--- a/pkg/virt-handler/vm.go
+++ b/pkg/virt-handler/vm.go
@@ -1875,13 +1875,7 @@ func (d *VirtualMachineController) calculateVmPhaseForStatusReason(domain *api.D
 			case api.ReasonCrashed, api.ReasonPanicked:
 				return v1.Failed, nil
 			case api.ReasonDestroyed:
-				// When ACPI is available, the domain was tried to be shutdown,
-				// and destroyed means that the domain was destroyed after the graceperiod expired.
-				// Without ACPI a destroyed domain is ok.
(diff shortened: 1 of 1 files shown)


stu-gott:
/cc @slintes 
It's worth double checking your opinion on this.

stu-gott:
@danielBelenky tests already exist. in tests/vmi_lifecycle_test.go

with ACPI and 0 grace period seconds: confirms failed phase
with ACPI and some grace period seconds: confirms succeeded phase

The specific condition the case corrected was almost impossible to test so wasn't present (which is why no tests changed)--however, since it's gone now, that's irrelevant.

rmohr:
As far as I know we don't support guest-agent shutdown right now, so i would actually assume that the only way that we can shutdown a VMI without ACPI is a forced shutdown which results in a "crash". Since it is the expected behaviour, I am not sure if we want to flood monitoring tools, logs and  events with warnings.

stu-gott:
Re-engaging this conversation. Thinking about this fact: if a VM is using a RerunOnFailure, this PR would prevent it from ever being shut down. Thus, I tend to agree this makes things worse.

PR #3364, opened 2020-04-29
Add skip condition for CPUManager=false

stu-gott:
Prior to this change, one CPUManager test assumed virt-handler would be
running on every node, including master nodes. In clusters where this
assumption is false, this test is invalid.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -1811,7 +1811,9 @@ var _ = Describe("Configurations", func() {
 				By("adding a cpumanger=true label to a node")
 				nodes, err := virtClient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: v1.CPUManager + "=" + "false"})
 				Expect(err).ToNot(HaveOccurred())
-				Expect(len(nodes.Items)).To(BeNumerically(">=", 1))
+				if len(nodes.Items) == 0 {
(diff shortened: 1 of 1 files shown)


tests/vmi_configuration_test.go
```diff
@@ -1811,7 +1811,9 @@ var _ = Describe("Configurations", func() {
 				By("adding a cpumanger=true label to a node")
 				nodes, err := virtClient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: v1.CPUManager + "=" + "false"})
 				Expect(err).ToNot(HaveOccurred())
-				Expect(len(nodes.Items)).To(BeNumerically(">=", 1))
+				if len(nodes.Items) == 0 {
+					Skip("Skip CPU manager test on clusters where virt-handler is not running on master nodes")
```
vladikr, line 1815: Technically it's true. But I think I would say something like "Skip CPU manager test on clusters where CPU manager is running on all worker/compute nodes" 
This test assumes that there is one node that is running the CPU manager and the other isn't. This is how kubevirtci is setup.
virt-handler will re-label the node according to the actual status of the cpu-manager. If the node which didn't run the cpu manger is labeled for whatever reason as cpumanager=true, virt-handler will relabel it back to  cpumanager=false. This is what this test verifies. On systems with a different setup - this tests will of course fail.

PR #3370, opened 2020-05-01
[WIP] Validate the presence of commands before using them

stu-gott:
Our test suite makes various assumptions about the images used in the test environment. However, the content of those images does change over time. It's also possible to test against different images than our CI environment does.

This PR introduces a mechanism to address this by explicitly checking for various commands presumed to be present in our images and clearly stating which command is missing from which image.

**Release note**:
```release-note
NONE
```


stu-gott:
This PR is marked as a WIP as I haven't yet covered all test modules, and might want to refactor some of the logic to be global to the entire test suite.

PR #3374, opened 2020-05-05
Supply virt-launcher an unique UUID for Kind Ipv6 

oshoval:
Libvirt require Unique UUID for each virt-launcher pod,
in order for migration to start.
Since we are using Kind provider for IPv6,
all the containers inherit the same UUID from DMI which is shared
between all the containers.

We limit in this PR the change for IPv6 kind provider,
as the task is urgent, and doesnt present a regression for
the other kind providers.

See
kubevirt/kubevirtci#353
once its done, the limitation can be removed, and more tests
can be enabled, in case they can be.

```release-note
None
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -57,6 +57,7 @@ import (
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	k8sv1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	settingsv1alpha1 "k8s.io/api/settings/v1alpha1"
 	storagev1 "k8s.io/api/storage/v1"
 	extclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -763,6 +764,10 @@ func BeforeTestSuitSetup() {
 	CreatePVC(osWindows, defaultWindowsDiskSize, Config.StorageClassWindows)
 	CreatePVC(osRhel, defaultRhelDiskSize, Config.StorageClassRhel)
 
+	if IsRunningOnKindInfraIPv6() {
+		createPodPreset("fix-node-uuid", "fake-product-uuid", "virt-launcher", "/kind/product_uuid", "/sys/class/dmi/id/product_uuid")
+	}
+
 	EnsureKVMPresent()
 
 	SetDefaultEventuallyTimeout(defaultEventuallyTimeout)
@@ -1244,6 +1249,48 @@ func cleanupSubresourceServiceAccount() {
 
 	err = virtCli.RbacV1().ClusterRoleBindings().Delete(SubresourceServiceAccountName, nil)
 	if !errors.IsNotFound(err) {
+		PanicOnError(err)
+	}
+}
+
+func createPodPreset(podPresetName, volName, label, srcPath, dstPath string) {
+	virtCli, err := kubecli.GetKubevirtClient()
+	PanicOnError(err)
+
+	hostPathType := k8sv1.HostPathFile
+	podPreset := settingsv1alpha1.PodPreset{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podPresetName,
+			Namespace: NamespaceTestDefault,
+		},
+		Spec: settingsv1alpha1.PodPresetSpec{
+			Selector: metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"kubevirt.io": label,
+				},
+			},
+			Volumes: []k8sv1.Volume{
+				{
+					Name: volName,
+					VolumeSource: k8sv1.VolumeSource{
+						HostPath: &k8sv1.HostPathVolumeSource{
+							Path: srcPath,
+							Type: &hostPathType,
+						},
+					},
+				},
+			},
+			VolumeMounts: []k8sv1.VolumeMount{
+				{
+					Name:      volName,
+					MountPath: dstPath,
+				},
+			},
+		},
+	}
+
+	_, err = virtCli.SettingsV1alpha1().PodPresets(NamespaceTestDefault).Create(&podPreset)
+	if !errors.IsAlreadyExists(err) {
 		PanicOnError(err)
 	}
 }
@@ -4582,6 +4629,11 @@ func getClusterDnsServiceIP(virtClient kubecli.KubevirtClient) (string, error) {
 func IsRunningOnKindInfra() bool {
 	provider := os.Getenv("KUBEVIRT_PROVIDER")
 	return strings.HasPrefix(provider, "kind")
+}
+
+func IsRunningOnKindInfraIPv6() bool {
+	provider := os.Getenv("KUBEVIRT_PROVIDER")
+	return strings.HasPrefix(provider, "kind-k8s-1.17.0-ipv6")
 }
 
 func SkipPVCTestIfRunnigOnKindInfra() {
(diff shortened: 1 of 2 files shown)


tests/utils.go
```diff
@@ -4584,6 +4631,11 @@ func IsRunningOnKindInfra() bool {
 	return strings.HasPrefix(provider, "kind")
 }
 
+func IsRunningOnKindInfraIPv6() bool {
```
maiqueb, line 4634: The issue your PR is trying to fix happens in all kind based providers, as stated in the [PR's description](https://github.com/kubevirt/kubevirt/pull/3374#issue-413443532).

As such, I don't think this function is required at all, and the `IsRunningOnKindInfra` function should be used instead.

tests/utils.go
```diff
@@ -763,6 +764,10 @@ func BeforeTestSuitSetup() {
 	CreatePVC(osWindows, defaultWindowsDiskSize, Config.StorageClassWindows)
 	CreatePVC(osRhel, defaultRhelDiskSize, Config.StorageClassRhel)
 
+	if IsRunningOnKindInfraIPv6() {
```
maiqueb, line 767: This issue happens in all kind based providers, as stated in the [PR's description](https://github.com/kubevirt/kubevirt/pull/3374#issue-413443532).

As such, I think the pod preset should be created on **all providers based on kind infra**, not on kind ipv6 provider.

oshoval: lets see please what team decide
Alona and Vladik in the current groom said its ok to take it afterwards.
one reason is that if kind has running already migration on other lanes it would failed if indeed it would already need it

i must focus on my tasks first to wrap them

EdDev: @oshoval , lets limit the discussion on the technical aspects of the review.

The reasoning of not failing for other kind tests is irrelevant, it may be the case that no migration tests are running on kind except the ipv6 ones.

You could argue that we want to first pass the IPv6 tests due to the urgency of the matter and therefore the **correct** solution is therefore better differed to a following PR.

tests/utils.go
oshoval: as explained in the previous commit,
if will break the other kind providers, since they don't have the file,
failure to mount, result in failure to run the launcher.

i will chance the PR description as Edy asked

tests/utils.go
```diff
@@ -1248,6 +1253,48 @@ func cleanupSubresourceServiceAccount() {
 	}
 }
 
+func createPodPreset(ppName, volName, label, srcPath, dstPath string) {
+	virtCli, err := kubecli.GetKubevirtClient()
+	PanicOnError(err)
+
+	hostPathType := k8sv1.HostPathFile
+	pp := settingsv1alpha1.PodPreset{
```
AlonaKaplan, line 1261: Please change pp to meaningful name like podPreset.

tests/utils.go
```diff
@@ -1248,6 +1253,48 @@ func cleanupSubresourceServiceAccount() {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ppName,
```
AlonaKaplan, line 1263: same

AlonaKaplan:
@oshoval please update the PR message.
It should be clear the change here is relevant to ipv6 kind provider only and that a follow-up change is needed to have the same fix for the other kind providers.

tests/utils.go
maiqueb: Obviously. That's is why I've stated I'm [my review comment](https://github.com/kubevirt/kubevirt/pull/3374#pullrequestreview-405725579) what I think should be done (i.e. generate the product id in the kind common script).

I'm OK with it being done in a follow-up PR.

AlonaKaplan:
Besides the same renaming comment /lgtm

oshoval:
> Please change pp to podPreset

sorry right changed the other places and forgot this
btw used pp , just because the other functions in this file used this naming convention

oshoval:
> Please change pp to podPreset

one note please if possible, i think you commented in the commit, then it doesnt show here the context, commenting from the view window, selecting specific commit, and not the commit link would show the context i think

EdDev:
> > Please change pp to podPreset
> 
> one note please if possible, i think you commented in the commit, then it doesnt show here the context, commenting from the view window, selecting specific commit, and not the commit link would show the context i think

It was commented inline: https://github.com/kubevirt/kubevirt/pull/3374#discussion_r420591885

oshoval:
> `vendor/modules.txt` changes seem to just be a reorder of the lines, which means nothing was added or removed from the dependencies.
> Therefore, this is not related to this change at all.
> 
> If you run the same on master, you will get this without your change.
> So, please do not add it to this PR or just extract it to its own commit.

done
but keep in mind that in my setup its auto generated by
make deps-update
if its not just my setup, its wrong imo to not push it
reasons:
1. every one will need to check out it 
2. merge conflicts
3. auto gen files / vendor folder need to be pushed , else even prow / travis can fail,
in case it checks the diff, as it does for some stuff like make generate

EdDev:
> but keep in mind that in my setup its auto generated by
> make deps-update
> if its not just my setup, its wrong imo to not push it
> reasons:
> 
>     1. every one will need to check out it

Current master is already in this state, so this this is true with or without this PR.
As I mentioned already, this can be done in a separate change.
> 
>     2. merge conflicts
> 
>     3. auto gen files / vendor folder need to be pushed , else even prow / travis can fail,
>        in case it checks the diff, as it does for some stuff like make generate

It needs to be pushed only if you add or remove a package there, as this change has not done it, pushing it is confusing the reviewer and unnecessary.

oshoval:
> > but keep in mind that in my setup its auto generated by
> > make deps-update
> > if its not just my setup, its wrong imo to not push it
> > reasons:
> > ```
> > 1. every one will need to check out it
> > ```
> 
> Current master is already in this state, so this this is true with or without this PR.
> As I mentioned already, this can be done in a separate change.
> 
> > ```
> > 2. merge conflicts
> > 
> > 3. auto gen files / vendor folder need to be pushed , else even prow / travis can fail,
> >    in case it checks the diff, as it does for some stuff like make generate
> > ```
> 
> It needs to be pushed only if you add or remove a package there, as this change has not done it, pushing it is confusing the reviewer and unnecessary.

thanks, i edited the comment but already to say that it looks like just my setup have this problem
i reverted it in this PR.

oshoval:
This PR was verified locally and also by PR which i work on that use it and fix more issues of the migration
https://github.com/kubevirt/kubevirt/pull/3386

tests/utils.go
```diff
@@ -4584,6 +4631,11 @@ func IsRunningOnKindInfra() bool {
 	return strings.HasPrefix(provider, "kind")
 }
 
+func IsRunningOnKindInfraIPv6() bool {
+	provider := os.Getenv("KUBEVIRT_PROVIDER")
+	return strings.HasPrefix(provider, "kind-k8s-1.17.0-ipv6")
```
vladikr, line 4636: @oshoval is it something specific for 1.17?

oshoval: its not specific, 
for the moment as we groomed with you and Alona and the team,
it was decided that indeed the change is better to exists, 
for future usage of kind that will need migration,
but will be done separately
https://github.com/kubevirt/kubevirtci/issues/353

vladikr: Yeah, I not against it :) 
Perhaps we shouldn't check for `1.17.0`, but rather check for prefix and suffix?

oshoval: kind*ipv6 if i understood right
hmm it means that in a case there will be other that have this pattern it would fail
unless the kubevirtci part would be added there as well.

also at the moment just this provider has the needed code in the kubevirtci,
the change there isnt hard and then the if will be indeed kind* only
so i believe its fine as its now imo ?

or i didnt understand and the reason is different ?

oshoval: glad to hear btw, thanks

vladikr:
Alright, seems like this is specifically needed for this version so I won't block.

/lgtm
/approve

oshoval:
Thanks Vladik and all
/retest

PR #3383, opened 2020-05-07
schema: Introduce an IPv6 CIDR for the vm network

EdDev:

When using the masquerade binding method, an internal IPv6 network is used
for the pod-vm communication.
This change introduces the ability to customize this network address in
scenarios where the default one cannot be used.

~~Tests are missing for both CIDR 4 and 6.~~

**Release note**:
```release-note
Add `vmIPv6NetworkCIDR` under `NetworkSource.pod` to support custom IPv6 CIDR for the vm network when using masquerade binding.
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -10765,6 +10765,10 @@
     "description": "Represents the stock pod network interface.",
     "type": "object",
     "properties": {
+     "vmIPv6NetworkCIDR": {
+      "description": "IPv6 CIDR for the vm network. Defaults to fd10:0:2::/120 if not specified.",
+      "type": "string"
+     },
      "vmNetworkCIDR": {
       "description": "CIDR for vm network. Default 10.0.2.0/24 if not specified.",
       "type": "string"
--- a/pkg/virt-operator/resource/generate/components/validations_generated.go
+++ b/pkg/virt-operator/resource/generate/components/validations_generated.go
@@ -2807,6 +2807,9 @@ var CRDsValidation map[string]string = map[string]string{
                       pod:
                         description: Represents the stock pod network interface.
                         properties:
+                          vmIPv6NetworkCIDR:
+                            description: IPv6 CIDR for the vm network. Defaults to fd10:0:2::/120 if not specified.
+                            type: string
                           vmNetworkCIDR:
                             description: CIDR for vm network. Default 10.0.2.0/24 if not specified.
                             type: string
@@ -4612,6 +4615,9 @@ var CRDsValidation map[string]string = map[string]string{
               pod:
                 description: Represents the stock pod network interface.
                 properties:
+                  vmIPv6NetworkCIDR:
+                    description: IPv6 CIDR for the vm network. Defaults to fd10:0:2::/120 if not specified.
+                    type: string
                   vmNetworkCIDR:
                     description: CIDR for vm network. Default 10.0.2.0/24 if not specified.
                     type: string
@@ -7191,6 +7197,9 @@ var CRDsValidation map[string]string = map[string]string{
                       pod:
                         description: Represents the stock pod network interface.
                         properties:
+                          vmIPv6NetworkCIDR:
+                            description: IPv6 CIDR for the vm network. Defaults to fd10:0:2::/120 if not specified.
+                            type: string
                           vmNetworkCIDR:
                             description: CIDR for vm network. Default 10.0.2.0/24 if not specified.
                             type: string
@@ -9306,6 +9315,9 @@ var CRDsValidation map[string]string = map[string]string{
                                   pod:
                                     description: Represents the stock pod network interface.
                                     properties:
+                                      vmIPv6NetworkCIDR:
+                                        description: IPv6 CIDR for the vm network. Defaults to fd10:0:2::/120 if not specified.
+                                        type: string
                                       vmNetworkCIDR:
                                         description: CIDR for vm network. Default 10.0.2.0/24 if not specified.
                                         type: string
--- a/staging/src/kubevirt.io/client-go/api/v1/schema.go
+++ b/staging/src/kubevirt.io/client-go/api/v1/schema.go
@@ -1348,6 +1348,10 @@ type PodNetwork struct {
 	// CIDR for vm network.
 	// Default 10.0.2.0/24 if not specified.
 	VMNetworkCIDR string `json:"vmNetworkCIDR,omitempty"`
+
+	// IPv6 CIDR for the vm network.
+	// Defaults to fd10:0:2::/120 if not specified.
+	VMIPv6NetworkCIDR string `json:"vmIPv6NetworkCIDR,omitempty"`
 }
 
 // Rng represents the random device passed from host
--- a/tests/network/vmi_networking.go
+++ b/tests/network/vmi_networking.go
@@ -627,7 +627,7 @@ var _ = SIGDescribe("[Serial][rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com]
 			return vmi
 		}
 
-		fedoraMasqueradeVMI := func(ports []v1.Port) (*v1.VirtualMachineInstance, error) {
+		fedoraMasqueradeVMI := func(ports []v1.Port, ipv6NetworkCIDR string) (*v1.VirtualMachineInstance, error) {
 
 			networkData, err := libnet.NewNetworkData(
 				libnet.WithEthernet("eth0",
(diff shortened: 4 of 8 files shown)


staging/src/kubevirt.io/client-go/api/v1/schema.go
```diff
@@ -1084,6 +1084,10 @@ type PodNetwork struct {
 	// CIDR for vm network.
 	// Default 10.0.2.0/24 if not specified.
 	VMNetworkCIDR string `json:"vmNetworkCIDR,omitempty"`
+
+	// IPv6 CIDR for the vm network.
+	// Defaults to fd10:0:2::/120 if not specified.
+	VMIpv6NetworkCIDR string `json:"vmIpv6NetworkCIDR,omitempty"`
```
rmohr, line 1090: It looks like this is not used anywhere. Is there maybe a commit missing?

EdDev: Originally I wanted to add a test, see it fails and then link it to the usage and see it passes.
But it seems I need to do some work on the network tests in general and introduce tests for these API/s afterwards.

I added the usage now, not to confuse anyone.

AlonaKaplan:
The change lgtm. Would love to see e2e test. Do you plan to add it in a separate PR?

EdDev:
> The change lgtm. Would love to see e2e test. Do you plan to add it in a separate PR?

Thanks!
I plan to add them here once I arrange the current mess.

EdDev:
#3460 introduces tests for the IPv4 CIDR scenario.
Unfortunately, doing the same for IPv6 is more challenging and requires modifications to the existing code (i.e. refactoring).

EdDev:
> Can you add tests?

I commented about it (see above). The current implementation does not allow doing it without a refactor effort. However, https://github.com/kubevirt/kubevirt/pull/3454 may do the job and then it will be easy.

rmohr:
> I commented about it (see above). The current implementation does not allow doing it without a refactor effort. However, #3454 may do the job and then it will be easy.

Ok.

/hold

Let's wait for tests. Feel free to cancel the hold once they are there.

tests/network/vmi_networking.go
```diff
@@ -729,28 +732,31 @@ var _ = SIGDescribe("[Serial][rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com]
...
+				clientVMI, err = fedoraMasqueradeVMI([]v1.Port{}, networkCIDR)
 				Expect(err).ToNot(HaveOccurred())
 				clientVMI, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(clientVMI)
 				Expect(err).ToNot(HaveOccurred())
 				clientVMI = tests.WaitUntilVMIReady(clientVMI, console.LoginToFedora)
 
-				Expect(configureIpv6(clientVMI)).To(Succeed(), "failed to configure ipv6 on client vmi")
+				if networkCIDR == "" {
+					networkCIDR = api.DefaultVMIpv6CIDR
+				}
```
maiqueb, line 748: IIUC, you're doing this to test the default value. Good.

I don't understand why you would change the value of an input parameter though. I'd have something like

```golang
expectedNetworkCIDR := networkCIDR
if networkCIDR == "" {
    expectedNetworkCIDR = api.DefaultVMIpv6CIDR
}
```
and compare against that new variable. It is more explicit and readable, does not change the value of an input parameter (something which I consider a bad practice), and cost a little bit of memory.

Will not make a fuss over this though.

api/openapi-spec/swagger.json
```diff
@@ -10765,6 +10765,10 @@
     "description": "Represents the stock pod network interface.",
     "type": "object",
     "properties": {
+     "vmIpv6NetworkCIDR": {
+      "description": "IPv6 CIDR for the vm network. Defaults to fd10:0:2::/120 if not specified.",
```
maiqueb, line 10769: The `if not specified` is redundant. `Defaults to` only has meaning when the parameter is not specified.

Not sure what's worse, being redundant or break consistency (the `vmNetworkCIDR` parameter features the same wording).

pkg/virt-operator/resource/generate/components/validations_generated.go
```diff
@@ -2807,6 +2807,9 @@ var CRDsValidation map[string]string = map[string]string{
                       pod:
                         description: Represents the stock pod network interface.
                         properties:
+                          vmIpv6NetworkCIDR:
+                            description: IPv6 CIDR for the vm network. Defaults to fd10:0:2::/120 if not specified.
```
maiqueb, line 2811: Ditto.

maiqueb (changes requested):
Must remove the focus instruction in the tests; all other comments are opinionated.

tests/network/vmi_networking.go
```diff
@@ -729,28 +732,31 @@ var _ = SIGDescribe("[Serial][rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com]
 				table.Entry("with custom CIDR [IPv4]", []v1.Port{}, "10.10.10.0/24"),
 			)
 
-			table.DescribeTable("IPv6", func(ports []v1.Port) {
+			table.FDescribeTable("IPv6", func(ports []v1.Port, networkCIDR string) {
```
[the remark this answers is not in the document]
EdDev: Done

tests/network/vmi_networking.go
EdDev: It is actually worse, I should not have used it here.
Fixed it now by passing it to where it is really needed. Still I update the input arg to a default, it feels natural to me, it is just a way to implement defaults in Go.

api/openapi-spec/swagger.json
EdDev: As this is part of the API and potential documentation is generated from it, I prefer to stay safe and use the current format which mimics IPv4. But I do agree with you.
I will try to propose the text change in a separate PR to align both and see how it goes.

maiqueb: Fair enough; thanks.

tests/network/vmi_networking.go
maiqueb: I don't think so. Mutating an input argument is, imo, a bad idea, and bad practice. 

But hey, if the language cared about that, they'd have not allowed for you to mutate variables that are not declared w/ var. 

I'd just prefer to have immutable constructs everywhere; as I said, your call, not making a fuss over this.

EdDev: It is one of the ways to implement a "default" and I have not seen anyone reasoning why this is not allowed.
In C/C++ you have the `const` keyword, in others you can have a default in the signature. Here we can do this.

If you think that it is bad practice, I would love to know the arguments for it. You can take this case as an example if you like.

staging/src/kubevirt.io/client-go/api/v1/schema.go
```diff
@@ -1348,6 +1348,10 @@ type PodNetwork struct {
...
```
rmohr, line 1354: nit: `json:vmIPv6NetworkCIDR` should be the more correct field name.

EdDev: I agree, but this is following the IPv4 format and I am not sure if it is better to break the consistency here.

tests/network/vmi_networking.go
maiqueb: > It is one of the ways to implement a "default" and I have not seen anyone reasoning why this is not allowed.
In C/C++ you have the const keyword, in others you can have a default in the signature. Here we can do this.

I think we're actually saying the same :) 

Above I wrote: "if the language cared about that, they'd have not allowed for you to mutate variables that are not declared w/ var."

> If you think that it is bad practice, I would love to know the arguments for it. You can take this case as an example if you like.

If you never mutate a variable, you will know at all times what it's value is. You can move code around more easily, since no matter where you put it, it'll have the same thing in it. Plenty more arguments about immutable variables / data structure can be found online.

But, again, your call. You want to mutate an input argument, I won't stand in your way. I am just pointing out that you could achieve the same thing in a safer way. Up to you.

EdDev: I agree that mutating an argument with different values in the same scope is problematic.
In this case, I think it is normalized in the begining and not in the middle, also not used for two different things (like the original code which was wrong).
In my opinion, with your reasoning in mind, the rule should be that a value of an argument should be stable from the first use. Until the first use, it can be adjusted, like with the default pattern.

staging/src/kubevirt.io/client-go/api/v1/schema.go
rmohr: I could not find an `Ipv4` field in the schema, could you point me to the field which you mean?

EdDev: Ahh... I am wrong here. I thought about something else completely. Sorry.
Changing...

tests/network/vmi_networking.go
maiqueb: > I agree that mutating an argument with different values in the same scope is problematic.
> In this case, I think it is normalized in the begining and not in the middle, also not used for two different things (like the original code which was wrong).
> In my opinion, with your reasoning in mind, the rule should be that a value of an argument should be stable from the first use. Until the first use, it can be adjusted, like with the default pattern.

I keep thinking it should be normalized at creation time, not when it is used. E.g. it's first use may move. And the function behavior will change. 

It's creation time may move, but the compiler will blow up on your face if you use it before creating it, so it is safer.

And IMHO, 'defaulting' things has a time do be done: at creation time, nor when you don't find a value in it. I guess this is another facet of the discussion we're having :) 

But again: your call; if the language was meant to be used as I'm describing, it would have a feature set to enforce it.

staging/src/kubevirt.io/client-go/api/v1/schema.go
EdDev: Done

rmohr: :+1:

tests/network/vmi_networking.go
EdDev: At this moment, I think it is fine as is. Having to name one variable/argument right is a hard, trying to name two that mean (almost) the same thing is even harder.

maiqueb: Fair enough; thanks.

maiqueb (approved):
Thanks.

PR #3386, opened 2020-05-07
Use IPv6 bind address instead of IPv4 one

oshoval:
Use IPv6 bind address instead of IPv4 one
    
Change migration virt-handler to virt-handler socket to listen
on [::] instead of 0.0.0.0, unless IPv6 is disabled.
    
According to standard, listening on [::] will also allow
incoming IPv4 connections, but not vice versa.
It does work without it because Golang net library use implicity
an IPv6 socket, but its not readable and not maintainable.
    
This PR still considers IPv4 mapping is enabled.
    
Depends on: #3392

Signed-off-by: Or Shoval oshoval@redhat.com

```release-note
None
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -1,6 +1,8 @@
 package util
 
 import (
+	"os/exec"
+
 	v1 "kubevirt.io/client-go/api/v1"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/util/util.go
```diff
@@ -29,3 +31,18 @@ func IsGPUVMI(vmi *v1.VirtualMachineInstance) bool {
 	}
 	return false
 }
+
+// IsIpv6Disabled returns if IPv6 is disabled according sysctl
+func IsIpv6Disabled() bool {
```
oshoval, line 36: Might need to be refactor as Vladik suggested on #3396
investigating still why using file read doesn't work on prow, 
and then its possible to move it to new net lib at util as Alona suggested
and there are some more re-factoring possible, that i saw but not related to this PR.
in any case, imo it shouldn't block. i will have more info soon, and act accordingly

oshoval:
Closing in favor of other PR, it will be handled there, with its counter part.

PR #3392, opened 2020-05-10
Fix IPv6 migration

oshoval:
When dealing with IPv6 addresses, the standard net package
is expecting an IPv6 address format that includes
square brackets (e.g [::1]).
By normalizing the IPv6 address, the whole migration flow succeeds
and the currently skipped functional tests are therefore run normally.

No unit test was added for the IPv6 scenario because
the existing tests are dependent on OS capabilities and the travis-ci
is not supporting IPv6 at the moment [1].
We are still safe as the IPv6 code is covered by the functional tests.

1. https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system

```release-note
None
```

### DIFF
--- a/pkg/virt-handler/migration-proxy/migration-proxy.go
+++ b/pkg/virt-handler/migration-proxy/migration-proxy.go
@@ -30,6 +30,8 @@ import (
 	"strings"
 	"sync"
 
+	netutils "k8s.io/utils/net"
+
 	"kubevirt.io/client-go/log"
 )
 
@@ -213,6 +215,10 @@ func (m *migrationProxyManager) StopTargetListener(key string) {
 func (m *migrationProxyManager) StartSourceListener(key string, targetAddress string, destSrcPortMap map[string]int, baseDir string) error {
 	m.managerLock.Lock()
 	defer m.managerLock.Unlock()
+
+	if netutils.IsIPv6String(targetAddress) {
+		targetAddress = "[" + targetAddress + "]"
+	}
 
 	isExistingProxy := func(curProxies []*migrationProxy, targetAddress string, destSrcPortMap map[string]int) bool {
 		if len(curProxies) != len(destSrcPortMap) {
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -406,8 +406,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 		})
 		Context("with a Cirros disk", func() {
 			It("[test_id:4113]should be successfully migrate with cloud-init disk with devices on the root bus", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
-
 				vmi := tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskCirros))
 				vmi.Annotations = map[string]string{
 					v1.PlacePCIDevicesOnRootComplex: "true",
@@ -450,8 +448,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 			})
 
 			It("[test_id:1783]should be successfully migrated multiple times with cloud-init disk", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
-
 				vmi := tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskCirros))
 				tests.AddUserData(vmi, "cloud-init", "#!/bin/bash\necho 'hello'\n")
 
@@ -504,8 +500,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 			})
 
 			It("[test_id:3237]should complete a migration", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
-
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
 				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse(fedoraVMSize)
 
@@ -540,7 +534,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 		})
 		Context("with setting guest time", func() {
 			It("[test_id:4114]should set an updated time after a migration", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
 				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse(fedoraVMSize)
 				vmi.Spec.Domain.Devices.Rng = &v1.Rng{}
@@ -681,8 +674,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				}
 			})
 			It("[test_id:3239]should reject a migration of a vmi with a non-shared data volume", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
-
 				dataVolume := tests.NewRandomDataVolumeWithHttpImport(tests.GetUrl(tests.AlpineHttpUrl), tests.NamespaceTestDefault, k8sv1.ReadWriteOnce)
 				vmi := tests.NewRandomVMIWithDataVolume(dataVolume.Name)
 
@@ -874,7 +865,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 			var pvName string
 			var vmi *v1.VirtualMachineInstance
 			BeforeEach(func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
+				tests.SkipNFSTestIfRunnigOnKindInfra()
 				pvName = "test-nfs" + rand.String(48)
 				// Prepare a NFS backed PV
 				By("Starting an NFS POD")
@@ -1082,6 +1073,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
 			It("[test_id:2227]should abort a vmi migration without progress", func() {
+				tests.SkipStressTestIfRunnigOnKindInfra()
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
 				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse("1Gi")
 
@@ -1203,8 +1195,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 			}
 
 			table.DescribeTable("should be able to cancel a migration", func(createVMI vmiBuilder) {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
-
 				vmi, dv := createVMI()
 				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse(fedoraVMSize)
 				defer deleteDataVolume(dv)
@@ -1246,8 +1236,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				table.Entry("[test_id:2731] with OCS Disk", newVirtualMachineInstanceWithFedoraOCSDisk),
 			)
 			It("[test_id:3241]should be able to cancel a migration right after posting it", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
-
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
 				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse(fedoraVMSize)
 
@@ -1285,8 +1273,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 	Context("with sata disks", func() {
 
 		It("[test_id:1853]VM with containerDisk + CloudInit + ServiceAccount + ConfigMap + Secret", func() {
-			tests.SkipMigrationTestIfRunnigOnKindInfra()
-
 			configMapName := "configmap-" + rand.String(5)
 			secretName := "secret-" + rand.String(5)
 
@@ -1376,8 +1362,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 			})
 
 			It("[test_id:3244]should block the eviction api while a slow migration is in progress", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
-
 				vmi = fedoraVMIWithEvictionStrategy()
 
 				By("Starting the VirtualMachineInstance")
@@ -1433,7 +1417,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 
 			Context("with node tainted during node drain", func() {
 				It("[test_id:2221] should migrate a VMI under load to another node", func() {
-					tests.SkipMigrationTestIfRunnigOnKindInfra()
 					tests.SkipIfVersionBelow("Eviction of completed pods requires v1.13 and above", "1.13")
 
 					vmi = fedoraVMIWithEvictionStrategy()
@@ -1486,7 +1469,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				})
 
 				It("[test_id:2222] should migrate a VMI when custom taint key is configured", func() {
-					tests.SkipMigrationTestIfRunnigOnKindInfra()
 					tests.SkipIfVersionBelow("Eviction of completed pods requires v1.13 and above", "1.13")
 
 					vmi = cirrosVMIWithEvictionStrategy()
@@ -1541,7 +1523,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				}, 400)
 
 				It("[test_id:2224] should handle mixture of VMs with different eviction strategies.", func() {
-					tests.SkipMigrationTestIfRunnigOnKindInfra()
 					tests.SkipIfVersionBelow("Eviction of completed pods requires v1.13 and above", "1.13")
 
 					vmi_evict1 := cirrosVMIWithEvictionStrategy()
@@ -1665,8 +1646,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 		Context("with multiple VMIs with eviction policies set", func() {
 
 			It("[test_id:3245]should not migrate more than two VMIs at the same time from a node", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
-
 				var vmis []*v1.VirtualMachineInstance
 				for i := 0; i < 4; i++ {
 					vmi := cirrosVMIWithEvictionStrategy()
--- a/pkg/virt-handler/migration-proxy/BUILD.bazel
+++ b/pkg/virt-handler/migration-proxy/BUILD.bazel
@@ -5,7 +5,10 @@ go_library(
     srcs = ["migration-proxy.go"],
     importpath = "kubevirt.io/kubevirt/pkg/virt-handler/migration-proxy",
     visibility = ["//visibility:public"],
-    deps = ["//staging/src/kubevirt.io/client-go/log:go_default_library"],
+    deps = [
+        "//staging/src/kubevirt.io/client-go/log:go_default_library",
+        "//vendor/k8s.io/utils/net:go_default_library",
+    ],
 )
 
 go_test(
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -4663,15 +4663,21 @@ func IsRunningOnKindInfraIPv6() bool {
 	return strings.HasPrefix(provider, "kind-k8s-1.17.0-ipv6")
 }
 
+func SkipStressTestIfRunnigOnKindInfra() {
+	if IsRunningOnKindInfra() {
+		Skip("Skip stress test till issue https://github.com/kubevirt/kubevirt/issues/3323 is fixed")
+	}
+}
+
 func SkipPVCTestIfRunnigOnKindInfra() {
 	if IsRunningOnKindInfra() {
 		Skip("Skip PVC tests till PR https://github.com/kubevirt/kubevirt/pull/3171 is merged")
 	}
 }
 
-func SkipMigrationTestIfRunnigOnKindInfra() {
+func SkipNFSTestIfRunnigOnKindInfra() {
 	if IsRunningOnKindInfra() {
-		Skip("Skip migration tests till PR https://github.com/kubevirt/kubevirt/pull/3221 is merged")
+		Skip("Skip NFS tests till issue https://github.com/kubevirt/kubevirt/issues/3322 is fixed")
 	}
 }
 


oshoval:
Imo this is the only thing that needs to be fixed for migration
The other issue of making the migration robust to cases of bindv6only,
by using ::1 and :: unless ipv6 is disabled (and then use ipv4 127.0.0.1 and 0.0.0.0)
should be handled in a separated PR which i work on right now

this links explains why ipv4 address work also for ipv6
https://stackoverflow.com/questions/6343747/ipv6-socket-creation
since ipv4 is a subset of ipv6

in any case i work on the other PR, at this very moment and this one is a cherry-pick of the base commit of that one, and which of the PR that will be chosen is the Team decision

Thanks

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -317,12 +323,16 @@ func NewTargetProxy(tcpBindAddress string, tcpBindPort int, serverTLSConfig *tls
 
 }
 
+func isLoopbackAddress(ipAddress string) bool {
+	return (strings.Contains(ipAddress, "127.0.0.1") || strings.Contains(ipAddress, "[::1]"))
```
ormergi, line 327: I am trying to remove hard-coded addresses form the code,
I will appreciate if you will export those addresses to variables :)

oshoval: np, to put them as a var in that specific function is good enough?

ormergi: Thanks!
We can add those ports to this file [const block ](https://github.com/kubevirt/kubevirt/blob/85435dd0711fb4773885ccfad27f9627ed325279/pkg/virt-handler/migration-proxy/migration-proxy.go#L38) and then pass them that function

AlonaKaplan: @oshoval please explain why this change is needed while the other changes related to using ::1 instead of 127.0.0.1 were dropped from the PR?

oshoval: Alona
Imo the function should be robust to both derivates of ::1 
which are both the same thing in different notation.

Also during development i saw a case where i had to use it, i will double check now, if its still necessary because there were changes from that point of time.

EdDev: > the other changes are purely for supported futuristic configuration bindv6only

Are these facts or assumptions? How do you know if the needed works with IPv6 and not IPv4?
You need to explain us how it works so we will be convinced that this point of yours is correct.
`bind6only` is totally irrelevant here, that is an implementation of socket details and you have no idea how the application is actually listening to the address (they may be listening to one or two sockets).

oshoval:
> It seems that this PR changes the proxy listener but it does not talk about the client side or the target VM.
> I am not familiar how this proxy works, but you need to explain and convince us why this makes sense to just listen to whatever was asked (I guess it is the IPv6 address).
> 
> I expect all this to be explained in the commit message so as a reviewer I'll understand why this is enough.
> 
> In addition, I see no unit tests here. Have there been unit tests covering this part? If so, an IPv6 flow is in order.

@EdDev 
It is another notation of the loopback,
just loopback are accepted without TLS
see please 
https://github.com/kubevirt/kubevirt/pull/3392/files#diff-e88e0fbecb2c3ab1c8c08d12a1c8b4f6L328

about
https://github.com/kubevirt/kubevirt/pull/3392#discussion_r422619425
I tested what Alona said, and indeed it isnt mandatory, it is needed when ::1 is used
(and i will double check again in that scope)
The reason was notation

All of the comments here of Alona / Edy / Or Mergi are about that function
so please decide if i should remove this function and address all of those comments in the PR that will relate to it ?

pkg/virt-handler/migration-proxy/migration-proxy.go
oshoval: see please my first post
https://stackoverflow.com/questions/6343747/ipv6-socket-creation
bindv6only is not just socket wise, its kernel wise
cat /proc/sys/net/ipv6/bindv6only

this link was groomed with several people
that ::1 can be used as long as bindv6only=0 
this is what was done on ovirt according Alona

and we dont have a configuration to test it,
imo it isnt related to this PR because its two different things:
1. support migration on ipv6 (rest of the changes here beside this function of isloopback that i can remove and move to the other PR)
2. the other PR is the one that will support a mode (bindv6only) that we even dont support and can test right now.

EdDev:
> > It seems that this PR changes the proxy listener but it does not talk about the client side or the target VM.
> > I am not familiar how this proxy works, but you need to explain and convince us why this makes sense to just listen to whatever was asked (I guess it is the IPv6 address).
> > I expect all this to be explained in the commit message so as a reviewer I'll understand why this is enough.
> > In addition, I see no unit tests here. Have there been unit tests covering this part? If so, an IPv6 flow is in order.
> 
> @EdDev
> It is another notation of the loopback,
> just loopback are accepted without TLS
> see please
> https://github.com/kubevirt/kubevirt/pull/3392/files#diff-e88e0fbecb2c3ab1c8c08d12a1c8b4f6L328

I do not understand to which of the points you answered here.

> 
> about
> [#3392 (comment)](https://github.com/kubevirt/kubevirt/pull/3392#discussion_r422619425)
> I tested what Alona said, and indeed it isnt mandatory, it is needed when ::1 is used
> (and i will double check again in that scope)
> The reason was notation
> 
> All of the comments here of Alona / Edy / Or Mergi are about that function
> so please decide if i should remove this function and address all of those comments in the PR that will relate to it ?

As we do not understand "why" this whole thing works, we cannot recommend anything.
If you just fix the `[]` addition, you will still need to explain why that is the only thing needed. Why the loopback part does not need changing and why the other places where `127.0.0.1` is mentioned is not requiring changes as well.

oshoval:
> > > It seems that this PR changes the proxy listener but it does not talk about the client side or the target VM.
> > > I am not familiar how this proxy works, but you need to explain and convince us why this makes sense to just listen to whatever was asked (I guess it is the IPv6 address).
> > > I expect all this to be explained in the commit message so as a reviewer I'll understand why this is enough.
> > > In addition, I see no unit tests here. Have there been unit tests covering this part? If so, an IPv6 flow is in order.
> > 
> > 
> > @EdDev
> > It is another notation of the loopback,
> > just loopback are accepted without TLS
> > see please
> > https://github.com/kubevirt/kubevirt/pull/3392/files#diff-e88e0fbecb2c3ab1c8c08d12a1c8b4f6L328
> 
> I do not understand to which of the points you answered here.

It answers about
" I am not familiar how this proxy works, but you need to explain and convince us why this makes sense to just listen to whatever was asked (I guess it is the IPv6 address)."
only loobpack is the one that is acceptable without TLS,
and the link show the if conditions and the error in case its not TLS
and not loopback.
> 
> > about
> > [#3392 (comment)](https://github.com/kubevirt/kubevirt/pull/3392#discussion_r422619425)
> > I tested what Alona said, and indeed it isnt mandatory, it is needed when ::1 is used
> > (and i will double check again in that scope)
> > The reason was notation
> > All of the comments here of Alona / Edy / Or Mergi are about that function
> > so please decide if i should remove this function and address all of those comments in the PR that will relate to it ?
> 
> As we do not understand "why" this whole thing works, we cannot recommend anything.
> If you just fix the `[]` addition, you will still need to explain why that is the only thing needed. Why the loopback part does not need changing and why the other places where `127.0.0.1` is mentioned is not requiring changes as well.

for this see please for that
https://github.com/kubevirt/kubevirt/pull/3392#discussion_r422623830

pkg/virt-handler/migration-proxy/migration-proxy.go
EdDev: > see please my first post and this link there
> https://stackoverflow.com/questions/6343747/ipv6-socket-creation
> bindv6only is not just socket wise, its kernel wise
> cat /proc/sys/net/ipv6/bindv6only

That system wide parameter overwrites the ability for a socket to be opened as a dual socket. How the application opens the socket is not explaining me how all the dots connect in this flow.

> 
> this link was groomed with several people
> that ::1 can be used as long as bindv6only=0
> this is what was done on ovirt according Alona
> (use ::1 unless ipv6 is disabled, hence also case of bindv6only=1
> is supported, because we wont use 127.0.0.1 / 0.0.0.

`bind6only` is irrelevant IMO. The application may open sockets as a single stack or as dual stack, either way, it should work with IPv6.

> 
> and we dont have a configuration to test it,
> imo it isnt related to this PR because its two different things:
> 
>     1. support migration on ipv6 (rest of the changes here beside this function of isloopback that i can remove and move to the other PR)
> 
>     2. the other PR is the one that will support a mode (bindv6only) that we even dont support and can test right now.

I only care at the moment about the correct support of IPv6 in the migration flow.
The current change normalizes the ipv6 address for the listener socket and adds loopback detection in case it is an IPv6 loopback.
It does all this in the proxy listener and ignores all other code that uses IPv4 only addresses (like 127.0.0.1).
I am looking for an explanation why the proxy listener is the only one needing change.
I think you also mentioned that even without the loopback change it works, so the same questions pops up for that.

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -214,6 +216,10 @@ func (m *migrationProxyManager) StartSourceListener(key string, targetAddress st
 	m.managerLock.Lock()
 	defer m.managerLock.Unlock()
 
+	if netutils.IsIPv6String(targetAddress) {
```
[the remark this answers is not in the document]
oshoval: @AlonaKaplan 
@EdDev 
for the 2nd PR this function is mandatory of course because we do use there ::1
which is different notation of 127.0.0.1 as 
IPv4 is subset of IPv6

i rechecked there that its indeed mandatory and it is
according the decisions on this PR
i will know if to understand all the comments and where
but first lets please focus on those questions

pkg/virt-handler/migration-proxy/migration-proxy.go
oshoval: the virt-handlers are the only place in migration that use IPv6
there are address of the pods that are used in the migration
and they are the ones that need the normalize [].

i can fail the migration and show that error directly,
i will enclose soon if needed.

indeed the loopback detection isnt needed, because we still use 127.0.0.1 in this PR
and as 127.0.0.1 is subset of IPv6 it is fine.

Is it fine that i will remove that func from this PR, and focus on the normalizing ?
The func will be mandatory in the 2nd PR (verified right now), and then i will handle all the comments of it.

oshoval:
```
{"component":"virt-handler","level":"error","msg":"unable to create outbound leg of proxy to host fd00:10:244::37:39173","pos":"migration-proxy.go:392","reason":"dial tcp: address fd00:10:244::37:39173: too many colons in address","timestamp":"2020-05-10T12:42:02.271518Z"}
```
this is the error that line
https://github.com/kubevirt/kubevirt/pull/3392/files#diff-e88e0fbecb2c3ab1c8c08d12a1c8b4f6R220
fixed when handler of the source migration node tries to connect to the dst migration handler

oshoval: https://access.redhat.com/solutions/4291941
does this answer please ?

AlonaKaplan:
@oshoval I'm very confused. But if I understand you correctly, you want this PR to deal only with the missing []. If so, please update the PR message and the code accordingly. 
As the whole discussion regarding 127.0.0.1 and 0.0.0.0 addresses the migration proxy is listening to is in this PR and we don't fully understand how ipv6 migration works with keeping those addresses in ipv4 format (especially I don't understand how the proxy that is listening to 0.0.0.0 catches ipv6 requests) I think it is a shame to move it to a different PR. But if you still decide to do it please update the PR code, message and title accordantly.

oshoval:
> @oshoval I'm very confused. But if I understand you correctly, you want this PR to deal only with the missing []. If so, please update the PR message and the code accordingly.
> As the whole discussion regarding 127.0.0.1 and 0.0.0.0 addresses the migration proxy is listening to is in this PR and we don't fully understand how ipv6 migration works with keeping those addresses in ipv4 format (especially I don't understand how the proxy that is listening to 0.0.0.0 catches ipv6 requests) I think it is a shame to move it to a different PR. But if you still decide to do it please update the PR code, message and title accordantly.

I dont want to move it, in a "shame" mode, i truly believe its 2 different discussions,
thats all, i will shortly write more info, and according your decision will do what needed.

oshoval:
> > @oshoval I'm very confused. But if I understand you correctly, you want this PR to deal only with the missing []. If so, please update the PR message and the code accordingly.
> > As the whole discussion regarding 127.0.0.1 and 0.0.0.0 addresses the migration proxy is listening to is in this PR and we don't fully understand how ipv6 migration works with keeping those addresses in ipv4 format (especially I don't understand how the proxy that is listening to 0.0.0.0 catches ipv6 requests) I think it is a shame to move it to a different PR. But if you still decide to do it please update the PR code, message and title accordantly.
> 
> I don't want to move it, in a "shame" mode, i truly believe and know its 2 different independent discussions,
> thats all, i will shortly write more info, and according your decision will do what needed.
> I even divided it in order to reduce noise and that the parts would be as clear as possible,
> else some questions might wont be raised.

This is what i know so far on why we / i might need to rethink about the changes
in the following PRs, hope it sheds some light.

There are multiple communication channels in the migration process.
Some are unix socket, which arent affect in case the pods are IPv4 or IPv6
there are used for internally communicate in the node, and to connect to the libvirtd,
which actually perform the migration once all the data is set.

https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-handler/migration-proxy/migration-proxy.go#L133
This is the function that use 0.0.0.0, and this is its comment.
// Target proxy listens on a tcp socket and pipes to a libvirtd unix socket
See please the next section about manager.go for more info.

This is about the 0.0.0.0
It isnt affected by the pod IP from what i understood.

Lets leave a moment isLoopbackAddress
https://github.com/kubevirt/kubevirt/pull/3392/files#diff-e88e0fbecb2c3ab1c8c08d12a1c8b4f6R326
Assuming its needed only for the robustness feature, which i am trying to say is not part of this PR,
just because its a new feature and not a fix of this PR scope.

The occurences at
https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/manager.go#L400
are for overcoming issue of libvirt and unix sockets, but still they are for local communication,
Not virt-handler to virt-handler communication which is the one that is sensitive to the IPv6 or IPv4 of the pods.
// Start local migration proxy.
// Right now Libvirt won't let us perform a migration using a unix socket, so
// we have to create a local host tcp server (on port 22222) that forwards the traffic
// to libvirt in order to trick libvirt into doing what we want.
// This also creates a tcp server for each additional direct migration connections
// that will be proxied to the destination pod

The 4 places in this function is for serving the above comment, and for connecting to the hypervisor,
which is again not connected to the virt-handler IP family.

Two more places which do need invetigation but not as part of this PR imo, 
are
https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/manager.go#L749
https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/manager.go#L857
it updates the /etc/hosts file, which i saw that have the IPv4 / IPv6 localhost address already, need to understand how,
and if there are cases when they will both be needed, from what i started to read, they both needed,
so the change there need to be researched. However again its for hostname resolving of localhost, not for the 
virt-handler that are the IP sensitive.

The last place of 127.0.0.1 is again not connected to virt-handlers communication (the IP sensitive one),
but to local libvirt communication via the proxy
https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/manager.go#L866

oshoval:
> > > @oshoval I'm very confused. But if I understand you correctly, you want this PR to deal only with the missing []. If so, please update the PR message and the code accordingly.
> > > As the whole discussion regarding 127.0.0.1 and 0.0.0.0 addresses the migration proxy is listening to is in this PR and we don't fully understand how ipv6 migration works with keeping those addresses in ipv4 format (especially I don't understand how the proxy that is listening to 0.0.0.0 catches ipv6 requests) I think it is a shame to move it to a different PR. But if you still decide to do it please update the PR code, message and title accordantly.
> > 
> > 
> > I don't want to move it, in a "shame" mode, i truly believe and know its 2 different independent discussions,
> > thats all, i will shortly write more info, and according your decision will do what needed.
> > I even divided it in order to reduce noise and that the parts would be as clear as possible,
> > else some questions might wont be raised.
> 
> This is what i know so far on why we / i might need to rethink about the changes
> in the following PRs, hope it sheds some light.
> 
> There are multiple communication channels in the migration process.
> Some are unix socket, which aren't affected in case the pods are IPv4 or IPv6.
> They are used for internally communicate in the node, and to connect to the libvirtd,
> which actually perform the migration once all the data is set.
> 
> https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-handler/migration-proxy/migration-proxy.go#L133
> This is the function that use 0.0.0.0, and this is its comment.
> // Target proxy listens on a tcp socket and pipes to a libvirtd unix socket
> See please the next section about manager.go for more info.
> 
> This is about the 0.0.0.0
> It isnt affected by the pod IP from what i understood.
> 
> Lets leave a moment isLoopbackAddress
> https://github.com/kubevirt/kubevirt/pull/3392/files#diff-e88e0fbecb2c3ab1c8c08d12a1c8b4f6R326
> Assuming its needed only for the robustness feature, which i am trying to say is not part of this PR,
> just because its a new feature and not a fix of this PR scope.
> 
> The occurrences at
> https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/manager.go#L400
> are for overcoming issue of libvirt and unix sockets, but still they are for local communication,
> Not virt-handler to virt-handler communication which is the one that is sensitive to the IPv6 or IPv4 of the pods.
> // Start local migration proxy.
> // Right now Libvirt won't let us perform a migration using a unix socket, so
> // we have to create a local host tcp server (on port 22222) that forwards the traffic
> // to libvirt in order to trick libvirt into doing what we want.
> // This also creates a tcp server for each additional direct migration connections
> // that will be proxied to the destination pod
> 
> The 4 places in this function is for serving the above comment, and for connecting to the hypervisor,
> which is again not connected to the virt-handler IP family.
> 
> Two more places which do need invetigation but not as part of this PR imo,
> are
> https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/manager.go#L749
> https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/manager.go#L857
> it updates the /etc/hosts file, which i saw that have the IPv4 / IPv6 localhost address already, need to understand how,
> and if there are cases when they will both be needed, from what i started to read, they both needed,
> so the change there need to be researched. However again its for hostname resolving of localhost, not for the
> virt-handler that are the IP sensitive.
> 
> The last place of 127.0.0.1 is again not connected to virt-handlers communication (the IP sensitive one),
> but to local libvirt communication via the proxy
> https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/manager.go#L866

https://github.com/kubevirt/kubevirt/blob/b488ebd49594a8c30f7bf7755f10255d7ab2425c/pkg/virt-handler/migration-proxy/migration-proxy.go#L310

See please that those that use this function are unix socket so IP doesn't matter at all here, does it ?

AlonaKaplan:
Thanks for the long explanation. But I still don't understand how it works. Afaik, please correct me if I'm wrong, if a server is listening on ipv4 address, you cannot connect to it using ipv6 address.

So to make sure keeping the ipv4 addresses is fine I need to see both of the sides. The client and the server.
I'm not fully familiar with the migration flow and why we are using a proxy. But the thing that bothers me the most is the proxy that is listening on 0.0.0.0. I would love to see the code of the client connection to this proxy to be sure it is always using ipv4 address as well, although we are in ipv6 cluster.

oshoval:
> Thanks for the long explanation. But I still don't understand how it works. Afaik, please correct me if I'm wrong, if a server is listening on ipv4 address, you cannot connect to it using ipv6 address.
> 
> So to make sure keeping the ipv4 addresses is fine I need to see both of the sides. The client and the server.
> I'm not fully familiar with the migration flow and why we are using a proxy. But the thing that bothers me the most is the proxy that is listening on 0.0.0.0. I would love to see the code of the client connection to this proxy to be sure it is always using ipv4 address as well, although we are in ipv6 cluster.

Thanks for reading, 
please see this
https://unix.stackexchange.com/questions/419880/connecting-to-ip-0-0-0-0-succeeds-how-why
i think the IP here isnt an issue due to this

Also lets assume a moment it was, since its on the same host (do we agree?)
and as long as IPv4-mapped address feature isn't disabled it would work
see please last paragram
https://access.redhat.com/solutions/4291941

AlonaKaplan:
> > Thanks for the long explanation. But I still don't understand how it works. Afaik, please correct me if I'm wrong, if a server is listening on ipv4 address, you cannot connect to it using ipv6 address.
> > So to make sure keeping the ipv4 addresses is fine I need to see both of the sides. The client and the server.
> > I'm not fully familiar with the migration flow and why we are using a proxy. But the thing that bothers me the most is the proxy that is listening on 0.0.0.0. I would love to see the code of the client connection to this proxy to be sure it is always using ipv4 address as well, although we are in ipv6 cluster.
> 
> Thanks for reading,
> please see this
> https://unix.stackexchange.com/questions/419880/connecting-to-ip-0-0-0-0-succeeds-how-why
> i think the IP here isnt an issue due to this

Please explain why the ip is not an issue due to this and what are you talking about when you say this. Iiuc, we are listening on 0.0.0.0. Which means (according to the link you provided) any ipv4 address will be accepted. And that's exactly what I'm worried about, do we connect to this proxy server using only ipv4 addresses? I couldn't find any reference in the link regarding connecting to 0.0.0.0 with ipv6 so I cannot understand what answer should I get in this link.
> 
> Also lets assume a moment it was, since its on the same host (do we agree?)
> and as long as IPv4-mapped address feature isn't disabled it would work
> see please last paragraph
> https://access.redhat.com/solutions/4291941
This link is saying that listening over ipv6 localhost (::1) or any host (::) will except ipv4 connections as well. Well, that's exactly the code you removed from the pr (using ::1 and :: for both ipv6 and ipv4). Currently the pr contains listening only on the ipv4 addresses. Does listening on 127.0.0.1 will receive connection to ::1? The link doesn't answer this question so I don't understand how it helps to our discussion.

Regarding the fact that it is on the same host, again I don't understand why it is relevant. If you have a process that is listening on 127.0.0.1:2222 and another process that is connecting to [::]:2222. Would it work?

oshoval:
Not quoting because it adds noise, referring to last topic, trying to wrap the answers,
please say what do you think, thanks

Communication channels in migration:
SRC POD ENV (migration unix socket) <-> HOST ENV (tcp client) <--*--> (tcp server) HOST ENV <-> TARGET POD ENV (libvirtd unix socket)

The reason 0.0.0.0 / 127.0.0.1 (handler to handler, marked with *) works is that golang net library,
when listening to "tcp" [4] which is our case, and ipv4 mapping enabled,
it creates IPv6 socket [1], and the address therefore act as INADDR6_ANY and so on.

Call stack if desired:
[2] https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-handler/migration-proxy/migration-proxy.go#L135
[3] https://github.com/kubevirt/kubevirt/blob/b488ebd49594a8c30f7bf7755f10255d7ab2425c/pkg/virt-handler/migration-proxy/migration-proxy.go#L426
[4] https://github.com/kubevirt/kubevirt/blob/b488ebd49594a8c30f7bf7755f10255d7ab2425c/pkg/virt-handler/migration-proxy/migration-proxy.go#L326
[5] https://golang.org/src/net/dial.go?s=9855:9903#L308

[1] https://github.com/golang/go/blob/440f7d64048cd94cba669e16fe92137ce6b84073/src/net/ipsock_posix.go#L121

Taking this under consideration, this PR can be standalone
As Edy suggested, even that golang act like this,

A seperate PR will deal with using IPv6 address (unless ipv6_disable=1) explicity in order to:
1. Make it explicit to use IPv6 addresses and not relay on internal go implementation.
2. Be robust to case of bindv6only, where the mapping is disabled, and therefore it might fail.

I will remove from this PR the part about isLoopbackAddress, because its not an intergral part.

AlonaKaplan:
> Not quoting because it adds noise, referring to last topic, trying to wrap the answers,
> please say what do you think, thanks
> 
> Communication channels in migration:
> SRC POD ENV (migration unix socket) <-> HOST ENV (tcp client) <--*--> (tcp server) HOST ENV <-> TARGET POD ENV (libvirtd unix socket)
> 
> The reason 0.0.0.0 / 127.0.0.1 (handler to handler, marked with *) works is that golang net library,
> when listening to "tcp" [4] which is our case, and ipv4 mapping enabled,
> it creates IPv6 socket [1], and the address therefore act as INADDR6_ANY and so on.
> 
> Call stack if desired:
> [2] https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-handler/migration-proxy/migration-proxy.go#L135
> [3]
> 
> https://github.com/kubevirt/kubevirt/blob/b488ebd49594a8c30f7bf7755f10255d7ab2425c/pkg/virt-handler/migration-proxy/migration-proxy.go#L426
> 
> [4]
> 
> https://github.com/kubevirt/kubevirt/blob/b488ebd49594a8c30f7bf7755f10255d7ab2425c/pkg/virt-handler/migration-proxy/migration-proxy.go#L326
> 
> [5] https://golang.org/src/net/dial.go?s=9855:9903#L308
> 
> [1] https://github.com/golang/go/blob/440f7d64048cd94cba669e16fe92137ce6b84073/src/net/ipsock_posix.go#L121
> 
> Taking this under consideration, this PR can be standalone
> As Edy suggested, even that golang act like this,
> 
> A seperate PR will deal with using IPv6 address (unless ipv6_disable=1) explicity in order to:
> 
>     1. Make it explicit to use IPv6 addresses and not relay on internal go implementation.
> 
>     2. Be robust to case of bindv6only, where the mapping is disabled, and therefore it might fail.
> 
> 
> I will remove from this PR the part about isLoopbackAddress, because its not an intergral part.

@oshoval thanks for the detailed response. Now it makes sense and I understand how come it works with the ipv4 addresses.

Your plan sounds reasonable to me.

oshoval:
Thanks Alona, just see please i edited and added my view,
i will ask a maintainer for his opinion just to see the direction you prefer is fine please

maiqueb:
> > Not quoting because it adds noise, referring to last topic, trying to wrap the answers,
> > please say what do you think, thanks
> > Communication channels in migration:
> > SRC POD ENV (migration unix socket) <-> HOST ENV (tcp client) <--*--> (tcp server) HOST ENV <-> TARGET POD ENV (libvirtd unix socket)
> > The reason 0.0.0.0 / 127.0.0.1 (handler to handler, marked with *) works is that golang net library,
> > when listening to "tcp" [4] which is our case, and ipv4 mapping enabled,
> > it creates IPv6 socket [1], and the address therefore act as INADDR6_ANY and so on.
> > Call stack if desired:
> > [2] https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-handler/migration-proxy/migration-proxy.go#L135
> > [3]
> > https://github.com/kubevirt/kubevirt/blob/b488ebd49594a8c30f7bf7755f10255d7ab2425c/pkg/virt-handler/migration-proxy/migration-proxy.go#L426
> > 
> > [4]
> > https://github.com/kubevirt/kubevirt/blob/b488ebd49594a8c30f7bf7755f10255d7ab2425c/pkg/virt-handler/migration-proxy/migration-proxy.go#L326
> > 
> > [5] https://golang.org/src/net/dial.go?s=9855:9903#L308
> > [1] https://github.com/golang/go/blob/440f7d64048cd94cba669e16fe92137ce6b84073/src/net/ipsock_posix.go#L121
> > Taking this under consideration, this PR can be standalone
> > As Edy suggested, even that golang act like this,
> > A seperate PR will deal with using IPv6 address (unless ipv6_disable=1) explicity in order to:
> > ```
> > 1. Make it explicit to use IPv6 addresses and not relay on internal go implementation.

Why not just listen (only) to ::0 instead ? (not sure this what you mean here; if so, sorry to point the obvious).

> > 
> > 2. Be robust to case of bindv6only, where the mapping is disabled, and therefore it might fail.
> > ```
> > 
> > 
> > I will remove from this PR the part about isLoopbackAddress, because its not an intergral part.
> 
> @oshoval thanks for the detailed response. Now it makes sense and I understand how come it works with the ipv4 addresses.
> 
> Your plan sounds reasonable to me.

Does sound reasonable. I'd just change from `0.0.0.0` to `::0`, making explicit that the proxy is binding to v4 / v6 addresses.

AlonaKaplan:
Interesting link @SchSeba sent me https://docs.google.com/document/d/1zznvkK87y7UvOJZqs2DM4oWzx94P8Lg6YUb43whil3g/edit#heading=h.23ehcjjvl9pd (Section - Some IPv6 Porting Best Practices).

oshoval:
@maiqueb 
> Why not just listen (only) to ::0 instead ? (not sure this what you mean here; if so, sorry to point the obvious).
Its interesting, syntactic it will achieve the desired readability.
But if ipv6 will be disabled (which i dont know if there will be a feature like that, we can ask if desired),
it would either not work, or do implicit magics, Edy voted down for it.
This also address your 2nd comment.

AlonaKaplan:
> Thanks Alona, just see please i edited and added my view,
> i will ask a maintainer for his opinion just to see the direction you prefer is fine please

@oshoval I liked your response before you edited it:). I prefer you to introduce the second PR that will substitute 127.0.0.0 with ::1 and 0.0.0.0 with :: (in case ip6 is not disabled in the kernel).
Although it works with the ipv4 addresses (and now I even understand why) I think it will make things more clear if the ipv6 addresses will be used.

But we can consult with the kubevirt community about it.

maiqueb:
> > Thanks Alona, just see please i edited and added my view,
> > i will ask a maintainer for his opinion just to see the direction you prefer is fine please
> 
> @oshoval I liked your response before you edited it:). I prefer you to introduce the second PR that will substitute 127.0.0.0 with ::1 and 0.0.0.0 with :: (in case ip6 is not disabled in the kernel).
> Although it works with the ipv4 addresses (and now I even understand why) I think it will make things more clear if the ipv6 addresses will be used.
> 
> But we can consult with the kubevirt community about it.

+1 on being explicit. Relying on the implementation of a private method is worrying.

oshoval:
> > Thanks Alona, just see please i edited and added my view,
> > i will ask a maintainer for his opinion just to see the direction you prefer is fine please
> 
> @oshoval I liked your response before you edited it:). I prefer you to introduce the second PR that will substitute 127.0.0.0 with ::1 and 0.0.0.0 with :: (in case ip6 is not disabled in the kernel).
> Although it works with the ipv4 addresses (and now I even understand why) I think it will make things more clear if the ipv6 addresses will be used.
> 
> But we can consult with the kubevirt community about it.

didn't meant otherwise, as i wrote.
one nit, the 127.0.0.1 i need to see if indeed we need it, because there are internal sockets which arent tcp even, i will first understand what i am doing.
and about the /etc/hosts that is updated by 127.0.0.1
its a WA for libvirt that doesnt allow migration to the same host, a feature Vladik wrote.
so each part need to be understanded before doing please.

pkg/virt-handler/migration-proxy/migration-proxy.go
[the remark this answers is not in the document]
oshoval: about the first comment

it cant, because line 223 creates a lookup table with targetAddress, 
and later on the key is the formatted targetAddress, which must be as the one that was used
to build that table.
moving it will corrupt the logic, i tried that in the past.

pkg/virt-handler/migration-proxy/migration-proxy.go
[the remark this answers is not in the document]
oshoval: 1. https://golang.org/src/net/ip.go
the whole 127/8 is considered loopback i think,
i will check if there are alternatives that fit

   120  // IsLoopback reports whether ip is a loopback address.
   121  func (ip IP) IsLoopback() bool {
   122  	if ip4 := ip.To4(); ip4 != nil {
   123  		return ip4[0] == 127
   124  	}
   125  	return ip.Equal(IPv6loopback)
   126  }

pkg/virt-handler/migration-proxy/migration-proxy.go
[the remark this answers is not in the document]
EdDev: The `[]` wrapper is an address formatting that the netutils package requires for working with sockets.
On the other hand, the lookup check is about the IP identifier, which does not have to require those braces (as it does not work directly with the netutils). The lookup seems independent in this regard and I see no reason to enforce using braces there.
I guess the other option would have been to require in the API to represent the IPv6 address in the correct format in the first place, but I am not sure about the implications of this. We could explore it in the future.

At the moment I think we should do the minimum needed without affecting anything else, i.e. normalize the IPv6 address where is it a must.

Is something breaking if you move this very close to the listener?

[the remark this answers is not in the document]
oshoval: yes it breaks, because the key need always to be with [ ] or without [ ]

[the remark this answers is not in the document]
EdDev: > yes it breaks, because the key need always to be with [ ] or without [ ]
> consistency,
> it means to have two instance, one with and one without
> i think its not needed, the key imo can be with [ ]
> its just a key in just this function, but whatever you want

I'm trying to understand how come the address may be added or checked against one with braces and once without.

The problem with normalizing in the middle is that there is a good chance to miss things or that someone that adds another path int he future will also add these braces again. Therefore, the solution usually is to do it either in the first place it is read or as late as possible.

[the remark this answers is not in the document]
oshoval: keep in mind please, this is not part of this PR, can we resolve, and i will copy the link to the PR that do matter about it ?
so it wont cause noise here
i already removed the code from this PR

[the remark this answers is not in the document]
oshoval: first place it is read is 
https://github.com/kubevirt/kubevirt/pull/3392/files#diff-e88e0fbecb2c3ab1c8c08d12a1c8b4f6R243
it must be consistent with 
https://github.com/kubevirt/kubevirt/pull/3392/files#diff-e88e0fbecb2c3ab1c8c08d12a1c8b4f6R257

in line 243 (the first link), it use the instance that was created by line 257 (2nd link)
in the previous run

so maybe should we move it just above 242 

i always look in the line numbers of the right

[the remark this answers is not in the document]
oshoval: we can use 
https://golang.org/pkg/net/#JoinHostPort
and normalize in 2 places
line below 
https://github.com/kubevirt/kubevirt/pull/3392/files#diff-e88e0fbecb2c3ab1c8c08d12a1c8b4f6R229
and line below
https://github.com/kubevirt/kubevirt/pull/3392/files#diff-e88e0fbecb2c3ab1c8c08d12a1c8b4f6R257

it takes ip:port and makes [ip]:port , which is that we need to "listen" 
and what it saves in the proxy map, which is later must be the key we used to lookup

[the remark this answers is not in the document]
EdDev: The location is not ideal but it seems that there is a deeper problem with moving this down the stack and close to the net.Listen call.

We should investigate how to make this normalization immediately before the net.Listen call and make sure we are consistent in the code flow with one format.
In any case, this can be handled in a separate PR as it has the potential of touching other areas of the code.

oshoval:
> When dealing with IPv6 addresses, the standard net package is expecting an IPv6 address format that includes square brackets (e.g [::1]).
> By normalizing the IPv6 address, the whole migration flow succeeds and the currently skipped functional tests are therefore ran normally.

Uncle bob said 72 chars xD

oshoval:
/hold
adding a revert to check that the ipv6 lane pass without it, 
so putting this hold guard

oshoval:
the summary of this PR with the revert of fc32, there are 4 failing tests, trying to understand if they are flaky
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3392/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1260907237409296384

EdDev:
Can the functional tests that have been un-skipped passing now on CI? Is the current failing IPv6 lane affecting these tests?

oshoval:
/assign @vladikr 
Hi Vladik, can you review this one please?
This one fixes the migration, by normalization the IPv6 address.
A following PR will make the code more explicit about using IPv4 or IPv6 bind address,
without letting the Golang lib to implicity choose the socket family. 

Thanks

oshoval:
> Are the functional tests that have been un-skipped passing now on CI? Is the current failing IPv6 lane affecting these tests?

The functional ones that were caused by this bug, are unskipped and passing,
All the non passing are connected to Agent problems,
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3392/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1261737473550061569

tests/migration_test.go
```diff
@@ -1084,6 +1075,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 			})
 			It("[test_id:2227]should abort a vmi migration without progress", func() {
 				tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
+				tests.SkipStressTestIfRunnigOnKindInfra()
```
AlonaKaplan, line 1078: Can you please remove one of the skips? I guess the 'SkipMigrationTestIfRunnigOnKindInfra' should be removed.

oshoval: there are 2 issues that need to be fixed, they are independent 
this is why there are 2 lines, is it ok ?

AlonaKaplan: I added the 'SkipIpv6TestsFailingAfterVmImageUpgradeToFc32' since I thought it was caused by the upgrade. But since I see now you had this skip even before the upgrade, I understand now the test is flaky, unrelated to the upgrade.
I'm afraid that if the pressure issue would be fixed before the upgrade, the test will still be skipped (although there is a chance the upgrade didn't harm the test).

But since we don't actually know, you can keep both of them, up to you.

oshoval:
This PR needs #3429 Fix, so some of the tests will fail over and over,
this is why the bot is busy retesting

EdDev:
> This PR needs #3429 Fix, so some of the tests will fail over and over,
> this is why the bot is busy retesting

It was finally merged. You will need now to rebase.

EdDev:
Holding until a rebase.
/hold

oshoval:
@EdDev 
Rebased, removing hold, thanks
/hold cancel

oshoval:
Thanks Edy
@vladikr 
can you please re-approve ? a rebase was needed
thanks

oshoval:
/override pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6
checking if the bot allow me

PR #3396, opened 2020-05-11
Use IPv6 bind and loopback addresses instead of IPv4 ones

oshoval:
Add IP library
IP library implements API that allow supporting
IPv4 and IPv6 side by side, in an explicit and robust manner.

Change migration virt-handler to virt-handler socket to listen
on [::] instead of 0.0.0.0, unless IPv6 is disabled.

Change migration proxy socket to listen on [::1]
instead of 127.0.0.1, unless IPv6 is disabled.

According to standard, listening on [::] / [::1] will also allow
incoming IPv4 connections, if IPv4 mapping is enabled,
but not vice versa.
It does work without it because Golang net library use implicity
an IPv6 socket, but its not readable and not maintainable.

This change considers IPv4 mapping is enabled.

Depends on: #3392, #3429 

```release-note
None
```


davidvossel:
hey, can we get some more context on what we're aiming to solve with this PR?

oshoval:
> hey, can we get some more context on what we're aiming to solve with this PR?

Sure, 
https://github.com/kubevirt/kubevirt/pull/3392#issuecomment-626559676

The reason IPv6 can connect to a socket that is listening to 0.0.0.0 / 127.0.0.1
is that golang net library, has an abstraction (details in the link).
The team asked to use IPv6 so the code will be explicit, and handle the case,
where IPv4 Mapping is disabled.

This PR is for focusing on the 127.0.0.1 parts, which runs locally  between the virt-launcher and the virt-handler, in order to create the migration-proxy and update /etc/hosts, as a WA
for libvirt hostname resolving.

See please the whole discussion if needed, or i am here if needed.
Thanks

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -497,6 +500,22 @@ func (l *LibvirtDomainManager) SetGuestTime(vmi *v1.VirtualMachineInstance) erro
 	return nil
 }
 
+func (l *LibvirtDomainManager) GetLoopbackAddress() string {
+	//DEBUG
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Println("DBGDBG Recovered in f", r)
+		}
+	}()
```
vladikr, line 509: I suppose you will remove this later? ;)

oshoval: Hehe ye, trying to debug how to fix the travis failure and the mock

pkg/util/util.go
```diff
@@ -29,3 +31,18 @@ func IsGPUVMI(vmi *v1.VirtualMachineInstance) bool {
 	}
 	return false
 }
+
+// IsIpv6Disabled returns if IPv6 is disabled according sysctl
+func IsIpv6Disabled() bool {
+	ipv6Disabled, err := exec.Command("cat", "/proc/sys/net/ipv6/conf/default/disable_ipv6").Output()
```
vladikr, line 37: Perhaps, I'll be better if you'll open and read the file instead of calling `cat`
See an example.. here maybe: https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/util/cpu_utils.go#L30

oshoval: Thanks Vladik, on it

oshoval: ah btw this is from other PR (its cherry-picked), i will try it on both

pkg/virt-launcher/virtwrap/manager_test.go
```diff
@@ -574,13 +583,22 @@ var _ = Describe("Manager", func() {
...
+				client.Close()
+			*/
 			options := &cmdclient.MigrationOptions{
 				Bandwidth:               resource.MustParse("64Mi"),
 				ProgressTimeout:         150,
 				CompletionTimeoutPerGiB: 300,
 			}
 			err = manager.MigrateVMI(vmi, options)
 			Expect(err).To(BeNil())
+			_ = manager.GetLoopbackAddress()
```
oshoval, line 601: to remove

pkg/util/util.go
oshoval: Sweet man, it fixed Travis
Thanks a bunch

oshoval:
/hold
Added here a revert of a commit that cause IPv6 master to be broken,
i need the lane to work for my changes, but dont want this revert to be pushed.

oshoval:
/hold
adding fc32 revert again, to check the lanes

pkg/util/net/ip/ip_utils.go
```diff
@@ -0,0 +1,67 @@
...
+const DISABLE_IPV6_PATH = "/proc/sys/net/ipv6/conf/default/disable_ipv6"
+const IPV4_BIND_ADDRESS = "0.0.0.0"
+const IPV6_BIND_ADDRESS = "[::]"
+const IPV4_LOOPBACK_ADDRESS = "127.0.0.1"
+const IPV6_LOOPBACK_ADDRESS = "[::1]"
+
+// GetIPBindAddress returns IP bind address (either 0.0.0.0 or [::] according sysctl disable_ipv6)
+func GetIPBindAddress() string {
+	var disableIPv6 string
+	file, err := os.Open(DISABLE_IPV6_PATH)
```
AlonaKaplan, line 18: What is the reason you are not using exec.Command("sysctl", "net.ipv6.conf.default.disable_ipv6").CombinedOutput()?
(I remember we talked about it, so I guess there is a reason).

pkg/util/net/ip/ip_utils.go
```diff
@@ -0,0 +1,67 @@
...
+	if disableIPv6 == "1" {
+		return IPV4_BIND_ADDRESS
+	}
+
+	return IPV6_BIND_ADDRESS
+}
+
+// GetLoopbackAddress returns loopback IP (either 127.0.0.1 or [::1] according sysctl disable_ipv6)
+func GetLoopbackAddress() string {
+	var disableIPv6 string
```
AlonaKaplan, line 41: Please extract to a method the logic for 'isIpv6Disabled' and use it in both GetIPBindAddress and GetIPBindAddress.

pkg/util/net/ip/ip_utils.go
```diff
@@ -0,0 +1,67 @@
...
+	if disableIPv6 == "1" {
+		return IPV4_LOOPBACK_ADDRESS
+	}
+
+	return IPV6_LOOPBACK_ADDRESS
+}
+
+// IsLoopbackAddress checks if the address is IPv4 / IPv6 loopback address
+func IsLoopbackAddress(ipAddress string) bool {
+	Loopback := net.ParseIP(ipAddress)
```
AlonaKaplan, line 65: Why does the 'Loopback' variable start with capital letter?

pkg/util/net/ip/ip_utils.go
```diff
@@ -0,0 +1,67 @@
...
+	return Loopback.IsLoopback() || ipAddress == IPV6_LOOPBACK_ADDRESS
```
AlonaKaplan, line 66: IP.IsLoopback should work for ipv6 address as well. I think the second part of the condition is redundant.

pkg/util/net/ip/ip_utils.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+const DISABLE_IPV6_PATH = "/proc/sys/net/ipv6/conf/default/disable_ipv6"
+const IPV4_BIND_ADDRESS = "0.0.0.0"
+const IPV6_BIND_ADDRESS = "[::]"
+const IPV4_LOOPBACK_ADDRESS = "127.0.0.1"
+const IPV6_LOOPBACK_ADDRESS = "[::1]"
+
+// GetIPBindAddress returns IP bind address (either 0.0.0.0 or [::] according sysctl disable_ipv6)
+func GetIPBindAddress() string {
```
AlonaKaplan, line 16: Does 'IPBind' id some known terminology to "0.0.0.0"? (I found INADDR_ANY)

pkg/util/net/ip/ip_utils.go
oshoval: Vladik suggested to use file open as other sysfs files are dealt with in the project, moreover it doesn't spawn a thread, travis is sensitive for threads in its goveralls,
and also its lighter.

pkg/util/net/ip/ip_utils.go
oshoval: it works for ::1 but we have  [::1]

pkg/util/net/ip/ip_utils.go
oshoval: I think this was suggestion of Miguel
indeed the define is right INADDR_ANY
but using GetIPAnyAddress and so on is confusing, this what Miguel said once

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,81 @@
...
+		return IPV4_LOOPBACK_ADDRESS
+	}
+
+	return IPV6_LOOPBACK_ADDRESS
+}
+
+// IsLoopbackAddress checks if the address is IPv4 / IPv6 loopback address
+func IsLoopbackAddress(ipAddress string) bool {
+	loopback := net.ParseIP(ipAddress)
+	return loopback.IsLoopback() || ipAddress == IPV6_LOOPBACK_ADDRESS
```
AlonaKaplan, line 80: I cannot find my previous comment and your response. You said you cannot use just loopback.IsLoopback since the ipv6 address contains the brackets. As Eddy mentioned in the comment (https://github.com/kubevirt/kubevirt/pull/3392/commits/fde14127e7e5b3035167a1a7a726c9a387fa7fa8#r422618271). Adding the brackets should be done most close as possible to creating the socket.
IMO,
* `GetIPBindAddress` and `GetLoopbackAddress` should return a raw address. Without the brackets.
* The `tcpBindAddress` passed to `NewTargetProxy` shouldn't have brackets. Add the brackets when you need them, and that's when creating a socket.
* BTW, with your current code, having the brackets in the `tcpBindAddress` the code of [PR 3396](https://github.com/kubevirt/kubevirt/pull/3396/commits/d5fd692c7afde5e2935074a269a4e31d66d063f0#diff-e88e0fbecb2c3ab1c8c08d12a1c8b4f6R220) is redundant.

oshoval: This is what i did in the beginning and then was asked to move from there,
anyway i will try and see what it involves, i assume you want it as part of this PR,
just saying we already said lgtm in the other one, and decided it will be in other PR...

oshoval: About "GetIPBindAddress and GetLoopbackAddress should return a raw address. Without the brackets." it might add more places that need normalization, i prefer if we can stick to the original plan and please do it in a separate PR ?
I already started couple other things, and also this investigation, but planning and order is better when possible please.

oshoval: keep in mind please that the condition i added can be easily removed if i don't use the golib, but have the two comparison, and then the code would be great, and the 2nd issue Edy and you raised can be fixed in a following PR imo.

https://github.com/kubevirt/kubevirt/pull/3396/files#diff-4713d8334b0e7a78d689e3a725ab753fR80

WORKSPACE
```diff
@@ -142,19 +142,37 @@ http_file(
 
 http_file(
     name = "fedora_image",
-    sha256 = "423a4ce32fa32c50c11e3d3ff392db97a762533b81bef9d00599de518a7469c8",
+    sha256 = "a30549d620bf6bf41d30a9a58626e59dfa70bb011fd7d50f6c4511ad2e479a39",
     urls = [
-        "https://download.fedoraproject.org/pub/fedora/linux/releases/32/Cloud/x86_64/images/Fedora-Cloud-Base-32-1.6.x86_64.qcow2",
-        "https://storage.googleapis.com/builddeps/423a4ce32fa32c50c11e3d3ff392db97a762533b81bef9d00599de518a7469c8",
+        "https://download.fedoraproject.org/pub/fedora/linux/releases/29/Cloud/x86_64/images/Fedora-Cloud-Base-29-1.2.x86_64.qcow2",
```
vladikr, line 147: we've switched to fedora 32, You probably shouldn't change this. Or has it been done on purpose?

oshoval: I am only reverting this commit as a whole, in this PR, in order to test migration,
since this PR made guest agent connection problems.
There is a hold on the PR so it wont be merged, once the code is fine, i will revert this commit and remove the lgtm.
There are ongoing PRs by the team to fix the problem fc32 presented meanwhile.

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -328,9 +322,9 @@ func (m *migrationProxy) createTcpListener() error {
 	var listener net.Listener
 	var err error
 	if m.serverTLSConfig != nil {
-		listener, err = tls.Listen("tcp", fmt.Sprintf("%s:%d", m.tcpBindAddress, m.tcpBindPort), m.serverTLSConfig)
+		listener, err = tls.Listen("tcp", fmt.Sprintf("%s:%d", ip.NormalizeIPAddress(m.tcpBindAddress), m.tcpBindPort), m.serverTLSConfig)
```
AlonaKaplan, line 325: Why don;t you use net.JoinHostPort here as well?

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -328,9 +322,9 @@ func (m *migrationProxy) createTcpListener() error {
...
 	} else if ip.IsLoopbackAddress(m.tcpBindAddress) {
-		listener, err = net.Listen("tcp", fmt.Sprintf("%s:%d", m.tcpBindAddress, m.tcpBindPort))
+		listener, err = net.Listen("tcp", fmt.Sprintf("%s:%d", ip.NormalizeIPAddress(m.tcpBindAddress), m.tcpBindPort))
```
AlonaKaplan, line 327: same

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -436,8 +438,9 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
 		}
 
 		// For a tunnelled migration, this is always the uri
-		dstUri := fmt.Sprintf("qemu+tcp://%s:%d/system", loopbackAddress, LibvirtLocalConnectionPort)
```
AlonaKaplan: same

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -436,8 +438,9 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
...
-		migrUri := "tcp://" + loopbackAddress
+		normalizedLoopbackAddress := ip.NormalizeIPAddress(loopbackAddress)
+		dstURI := fmt.Sprintf("qemu+tcp://%s:%d/system", normalizedLoopbackAddress, LibvirtLocalConnectionPort)
```
AlonaKaplan, line 442: same

oshoval:
> /lgtm - please remove the non-relevant commits.

thanks, just please keep in mind that after i do it , i will need #3429 merged , or cherry-picked

oshoval:
@AlonaKaplan 
can you please lgtm again so i will ask a maintainer to review?

oshoval:
/hold cancel
removed the revert, tests might fail as long as we dont have #3429

WORKSPACE
oshoval: reverted

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Special IPs
+	IPV4_BIND_ADDRESS     = "0.0.0.0" //INADDR_ANY
+	IPV4_LOOPBACK_ADDRESS = "127.0.0.1"
+	IPV6_BIND_ADDRESS     = "[::]" //INADDR6_ANY
+	IPV6_LOOPBACK_ADDRESS = "[::1]"
+)
+
+func readSysfsFile(path string) (string, error) {
+	var output string = ""
```
EdDev, line 41: This can be shorten by one of these options:
- Name it as a return value, making this line redundant: `func foo() (output string, err error)`
- `var output string`
- `output := ""`

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,81 @@
...
+	defer file.Close()
+	scanner := bufio.NewScanner(file)
+	if scanner.Scan() {
+		output = scanner.Text()
+	}
+	if err := scanner.Err(); err != nil {
+		return output, err
+	}
+
+	return output, nil
```
EdDev, line 56: I think this should work:
```
data, err := ioutil.ReadAll(path)
if err != nil {
    return ...
}
return string(data), nil
```

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,81 @@
...
+	DISABLE_IPV6_PATH = "/proc/sys/net/ipv6/conf/default/disable_ipv6"
+
+	// Special IPs
+	IPV4_BIND_ADDRESS     = "0.0.0.0" //INADDR_ANY
+	IPV4_LOOPBACK_ADDRESS = "127.0.0.1"
+	IPV6_BIND_ADDRESS     = "[::]" //INADDR6_ANY
+	IPV6_LOOPBACK_ADDRESS = "[::1]"
+)
+
+func readSysfsFile(path string) (string, error) {
```
EdDev, line 40: Please place callers before callee for easier reading (I prefer to read from top down, details revealing each step).

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	if err := scanner.Err(); err != nil {
+		return output, err
+	}
+
+	return output, nil
+}
+
+// GetIPBindAddress returns IP bind address (either 0.0.0.0 or [::] according sysctl disable_ipv6)
+func GetIPBindAddress() string {
```
EdDev, line 60: Why do you call it `bind`? These are the `zero` addresses.

pkg/util/net/ip/ip.go
EdDev: Why do you need `|| ipAddress == IPV6_LOOPBACK_ADDRESS` ?

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,81 @@
...
+
+const (
+	// sysfs
+	DISABLE_IPV6_PATH = "/proc/sys/net/ipv6/conf/default/disable_ipv6"
+
+	// Special IPs
+	IPV4_BIND_ADDRESS     = "0.0.0.0" //INADDR_ANY
+	IPV4_LOOPBACK_ADDRESS = "127.0.0.1"
+	IPV6_BIND_ADDRESS     = "[::]" //INADDR6_ANY
+	IPV6_LOOPBACK_ADDRESS = "[::1]"
```
EdDev: All these exist in the net package in the IP struct format.
https://golang.org/pkg/net/#pkg-variables

pkg/util/net/ip/ip.go
oshoval: outdated, please see refactor commit

pkg/util/net/ip/ip.go
oshoval: how to call it then?
This is what Miguel suggested, since "Any" is confusing

pkg/util/net/ip/ip.go
oshoval: the snippet i use, is the one that other places in the code use,
do we want to change it ?

pkg/util/net/ip/ip.go
EdDev: ZeroAddress ?

pkg/util/net/ip/ip.go
EdDev: So you are saying that we need to use this because it appears elsewhere in the code? I do not understand your point here.

oshoval:
Using #3429 
And the guest-agent migration 5 tests pass
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3396/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1263340949879132160

pkg/util/net/ip/ip.go
oshoval: Lets look on 
https://github.com/kubevirt/kubevirt/blob/279617274c7c1fb6e16b9bb325b94bf41841acb9/pkg/virt-launcher/virtwrap/util/cpu_utils.go

IMO we shouldn't act in different logics, in different places,
This isn't the only place in the code that already use this logic,
Maybe there are reasons the developers chosen that way, maybe its just a matter of cosmetics and you are right, 
The most important thing is its would be in a one common way.

pkg/util/net/ip/ip.go
oshoval: I can, i just less like it, its isnt meaningful as the Bind (BindAny maybe) imo

pkg/util/net/ip/ip.go
oshoval: No problem but lets keep in mind they are []byte, i would need to cast to string
Again this is the outdated version, the new one is without [ ] , but the bottom line is the same.

pkg/util/net/ip/ip.go
EdDev: That is not a valid reason, sorry. If you would have used a function that all are using, then it would have been a stronger argument, but this is not the case.

If someone did something wrong, it does not need to be duplicated in all the project.
I would suggest using a technical reasoning on why this approach is better than the simpler one.

pkg/util/net/ip/ip.go
EdDev: `bing` is incorrect naming. You do not give a name to a function based on where it is used, but based on what it represent.
This one returns a zeroed address and it is used at the caller to bind the socket to it.

pkg/util/net/ip/ip.go
oshoval: The reason is that the approach i suggest is preparation for a smoother refactor, that we can do in another PR imo, since its not belong to this scope.
The use of this code in this manner was also suggested by a maintainer.
The code at the moment doesn't have one function that is common in all the other places,
creating various delicates of the same functionality is a cookbook for problems imo.

pkg/util/net/ip/ip.go
oshoval: it represent the bind all address, but i will change, i disagree here and Miguel was suggested the current name as well.

pkg/util/net/ip/ip.go
oshoval: done, just 127.0.0.1 wasnt tthere

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func readSysfsFile(path string) (string, error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
+
+	data, err := ioutil.ReadAll(file)
```
EdDev, line 78: Please use `ioutil.ReadFile(path)`.

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -410,10 +414,12 @@ func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, opti
 		isBlockMigration := (vmi.Status.MigrationMethod == v1.BlockMigration)
 		migrationPortsRange := migrationproxy.GetMigrationPortsList(isBlockMigration)
 
+		loopbackAddress := l.GetLoopbackAddress()
```
EdDev, line 417: I suggests you do this outside the go routine.

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -95,6 +98,7 @@ type DomainManager interface {
 	GetUsers() ([]v1.VirtualMachineInstanceGuestOSUser, error)
 	GetFilesystems() ([]v1.VirtualMachineInstanceFileSystem, error)
 	SetGuestTime(*v1.VirtualMachineInstance) error
+	GetLoopbackAddress() string
```
EdDev, line 101: Why the loopback getter belongs to the DomainManager? Is this just for allowing to inject mocks?

oshoval: yes, i now re-factored it, see please last commit

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -499,6 +503,8 @@ func (l *LibvirtDomainManager) SetGuestTime(vmi *v1.VirtualMachineInstance) erro
 	return nil
 }
 
+var GetLoopbackAddress = ip.GetLoopbackAddress
```
vladikr, line 506: I think it doesn't have to be exportable, please change to it `getLoopbackAddress`
Also, what's the difference between this and `loopbackAddress`?
Perhaps you can simply get the address once?

oshoval: done, thanks

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,76 @@
...
+	"net"
+
+	netutils "k8s.io/utils/net"
+)
+
+const (
+	// sysfs
+	DISABLE_IPV6_PATH = "/proc/sys/net/ipv6/conf/default/disable_ipv6"
+
+	IPV4_LOOPBACK_ADDRESS = "127.0.0.1"
```
EdDev, line 32: I think that even for constants the capital letter applies and the convention of camel-case over underscores. No?

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// GetIPZeroAddress returns INADDR_ANY or INADDR6_ANY (according sysctl disable_ipv6)
+func GetIPZeroAddress() string {
```
EdDev, line 36: nit: Seems it is encouraged to avoid `Get` prefixes (I personally dislike get*, that is why I point this out).
ref: https://golang.org/doc/effective_go.html#Getters

Not a must at the moment.

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -396,7 +399,6 @@ func getDiskTargetsForMigration(dom cli.VirDomain, vmi *v1.VirtualMachineInstanc
 }
 
 func (l *LibvirtDomainManager) asyncMigrate(vmi *v1.VirtualMachineInstance, options *cmdclient.MigrationOptions) {
-
```
EdDev, line 399: nit: Not a related change.

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -499,6 +501,8 @@ func (l *LibvirtDomainManager) SetGuestTime(vmi *v1.VirtualMachineInstance) erro
 	return nil
 }
 
+var loopbackAddress = ip.GetLoopbackAddress()
```
AlonaKaplan, line 504: Can you please move it upper in the file, before the func start (~line 130).

EdDev, line 504: Global alert, global alert!
First, please locate global declarations at the top of the file, it took me a while to detect it.

Second, you need to really reason on why you use a global here, especially that this will run pre-main even which sounds really bad to me.

oshoval: This is the method that Vladik suggested to use, 
it appears on more places in the manager.go official code
We need this because Interface mocking is for RPC only according Vladik,
and the unit tests which doesn't have a cluster fails,
I will investigate, it might be that a delay would be enough, or increasing timeouts in the unit tests.

pkg/util/net/ip/ip.go
oshoval: this is the convention i saw on other files of the project
i have go lint installed, 
which say
```
don't use ALL_CAPS in Go names; use CamelCasego-lint
```
but generally we dont obey go lint atm

so question if we use convention or create a file which is little different

EdDev: "Because others do it like that" is not a good reasoning.
Technically, you should obey the convention and make sure you do not export constants when it is not needed.

pkg/virt-launcher/virtwrap/manager.go
oshoval: Debugging i see the following error
https://travis-ci.org/github/kubevirt/kubevirt/builds/690593931#L884

{"component":"","level":"error","msg":"failed to create unix socket for proxy service","pos":"migration-proxy.go:334","reason":"listen tcp [::1]:49152: bind: cannot assign requested address","timestamp":"2020-05-24T12:37:17.008418Z"}

The function when running "make goveralls"
runs on the travis Virtual machine which doesnt support IPv6, but according the above error
the VM have ipv6_disable=0 in the sysfs.

We should decide on a way to stub this function for the unit tests which doesnt have a real cluster.
I would suggest to stick with the current method, either a variable or a function, as this is the common method
on production code, and was suggested by Vladik as well.
We see its a must to stub it out.
Vladik asked that it wont be a function if possible (which what i done already),
so lets please talk tommorow with him,
the 4 of us, or decide one direction and unblock this PR.
We can change it, if needed, in a following PR.
If its fine with you, i will do rest of the nits, so we are ready with lgtm for Vladik's approval.

Thanks

EdDev: Unit tests cannot run functions on the OS (including reading from sysfs).
It needs to be replaced in the test through a mock/stub.

AFAIK, doing that requires function pointers on a struct that represents the parts that needs replacements.
The mocking mechanics is one way to do this, but it requires an interfaces and does a bit more than is needed IMO. I would just do it manually here instead of that automatic mocking.

oshoval: I did it manually here, this is what you mean that you suggest ?
Interface as DomainManager is for RPC as Vladik explained.

EdDev: Github is swallowing such threads when code is changed and this thread merits a reset anyway.
I'm moving the discussion to the new location where the global was moved.

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -124,6 +127,8 @@ type pausedVMIs struct {
 	paused map[types.UID]bool
 }
 
+var loopbackAddress = ip.GetLoopbackAddress()
```
EdDev, line 130: Executing io actions at pre-main is inappropriate and having such a global is a high risk (like any global).
This may also be logically incorrect as the state of the pod when the program starts may not be the same later on during its lifetime.

There are a few options here, usually the nicer ones are also more complex. The inability to stub it easily is a result of sub-optimal design of the current code.
Here we go:
- Global function pointer at the (new) `ip` package: A test should be able to overwrite it by placing a stub.
  I do not like it much, but it is much better than the current option.
- Inject a `IpInfo` handler/object into the `LibvirtDomainManager`, used to fetch IP information during its operation. The loopback function can be placed in it.
- The previous options may come with an interface of `ip`, such that it can be substituted in the test.

Note: There may be other options but I found little on the subject in general.

We can go with the first, reasoning that the current code is too messy to adjust to fit the other options. But this is a smell that things are not in a good shape at the moment.

vladikr: > This may also be logically incorrect as the state of the pod when the program starts may not be the same later on during its lifetime.

how does the loopback address changes during the lifetime of the pod? I don't see how this is a high risk

vladikr: Yes, executing this here is probably not a good idea, but this variable can be populated from somewhere else.
I don't this we should overengineer this. It looks very straight forward and this address shouldn't change during the lifetime of the pod - I don't see a reason why would it.

EdDev: Enabling and disabling IPv6 may be actively changed during the lifetime of the pod.
The inability to run parallel tests is also a concern and a limitation of globals.
Globals look innocent and simple solutions at first, then they mess up and slow down progress in various ways. Please strive not to use them and when you do, feel bad about it (i.e. feel the urge to refactor).

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,76 @@
...
+// NormalizeIPAddress returns normalized IP, adding square brackets for IPv6 if needed
+func NormalizeIPAddress(ipAddress string) string {
+	if netutils.IsIPv6String(ipAddress) {
+		return "[" + ipAddress + "]"
+	}
+
+	return ipAddress
+}
+
+func readSysfsFile(path string) (string, error) {
```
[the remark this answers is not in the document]
oshoval: I dont think its redundant.
It does more than just ReadFile logic, and can be used in several places.
The less code people rewrite on their own, the better, and less potential bugs occur,
plus less time wasting, we dedicated time, so i would like it to be worth it please.

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,76 @@
...
+const (
+	// sysfs
+	disableIPv6Path = "/proc/sys/net/ipv6/conf/default/disable_ipv6"
+
+	ipv4Loopback = "127.0.0.1"
+)
+
+// GetIPZeroAddress returns INADDR_ANY or INADDR6_ANY (according sysctl disable_ipv6)
+func GetIPZeroAddress() string {
+	disableIPv6, err := readSysfsFile(disableIPv6Path)
```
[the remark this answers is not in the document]
oshoval: I dont see why, i added prints and checked exactly what happens,
also the only case where ipv6 is choosen is if all went well, and the value isnt "1"
i checked its indeed "0" exactly

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,76 @@
...
+	"net"
+
+	netutils "k8s.io/utils/net"
+)
+
+const (
+	// sysfs
+	disableIPv6Path = "/proc/sys/net/ipv6/conf/default/disable_ipv6"
+
+	ipv4Loopback = "127.0.0.1"
```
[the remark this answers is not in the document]
oshoval: Each of them compiles and work, its true that one is mixed, i will fix it.

pkg/util/net/ip/ip.go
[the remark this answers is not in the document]
EdDev: I do not know to what you refer, I meant unit testing this.

pkg/virt-launcher/virtwrap/manager.go
oshoval: I vote as well not to over-engineer please, i did in this PR enough out of the box.
PR scopes should be limited in their scope.
Its not that i violated a rule, and cant block this PR for inventing new ways.
We should consider that soon there will be dual-stack and the code might going to change and dictates new requirements,
We are the ones that control the pod settings.
For now we there isn't in the agenda a dynamic pod changes.
Adding more and more code without a feature need and a way to test it naturally in a real test case, isn't the right way imo.
We should focus on creating value, each one see value in a different way.

pkg/util/net/ip/ip.go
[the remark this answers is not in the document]
EdDev: Please be concrete and explain what “more” it does. I see only a string wrapper.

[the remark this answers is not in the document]
oshoval: Run function, check for error, cast
To recreate the flow, instead using a function takes time, and testing.
More reasons written above,

pkg/util/net/ip/ip.go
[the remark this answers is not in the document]
oshoval: readSysfsFile as you said shouldnt be used in unittest 
because its reading a sysfs file

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,69 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2020 Red Hat, Inc.
+ *
+ */
+package ip
```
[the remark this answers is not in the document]
oshoval: sysfs oriented,  so unit test is pretty problematic as you also said once

pkg/util/net/ip/ip.go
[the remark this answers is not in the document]
AlonaKaplan: The error is rechecked again when using `readSysfsFile` so I tend ot agree with Eddy. It is just a casting. I would change the method to `isIpv6Disabled`.

pkg/util/net/ip/ip.go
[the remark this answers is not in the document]
EdDev: It can be injected as a stub in tests.
The problem here is that logic is mixed with it, therefore not testing means logic is not checked.

Code must take into account tests and needs to be designed in a testable way.
I will not block on this one, but you should note and recognize the problem.

pkg/util/net/ip/ip.go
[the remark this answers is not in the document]
EdDev: Go is not supporting casting, only conversion. This is only adding the conversion from bytes to string, erros are checked at the caller anyway.
For conversion, you do not need 6 lines of code.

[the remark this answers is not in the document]
oshoval: btw Scanner method might be lighter than ReadFile,
we can benchmark it in the future.
or find an article

pkg/virt-launcher/virtwrap/manager.go
EdDev: I disagree and feel uncomfortable that I need to argue against global/s repeatedly.
I can think of several globals that make sense to me, this is not one of them.

oshoval: I wrote more reasons beside the issue if its global or not.
I am retesting and falling about blockers again and again because the PR is stalling,
each iterations with new comments, some even from the previous iterations,
please consider that there are ways to be more efficient for me, and not just for me.
we can take it into another PR, in a fresh clean way please if needed.
On this PR i ask and vote to wrap up please.

oshoval: There is a bug now in ipv4 that i must first take care of, those iterations cost a lot, we did around 3 variations of the code. Its more than enough.

oshoval: what about adding changing the condition to the function that only if
KUBERNETES_SERVICE_HOST has an IPv6 value
then use IPv6 address ?

This value exists in the pod env, as mandatory for K8s pods
and is even better, because it will handle case of IPv4 mapping disabled
And will work for both the unit test and the official code

EdDev: Reading the capabilities from the stack is the correct  approach for the production code.
You can use such techniques for tests, but I do not think it will help here.

oshoval: We will have a problem when IPv4 mapping is disable, then we will must need to determine what IP to use, according the pod network, this is outside of this PR scope.
But we could use this in order to reduce the need of stubbing.

Anyway i will first work on making this work again, something broken lately.
And then refocus on this.

pkg/virt-launcher/virtwrap/manager_test.go
```diff
@@ -474,6 +475,12 @@ var _ = Describe("Manager", func() {
 	})
 
 	Context("on successful VirtualMachineInstance migrate", func() {
+		BeforeEach(func() {
+			ip.GetLoopbackAddress = func() string {
+				return "127.0.0.1"
```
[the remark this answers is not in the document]
oshoval: this is the stub, you cant return here ::1 for travis

pkg/virt-launcher/virtwrap/manager_test.go
```diff
@@ -491,7 +498,6 @@ var _ = Describe("Manager", func() {
 		})
 		It("should verify that migration failure is set in the monitor thread", func() {
 			isMigrationFailedSet := make(chan bool, 1)
-
```
AlonaKaplan, line 494: Why did you remove the empty line?

oshoval: due to iterations mistake, 
i saw it , didn't want to push yet because wanted the test to run.

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,89 @@
...
+	"io/ioutil"
+	"net"
+
+	netutils "k8s.io/utils/net"
+)
+
+type readFileFunc func(filename string) ([]byte, error)
+
+var GetLoopbackAddress = getLoopbackAddress
+var readFile = ioutil.ReadFile
```
EdDev, line 31: I compromized for the `GetLoopbackAddress` global because it is hard to do it right with the current production code, requiring a lot of refactoring.

But here we have new code, which can be easily built right with tests without global hacks.
Here is the option you could use to have this sorted out without globals:
```
func GetIPZeroAddress() string {
	return getIPZeroAddress(isIPv6Disabled(disableIPv6Path))
}
func getIPZeroAddress(disableIPv6Disabled bool) string {
	if disableIPv6Disabled {
		return net.IPv4zero.String()
	}
	return net.IPv6zero.String()
}
```

It separates the io part for the logic part (although it can be done even more, but this seems enough).
Allowing you to test `getIPZeroAddress`.

pkg/util/net/ip/ip_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package ip_test
```
EdDev, line 1: Shouldn't this also be set with package `ip` like the test file?

oshoval: the file was auto generated, i will verify

pkg/util/net/ip/ip.go
oshoval: see please that i did not override the var from outside directly, it isnt exported even
i added function that do it.
and this method that i did even can remove the need of that other global that already exists.

The code you suggested indeed separate the io from the logic, 
but the code becomes, imo, less readable and i feel that the production code should be readable and easier as it can,

but i will change if you think otherwise, this is my opinion at least.

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,89 @@
...
+	data, err := readFile(filename)
+	if err != nil {
+		return true
+	}
+
+	return data[0] == '1'
+}
+
+// StubReadFile replaces readFile for unit tests
+func StubReadFile(f readFileFunc) {
```
oshoval, line 82: @EdDev 
making sure that you saw please that i don't use the global directly

pkg/virt-handler/migration-proxy/migration-proxy.go
```diff
@@ -132,7 +131,7 @@ func (m *migrationProxyManager) StartTargetListener(key string, targetUnixFiles
 	proxiesList := []*migrationProxy{}
 	for _, targetUnixFile := range targetUnixFiles {
 		// 0 means random port is used
-		proxy := NewTargetProxy("0.0.0.0", 0, m.serverTLSConfig, m.clientTLSConfig, targetUnixFile)
+		proxy := NewTargetProxy(ip.GetIPZeroAddress(), 0, m.serverTLSConfig, m.clientTLSConfig, targetUnixFile)
```
EdDev, line 134: I think you can do it outside the loop to save io.

pkg/virt-launcher/virtwrap/manager_test.go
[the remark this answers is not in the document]
EdDev: This is a unit test, I do not expect it to read from the sysfs.
I guess we can live with this, but it is wrong if it does.

pkg/util/net/ip/ip_test.go
```diff
@@ -0,0 +1,97 @@
...
+			Expect(result).To(BeTrue(), "Is Loopback")
+		})
+
+		It("should detect IPv6 non loopback address", func() {
+			result := IsLoopbackAddress("fd00:10:244:0:1::e")
+			Expect(result).To(BeFalse(), "Is not Loopback")
+		})
+	})
+
+	Context("NormalizeIPAddress", func() {
```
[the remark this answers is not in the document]
oshoval: i added that if the address is already normalized it wont touch it
and added that it wont touch ipv4 (which is non valid ipv6), 
i think its enough, but will add ::x or something if needed.

pkg/util/net/ip/ip.go
```diff
@@ -0,0 +1,85 @@
...
+	return net.IPv6loopback.String()
+}
+
+// IsLoopbackAddress checks if the address is IPv4 / IPv6 loopback address
+func IsLoopbackAddress(ipAddress string) bool {
+	loopback := net.ParseIP(ipAddress)
+	return loopback.IsLoopback()
+}
+
+// NormalizeIPAddress returns normalized IP, adding square brackets for IPv6 if needed
```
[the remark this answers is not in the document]
oshoval: the code checks if and only if its ipv6 , [ipv6] isnt ipv6 
but i will need, i just think its self explanatory

pkg/util/net/ip/ip_test.go
```diff
@@ -0,0 +1,97 @@
...
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("IP utils test", func() {
+
+	Context("IsLoopbackAddress", func() {
+
+		It("should detect IPv4 loopback address", func() {
+			result := IsLoopbackAddress(IPv4Loopback)
+			Expect(result).To(BeTrue(), "Is Loopback")
```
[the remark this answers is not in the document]
oshoval: will remove it, saw other test use the "expected case" in the line , so did according it

pkg/util/net/ip/ip.go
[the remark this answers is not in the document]
EdDev: The user of this API will not want to look into the implementation, only at the description of it.
That is why you add a text in the first place.

pkg/util/net/ip/ip_test.go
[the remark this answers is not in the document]
EdDev: You have not tested a non valid IP, this is what I am saying.
I do not understand your argument.

oshoval:
Thanks Edy.

@vladikr 
Can you review and approve please ?

oshoval:
Prow has a generic problem now, so the failing lane wont even start.
as can see on
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3396/pull-kubevirt-e2e-k8s-1.17/1265713623297167360

PR #3413, opened 2020-05-14
travis-ci: Fail on first error in script

EdDev:

The CI is passing several scripts of tests where each may fail.
When a script line fails, travis-ci is not existing but continues with the
next one.

This behavior makes the error hard to detect, wastes time and
resources.
Therefore, the scripts are wrapped with `set {-,+}e` for early exit on
failure.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ install:
 - git reset --hard
 
 script:
+- set -e
 - make generate
 - if [[ -n "$(git status --porcelain)" ]] ; then echo "It seems like you need to run
   `make generate`. Please run it and commit the changes"; git status --porcelain; false; fi
@@ -36,6 +37,7 @@ script:
 - make manifests DOCKER_PREFIX="docker.io/kubevirt" DOCKER_TAG=$TRAVIS_TAG # skip getting old CSVs here (no QUAY_REPOSITORY), verification might fail because of stricter rules over time; falls back to latest if not on a tag
 - make olm-verify
 - if [[ $TRAVIS_CPU_ARCH == "amd64" ]]; then make prom-rules-verify; fi
+- set +e
 
 deploy:
 - provider: script


.travis.yml
```diff
@@ -15,6 +15,7 @@ install:
 - git reset --hard
 
 script:
+- set -e
 - make generate
 - if [[ -n "$(git status --porcelain)" ]] ; then echo "It seems like you need to run
   `make generate`. Please run it and commit the changes"; git status --porcelain; false; fi
```
oshoval, line 21: Nice , i would suggest to add some special reserved word like
TRAVIS_ERROR on each of the error cases,
So in case we will back to "run all" in the future, it will be still easier to track the error,
Running all cost resources, hard to find (unless there is a special word),
but takes less iteration to see all the errors at once, if we run just one by one, it might take more iterations,
to see all the error, and sometimes you do want to know all at once, to get the whole picture imo.

EdDev: Next time please comment on the reasoning in the first PR comment.

If this change will be accepted, the future PR which will want to add "run all" will need to explain and resolve the problems it introduces.
I want to solve some of the problems this current "run all" causes.

oshoval: np, if i understand you right, i marked this line
so we can add TRAVIS_ERROR on each error, but up to you.

slintes:
Note that not failing on 1st error has an advantage: you see all errors in one test run if you have multiple of them. And you can fix all of them in one go.

enp0s3:
> Note that not failing on 1st error has an advantage: you see all errors in one test run if you have multiple of them. And you can fix all of them in one go.

@slintes That depends on how much actions are done by the script overall and how big is the log output at the end.

EdDev:
> Note that not failing on 1st error has an advantage: you see all errors in one test run if you have multiple of them. And you can fix all of them in one go.

Yes, that is the thing that is lost by this, but considering the following points, I think it is worth the lost:
- The log is huge and failing in the middle make it very hard to detect the problem.
- As these scripts sit one over the other and not as stand-alone, one fail may make others irrelevant and mislead when investigating.
   - If there are stand-alone checks, they can be ran as a separate job in parallel.
- Failing on the first is usually enough to detect the problem, especially if the change is small enough (hopefully). Other fails in many cases are a consequence of the same issue.
- The CI is not a developer environment, the developer is expected to run the tests locally before engaging the CI (i.e. pushing the change). The CI is there to protect and level the quality expected, it is not an alternative to one running and debugging locally.

rmohr:
/lgtm
/approve

I normally prefer to run everything till the end but the Travis output really is too big right now.

jean-edouard:
`set -e` is known to not play well with travis-ci... It's briefly discussed there for example:
https://github.com/travis-ci/docs-travis-ci-com/issues/1672
I submitted a PR that needed a `make generate`, and got that seemingly errorless output:
https://travis-ci.org/github/kubevirt/kubevirt/builds/692534118
When `set -e` is required, it's recommended to use a script instead.

EdDev:
> `set -e` is known to not play well with travis-ci... It's briefly discussed there for example:
> [travis-ci/docs-travis-ci-com#1672](https://github.com/travis-ci/docs-travis-ci-com/issues/1672)
> I submitted a PR that needed a `make generate`, and got that seemingly errorless output:
> https://travis-ci.org/github/kubevirt/kubevirt/builds/692534118
> When `set -e` is required, it's recommended to use a script instead.

I agree it needs to be in a script and not in the yaml file (not only because of this issue).
But the previous behavior is worse IMO, making it very hard to detect what failed.

How often is this happening here?

Issue #3421, opened 2020-05-14
Ipv6 lane failures after upgrading vm image to fc32


AlonaKaplan:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
/kind bug
> /kind enhancement

**What happened**:
The fedora image used by the e2e tests was upgraded to fedora32 on PR https://github.com/kubevirt/kubevirt/pull/3372.
This change cause several tests to start failing on ipv6 lane (for example - https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3319/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1260750823483248640)

**What you expected to happen**:
The test should pass

**How to reproduce it (as minimally and precisely as possible)**:
Run ip6 test lane

**Anything else we need to know?**:

**Environment**:
- KubeVirt version (use `virtctl version`):
- Kubernetes version (use `kubectl version`):
- VM or VMI specifications:
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

maiqueb:
@AlonaKaplan can we close this ?

AlonaKaplan:
@maiqueb if you can attach the pr that fixed the issue before closing, it can be great.

maiqueb:
They're already mentioned above:
  - #3454 
  - #3429 

I fail to see a button to close this issue, hence I'm reaching out for you to hit it (I'm assuming closing issues require approver rights).

maiqueb:
/close 

Fixed by #3454 , #3429 .

PR #3429, opened 2020-05-15
functest: Fix VMI IPv6 configuration

EdDev:

On an IPv6 deployment, IPv6 configuration is set on boot through
(injected) scripts in order to enable the VM connectivity.
The configuration includes an IP address, default route and DNS.
Implemented through the `ip` tool.

The IP configuration failed frequently due to the async behavior of
the IP addition operation. There is a need to give it time before
proceeding to add the route (which cannot be added if the previous
action has not completed).

This change adds a small sleep between the addition of the IP address
and the route.
In addition, in order to avoid NetworkManager to overtake and overwrite
the IP configuration, disable it.

**Fixes** #3421

The IPv6 lane (pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6) should pass.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -971,7 +971,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
 			It("[test_id:2303][posneg:negative] should secure migrations with TLS", func() {
-				tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
 				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse(fedoraVMSize)
 
@@ -1083,7 +1082,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
 			It("[test_id:2227]should abort a vmi migration without progress", func() {
-				tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
 				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse("1Gi")
 
--- a/tests/pausing_test.go
+++ b/tests/pausing_test.go
@@ -391,7 +391,6 @@ var _ = Describe("[rfe_id:3064][crit:medium][vendor:cnv-qe@redhat.com][level:com
 		}
 
 		It("[test_id:3090]should be continued after the VMI is unpaused", func() {
-			tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
 			By("Starting a Fedora VMI")
 			vmi := tests.NewRandomFedoraVMIWitGuestAgent()
 			vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse(fedoraVMSize)
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -4675,12 +4675,6 @@ func SkipMigrationTestIfRunnigOnKindInfra() {
 	}
 }
 
-func SkipIpv6TestsFailingAfterVmImageUpgradeToFc32() {
-	if IsRunningOnKindInfraIPv6() {
-		Skip("Skip ipv6 tests started to fail after https://github.com/kubevirt/kubevirt/pull/3372 was merged (issue: https://github.com/kubevirt/kubevirt/issues/3421)")
-	}
-}
-
 func IsUsingBuiltinNodeDrainKey() bool {
 	return GetNodeDrainKey() == "node.kubernetes.io/unschedulable"
 }
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -2084,12 +2084,14 @@ func GetGuestAgentUserData() string {
 	ipv6UserDataString := ""
 	if netutils.IsIPv6String(dnsServerIP) {
 		ipv6UserDataString = fmt.Sprintf(`sudo ip -6 addr add fd10:0:2::2/120 dev eth0
-                             sudo ip -6 route add default via fd10:0:2::1 src fd10:0:2::2
+                             for i in {1..20}; do sudo ip -6 route add default via fd10:0:2::1 src fd10:0:2::2 && break || sleep 0.1; done
                              echo "nameserver %s" >> /etc/resolv.conf`, dnsServerIP)
 	}
 	return fmt.Sprintf(`#!/bin/bash
                 echo "fedora" |passwd fedora --stdin
-				%s
+                systemctl disable NetworkManager
+                systemctl stop NetworkManager
+                %s
                 mkdir -p /usr/local/bin
                 curl %s > /usr/local/bin/qemu-ga
                 chmod +x /usr/local/bin/qemu-ga
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -1108,7 +1108,6 @@ var _ = Describe("Configurations", func() {
 			})
 
 			It("[test_id:1677]VMI condition should signal agent presence", func() {
-				tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
 				agentVMI := prepareAgentVM()
 				getOptions := metav1.GetOptions{}
 
@@ -1124,7 +1123,6 @@ var _ = Describe("Configurations", func() {
 			})
 
 			It("should remove condition when agent is off", func() {
-				tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
 				agentVMI := prepareAgentVM()
 				getOptions := metav1.GetOptions{}
 
@@ -1165,7 +1163,6 @@ var _ = Describe("Configurations", func() {
 				})
 
 				It("[test_id:]VMI condition should signal unsupported agent presence", func() {
-					tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
 					agentVMI := prepareAgentVM()
 					getOptions := metav1.GetOptions{}
 
@@ -1187,7 +1184,6 @@ var _ = Describe("Configurations", func() {
 			})
 
 			It("should have guestosinfo in status when agent is present", func() {
-				tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
 				agentVMI := prepareAgentVM()
 				getOptions := metav1.GetOptions{}
 				var updatedVmi *v1.VirtualMachineInstance
@@ -1207,7 +1203,6 @@ var _ = Describe("Configurations", func() {
 			})
 
 			It("should return the whole data when agent is present", func() {
-				tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
 				agentVMI := prepareAgentVM()
 
 				By("Expecting the Guest VM information")
@@ -1229,7 +1224,6 @@ var _ = Describe("Configurations", func() {
 			})
 
 			It("should not return the whole data when agent is not present", func() {
-				tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
 				agentVMI := prepareAgentVM()
 
 				By("Expecting the VirtualMachineInstance console")
@@ -1255,7 +1249,6 @@ var _ = Describe("Configurations", func() {
 			})
 
 			It("should return user list", func() {
-				tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
 				agentVMI := prepareAgentVM()
 
 				expecter, err := tests.LoggedInFedoraExpecter(agentVMI)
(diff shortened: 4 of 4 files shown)


tests/utils.go
```diff
@@ -2084,12 +2084,15 @@ func GetGuestAgentUserData() string {
 	ipv6UserDataString := ""
 	if netutils.IsIPv6String(dnsServerIP) {
 		ipv6UserDataString = fmt.Sprintf(`sudo ip -6 addr add fd10:0:2::2/120 dev eth0
+                             sleep 2
```
phoracek, line 2087: Any option not to make our slow tests even slower with sleeps? Plus it may cause (keep) flakiness on veeery slow systems with emulation enabled.

EdDev: It should take a few msec, I took 2sec as a precaution.
Compared to hundreds of sec it takes this test, this is negligible in terms of time.
But if you have a better solution, please share.

I was thinking of using NM instead of the `ip` command, but that may fit a different change.

phoracek: An alternative to sleep would be polling or monitoring.

>  Compared to hundreds of sec it takes this test, this is negligible in terms of time.

Is not this command executed in virtually every single test on IPv6 lane? Also, this is a dangerous precedent.

EdDev: Ok then, we will retry. Lets see if it works.

tests/utils.go
```diff
@@ -2084,12 +2084,14 @@ func GetGuestAgentUserData() string {
 	ipv6UserDataString := ""
 	if netutils.IsIPv6String(dnsServerIP) {
 		ipv6UserDataString = fmt.Sprintf(`sudo ip -6 addr add fd10:0:2::2/120 dev eth0
-                             sudo ip -6 route add default via fd10:0:2::1 src fd10:0:2::2
+                             for i in {1..20}; do sudo ip -6 route add default via fd10:0:2::1 src fd10:0:2::2 && break || sleep 0.1; done
```
phoracek, line 2087: Do you mind adding a comment explaining any random reader what is this for? Yeah, they can read git history, but the info there can get lost after the function is moved.

EdDev: Comments tend to rot, developers should track down the git history for reasoning and intent.
The history cannot be lost, it may just be harder to track.

Based on the comment pattern, we will see more stories than code. For this specific script, one can also be interest to know why NM is disabled, why IPv6 is added here, why the route has a source address, etc.. this is impractical IMO.

phoracek: I think that if there is a code that cannot be understood from the codebase, knowledge of the language and understanding of the topic ... should be documented. But it is debatable and your point is fair.

oshoval: configureIPv6OnVMI this function also have same logic

AlonaKaplan:
Can you please mention the PR fixes https://github.com/kubevirt/kubevirt/issues/3421 and why it fixes it (in fc32 setting the ip address takes more time?).

EdDev:
> Can you please mention the PR fixes #3421 and why it fixes it (in fc32 setting the ip address takes more time?).

I added the ref to the issue above.
The "why" is generic, it was wrong originally but just worked. I explained all I could in the commit.
I think that with the fc32 it just revealed it due to timing.

tests/utils.go
AlonaKaplan: https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L2842

> configureIPv6OnVMI this function also have same logic, in case it should be updated please

configureIPv6OnVMI already has a sleep [1]. Here it was missed.

Anyway, I would change the way we do the sleep in both of the places to be the same.

[1] https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L2842

EdDev: Yes, it needs to be the same... but I think the current code in general is bad, so we will try to work on it as part of a refactoring effort.

tests/utils.go
```diff
@@ -2084,12 +2084,14 @@ func GetGuestAgentUserData() string {
...
                              echo "nameserver %s" >> /etc/resolv.conf`, dnsServerIP)
 	}
 	return fmt.Sprintf(`#!/bin/bash
                 echo "fedora" |passwd fedora --stdin
-				%s
+                systemctl disable NetworkManager
```
ormergi, line 2092: Hey I think we shouldnt disable NM, if NM takes control and overwrite addresses 
we should first check if NM is installed and if it is then configure what needed though NM. 
We might miss issues we want to take care of caused by NM with kubevirt

EdDev: The need here is to have connectivity, or more accurately, to solve a problem of connectivity.
Starting to work with NM at this stage is an overkill and not needed.

The real solution is to use cloud-init network options, which will probably translate to NM ifcfg files. But this needs more work and investigation.

tests/utils.go
```diff
@@ -2084,12 +2084,14 @@ func GetGuestAgentUserData() string {
...
+                systemctl stop NetworkManager
```
AlonaKaplan, line 2093: Same here, if stopping the NetworkManager helps the test to be less flaky it should be done in configureIPv6OnVMI as well.

tests/utils.go
AlonaKaplan: I think it is unrelated. You're fixing here something that was already done in 'configureIPv6OnVMI' but was missed here. The fix should be the same.
If we had a way to reuse the same code here and in 'configureIPv6OnVMI' it would be the best. But I don't have an idea how. And if you're going to refactor it anyway, maybe it doesn't worth the effort.

oshoval: For the refactor,
Keep in mind please that the two places (cloud-init and configureIPv6OnVMI),
happen in a different stage of the running, one upon boot, and one in the tests,
does it affect behaviour of NM ?
i mean, would disabling it at configureIPv6OnVMI can be too late compared to disable it in the cloud-init ?

Moreover from what i rememeber, we added the logic configureIPv6OnVMI as a futuristic need,
not because we actually saw a need in a test, i am not sure this is the right way to do so.
Mostly in case the issue above is real.
maybe we need to remove it from there, and to make sure what tests need it,
and for them have it in userData so we need to maintain just one place ?

tests/vmi_configuration_test.go
```diff
@@ -1165,7 +1164,7 @@ var _ = Describe("Configurations", func() {
 				})
 
 				It("[test_id:]VMI condition should signal unsupported agent presence", func() {
-					tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
+
```
oshoval, line 1167: there are some blank lines that were added in this file, if you want to remove please

tests/pausing_test.go
```diff
@@ -391,7 +391,7 @@ var _ = Describe("[rfe_id:3064][crit:medium][vendor:cnv-qe@redhat.com][level:com
 		}
 
 		It("[test_id:3090]should be continued after the VMI is unpaused", func() {
-			tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
+
```
oshoval, line 394: blank , if you want to remove please

tests/migration_test.go
```diff
@@ -971,7 +971,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
 			It("[test_id:2303][posneg:negative] should secure migrations with TLS", func() {
-				tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
+				tests.SkipMigrationTestIfRunnigOnKindInfra()
```
oshoval, line 974: i am not sure my PR fix those 2, we need to verify
was SkipMigrationTestIfRunnigOnKindInfra were on master before SkipIpv6TestsFailingAfterVmImageUpgradeToFc32 was added ?

in case it was, its not right imo to replace skips, but need to cascade them,
i dont see i removed them from my PR #3992

tests/migration_test.go
```diff
@@ -1083,7 +1083,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
 			It("[test_id:2227]should abort a vmi migration without progress", func() {
-				tests.SkipIpv6TestsFailingAfterVmImageUpgradeToFc32()
+				tests.SkipMigrationTestIfRunnigOnKindInfra()
```
oshoval, line 1086: this should be here please, not migration
				tests.SkipStressTestIfRunnigOnKindInfra()

tests/migration_test.go
oshoval: https://github.com/kubevirt/kubevirt/blob/fde14127e7e5b3035167a1a7a726c9a387fa7fa8/tests/migration_test.go#L964

see please that there wasnt a line there

tests/utils.go
EdDev: > I think it is unrelated. You're fixing here something that was already done in 'configureIPv6OnVMI' but was missed here. The fix should be the same.
> If we had a way to reuse the same code here and in 'configureIPv6OnVMI' it would be the best. But I don't have an idea how. And if you're going to refactor it anyway, maybe it doesn't worth the effort.

These are not the same, one seems to work on VMI boot (through cloud-init) and the other while it is running.
Therefore, I would not recommend mixing the two.
I do not even understand why there is a need to configure IPv6 in two places (but I do not expect to discuss this here).

oshoval, line 2087: 20 x 0.1 = 2 second, 
2 seconds are enough ?
we used 5 so far, maybe worth to be on the safe side, giving more time ?

tests/utils.go
EdDev: Same answer here. The other is not the same in the sense of when this occurs.
I do not know the consequences of touching NM there, perhaps some existing configuration is removed or dependent on it.
We should discuss merging these two in a separate task.

tests/vmi_configuration_test.go
EdDev: Done

tests/utils.go
EdDev: Yes, it is more than enough per my tests.
If it is not enough, we should see the tests failing.

tests/migration_test.go
oshoval: right this why added the Edit above,
not all migration, just the ones that i didnt add a new skip for them, 
due to stressTest or to NFS 
https://github.com/kubevirt/kubevirt/pull/3392/files#diff-a5a178b4c66835110ceff63c61d28c96R4664

see the two skipper here please, i have added an issue for them

tests/migration_test.go
oshoval: it need just stressTest, but i understand what was confusing me.
my PR adds the stress Skip
in your it isnt available yet, so its ok that you put migration, sorry for the noise

here my PR adds the stress skip
https://github.com/kubevirt/kubevirt/pull/3392/files#diff-a5a178b4c66835110ceff63c61d28c96R4664
it was groomed and researched as part of my PR

EdDev: Ok, now it is clear.
So if your PR gets in first, then on rebase of this one the new skip should appear.
And as `SkipMigrationTestIfRunnigOnKindInfra` will be gone, we will not forget it for sure.

tests/migration_test.go
AlonaKaplan: You don't need to add the migration skip, since this test is not failing on the issues the migration PR fixes (it is a negative test).

EdDev: > You don't need to add the migration skip, since this test is not failing on the issues the migration PR fixes (it is a negative test).

I see this line: `if strings.Contains(err.Error(), "remote error: tls: bad certificate")`
Doesn't that mean it expects to fail on a specific tls issue? For that a connection needs to be established (or attempted) and failing with that error.
Or do I understand this whole story wrongly?

AlonaKaplan: I don't know the story. But currently the lane is green without that test being skipped. Maybe the test has a bug and doesn't actually check what it has to check. But adding a skip and then removing it in the migration PR won't fix it.
Let's open a separate discussion to understand this test.

EdDev: I do not like it but I will comply if the test passes with this change.
Checking locally and will update accordingly.

EdDev: It passed locally on the ipv6 lane.

oshoval:
once you have a lgtm approve from reviewer / maintainer , the bot will do the retest itself,
i believe this PR fixes some flakes that were not skipped as well
so it will be great if we can advance it please

oshoval:
/lgtm
please ask a reviewer with lgtm permissions to lgtm and then a maintainer
then you wont need to retest manually, when you know all the non passing is true flakes,
because the bot will do that for you
i believe your PR fixes those 5 flakes as well
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3392/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1263001457066512384

tests/utils.go
maiqueb, line 2093: you could avoid this instruction by doing `systemctl disable --now NetworkManager` above, but it is not reason to block.

EdDev: Considering the fact that I want to take this out completely, I think it is not worth the change.
I can do it if some other change is requested.

maiqueb: Fair enough.

PR #3422, opened 2020-05-14
Preparing kubevirt to make the ipv6 lane mandatory

AlonaKaplan:
**What this PR does / why we need it**:
Since ipv6 support was merged to kubevirt there were servel PRs that broke
the ipv6 lane.
Since the lane is not mandaroty it is very hard to track and to catch
those failures on time.

To avoid this situation in the future. The current failing tests on the
ipv6 lane are skipped so the lane can become green.
Once this PR will be merged, the ipv6 lane can be moved to be
mandatory.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
"NONE"
```

PR #3432, opened 2020-05-17
Revert ipv6 skip tests

EdDev:

Remove the skipped tests introduced in #3422 

```release-note
NONE
```

Depends on #3429

### DIFF
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -540,7 +540,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 		})
 		Context("with setting guest time", func() {
 			It("[test_id:4114]should set an updated time after a migration", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
 				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse(fedoraVMSize)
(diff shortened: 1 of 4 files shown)


tests/migration_test.go
```diff
@@ -540,7 +540,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 		})
 		Context("with setting guest time", func() {
 			It("[test_id:4114]should set an updated time after a migration", func() {
-				tests.SkipMigrationTestIfRunnigOnKindInfra()
```
AlonaKaplan, line 543: Does https://github.com/kubevirt/kubevirt/pull/3429 fixes this test?
This test will be fixed by the migration PR (https://github.com/kubevirt/kubevirt/pull/3392).

EdDev: I though you skipped only the ones who had to do with fc32 change.
If you originally skipped more, then it will probably not solve it and we need to depend also on the migration fix.

Do you prefer to add it as a dependency or split the PR?

AlonaKaplan: The migration PR removes the skip fro the migration tests.
I would add the removal on the "fc32 upgrade" skips as a commit to the PR solving the issue.

EdDev: This one actually fails due to `SkipIpv6TestsFailingAfterVmImageUpgradeToFc32` as well.
All tests here fail because of the GuestAgent and the Migration with IPv6 problems.

EdDev:
Changes are moved to #3429

PR #3433, opened 2020-05-17
functests: Refactor VMI helpers

EdDev:

The existing helper functions in tests/utils.go cover all functional
tests helpers, including the VMI ones.

This change extracted the VMI helpers into a dedicated module and placed
under a new `lib` package.

In order to show the usage, the `NewRandomFedora32VMIWithFedoraUser`
helper function has been adjusted to work with the new package.

Following changes will start using the new package for VMI processing.
It aims making VMI creation less convoluted with explicit and clear
properties. Unlike the current hierarchical structure, the new VMI helpers
will not hide details under a chain of functions, showing exactly of what
a VMI is composed of.

Draft.
Needs to be split in order not to add functionality which is not yest used from `utils`.

**Release note**:
```release-note
NONE
```


EdDev:
/cc @phoracek 
I could not resist :)

phoracek:
@EdDev beautiful!

tests/lib/vmi.go
```diff
@@ -0,0 +1,332 @@
...
+
+	k8sv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	k8smetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	kvirtv1 "kubevirt.io/client-go/api/v1"
+)
+
+type vmiOptions struct {
```
phoracek, line 33: Why do you add this level of indirection? We can use `With*` functions to directly manipulate `VirtualMachine`.

phoracek: `New` would just bloat when we add more and more options.

EdDev: > Why do you add this level of indirection? We can use `With*` functions to directly manipulate `VirtualMachine`.

You will have to say more than that for me to understand.
The `With*` pattern here builds a collection of properties with a single interface: `Option` (which is a function with the `*vmiOptions` as its argument). The collection is then processed by the constructor by calling the functions which in turn populates the options.

If you have another proposal, please elaborate (i.e. provide an example).

BTW, in no way this is my invention.

EdDev: > `New` would just bloat when we add more and more options.

Yes, I was looking to have a dedicated namespace for VMI and potentially even extract out the storage parts, but I failed to name the package by the file name. Perhaps you know how to make this happen.
I want to have under `lib/` multiple namespaces for each group.
Then we can do this: `libvmi.New()`.

phoracek: In the current code, `With*` functions are modifying `defaultVMIOptions`. Those options are then used to control modifications on `vmi`.

I suggest you use `With*` callbacks to manipulate `vmi` directly. e.g. the callback returned by `WithCloudInitConfigDriveNetworkData` would take `vmi` as an attribute and set its parameters properly to include the cloud init section.

This way the `New` function does not need to carry the modification logic, only execute callbacks. And writing/reading og those callback `With*` functions would be more intuitive IMO.

The only drawback I see is that this way of handling would not allow you to easily handle conflicts and dependencies. But maybe it would not be an issue if all callbacks are idempotent.

EdDev: You are suggesting a plug-able design. It would have been very nice, especially the notion that every operation is self contained and one can add operations without touching the others (unless dependencies exists).

Unfortunately, this seems to be challenging due to the dependencies between the operation.
- The order is important for some options but the collection is unordered.
- Other options collide with each other.

We could change the options structure to include a handler function, but I am not sure if it is worth it.
I suggest we start with one refactoring and consider the next one per the outcome.

enp0s3: @EdDev @phoracek Regarding the file hierarchy, what about "tests/testlibs/libvmi/*.go"?

EdDev: I had a quick discussion on this with @qinqon and we negotiated this: `tests/libvmi/vmi.go`
It seems similar to your suggestion here except the grouping of all helpers.
I think we do not need it now, but if we will, then change will be minimal in the future.

What do you think?

enp0s3: @EdDev Sure, also LGTM

tests/libvmi/vmi.go
```diff
@@ -0,0 +1,240 @@
...
+			o.cloudInitNoCloudNetworkData = data
+		}
+	}
+}
+
+func New(namespace, name string, opts ...Option) *kvirtv1.VirtualMachineInstance {
+	options := defaultVmiOptions
+	for _, opt := range opts {
+		opt(&options)
+
```
phoracek, line 118: Please drop the empty line.

tests/lib/vmi.go
phoracek: @EdDev sorry to keep bringing this up, but would you raise one example when my suggestion would not work? For better sleep.

phoracek:
> > sorry to keep bringing this up, but would you raise one example when my suggestion would not work? For better sleep.
> 
>     * Name needs to be randomized or not before creating the vmi object.
> 
>     * Either data-volume or container-image needs to be set, but not both.
> 
>     * Either cloudInitNoCloudSource or cloudInitConfigDriveSource needs to be set, but not both.
> 

All three of those can be just left to the user. If I am creating a test, I should know what VM definition I want. Once I know that, I can build it with `With*` functions. I should know that I can't use incompatible options. KubeVirt has validation set in place that makes sure that VM has proper settings. We should not duplicate that logic in test helpers.

> 
> These are the only cases I cover at the moment, I need to add the rest and see if there are other such cases.
> I may find that more combinations are not possible, which means there is a need to examine the existing vmi state before taking actions.

Then we can be reactive. This Option based approach adds 3 times as much code as the simple VMI callback one. It would be easier to switch from the callback one (where you would prepare functions close to e.g. `primaryInterfaceDevice`). If we start we complex, there is a risk that people start using it, it grows too much and even if there is no blocker to make it simple, there will be no will.

> 
> > I really hate that we introduce a new Options API which is basically slightly simplified VirtualMachine. The New function would grow and grow until it becomes impossible to extend options anymore without a lot of `ifs`. And the complexity would scare of anybody who would dare to extend it. I'm afraid we are inventing a new junkyard of not understood helpers here.
> 
>     * The current code adds things in layers (functions over functions over functions, each adding something, sometime contradicting the previous).
> 
>     * This proposal asks to flatten this both at the caller and at the implementation. This makes it clear what happens and avoids overwrites and duplication.
> 
>     * Complexity is relative. Now, each options has an action that is added serially and ordered. It is now visible and transparent, without abstracting too much. It also give more freedom to make combinations without adding anything.

It is scritpy and once it grows out of a single screen, people may be tempted to just place there code somewhere in the middle. That is my worry.

> 
>     * The grow you mention is limited to the `vmi` and not to everything.
> 
> 
> As I mentioned before, I do think your idea is great and if we manage to have something like that it may be really nice. But moving from one extreme to the other is not the best approach, refactoring can be done in small steps. If at the end we will see this is indeed possible, it will be much easier to move to your proposal than from what we have now.

If my arguments above don't convince you, I won't fight for it.

EdDev:
> > ```
> > * Name needs to be randomized or not before creating the vmi object.
> > 
> > * Either data-volume or container-image needs to be set, but not both.
> > 
> > * Either cloudInitNoCloudSource or cloudInitConfigDriveSource needs to be set, but not both.
> > ```
> 
> All three of those can be just left to the user. If I am creating a test, I should know what VM definition I want. Once I know that, I can build it with `With*` functions. I should know that I can't use incompatible options. KubeVirt has validation set in place that makes sure that VM has proper settings. We should not duplicate that logic in test helpers.

I think you miss the point that one overwrites the other, the user will not know and I am not even sure if the order will be consistent between calls (last wins?).
But this we can ignore if you want, leaving the problem to the user or later for checking collisions.

How do you solve the random name? Will it always be set to true?
Editing the name does not seem easy.

> > These are the only cases I cover at the moment, I need to add the rest and see if there are other such cases.
> > I may find that more combinations are not possible, which means there is a need to examine the existing vmi state before taking actions.
> 
> Then we can be reactive. This Option based approach adds 3 times as much code as the simple VMI callback one. It would be easier to switch from the callback one (where you would prepare functions close to e.g. `primaryInterfaceDevice`). If we start we complex, there is a risk that people start using it, it grows too much and even if there is no blocker to make it simple, there will be no will.

This PR shows there is a will, so you should expect it to continue.
Serializing options is simpler than callbacks, but I am willing to try. Just suggest the above random name blocker.

> It is scritpy and once it grows out of a single screen, people may be tempted to just place there code somewhere in the middle. That is my worry.

Nothing will replace a good reviewer and maintainer.
I agree with your reasoning, but every path has a cost.
 
> If my arguments above don't convince you, I won't fight for it.

You should fight and keep suggesting solutions to the problems raised.

phoracek:
> How do you solve the random name? Will it always be set to true?

`New` first initializes with `basicVmi` and then iterates `With*` functions.

For static:

```
vmi.New("namespace", "name")
```

That would simple intialize with `baseVmi()`.

For random:

```
vmi.New("namespace", "name", WithRandomSuffix())`
```

Where `WithRandomSuffix` would return a callback that would do roughly following `vmi.Metadata.Name += rand.String(48)`.

Maybe I'm missing the issue you are pointing out.

EdDev:
> > How do you solve the random name? Will it always be set to true?
> 
> `New` first initializes with `basicVmi` and then iterates `With*` functions.
> 
> For static:
> 
> ```
> vmi.New("namespace", "name")
> ```
> 
> That would simple intialize with `baseVmi()`.
> 
> For random:
> 
> ```
> vmi.New("namespace", "name", WithRandomSuffix())`
> ```
> 
> Where `WithRandomSuffix` would return a callback that would do roughly following `vmi.Metadata.Name += rand.String(48)`.
> 
> Maybe I'm missing the issue you are pointing out.

Yes, you are: https://github.com/kubevirt/kubevirt/blob/8ba7e69e6f26efd1f5b33589e7163496ee898c47/staging/src/kubevirt.io/client-go/api/v1/types.go#L543

Note that we will need to run defaults beyond the name... like memory and potential others.

phoracek:
> Note that we will need to run defaults beyond the name... like memory and potential others.

I don't think we need any defaults. Memory should not be needed and even the minimal disk configuration could be left empty. The most basic helpers using this new library would decide which kind of disks they want to use.

For the SelfLink, I don't think you need to set it as a client. It is added by API while you create an object.

I know you will not like the sound of it, but with simple callback, we basically form a "initializer train". Having the train API would make it more clear that those adjustments are made sequentially and may overwrite one another.

```go
vmi.New("namespace", "name", WithRandomSuffix(), WithSeocndaryNetwork())
```

vs.

```go
vmi.New("namespace", "name").WithRandomSuffix().WithSeocndaryNetwork()
```

EdDev:
> > Note that we will need to run defaults beyond the name... like memory and potential others.
> 
> I don't think we need any defaults. Memory should not be needed and even the minimal disk configuration could be left empty. The most basic helpers using this new library would decide which kind of disks they want to use.

This is not the current state, and I am following the existing logic as much as possible.
The memory is 8M, changing it may change tests outcome. I do not want to do that here, it is unsafe.
Declaring that there are no default is hiding the problem... please assume there are for this discussion.

> For the SelfLink, I don't think you need to set it as a client. It is added by API while you create an object.

But it is set, this is what exists in the tests.
 
> I know you will not like the sound of it, but with simple callback, we basically form a "initializer train". Having the train API would make it more clear that those adjustments are made sequentially and may overwrite one another.
> 
> ```go
> vmi.New("namespace", "name", WithRandomSuffix(), WithSeocndaryNetwork())
> ```
> 
> vs.
> 
> ```go
> vmi.New("namespace", "name").WithRandomSuffix().WithSeocndaryNetwork()
> ```

You are right, this will not happen on my watch. You can assume an automatic nack when seeing this.

phoracek:
> > > Note that we will need to run defaults beyond the name... like memory and potential others.
> > 
> > 
> > I don't think we need any defaults. Memory should not be needed and even the minimal disk configuration could be left empty. The most basic helpers using this new library would decide which kind of disks they want to use.
> 
> This is not the current state, and I am following the existing logic as much as possible.
> The memory is 8M, changing it may change tests outcome. I do not want to do that here, it is unsafe.
> Declaring that there are no default is hiding the problem... please assume there are for this discussion.

Set it in the basicVmi constructor then. If you find defaults you cannot handle that way, you may need to add explicit defaulting after the iteration of constructors. But it should be an exception.

> 
> > For the SelfLink, I don't think you need to set it as a client. It is added by API while you create an object.
> 
> But it is set, this is what exists in the tests.


Then you can recreate that in `WithRandomSuffix` as well.
> 
> > I know you will not like the sound of it, but with simple callback, we basically form a "initializer train". Having the train API would make it more clear that those adjustments are made sequentially and may overwrite one another.
> > ```go
> > vmi.New("namespace", "name", WithRandomSuffix(), WithSeocndaryNetwork())
> > ```
> > 
> > 
> > vs.
> > ```go
> > vmi.New("namespace", "name").WithRandomSuffix().WithSeocndaryNetwork()
> > ```
> 
> You are right, this will not happen on my watch. You can assume an automatic nack when seeing this.

Gotta get those in before you get a nack badge.

EdDev:
@phoracek , please revisit this one.
I tried to answer both needs, hope you like this more.

tests/libvmi/vmi.go
```diff
@@ -0,0 +1,151 @@
...
+// building its properties based on the specified With* options.
+func New(namespace, name string, opts ...Option) *kvirtv1.VirtualMachineInstance {
+	options := defaultVmiOptions
+	for _, opt := range opts {
+		opt(options)
+	}
+
+	// Randomizing the VMI name is a special case which must occur before
+	// creating the VMI.
+	if _, ok := options[optEnableRandomName]; ok {
```
qinqon, line 120: Why don't we implement this as an normal option at default ones ? `WIthNameSuffix` at default, so is all at the same place.

tests/utils.go
```diff
@@ -2050,11 +2052,17 @@ func AddEphemeralCdrom(vmi *v1.VirtualMachineInstance, name string, bus string,
 }
 
 func NewRandomFedora32VMIWithFedoraUser() *v1.VirtualMachineInstance {
-	vmi := NewRandomVMIWithEphemeralDiskAndUserdata(ContainerDiskFor(ContainerDiskFedora), `#!/bin/bash
-	    echo "fedora" |passwd fedora --stdin
-        echo `)
-	vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse("512M")
-	return vmi
+	userData := `#!/bin/bash
+		echo "fedora" |passwd fedora --stdin
```
qinqon, line 2056: Remove tabs there they are inserted at the userData

tests/libvmi/vmi.go
EdDev: We need this examined before the vmi object is created, because it is not that straight forward to update a given name.

tests/utils.go
EdDev: Done

tests/libvmi/cloudinit.go
```diff
@@ -0,0 +1,61 @@
...
+package libvmi
+
+import (
+	"encoding/base64"
+
+	kvirtv1 "kubevirt.io/client-go/api/v1"
+)
+
+// WithCloudInitNoCloudUserData adds could-ini no-cloud user data.
+func WithCloudInitNoCloudUserData(data string, b64Encoding bool) Option {
```
qinqon, line 29: You can pass a function at data so you don't need the if statement, also the if statement is only needed at conversion and field.

tests/libvmi/vmi.go
```diff
@@ -0,0 +1,151 @@
...
+	},
+	optResourceMemory: func(vmi *kvirtv1.VirtualMachineInstance) {
+		vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{
+			k8sv1.ResourceMemory: resource.MustParse("8192Ki"),
+		}
+	},
+}
+
+// WithRandomName adds a random text to the vmi name.
+func WithRandomName(enableRandomName bool) Option {
```
qinqon, line 80: Better two functions to have some syntactic sugar and remove f
Flag, With... and Without

tests/libvmi/cloudinit.go
EdDev: Please be explicit, how will you like this to look from the caller perspective?

tests/libvmi/vmi.go
EdDev: `WithoutRandomName` ? This is confusing and ugly IMO.
Per what I see, no one really uses it anyway (it is always random), I just added it to be in sync with the current code.

tests/libvmi/cloudinit.go
qinqon: So I think we have two options
- Keep function firm like this but reduce the if statement scope to just the base64 encoding
- Do something like `WithCloudInitNoCloudUserData(data string, func (source *CloudInitNoCloudSource, data string))`  and from there you can creat base64 encoders and all

tests/libvmi/vmi.go
qinqon: Well it's is always random, maybe we can just remove the optionallity of it so we reduce stuff.

tests/libvmi/cloudinit.go
EdDev: The second option exposes a detail (CloudInitNoCloudSource) which is now abstracted away.
The user just wants to use the data with a specific encoding.

I am also unsure how I can implement the first, I will still need two functions due to the different members assigned  (`UserDataBase64`, `UserDataBase64`).

qinqon: Like the following: 

```golang
func(o vmiOptions) {                                                    
    o[optCloudInitNoCloudUserDataB64] = func(vmi *kvirtv1.VirtualMachineInstance) {
        diskName, bus := "disk1", "virtio"                                      
        addDiskVolumeWithCloudInitNoCloud(vmi, diskName, bus)                   
        volume := getVolume(vmi, diskName)                                      
        if b64Encoding {                                                        
            encodedData := base64.StdEncoding.EncodeToString([]byte(data))      
            volume.CloudInitNoCloud.UserDataBase64 = encodedData                
        } else {                                                                
            volume.CloudInitNoCloud.UserData = data                             
        }                                                                       
    }                                                                           
}                                           
```

EdDev: Aha, you want to inject the flag in...
I was looking to create the function so it will have no conditions internally.

As we have no performance issue and your code has fewer lines of code, I will comply.

EdDev:
@phoracek , could you please revisit this PR?

tests/libvmi/vmi.go
```diff
@@ -0,0 +1,132 @@
...
+	vmi := kvirtv1.NewVMIReferenceFromNameWithNS(namespace, name)
+	vmi.Spec = kvirtv1.VirtualMachineInstanceSpec{Domain: kvirtv1.DomainSpec{}}
+	vmi.TypeMeta = k8smetav1.TypeMeta{
+		APIVersion: kvirtv1.GroupVersion.String(),
+		Kind:       "VirtualMachineInstance",
+	}
+	return vmi
+}
+
+func primaryInterfaceDevice() kvirtv1.Interface {
```
phoracek, line 125: There is storage.go, why not network.go?

tests/libvmi/vmi.go
```diff
@@ -0,0 +1,132 @@
...
+
+	for _, f := range options {
+		f(vmi)
+	}
+
+	return vmi
+}
+
+func baseVmi(namespace, name string) *kvirtv1.VirtualMachineInstance {
+	vmi := kvirtv1.NewVMIReferenceFromNameWithNS(namespace, name)
```
phoracek, line 116: Is not that function what you want to get rid of?

tests/libvmi/vmi.go
```diff
@@ -0,0 +1,132 @@
...
+				k8sv1.ResourceMemory: resource.MustParse(value),
+			}
+		}
+	}
+}
+
+// New instantiates a new VMI configuration,
+// building its properties based on the specified With* options.
+func New(namespace, name string, opts ...Option) *kvirtv1.VirtualMachineInstance {
+	options := defaultVmiOptions
```
phoracek, line 100: So we have a set of `With*` functions that register a callback to vmiOptions and then we iterate these options and use registered callbacks.

Why do we need that intermediate options structure? Can't just `WithRng` return the callback directly?

tests/libvmi/cloudinit.go
```diff
@@ -0,0 +1,55 @@
...
+
+package libvmi
+
+import (
+	"encoding/base64"
+
+	kvirtv1 "kubevirt.io/client-go/api/v1"
+)
+
+// WithCloudInitNoCloudUserData adds could-ini no-cloud user data.
```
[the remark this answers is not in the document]
EdDev: Done

tests/libvmi/vmi.go
EdDev: You have just pointed your comment on the reason.

It is needed in order to support defaults and the ability to overwrite them (e.g `ResourceMemory`).
It also provides the ability to have control over the options in case order or some special behavior is needed.  A previous version used it to handle the random name before creating the vmi, but it was dropped in favor of not having the option of non-random at all. But this mat change if it will be needed later on (I expect the scenario to occur, as testing that two vmi/s with the same name cannot co-exist is there somewhere).

tests/libvmi/vmi.go
EdDev: Not sure I understand the question, but if you are asking if we should implement here the function that we call, then the answer is per usage. This `NewVMIReferenceFromNameWithNS` is used in production as well, so we cannot drop it if we implement it here.

In general, the same pattern we us in tests now could be implemented in the production code.

tests/libvmi/vmi.go
EdDev: There is not enough reasoning to do it yet. We will create it once another network related options will present itself.

phoracek: One would argue that that is how utils junkyard starts. We gotta keep an eye on those new methods appearing here.

tests/libvmi/vmi.go
phoracek: I see

tests/libvmi/vmi.go
phoracek: You can handle defaults in the `baseVmi` function if you really want some test defaults. It sounds redundant and implicit. There already are defaults implemented in KubeVirt. For memory, this depends on the selected system, so it would make sense to set the default based on that. Or just let users set it.

If you want two different names, you use two differnt `WithName("FOO")`. Correct me if I'm wrong or provide a different example of why we cannot do that.

You want to replace the current helper code with something new. I'm all for that. But this way of doing things looks like a regression from simplicity/readability perspective to me.

EdDev: > You can handle defaults in the `baseVmi` function if you really want some test defaults.

Not if I want to handle all options in the same place, which is the case now.
When you handle them in a different place with different mechanics, it becomes less obvious and harder
to detect. When there is an opertunity to do it in the sample place, I take it.
Also, overwriting an already defined option is not a good path in my view.

> It sounds redundant and implicit. There already are defaults implemented in KubeVirt. For memory, this depends on the selected system, so it would make sense to set the default based on that. Or just let users set it.

Tests defaults do not and usually are not related to production ones. The random name of a vmi is a good example of it.
I used the memory as the existing code logic, it was not my plan to change logic with this refactor. My plan is to make it easier to read, understand and use over the current implementation.
If one wants to optimize it further, it can be done and should be done with time.

> If you want two different names, you use two differnt `WithName("FOO")`. Correct me if I'm wrong or provide a different example of why we cannot do that.

The name is now always random and follows the existing logic.
If one wants a non random name in the future, there will be a need to process the name before creating the vmi.

> You want to replace the current helper code with something new. I'm all for that. But this way of doing things looks like a regression from simplicity/readability perspective to me.

If this is a regression compared to the current code, then indeed you should nack this change.
IMO this is an improvement to the existing code, otherwise I would not have suggested it.
I am not convinced this should be further simplified at the moment and if you or someone else think it should, I welcome you to post such an improvement now or in the future.

tests/libvmi/vmi.go
EdDev: This is true for any code, bad or good. There is always an option to leave code worse or better, no matter what language, algorithm or structure it has.
With every addition you should question if cleaning and refactoring is needed, so I see not problem with moving this in the future when another network related option is introduced. I am not in favor of opening a file just for two defaults.

tests/libvmi/vmi.go
phoracek: I appreciate the effort, but I will not ack this. Despite the code is valid, it seems too complex to me. Writing tests should be as simple as possible and this code is raising the bar for people that would need to extend it.

I would be happy to hear from other reviewers what do they think.

tests/libvmi/vmi.go
maiqueb: It can go either way really; let's first understand if this is the direction we want to go.

tests/libvmi/vmi.go
```diff
@@ -0,0 +1,132 @@
...
+			vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{
+				k8sv1.ResourceMemory: resource.MustParse(value),
+			}
+		}
+	}
+}
+
+// New instantiates a new VMI configuration,
+// building its properties based on the specified With* options.
+func New(namespace, name string, opts ...Option) *kvirtv1.VirtualMachineInstance {
```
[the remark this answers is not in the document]
EdDev: It will look utterly weird and non understood.
You could argue that the name should come in randomized or not already, done by a helper the user can call.

maiqueb:
> > I feel inclined to agree with @phoracek ; the defaults (other than the name) seem superfluous - I don't see why your defaults are better than the ones KubeVirt has.
> 
> They are not better, they where the before and I do not want to change that.
> The intention was to refactor without logic change.

Understood, fair enough.

> 
> > As such, I would prefer to have a layered solution; the lower layer would be stripped of all the defaults, and the users of that bare API would be forced to be explicit. In an upper layer, you could introduce all the relevant defaults for a specific distro - e.g. BasicFedoraVMI would have proper memory defined, RNG, etc. The user would then either adapter the 'standard' VMIs provided by this upper layer, or manipulate the lower one directly.
> > What do you think (I'm aware this is a bit what we have now, but with a builder like api below... ) ?
> 
> Those builders are causing a mess in the current code, they have bad names and require each permutation to create a new builder. I won't be surprise if there are duplicates.
> Adding another option to all, means changing all their names or not mentioning it at all.
> It seems bad to me, hard to maneuver and potentially (it is already IMO) a pile of patching one over the other.

I'm trying to hit the sweet spot between flexibility and code readability. If we have 10 layers, I'm forced to agree that it's a mess. If we have 2 layers (one that gets me pre-defined spec known to work; another that I'll then customize according to whatever weird scenario I'm in) I think it's fairly readable.

As you say, we may end up with duplicated code - that's means we have just found a pattern that is more common than anticipated. We should then evaluate if that should be exported to a common place (e.g. why is it common ? )

> 
> > Keep in mind that it's very, very simple to just create a nameless function close to your code that [gets you the VMI you need](https://github.com/kubevirt/kubevirt/blob/f34ab0ccbf9f1ffe55c049ea5bdb2d84bfd3acc8/tests/vmi_multus_test.go#L102-L112); Your API is competing against pure laziness.
> 
> When I need to read the code I feel differently. I feel I need to be careful not to overwrite someones properties, not to be surprised that some property was added on the way and I mainly feel there is a need to walk all the way to the creation of the object to make sure I get what I hope to get.

If you're working on a nameless function next to the tests that use it, you can change that function without any fear whatsoever; you're the only user of that. 
 
> 
> > I think the most important requirement on it is how simple it is to extend (since every test will require something _a bit_ different), and the currently half unreadable / half surprising approach we have is just simpler to extend.
> 
> Yes, it is very simple to add another function every-time you need a tiny thing. It is also the outcome of this approach that the code base looks as it looks now.

And yet I think the solution is to distribute the spec generation among all the tests, keeping it as concise and specific to the use case at hand as possible. Having a bullet proof spec-generator seems too complex to be of use.

EdDev:
Maybe you should recheck the number of builders there are now, and some do not even express what they are all about beneath. The replaced one is a good example.

I do not have a problem with test local builders which use my suggested approach.
I do not even mind having a very few available for common use.

I do have a problem with Petr perfect ideal non ordered, non default and basically non controlled approach.
But you seem to go a step further and suggesting keeping use of the current function pattern.

Structure is more important than ease, unless it adds too much burden.

maiqueb:
> I find it odd that the preference is to keep on adding endless code to that utils, with no namespace, no sense of structure and no way to converge this into something better.
> 
> I hope one of you will take the step to propose something else, sooner than later. The current codebase, especially that utils file is unmaintainable and very costly for newcomers.

You're getting me wrong; I don't want to add stuff to utils. In fact, I'm 95% OK with this PR. I just think you should remove the defaults from there, and force the user be explicit about what he wants.

Then, have a separate namespace with 'recipes' for a 1 basic VMI spec for each VM distribution.

Finally, every test context would get the basic spec, and update that according to whatever its needs are **in the test**. 

None of those steps add anything to the utils file. 

Let me assure you I want to start removing things from there almost as much as you.

Let's take a step back here, and focus on what we agree; I think the 3 of us agree that the builder pattern approach here is perfectly fine. Would you consider making a PR just with that ?

EdDev:
@phoracek , @maiqueb , hope this version is acceptable to you.

maiqueb:
/lgtm

Thanks.

tests/libvmi/vmi.go
```diff
@@ -0,0 +1,108 @@
...
+// Default VMI values
+const (
+	DefaultTestGracePeriod int64 = 0
+	DefaultVmiName               = "testvmi"
+	DefaultResourceMemory        = "8192Ki"
+)
+
+// NewDefault instantiates a new default VMI configuration,
+// building its extra properties based on the specified With* options.
+func NewDefault(namespace, name string, opts ...Option) *kvirtv1.VirtualMachineInstance {
```
AlonaKaplan, line 43: s/name/namePrefix

tests/utils.go
```diff
@@ -2096,11 +2097,17 @@ func AddEphemeralCdrom(vmi *v1.VirtualMachineInstance, name string, bus string,
...
 func NewRandomFedora32VMIWithFedoraUser() *v1.VirtualMachineInstance {
-	vmi := NewRandomVMIWithEphemeralDiskAndUserdata(ContainerDiskFor(ContainerDiskFedora), `#!/bin/bash
-	    echo "fedora" |passwd fedora --stdin
-        echo `)
-	vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse("512M")
-	return vmi
+	userData := `#!/bin/bash
+echo "fedora" |passwd fedora --stdin
+echo `
+	return libvmi.NewDefault(
```
AlonaKaplan, line 2103: 1. Having a method called `NewDefault` that gets parameters is very confusing (especially when two of the params passed here have "default" in their name). I would expect `NewDefault` to create a default vmi a not getting any param.
Something like this - `NewDefault().WithResourceMemory("512M").With(Rng())...CreateInstance()
Or at least -
baseVmi := libvmi.NewDefault()
libvmi.CreateVmi(baseVmi, libvmi.WithResourceMemory("512M"), libvmi.WithRng() ...)
2. I understand the logic behind this PR and I think it can be very useful that our vmi factory methods will look like the PR suggests, with the `with` building blocks. And if a special vmi is needed it can be easily constructed with this building blocks.
However NewRandomVMIWithEphemeralDiskAndUserdata is not special. It has 101 use cases in the tests.
All those use cases don't care about the Rng, the name, the namespace and the b64Encoding of the userData.
I would expect the code here to look something like this -
baseVmi = NewRandomVMIWithEphemeralDiskAndUserdata(ContainerDiskFor(ContainerDiskFedora), `#!/bin/bash
        echo `)
return libvmi.CreateVmi(baseVmi, libvmi.WithResourceMemory("512M"))

EdDev: >     1. Having a method called `NewDefault` that gets parameters is very confusing (especially when two of the params passed here have "default" in their name). I would expect `NewDefault` to create a default vmi and not getting any param.
>        Something like this - `NewDefault().WithResourceMemory("512M").With(Rng())...CreateInstance()
>        Or at least -
>        baseVmi := libvmi.NewDefault()
>        libvmi.CreateVmi(baseVmi, libvmi.WithResourceMemory("512M"), libvmi.WithRng() ...)

The defaults represents the current code, I have not changed that logic.
They come from the common of all VMI creations. The fact that you can recognize this here is the whole point of this change.
I think it will be easier to convert other usages this way and only then take the `NewDefault` out of the picture.
Refactoring cannot be done in one shot.

Not accepting any parameters and then building it with some other builder functions is exactly what I wanted to avoid by this change. This is how things are built now and it looks terrible and very hard to read... each layer adds its own stuff and to understand what we have one needs to go all the way down the stack.
I see no problem with creating the default VMI and adding to it all options one needs. This explicitly causes the caller to build the VMI in one place for the reader to see.

> 
>     2. I understand the logic behind this PR and I think it can be very useful that our vmi factory methods will look like the PR suggests, with the `with` building blocks. And if a special vmi is needed it can be easily constructed with this building blocks.
>        However NewRandomVMIWithEphemeralDiskAndUserdata is not special. It has 101 use cases in the tests.
>        All those use cases, currently, don't need care about the Rng, the name, the namespace and the b64Encoding of the userData.

The users have no idea what they get with that factory function, so they explicitly overwrite options because they do not trust what is configured down the stack.
If one wants to use such a function, they can still do it and explicitly build the VMI with all the options, including the RNG, hiding it from the user. But if the user wants to see what it means, he just needs to look one layer down and not 10.

> 
> 
> I would expect the code here to look something like this -
> 
> ```
> baseVmi = NewRandomVMIWithEphemeralDiskAndUserdata(ContainerDiskFor(ContainerDiskFedora), `#!/bin/bash
> 	    echo "fedora" |passwd fedora --stdin
>         echo `)
> return libvmi.CreateVmi(baseVmi, libvmi.WithResourceMemory("512M"))
> ```

This will keep the current state and make this change irrelevant. Maybe a bit nicer, but not changing how I expect developers to use it.
I expect users to build specific factories that use the `libvirt.New` with the `With*` options.
I do not want them to edit the VMI after through a different API.
The factories, like the `NewDefault` can collect options on their way, but I think it will be much nicer to collect a collection of options and name them so factories or tests directly can use them.

EdDev:
@AlonaKaplan , the changes have been adjusted per the last review discussion.

- A factory file contains helpers to create dedicated VMI configurations.
- The NewDefault factory has been removed in favor of specific ones.

tests/config.go
```diff
@@ -23,6 +23,8 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"os"
+
+	"kubevirt.io/kubevirt/tests/flags"
```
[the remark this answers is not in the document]
EdDev: This is actually related to the dependent pr and not this one.
I will give an initial answer here but if you want to continue the discussion I suggest we take it to that PR.

Golang flags is a std lib that is used to manage command line arguments/options and each application/lib can define with it its own flags. So this package now centralizes all interpreted cli arguments and sets the globals. The name is from golang, it confused me as well initially until I met “flags” lib.

BTW, the names you mention are not functions.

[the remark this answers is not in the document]
RamLavi: you're right. thanks. so this means that all these params are now received via argument params?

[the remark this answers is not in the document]
EdDev: They were always received that way, I just extracted them from utils to a separate file and later made it  a package.

tests/libvmi/storage.go
```diff
@@ -0,0 +1,109 @@
...
+// WithContainerImage specifies the name of the container image to be used.
+func WithContainerImage(name string) Option {
+	return func(vmi *kvirtv1.VirtualMachineInstance) {
+		diskName, bus := "disk0", "virtio"
+		vmi.Spec.Domain.Devices.Disks = append(vmi.Spec.Domain.Devices.Disks, newDisk(diskName, bus))
+		vmi.Spec.Volumes = append(vmi.Spec.Volumes, newContainerVolume(diskName, name))
+	}
+}
+
+func addDisk(vmi *kvirtv1.VirtualMachineInstance, disk kvirtv1.Disk) {
```
phoracek, line 35: So many helpers just for storage. And it is not even hiding too much logic, just the structure.

Feels too low of an abstraction to be worth so many functions. For somebody who knows basics of VMI API it would be easier to read the structure alone. Jumping between these functions is tiresome.

Don't you think we should put the bar higher, don't put these simple assignments to functions and only abstract higher level logic?

EdDev:
Rebase over master with no other changes.

tests/libvmi/network.go
```diff
@@ -0,0 +1,50 @@
...
+ *
+ */
+
+package libvmi
+
+import (
+	kvirtv1 "kubevirt.io/client-go/api/v1"
+)
+
+// WithInterface adds a Doman Device Interface.
```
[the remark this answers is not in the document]
EdDev: Done

tests/libvmi/network.go
```diff
@@ -0,0 +1,50 @@
...
+func WithInterface(iface kvirtv1.Interface) Option {
+	return func(vmi *kvirtv1.VirtualMachineInstance) {
+		vmi.Spec.Domain.Devices.Interfaces = append(
+			vmi.Spec.Domain.Devices.Interfaces, iface,
+		)
+	}
+}
+
+// WithNetwork adds a network objec.
```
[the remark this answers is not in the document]
EdDev: Done

EdDev:
Changes: Typo fixes (based on recent review).

tests/libvmi/cloudinit.go
```diff
@@ -0,0 +1,53 @@
...
+package libvmi
+
+import (
+	"encoding/base64"
+
+	kvirtv1 "kubevirt.io/client-go/api/v1"
+)
+
+// WithCloudInitNoCloudUserData adds cloud-init no-cloud user data.
+func WithCloudInitNoCloudUserData(data string, b64Encoding bool) Option {
```
phoracek, line 29: Each of these Options should have a unit test making sure it does what we expect and documenting what is the result VMI for reviewer's comfort.

EdDev: Sorry, I do not intend to unit test a test library. It is exercised by tests with clear expectations, therefore it should be safe as is.
Test code is know for its limitations, this is one of them.

phoracek: Or it may cause all e2e tests report false positives testing one thing while we expect another.

If you refuse to write tests documenting what is the result VMI, document it in a docstring. It may get outdated when one of the many helper functions gets changed, but it is the second best thing after unit tests to keep this readable.

EdDev: Yes, that is the risk of tests. Production code is protected by tests, and tests are protected by always running the scenarios and passing reviews. I am not familiar with tests having tests.
That said, a common pattern (which I see not implemented much) is to make sure the test fails before it passes.

It is also interesting that you think the lib is not tested, when the tests are using it and in practice document their usage.

Regarding documentation, I am fine with adding it as long as it helps the tester. Please be specific and elaborate here what you expect to see on the docstring that is missing. This specific one is pretty straight forward to me, but maybe I am missing something.

phoracek: > Yes, that is the risk of tests. Production code is protected by tests, and tests are protected by always running the scenarios and passing reviews. I am not familiar with tests having tests.

You haven't wrote a test. You wrote a library working with API that is used in tests. When I add a new `With` function that consists of several helper function I would first write a test for it to make sure it does what I expect it to do. i.e. the result matches documented KubeVirt API I intend to use.

> That said, a common pattern (which I see not implemented much) is to make sure the test fails before it passes.
> 
> It is also interesting that you think the lib is not tested, when the tests are using it and in practice document their usage.

You suggest I should go over our e2e tests to learn how it is used? Embrace the copy-paste.

If one of the e2e tests fails, I want to be sure that it failed because of a change in production code, not due to helpers. That can be easily done by testing the helpers with unit tests before using them in e2e.

> 
> Regarding documentation, I am fine with adding it as long as it helps the tester. Please be specific and elaborate here what you expect to see on the docstring that is missing. This specific one is pretty straight forward to me, but maybe I am missing something.

It would certainly help me and many others who are used to read the YAML API of KubeVirt. Unlike in other projects, the YAML is here the main entrypoint for users. I expect everyone to see it on daily basis.

I want to see a doctring that shows the VMI YAML created with this option. A truncated VMI manifest with parts added by this function and all its subfunctions.

EdDev: > > Yes, that is the risk of tests. Production code is protected by tests, and tests are protected by always running the scenarios and passing reviews. I am not familiar with tests having tests.
> 
> You haven't wrote a test. You wrote a library working with API that is used in tests. When I add a new `With` function that consists of several helper function I would first write a test for it to make sure it does what I expect it to do. i.e. the result matches documented KubeVirt API I intend to use.

We have `myTest` testing `Foo`, now we refactored this foo and the test is now calling `BetterFoo`.
If the test was right, if we can make it fail with bad input, if we can trust it, then we can use a different helper and we can trust that helper because the test is still passing. So I think it is fine to consider that lib is tested, because it is used.

> > It is also interesting that you think the lib is not tested, when the tests are using it and in practice document their usage.
> 
> You suggest I should go over our e2e tests to learn how it is used? Embrace the copy-paste.

No, I expect reviewers to make sure any functionality that is added to the lib is indeed used by the tests. If something was added that is not used, it means we do not know if it is working or not.

> 
> If one of the e2e tests fails, I want to be sure that it failed because of a change in production code, not due to helpers. That can be easily done by testing the helpers with unit tests before using them in e2e.

It may be true that it may be useful to unit tests lib-tests, but this is not a common practice. Reasons why it is not a common practice was outlined earlier.

> 
> > Regarding documentation, I am fine with adding it as long as it helps the tester. Please be specific and elaborate here what you expect to see on the docstring that is missing. This specific one is pretty straight forward to me, but maybe I am missing something.
> 
> It would certainly help me and many others who are used to read the YAML API of KubeVirt. Unlike in other projects, the YAML is here the main entrypoint for users. I expect everyone to see it on daily basis.

It is not useful to me, and I did not see that a yaml was used in the code. Structures have been used, and if one was interested to see the details he could have gone to the schema that outlines all structures.
Personally I have no interest to see all VMI schema details or know about them.

When I want to read or write a test, I do not what to think about the yaml, I just want to structure the configuration in the easiest and faster way possible. If I am required to know all the details of how the yaml needs to look, it means it will take me more brain power to read and write tests.
The helpers are there to help avoid that formatting knowledge. Current helpers do this as well, maybe at higher levels but still do it. The fact that this is now used in lower levels as well is the same to me and helps me.

I do expect the test writer to trust these helpers, otherwise he should not use them and he can just create it himself from scratch.

> 
> I want to see a doctring that shows the VMI YAML created with this option. A truncated VMI manifest with parts added by this function and all its subfunctions.

If the tester is interested in such a detail data, he can dive in and read how it is build or just go to the schema (per the returned structure type) and examine how it looks.

phoracek: I don't mind if it is Go struct instead of yaml.

I will not lgtm this PR if it will use so much indirection while building structures without documentation of the result. When I review somebody's code using this new library, I want to know what is being created.

YAML manifests is what users will be using to create a VM. I want to be sure we are testing what a user would use.

I am not going to build the project and run the helper function to see what is the output.

EdDev: > I will not lgtm this PR if it will use so much indirection while building structures without documentation of the result. When I review somebody's code using this new library, I want to know what is being created.

You are expected to trust the helpers, if you do not, you should not use them.
My goal is to make the test read well so one can understand what it does, not to make the yaml visible (imp details).
Thank you for reviewing.

phoracek: Public API, that is this test suite exercising, is not an implementation detail. Hopefully you'll come around once you start using KubeVirt.

EdDev: Sorry, but testers do not need to know these details. If they do, they get slower.
If they have helpers that simplifies this for them, they can avoid this brain energy waste.

It is also what really happens in other places, but at different levels of abstraction.
You are trying to insist that higher level abstraction has different rules from lower ones. I on the other hand think it does not matter at what level I am, abstraction and readability fits everywhere.

phoracek (approved):
/lgtm

You should feel bad for the lack of docs, @EdDev.

It is still better than what we have now, +1.

tests/libvmi/storage.go
```diff
@@ -0,0 +1,109 @@
...
+	if !diskExists(vmi, disk) {
```
xpivarc, line 36: @EdDev Any reason why to silently ignore options? Isn't it better to panic?

EdDev: @xpivarc , the `addDisk` (and probably also volume) are serving at the current implementation the could-init storage.
These are called from two functions, the one that handles the [user-data](https://github.com/kubevirt/kubevirt/blob/a9292cdb466d4e852299d809bc36b1edb350d35c/tests/libvmi/cloudinit.go#L29) and the one that uses the [network-data](https://github.com/kubevirt/kubevirt/blob/a9292cdb466d4e852299d809bc36b1edb350d35c/tests/libvmi/cloudinit.go#L45).

Therefore, the original idea was that if one option is added, the other should just re-use the resources that are already there.
Currently, the disks and volumes have been fixed per the task in hand. It also reflected the actual code that existed, perhaps with some exceptions.

PR #3434, opened 2020-05-18
Fix tests which are related to cpu model / features

### LINKED ISSUE #3390: [flaky test] Host CPU does not provide required features: svm

>

/kind bug

> /kind enhancement

**What happened**:

Flakefinder shows on the report from the last [7 days](https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2020-05-07-168h.html) that

```
[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:component]VMIlifecycle [rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:component]Creating a VirtualMachineInstance with node feature discovery [test_id:1639]the vmi with cpu.model matching a nfd label on a node should be scheduled
```

fails on 

```
/root/go/src/kubevirt.io/kubevirt/tests/vmi_lifecycle_test.go:909
Unexpected Warning event received: testvmibrjjk7r4fzt8bfn48gv4vsvnjl2m28v2vbqkfcstxs9rgh7w,6b308098-de40-49c1-99cc-b400b23af177: server error. command SyncVMI failed: "LibvirtError(Code=91, Domain=31, Message='the CPU is incompatible with host CPU: Host CPU does not provide required features: svm')"
Expected
    <string>: Warning
not to equal
    <string>: Warning
/root/go/src/kubevirt.io/kubevirt/tests/utils.go:390
```

An example can be found here: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3364/pull-kubevirt-e2e-k8s-1.15-ceph/1255825501830778880.

Note that the [node dump](https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3364/pull-kubevirt-e2e-k8s-1.15-ceph/1255825501830778880/artifacts/1_nodes.log) shows this label:

```
                    "feature.node.kubernetes.io/cpu-feature-svm": "true",
```

So kubevirt seems to think that the flag is there, but libvirt seems to think that it is not there.

One of our components is wrong. Either we should report the flag as "false", or libvirt should accept the cpu.

**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ksimon1:
Virsh reports some cpu models as supported, but libvirt refuses to run vms with that cpu model/feature.
This PR adds cpu model / feature black list to tests. These tests will not run with qemu64, Opteron_G2 cpu models and svm feature.

Fixes #3390

**Release note**:
```NONE

```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -1859,6 +1859,39 @@ func NewRandomVMIWithNS(namespace string) *v1.VirtualMachineInstance {
 	return vmi
 }
 
+func GetSupportedCPUFeatures(node k8sv1.Node) []string {
+	var featureBlackList = map[string]bool{
+		"svm": true,
+	}
+	features := make([]string, 0)
+	for key := range node.Labels {
+		if strings.Contains(key, services.NFD_CPU_FEATURE_PREFIX) {
+			feature := strings.TrimPrefix(key, services.NFD_CPU_FEATURE_PREFIX)
+			if _, ok := featureBlackList[feature]; !ok {
+				features = append(features, feature)
+			}
+		}
+	}
+	return features
+}
+
+func GetSupportedCPUModels(node k8sv1.Node) []string {
+	var cpuBlackList = map[string]bool{
+		"qemu64":     true,
+		"Opteron_G2": true,
+	}
+	cpus := make([]string, 0)
+	for key := range node.Labels {
+		if strings.Contains(key, services.NFD_CPU_MODEL_PREFIX) {
+			cpu := strings.TrimPrefix(key, services.NFD_CPU_MODEL_PREFIX)
+			if _, ok := cpuBlackList[cpu]; !ok {
+				cpus = append(cpus, cpu)
+			}
+		}
+	}
+	return cpus
+}
+
 func NewRandomVMIWithDataVolume(dataVolumeName string) *v1.VirtualMachineInstance {
 	vmi := NewRandomVMI()
 
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
+	k8sv1 "k8s.io/api/core/v1"
 	kubev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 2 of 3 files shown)


enp0s3:
@ksimon1 IIUC the new helper functions that you've introduced return the first supported CPU model and first supported CPU feature that is fit the filters. Since the return value isn't consistent but relevant to the test, maybe it is worth to define these functions in the scope of the test itself?

tests/utils.go
```diff
@@ -1859,6 +1859,37 @@ func NewRandomVMIWithNS(namespace string) *v1.VirtualMachineInstance {
 	return vmi
 }
 
+func GetSupportedCPUFeature(node k8sv1.Node) string {
```
enp0s3, line 1862: @ksimon1 See my comment in the conversation

tests/utils.go
```diff
@@ -1859,6 +1859,37 @@ func NewRandomVMIWithNS(namespace string) *v1.VirtualMachineInstance {
...
+			feature := strings.TrimPrefix(key, services.NFD_CPU_FEATURE_PREFIX)
+			if _, ok := featureBlackList[feature]; !ok {
+				return feature
+			}
+		}
+	}
+	return ""
+}
+
+func GetSupportedCPUModel(node k8sv1.Node) string {
```
enp0s3, line 1877: @ksimon1 Same

tests/utils.go
ksimon1: I wanted to avoid code duplicate, because the same functions are called from two different files.

enp0s3: @ksimon1 Oh I understand. Maybe for a matter of consistency return array of strings that represent non blacklisted models/features and in the test code take whatever item you want.

ksimon1:
retest

ksimon1:
So what about adding those to cpus to black list?

tests/vmi_configuration_test.go
```diff
@@ -1350,15 +1350,15 @@ var _ = Describe("Configurations", func() {
...
-		libvirtCPUFeatureRegexp := regexp.MustCompile(`<feature name='(\w+)'/>`)
 		cpuModelNameRegexp := regexp.MustCompile(`Model name:\s*([\s\w\-@\.\(\)]+)`)
 
 		var libvirtCpuModel string
-		var libvirtCpuVendor string
 		var cpuModelName string
-		var cpuFeatures []string
 		var cpuVmi *v1.VirtualMachineInstance
 
+		nodes := tests.GetAllSchedulableNodes(virtClient)
```
rmohr, line 1359: Can you move that to a BeforeAll or BeforeEach?

rmohr:
> It's unclear how this is fixing the underlying issue in #3390. From what I understand, the issue is we're reporting cpu features/models that can't actually be used. So rather than having the tests ignore these labels that can't be used, why are we not focusing on making the labels accurate to begin with?

@davidvossel the labels are accurate. We detect the right thing, but libvirt has a bug where it blocks the cpus due to wrong internal checks. The is a workaround in #3403. There we disable the libvirt checks and check in virt-handler if we found out that it works. Since we did not come to an agreement, but tests continue to fail, this PR should fix the tests in the meanwhile.

rmohr:
/hold

Things look strange

rmohr:
/hold cancel

Not sure what was wrong before, but the code looks sane. Removing the hold.

rmohr:
Relevant test results are good. Going to merge this manually in order to reduce the failure rate to stabilize CI.

PR #3437, opened 2020-05-18
virt-operator, tests: Fix equality of slices

EdDev:

When comparing slices, the two need to be compared without order.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/virt-operator/util/config_test.go
+++ b/pkg/virt-operator/util/config_test.go
@@ -21,7 +21,6 @@ package util
 import (
 	"fmt"
 	"os"
-	"reflect"
 
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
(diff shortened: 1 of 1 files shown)


pkg/virt-operator/util/config_test.go
```diff
@@ -176,6 +177,14 @@ var _ = Describe("Operator Config", func() {
...
 
+			sortEnvVars := func(envVars []k8sv1.EnvVar) {
+				sort.Slice(
+					envVars,
+					func(i, j int) bool { return envVars[i].Name > envVars[j].Name },
+				)
+			}
+			sortEnvVars(*envObjects)
+			sortEnvVars(expected)
 			Expect(reflect.DeepEqual(*envObjects, expected)).To(BeTrue())
```
rmohr, line 188: Can you please use `Expect(envVars).To(ConsistOf(expected))`? That will do the right thing.

EdDev: Will it do a "deep" equality?

I will argue that that the source of the problem is the use of `rand` in the first place here. That is not recommended in tests.

EdDev: Done

rmohr: Yes it will: https://onsi.github.io/gomega/ (you have to search for ConsistOf there, can't find anchors).

PR #3438, opened 2020-05-18
Add missing roles to csv-generator

danielBelenky:
With recent changes, we now require extra roles to properly operator
virt-operator. Some of the roles are namespaced. This commit syncs the
namespaced roles to the generated CSV.

```release-note
NONE
```


rmohr:
@tiraboschi I am a little bit confused. I thought that you said that HCO is using `tools/manifest-templator`. Here we only fix `tools/csv-generator`. Is that still already good enough?

PR #3448, opened 2020-05-19
In the Development Guide, minimum compatible version in the cluster folder should be 1.14.

pnavarro:
Minimum compatible version in the cluster folder is 1.14 not 1.11.0


rmohr:
Thanks for the PR. I think it is 1.15 as we seek.

pnavarro:
Ok, Thanks Roman, I will close this one and I will open another one with the sign-off

Issue #3187, opened 2020-03-21
panic: Search domain is not valid: 'DHCP'


watermelon-brother:
**What happened**:
when I run the vm ,define below
```yaml
apiVersion: kubevirt.io/v1alpha3
kind: VirtualMachine
metadata:
  labels:
    kubevirt.io/vm: testvm1
  name: testvm1
spec:
  running: false
  template:
    metadata:
      labels:
        kubevirt.io/vm: testvm1
    spec:
      domain:
        devices:
          disks:
          - disk:
              bus: virtio
            name: containerdisk
          - disk:
              bus: virtio
            name: cloudinitdisk
        machine:
          type: ""
        resources:
          requests:
            memory: 64M
      terminationGracePeriodSeconds: 0
      volumes:
      - name: containerdisk
        containerDisk:
          image: kubevirt/cirros-container-disk-demo:latest
      - cloudInitNoCloud:
          userDataBase64: IyEvYmluL3NoCgplY2hvICdwcmludGVkIGZyb20gY2xvdWQtaW5pdCB1c2VyZGF0YScK
        name: cloudinitdisk
```
the vmi virt-launcher-testvm-vc7g9 states:
```shell
NAME                         READY   STATUS    RESTARTS   AGE
test-7b48b75784-79sv7        1/1     Running   0          112m
virt-launcher-testvm-vc7g9   0/2     Error     0          34s
```
locking in the logs:
```shell
{"component":"virt-launcher","kind":"","level":"info","msg":"bridge pod interface: VIF: { Name: eth0, IP: 10.42.0.54, Mask: ffffff00, MAC: 0a:58:0a:2a:00:36, Gateway: 10.42.0.1, MTU: 1450}","name":"vm-http-datavolume","namespace":"default","pos":"podinterface.go:258","timestamp":"2020-03-21T08:07:35.905419Z","uid":"3c3582a6-edfe-4ab1-8d67-2b527998a0b1"}
{"component":"virt-launcher","level":"info","msg":"Found nameservers in /etc/resolv.conf: \n+\u0000\n","pos":"converter.go:1508","timestamp":"2020-03-21T08:07:35.905569Z"}
{"component":"virt-launcher","level":"info","msg":"Found search domains in /etc/resolv.conf: default.svc.cluster.local svc.cluster.local cluster.local DHCP","pos":"converter.go:1509","timestamp":"2020-03-21T08:07:35.905640Z"}
[...]

PR #3449, opened 2020-05-19
Tolerate uppercase letters in "search SomeDomain".

maya-r:
Domain names are case insensitive.

This affects my own networking setup, where an ISP sometimes
adds "search Home" to /etc/resolv.conf.

```release-note
NONE
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
+++ b/pkg/virt-launcher/virtwrap/network/dhcp/dhcp.go
@@ -46,7 +46,7 @@ const (
 
 // simple domain validation regex. Put it here to avoid compiling each time.
 // Note this requires that unicode domains be presented in their ASCII format
-var searchDomainValidationRegex = regexp.MustCompile(`^(?:[_a-z0-9](?:[_a-z0-9-]{0,61}[a-z0-9])?\.)*(?:[a-z](?:[a-z0-9-]{0,61}[a-z0-9])?)?$`)
+var searchDomainValidationRegex = regexp.MustCompile(`^(?:[_a-zA-Z0-9](?:[_a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)*(?:[a-zA-Z](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)?$`)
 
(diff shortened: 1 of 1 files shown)


rmohr:
A nice test might be to add an uppercase domain to the VMIs custom DNS configuration and check if nothing blows up.

dhiller:
/hold

waiting for tests. Please remove the tests then

maya-r:
Having some trouble writing a test because we use kubernetes/apimachinery's "DNS RFC 1123 validator" in more places, and it also doesn't allow uppercase domains.

[This is our use of it](https://github.com/kubevirt/kubevirt/blob/5de00f252e978fc312dfa2573110ca59da9d998d/pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go#L1077-L1085) and this is the [upstream validation rule](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L172)

The few tests I was running at the time started passing, but injecting it via dnsConfig.Searches triggers the validation.

rmohr:
@maya-r could we at least add a few unit tests?

rmohr:
@maya-r we also had reports from different managed k8s distrubitions that some uppercase letters are present. Would be great if we can bring this in.

vasiliy-ul:
> This affects my own networking setup, where an ISP sometimes
> adds "search Home" to /etc/resolv.conf.

Looks like I am in the same situation with `search Home` in my `resolv.conf`. I wonder, is it possible to move forward with this PR? I have found a couple of related issues https://github.com/kubernetes/kubernetes/issues/71140 and https://github.com/kubernetes/kubernetes/issues/94088. Apparently the DNS validator in `kubernetes/apimachinery` is implemented incorrectly. I assume changing the regexp in kubevirt to allow `A-Z` or normalizing always to lower-case should be safe enough.

oshoval:
Is this PR still needed ?
and if so, it might fix https://github.com/kubevirt/kubevirt/issues/3187 ?
or https://github.com/kubevirt/kubevirt/pull/4817 was sufficient

maya-r:
Yeah that fix is sufficient.

PR #3452, opened 2020-05-20
Improve lane selection for repeated tests

dhiller:
The list of kubevirt providers for the flakiness check lane was hard coded, this is now replaced by fetching the available k8s providers from the `cluster-up` directory, sorting them and picking the three most recent ones.

Fixes #

/cc @oshoval 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/repeated_test.sh
+++ b/automation/repeated_test.sh
@@ -51,6 +51,22 @@ if (( $# > 0 )); then
     fi
 fi
 
+if (( $# > 0 )); then
+    declare -a TEST_LANES
+    max="$#"
+    while [ ! $max -lt 1 ]; do
+        max=$((max-1))
+        TEST_LANES[$max]="$1"
+        shift
+    done
+else
+    mapfile -t TEST_LANES < <( find cluster-up/cluster -name 'k8s-[0-9].[0-9][0-9]' -print | sort -rV | grep -oE 'k8s.*' | head -3 )
+fi
+echo "Test lanes: ${TEST_LANES[*]}"
+for lane in "${TEST_LANES[@]}"; do
+    [ -d "cluster-up/cluster/$lane" ] || ( echo "provider $lane does not exist!"; exit 1 )
+done
+
 if [[ -z ${TARGET_COMMIT-} ]]; then
     # if there's no commit provided default to the latest merge commit
     TARGET_COMMIT=$(git log -1 --format=%H --merges)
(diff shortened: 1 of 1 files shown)


automation/repeated_test.sh
```diff
@@ -80,7 +80,7 @@ if (( $# > 0 )); then
         shift
     done
 else
-    TEST_LANES=( 'k8s-1.14' 'k8s-1.15' 'k8s-1.16' )
+    TEST_LANES=( "$( find cluster-up/cluster -name 'k8s-[0-9].[0-9][0-9]' | sort -r | grep -oE 'k8s.*' | head -3 )" )
```
oshoval, line 83: would it catch kind ones ? do we want to catch them ?

dhiller: no, we only want the k8s lanes. The output on my system is as follows:

```
> find cluster-up/cluster -name 'k8s-[0-9].[0-9][0-9]'
cluster-up/cluster/k8s-1.17
cluster-up/cluster/k8s-1.16
cluster-up/cluster/k8s-1.15
cluster-up/cluster/k8s-1.14
```

oshoval: ok thanks, and just 3 , well maybe we need 4 ?
because David Vossel said once that kubevirt should support 2 back, and sometimes we are already developing one forward, so its 4

dhiller:
/hold

there's a test commit that I needed to check the run on ci, will remove that later on.

automation/repeated_test.sh
dhiller: For the use case of finding flaky tests three should be good enough, I think. If required we coul djust adjust the parameters. Please bear in mind that in this lane the tests are run three times (by default, see [docs](https://github.com/kubevirt/kubevirt/blob/master/automation/repeated_test.sh#L7) for options) against each of the providers. The other lanes already test all the other k8s versions and flavors.

oshoval: sgtm
/lgtm
Thanks

tests/config_test.go
```diff
@@ -32,7 +32,7 @@ import (
 	"kubevirt.io/kubevirt/tests"
 )
 
-var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:component]Config", func() {
+var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:component]Config ", func() {
```
oshoval, line 35: need space for the next string i guess ?

dhiller: this is the test commit I meant. will remove this later.

automation/repeated_test.sh
jean-edouard: Nit: maybe `sort -rV`?

dhiller:
/hold cancel

removed the test commit, looks good so far.

automation/repeated_test.sh
dhiller: Good point!

dhiller:
@jean-edouard @oshoval had to resort to moving this up a little in order to fail earlier if the provider is not found, furthermore I needed to use mapfile. PTAL thanks!

oshoval:
nice
thanks
/lgtm

PR #3455, opened 2020-05-22
Use KubeVirt CR to store configuration

ashleyschuett:
Signed-off-by: Ashley Schuett <ashleyns1992@gmail.com>

This allows all configuration values to be configurable from the KubeVirt CR. This PR is an enhancement  of the original PR https://github.com/kubevirt/kubevirt/pull/2996

Fixes #

This PR moves all configuration fields to come from the KubeVirt CR. It will ignore the configuration in the KubeVirt CR if the configmap is present for backwards compatibility. 

**Release note**:

```release-note
Add KubevirtConfiguration, MigrationConfiguration, DeveloperConfiguration and NetworkConfiguration to API-types
```


pkg/virt-config/config-map.go
```diff
@@ -427,35 +403,87 @@ func setConfig(config *Config, configMap *k8sv1.ConfigMap) error {
...
+	log.DefaultLogger().Infof("Updating cluster config to resource version '%s'", config.ResourceVersion)
+	c.lastValidConfig = config
+	return c.lastValidConfig
+}
+
+func (c *ClusterConfig) getConfigFromKubeVirtCR() *v1.KubeVirt {
+	// TODO: I'm unsure how we know which kubevirt CR is the one in use
+	// once it is created it will have the status "DEPLOYED" but what about
+	// when it is still initalizing?
+	if obj, exists, err := c.kubeVirtInformer.GetStore().GetByKey(c.namespace + "/" + "kubevirt"); err != nil {
```
ashleyschuett, line 478: Is there a way to know which KubeVirt CR creation triggered the creation of the `virt-api` and `virt-controller`? If so I could then pass that name to the `NewClusterConfig` along with the namespace. But i've been unable to figure out how?  or maybe there is a better idea for this?

ashleyschuett:
@rmohr based on our discussion does this look like the right path forward or should we discuss this more?

pkg/virt-config/config-map.go
rmohr: If there are multiple CRs, the kubevirt-operator should stop doing anything. Therefore you should at this stage only see one CR with a status which is indicating that it installed kubevirt. That is the right one. So the status on a second CR should pretty much not even show that it is initializing.

staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -1275,3 +1281,51 @@ type RenameOptions struct {
 	NewName         string  `json:"newName"`
 	OldName         *string `json:"oldName,omitempty"`
 }
+
+// KubeVirtConfiguration holds all kubevirt configurations
+// +k8s:openapi-gen=true
+type KubeVirtConfiguration struct {
+	ResourceVersion             string                  `json:"resourceVersion,omitempty"`
```
rmohr, line 1288: The ResourceVersion should be the one from the ConfigMap or the KubeVirtCR. Do we have a way to not have it included here?

ashleyschuett:
@rmohr this has been updated to remove the `ResourceVersion` under the Configuration and I have updated the CR detection.

pkg/virt-config/config-map.go
```diff
@@ -434,35 +415,88 @@ func setConfig(config *Config, configMap *k8sv1.ConfigMap) error {
...
+	var err error
+	if useConfigMap {
+		err = setConfigFromConfigMap(config, configMap)
+	} else {
+		err = setConfigFromKubeVirt(config, kv)
+	}
+
+	if err != nil {
+		c.lastInvalidConfigResourceVersion = resourceVersion
+		log.DefaultLogger().Reason(err).Errorf("Invalid cluster config with resource version '%s', falling back to last good resource version '%s'", resourceVersion, c.lastValidConfigResourceVersion)
```
rmohr, line 480: Can we differentiate in the error if it is an error coming from the CR or from the config-map? May make debugging easier.

pkg/virt-operator/creation/rbac/apiserver.go
```diff
@@ -148,6 +148,19 @@ func newApiServerClusterRole() *rbacv1.ClusterRole {
 					"watch",
 				},
 			},
+			{
+				APIGroups: []string{
```
rmohr, line 152: Do we miss the permissions for virt-controller? I could only find permissions for virt-api and virt-handler.

pkg/virt-config/config-map.go
```diff
@@ -67,56 +62,71 @@ const (
...
-	if err != nil {
-		panic(err)
-	}
-
-	return cfgMap
-}
+// Config has been deprecated in favor of the `KubeVirtSpec.Confguration` option on the
+// KubeVirt Spec. Please add new values there. This is only used for backward
+// compatibility until it can be deprecated
+// type Config struct {
```
rmohr, line 68: Were the struct definitions kept intentionally?

pkg/virt-config/config-map.go
ashleyschuett: absolutely! i have added the resource type to this message as well as the "getting resource from cache" error messages.

pkg/virt-config/config-map.go
ashleyschuett: I did so that the "deprecation" comment was kept. but on second thought it doesn't make much sense to keep it. I have now removed it.

pkg/virt-operator/creation/rbac/apiserver.go
ashleyschuett: `virt-controller` already has the proper permissions https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-operator/creation/rbac/controller.go#L135-L143

alonSadan:
Thank you @ashleyschuett  for the enhancement. looks very good!

pkg/virt-config/config-map.go
```diff
@@ -67,56 +62,37 @@ const (
...
+func NewClusterConfig(configMapInformer cache.SharedIndexInformer,
+	crdInformer cache.SharedIndexInformer,
+	kubeVirtInformer cache.SharedIndexInformer,
+	namespace string) *ClusterConfig {
+
+	defaultConfig := defaultClusterConfig()
+
+	objects := kubeVirtInformer.GetStore().List()
+	var kubeVirtName string
+	for _, obj := range objects {
```
rmohr, line 79: One lst point: I would sleep better, if on the long run our logic can deal with unusual but not impossible behaviour on the CR level. One can for instance delete the operator, force-delete the CR, recreate the operator and create a CR with a different name. It should be good enough if we just always list all elements from the cache and choose the CR which has a status.

ashleyschuett: Sorry, I'm not sure I understand. Do you mean this check should be done on every call to get the config to make sure we have the correct KV? So instead of getting it once during `NewClusterConfig`, instead check for the KV name on each run of `getConfigFromKubeVirtCR`? 

In the example it would mean the controller, api, and handler did not get deleted / restarted but he KV got renamed, if I am understanding the edge case correctly.

rmohr: Yes exactly.

rmohr: Wile absolutely an edge case I would not say there are no reasons to do such things at all. It would be great if we are as declarative and eventually consistent as possible. If we don't do this check, people would have to restart manually the components after such an undertaking.

ashleyschuett: makes sense to me! I have updated it now :)

tests/.github/ISSUE_TEMPLATE/__en-US.md
```diff
@@ -0,0 +1,25 @@
+---
```
rmohr, line 1: Did you add the issue templates intentionally in this PR?

pkg/virt-operator/creation/rbac/apiserver.go
rmohr: Great. Apparently they are too wide by default. We can consider moving the operator CR into another apiresource group when we go to v1, so that it does not get implicitly the rights to modify CRs, which it does right now.

rmohr:
/lgtm
/approve

@ashleyschuett would you mind going over the user-guide (https://github.com/kubevirt/user-guide) and update all relevant sections with the new improved way of configuring kubevirt?

staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -1277,3 +1283,51 @@ type RenameOptions struct {
 	NewName         string  `json:"newName"`
 	OldName         *string `json:"oldName,omitempty"`
 }
+
+// KubeVirtConfiguration holds all kubevirt configurations
+// +k8s:openapi-gen=true
+type KubeVirtConfiguration struct {
```
fabiand, line 1289: This could be renamed to VirtualizatioConfiguration

KubeVirtConfiguration leaves more the impression of "system wide settings", but many settings in this group are really virt specific.

But this is nit picking on a high level

rmohr:
@ashleyschuett another PR made it in before yours. Could you do a rebase?

rmohr:
Seems like the travis refactoring attempts still leave a situation where one can't see why travis fails. @ashleyschuett I think you need to run `make generate && make` and push the diff.

ashleyschuett:
@rmohr thank you! I wasn't sure what was failing. Seems to be okay now.

rmohr:
@EdDev could you check the travis flows in case of errors? Like if we see expected output when `make generate` is needed?

EdDev:
> @EdDev could you check the travis flows in case of errors? Like if we see expected output when `make generate` is needed?

Sure, will have a look first thing tomorrow morning.

rmohr:

fails on the known flaky migration test.

rmohr:
Test failures are unrelated. 

@ashleyschuett could you create  a follow-up issue? I just realized that most tests still use the config-map and not the new option.

fabiand:
Nice!

Can we please also have a user guide update explaining this whole topic?

PR #3456, opened 2020-05-22
Use virtio in tests where the disk order matters

rmohr:

Instead of sata use virtio on tests where we expect the same device
order like in the VMI spec. With sata we have no order-guarantee, with
virtio we have, if the VMI has the corresponding kernel drivers
installed.

This fixes flaky test

```
[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:component]Config With a Secret and a ConfigMap defined With a single volume [test_id:786]Should be that cfgMap and secret fs layout same for the pod and vmi
```

where now and then the order of some of the sata disks changes.

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/config_test.go
+++ b/tests/config_test.go
@@ -26,6 +26,8 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/pborman/uuid"
 
+	v1 "kubevirt.io/client-go/api/v1"
+
 	"kubevirt.io/client-go/kubecli"
 	"kubevirt.io/client-go/log"
 	"kubevirt.io/kubevirt/pkg/config"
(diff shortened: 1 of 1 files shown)


vladikr:
Personally, I didn't know that QEMU guarantees the order of devices :/
/lgtm
/approve

rmohr:
> Personally, I didn't know that QEMU guarantees the order of devices :/

Well I am not 100% sure, but afaik the `/dev/vd*` hints in the libvirt xml are taken into account by the virtio drivers of the kernel.

vladikr:
> Well I am not 100% sure, but afaik the `/dev/vd*` hints in the libvirt xml are taken into account by the virtio drivers of the guest kernel.

Yes, it is being taken into account, that's true. But I doubt that there is a guarantee, per se.
I think using virtio is more stable anyway. But if we will see issues we can always use `serial` for disks or even use `tags` and check against the devices metadata.

rmohr:
Relevant test results are good. Going to merge this manually in order to reduce the failure rate to stabilize CI.

PR #3457, opened 2020-05-22
tests: Do not trigger tests for documents changes

EdDev:
In case there are no changes or the only changes introduced
are of *.md files, skip testing.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,11 @@ bazel-push-images:
 push: bazel-push-images
 
 bazel-test:
-	hack/dockerized "hack/bazel-fmt.sh && hack/bazel-test.sh"
+	if git diff --name-only | grep -q -v -e "md$$"; then \
+	    hack/dockerized "hack/bazel-fmt.sh && hack/bazel-test.sh"; \
+	else \
+	    echo "No changes detected, skipping tests."; \
+	fi
 
 generate:
 	hack/dockerized "DOCKER_PREFIX=${DOCKER_PREFIX} DOCKER_TAG=${DOCKER_TAG} IMAGE_PULL_POLICY=${IMAGE_PULL_POLICY} VERBOSITY=${VERBOSITY} ./hack/generate.sh"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -23,7 +23,11 @@ bazel-push-images:
 push: bazel-push-images
 
 bazel-test:
-	hack/dockerized "hack/bazel-fmt.sh && hack/bazel-test.sh"
+	if git diff --name-only | grep -q -v -e "md$$"; then \
```
phoracek, line 26: I think we should match ".md", including the dot. To make sure we are not skipping some non-doc files. run_cmd, bin_amd (made those up).

Makefile
```diff
@@ -50,8 +54,12 @@ go-test: go-build
 test: bazel-test
 
 functest:
-	hack/dockerized "hack/build-func-tests.sh"
-	hack/functests.sh
+	if git diff --name-only | grep -q -v -e "md$$"; then \
```
phoracek, line 57: What if i want to run functoinal tests on top of master and the last merged commit happens to be a doc change? I think we should limit execution of tests only on CI, not per explicit request on make target.

Makefile
EdDev: Yes, I agree.
There is however a worse bug here.

Makefile
EdDev: This may present a real problem in general, even on CI.

Jobs that may periodically want to run tests will encounter this issue.
I guess to resolve this we can provide dedicated targets to always run the tests no matter what.

But I am not favoring adding logic just for CI or local user, all options should exists for all in order to have consistency. The current difference between the CI and local run is unfortunate IMO.

phoracek: Right. Or just keep `functest` as it is and add this doc skip to a new smoke tests target you've mentioned.

EdDev:
This needs more work before making it visible, I will re-think it.

PR #3460, opened 2020-05-24
Add masquerade ipv4 cidr test + refactor

EdDev:

Add masquerade network custom IPv4 CIDR test and refactor the existing masquerade test.

This PR is also a preparation for introducing IPv6 CIDR API with its own test.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/vmi_networking_test.go
+++ b/tests/vmi_networking_test.go
@@ -729,7 +729,7 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 				By("Checking traceroute from vmi to cluster nodes gateway")
 				// Cluster nodes subnet (docker network gateway)
 				// Docker network subnet cidr definition https://bit.ly/2wZTgMK
-				err := tests.CheckForTextExpecter(serverVMI, createExpectTraceroute6("2001:db8:1::1"), 180)
+				err := tests.CheckForTextExpecter(serverVMI, createExpectTraceroute6("2001:db8:1::1"), 30)
 				Expect(err).ToNot(HaveOccurred(), "Failed to traceroute to VMI %s within the given timeout", serverVMI.Name)
 			} else {
 				By("Checking ping to google")
--- a/tests/vmi_networking_test.go
+++ b/tests/vmi_networking_test.go
@@ -21,6 +21,7 @@ package tests_test
 
 import (
 	"fmt"
+	"net"
 	"strconv"
 	"strings"
 	"time"
@@ -42,6 +43,7 @@ import (
 	"kubevirt.io/client-go/kubecli"
 	"kubevirt.io/client-go/log"
 	"kubevirt.io/kubevirt/pkg/virt-controller/services"
+	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
 	"kubevirt.io/kubevirt/tests"
 )
 
@@ -691,29 +693,44 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 	})
 
 	Context("VirtualMachineInstance with masquerade binding mechanism", func() {
+		const (
+			defaultCIDR = false
+			customCIDR  = true
+		)
+
 		var serverVMI *v1.VirtualMachineInstance
 		var clientVMI *v1.VirtualMachineInstance
 
-		masqueradeVMI := func(Ports []v1.Port) *v1.VirtualMachineInstance {
+		masqueradeVMI := func(Ports []v1.Port, ipv4NetworkCIDR string) *v1.VirtualMachineInstance {
 			containerImage := tests.ContainerDiskFor(tests.ContainerDiskCirros)
 			userData := "#!/bin/bash\necho 'hello'\n"
 			vmi := tests.NewRandomVMIWithEphemeralDiskAndUserdata(containerImage, userData)
 			vmi.Spec.Domain.Devices.Interfaces = []v1.Interface{{Name: "default", Ports: Ports, InterfaceBindingMethod: v1.InterfaceBindingMethod{Masquerade: &v1.InterfaceMasquerade{}}}}
-			vmi.Spec.Networks = []v1.Network{*v1.DefaultPodNetwork()}
+			net := v1.DefaultPodNetwork()
+			if ipv4NetworkCIDR != "" {
+				net.NetworkSource.Pod.VMNetworkCIDR = ipv4NetworkCIDR
+			}
+			vmi.Spec.Networks = []v1.Network{*net}
 
 			return vmi
 		}
 
-		table.DescribeTable("[test_id:1780]should allow regular network connection", func(ports []v1.Port) {
+		table.DescribeTable("[test_id:1780]should allow regular network connection", func(ports []v1.Port, withCustomCIDR bool) {
+			var ipv4NetworkCIDR string
+
+			if withCustomCIDR {
+				ipv4NetworkCIDR = "10.10.10.0/24"
+			}
+
 			// Create the client only one time
 			if clientVMI == nil {
-				clientVMI = masqueradeVMI([]v1.Port{})
+				clientVMI = masqueradeVMI([]v1.Port{}, ipv4NetworkCIDR)
 				_, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(clientVMI)
 				Expect(err).ToNot(HaveOccurred())
 				tests.WaitUntilVMIReady(clientVMI, tests.LoggedInCirrosExpecter)
 			}
 
-			serverVMI = masqueradeVMI(ports)
+			serverVMI = masqueradeVMI(ports, ipv4NetworkCIDR)
 			serverVMI.Labels = map[string]string{"expose": "server"}
 			_, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(serverVMI)
 			Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


tests/vmi_networking_test.go
```diff
@@ -691,29 +691,44 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+			if ipv4NetworkCIDR != "" {
+				net.NetworkSource.Pod.VMNetworkCIDR = ipv4NetworkCIDR
+			}
+			vmi.Spec.Networks = []v1.Network{*net}
 
 			return vmi
 		}
 
-		table.DescribeTable("[test_id:1780]should allow regular network connection", func(ports []v1.Port) {
+		table.FDescribeTable("[test_id:1780]should allow regular network connection", func(ports []v1.Port, withCustomCIDR bool) {
```
AlonaKaplan, line 716: I guess you forgot to remove the F.

EdDev: Ahh.. thanks.

tests/vmi_networking_test.go
```diff
@@ -863,3 +836,42 @@ func NewRandomVMIWithInvalidNetworkInterface() *v1.VirtualMachineInstance {
 	vmi.Spec.Domain.Devices.Interfaces[0].Model = "gibberish"
 	return vmi
 }
+
+func createExpectTraceroute6(address string) []expect.Batcher {
+	return []expect.Batcher{
+		&expect.BSnd{S: "\n"},
+		&expect.BExp{R: "\\$ "},
+		&expect.BSnd{S: "traceroute -6 " + address + " -w1 > tr\n"},
```
maiqueb, line 860: I personally find this more readable w/ `Sprintf`, but it's a personal preference, feel free to disregard.

tests/vmi_networking_test.go
```diff
@@ -863,3 +836,42 @@ func NewRandomVMIWithInvalidNetworkInterface() *v1.VirtualMachineInstance {
...
+		&expect.BSnd{S: "echo $?\n"},
+		&expect.BExp{R: "1"},
+	}
+}
+
+func createExpectStartTcpServer(port string) []expect.Batcher {
+	return []expect.Batcher{
+		&expect.BSnd{S: "\n"},
+		&expect.BExp{R: "\\$ "},
+		&expect.BSnd{S: "screen -d -m sudo nc -klp " + port + " -e echo -e 'Hello World!'\n"},
```
maiqueb, line 873: Ditto.

tests/vmi_networking_test.go
```diff
@@ -691,27 +691,44 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
 			return vmi
 		}
 
-		table.DescribeTable("[test_id:1780]should allow regular network connection", func(ports []v1.Port) {
+		table.DescribeTable("[test_id:1780]should allow regular network connection", func(ports []v1.Port, withCustomCIDR bool) {
+			var ipv4NetworkCIDR string
+
+			if withCustomCIDR {
+				ipv4NetworkCIDR = "10.10.10.0/24"
+			}
```
maiqueb, line 721: wouldn't it be better if you injected this from the caller ? e.g. instead of a `withCustomCIDR bool` , have a `customCIDR *string ` ?

EdDev: The test needs to check custom CIDR, so this flag is here for both IPv4 and IPv6.
It looked nicer to me to read the test this way, over passing in empty strings on the other two scenarios.

As there is no intention to test now permutations between CIDR with or without IPv6, not different addresses, this seemed better.

tests/vmi_networking_test.go
EdDev: `Sprintf` felt noisy to me, this is why I used this version.
I will change it only if you insist.

tests/vmi_networking_test.go
EdDev: Same, I will change it if you insist.

tests/vmi_networking_test.go
maiqueb: > The test needs to check custom CIDR, so this flag is here for both IPv4 and IPv6.

Hm, this might change my opinion; could you paste a snippet of how you envision the test body with IPv6 ? Just some high level pseudo-code is enough.

> It looked nicer to me to read the test this way, over passing in empty strings on the other two scenarios.

You don't have to pass empty strings around; define `customCIDR` as a pointer, and use nil.

> 
> As there is no intention to test now permutations between CIDR with or without IPv6, not different addresses, this seemed better.

EdDev:
```go
table.FDescribeTable("[test_id:1780]should allow regular network connection", func(ports []v1.Port, withCustomCIDR bool) {
        var ipv4NetworkCIDR string
        var ipv6NetworkCIDR string

        if withCustomCIDR {
                ipv4NetworkCIDR = "10.10.10.0/24"
                ipv6NetworkCIDR = "fd10:10:10::/120"
        }
```

maiqueb: I'm convinced.

EdDev:
> I am not entirely sure but it looks to me like, if the CIDR setting would have no effect on the VMI at all, that we would still see the tests passing. Do we need an extra check somewhere to see that CIDR had some impact on anything?

This is a good point.
I can add a a ping to the GW which is on the same network to validate it is indeed applied.

rmohr:
> This is a good point.
> I can add a a ping to the GW which is on the same network to validate it is indeed applied.

Sounds good.

tests/vmi_networking_test.go
```diff
@@ -863,3 +860,50 @@ func NewRandomVMIWithInvalidNetworkInterface() *v1.VirtualMachineInstance {
...
+		&expect.BExp{R: "\\$ "},
+		&expect.BSnd{S: fmt.Sprintf("echo test | nc %s %s -i 1 -w 1 1> /dev/null\n", serverIP, tcpPort)},
+		&expect.BExp{R: "\\$ "},
+		&expect.BSnd{S: "echo $?\n"},
+		&expect.BExp{R: expectResult},
+	}
+}
+
+// gatewayIpFromCIDR returns the first address of a network.
+func gatewayIPFromCIDR(cidr string) string {
```
AlonaKaplan, line 904: You can do something similar to `[GetHostAndGwAddressesFromCIDR`](https://github.com/kubevirt/kubevirt/blob/51ce9d8e6def33c0c260f0cdb994f08ae5b9938e/pkg/virt-launcher/virtwrap/network/common.go#L218) so it will works for both ipv4 and ipv6.

EdDev: My final version with IPv6 support for these tests was this one:
```
func gatewayIPFromCIDR(cidr string) string {
	ip, _, _ := net.ParseCIDR(cidr)
	if netutils.IsIPv6(ip) {
		ip = ip.To16()
		ip[15]++
	} else {
		ip = ip.To4()
		ip[3]++
	}
	return ip.String()
```

In order to use the one you mentioned I'll need to move it to some package that can be shared by all (production and tests). Copying it here seems an overkill, as we do not need to support all types of input.
Are you ok with the version mentioned here (will be added with IPv6 support) or you prefer to use the big one?

EdDev: Although, if I take only snippets from that method, it will look like this:
Seems nicer.

```
	ip, net, _ := net.ParseCIDR(cidr)
	ip = ip.Mask(net.Mask)
	oct := len(ip) - 1
	ip[oct]++
	return ip.String()
```

tests/vmi_networking_test.go
```diff
@@ -863,3 +860,51 @@ func NewRandomVMIWithInvalidNetworkInterface() *v1.VirtualMachineInstance {
...
+	ip, ipnet, _ := net.ParseCIDR(cidr)
+	ip = ip.Mask(ipnet.Mask)
+	oct := len(ip) - 1
+	ip[oct]++
```
AlonaKaplan, line 908: Looks nicer, thanks.
But if the last byte of the ip is 'ff' it won't work (please refer to [inc](https://github.com/kubevirt/kubevirt/blob/51ce9d8e6def33c0c260f0cdb994f08ae5b9938e/pkg/virt-launcher/virtwrap/network/common.go#L238)).

Anyway, I would pass to the test the desired ip address of the gateway explicitly. The goal of the test is to verify the flow works as we expect. By duplicating the code, I think the goal is somehow missed.

EdDev: We do not need to handle the edge cases, it is a helper for the test. If a bad input is given, the test will constantly fail.
BTW, because we increment only once, if someone will pass `fc10::ff/120` or something like that, the result will still be ending with `01` because of the mask.

The reason we need this is to fit both default and custom values and if the default change (like it already did with IPv6) I do not want to touch here. This is also minimal to worth it IMO.

AlonaKaplan: Well yes, here you increment only once, so it should be ok.

The default CIDR is something that is not expected to be changed. And if it is, it makes sense the test should be updated.
And the statement 'If a bad input is given, the test will constantly fail' is not valid. Since the test calculates the gateway, if the calculation has the same bug the code has, the test won't fail.

But since unit tests should catch wrong calculation of the gateway in the code and the logic is pretty simple. I won't block you on it.

I saw you were asked to include the ipv6 cidr test in this PR and not a separate one. So I'm waiting for it before approving the PR.

EdDev: > I saw you were asked to include the ipv6 cidr test in this PR and not a separate one. So I'm waiting for it before approving the PR.

Where I was asked to do that? And I cannot do it here because this is IPv4 only PR, the IPv6 ones will come with `https://github.com/kubevirt/kubevirt/pull/3383` or a following one that adds tests.

AlonaKaplan: You're right, my mistake. I was confused with #3383 that doesn't have the tests yet.

EdDev: I can add there tests when @maiqueb PR gets in (https://github.com/kubevirt/kubevirt/pull/3454).

AlonaKaplan: According to Miguel's comment on the PR. The PR is relevant for fedora VMs.
This test is using cirros image.

EdDev: I will change it :)

PR #3461, opened 2020-05-24
RFC: limit utils.WaitForAllPodsReady to kubevirt's own namespace

dankenigsberg:
WaitForAllPodsReady is used to validate that KubeVirt is properly
deployed before a test begins.

Recently, we noticed that if performance-addon-operator is installed on
the cluster, but is not deployed well, we cannot run our tests. This
patch ignores pods in unrelated namespaces; They are none of our
business.

This PR does have a downside: the tests make start, but then fail for
elaborate reasons, which would be harder to resolve.

Signed-off-by: Dan Kenigsberg <danken@redhat.com>

```release-note
NONE
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -478,7 +478,7 @@ func WaitForAllPodsReady(timeout time.Duration, listOptions metav1.ListOptions) 
 		virtClient, err := kubecli.GetKubevirtClient()
 		PanicOnError(err)
 
-		podsList, err := virtClient.CoreV1().Pods(k8sv1.NamespaceAll).List(listOptions)
+		podsList, err := virtClient.CoreV1().Pods(KubeVirtInstallNamespace).List(listOptions)
 		PanicOnError(err)
 		for _, pod := range podsList.Items {
 			for _, status := range pod.Status.ContainerStatuses {


dankenigsberg:
this happens to be the same as pull #2653, which is now closed; but now we have a reproducer. Where should we continue the discussion @slintes @petrkotas ?

fabiand:
@vatsalparekh @rmohr could you take a look?

tests/utils.go
```diff
@@ -478,7 +478,7 @@ func WaitForAllPodsReady(timeout time.Duration, listOptions metav1.ListOptions)
 		virtClient, err := kubecli.GetKubevirtClient()
 		PanicOnError(err)
 
-		podsList, err := virtClient.CoreV1().Pods(k8sv1.NamespaceAll).List(listOptions)
+		podsList, err := virtClient.CoreV1().Pods(KubeVirtInstallNamespace).List(listOptions)
```
nellyc, line 481: i suggest to print out any not running pods anyway, but only fail on specific pods:
1. pods under the kubevirt NS as was done in this PR
2. other NMs we care deeply about e.g etcd, mco, ca related etc (but im not sure how easy it is to do for different platforms)

maiqueb, line 481: I think this change impacts the name of the function - you're no longer "waiting for all pods to be ready", and as such should update the function name.

IMHO, if you preserve the name, you should pass the namespace as an argument to the function.

vatsalparekh: +1

vatsalparekh, line 481: To have the pods Running/Ready in all namespaces doesn't seem ideal to me, but we should have more checks to see if core functional pods are fine.
But it this is a hotfix, I would lean towards lgtm

vatsalparekh: +1

rmohr, line 481: The main issue I see here is that this is tied with a mechanism which allows deploying arbitrary manifests as additional testing infrastructure. These additional manifests are not tied to the kubevirt install namespaces.

rmohr: In this case the check would for instance not work anymore with the default testing mainfests in this repo. We would start with the test too early.

rmohr:
Can this be refined or closed? I think the best default is to only start testing on a cluster without obvious surprises. It can of course make senses to override that behaviour, but the current approach is too simplistic.

tests/utils.go
dankenigsberg: I would have to trust you here that we need to watch more than our own namespace, even though I don't understand what else we should be watching.

dankenigsberg:
I'll have to close it, as I don't have the capacity to work on this now.

PR #3462, opened 2020-05-25
Revert node-labeller

ksimon1:
Revert node-labeller from release-0.29

Since the node-labeller was only partially integrated in release-0.29, the full feature set is only available if the standalone node-labeller is deployed via its operator. This operator conflicts with the virt-operator. To avoid this conflict, we revert the node-labeller parts for release-0.29. This allows using the standalone-labeller in release-0.29.

**Release note**:
```
Remove node-labeller from kubevirt
```

/cc @rmohr, @tiraboschi


rmohr:
/lgtm
/approve
/hold

Need to enable CI for the branch.

rmohr:
/hold cancel

Tests are in place.

pkg/virt-operator/creation/rbac/handler.go
```diff
@@ -99,22 +99,6 @@ func newHandlerClusterRole() *rbacv1.ClusterRole {
...
-					"watch",
-					"get",
-				},
-			},
-			{
-				APIGroups: []string{
-					"",
-				},
-				Resources: []string{
-					"configmaps",
```
rmohr, line 112: You will have to keep all configmap logic in the  operator to allow people upgrading. Just remove the configmap.

ksimon1: Updated

cmd/virt-handler/virt-handler.go
```diff
@@ -180,19 +180,6 @@ func (app *virtHandlerApp) Run() {
 		cache.Indexers{},
 	)
 
-	nodeLabellerConfigMapSharedInformer := cache.NewSharedIndexInformer(
```
rmohr, line 183: We will have to keep this informer. Otherwise we can't react to it properly (for instance deletes).

PR #3464, opened 2020-05-26
Explicitly fail on Multus if network-attachment-definitions are missing

xpivarc:
Signed-off-by: L. Pivarc <lpivarc@redhat.com>

Explicitly fail on Multus tests if network attachment definitions are missing. This will be less confusing for new contributors which runs tests locally.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/vmi_multus_test.go
+++ b/tests/vmi_multus_test.go
@@ -133,6 +133,9 @@ var _ = Describe("Multus", func() {
 			RequestURI(fmt.Sprintf(postUrl, tests.NamespaceTestDefault, "linux-bridge-net-vlan100")).
 			Body([]byte(fmt.Sprintf(linuxBridgeConfCRD, "linux-bridge-net-vlan100", tests.NamespaceTestDefault))).
 			Do()
+		if result.Error() != nil && errors.IsNotFound(result.Error()) {
+			Expect(result.Error()).NotTo(HaveOccurred(), "Network attachment definition is missing. Install it or skip this suite.")
+		}
 		Expect(result.Error()).NotTo(HaveOccurred())
 
 		// Create ptp crds with tuning plugin enabled in two different namespaces


phoracek:
Implicit Skips may give us false positives. Maybe I want to test Mutlus, but my cluster is misconfigured. I want it to fail loudly.

What about returning an explicit error saying "Can't run suite X because of Y is missing. Please install Y or skip this suite with `flags or variables that can be set to skip this suite`".

xpivarc:
> Implicit Skips may give us false positives. Maybe I want to test Mutlus, but my cluster is misconfigured. I want it to fail loudly.
> 
> What about returning an explicit error saying "Can't run suite X because of Y is missing. Please install Y or skip this suite with `flags or variables that can be set to skip this suite`".

I see your point. I believe that everyone is looking at skipped tests. So they would see that Multus is skipped but they didn't want that.... I am fine with explicit error as it improves the experience for newcomers.

phoracek:
> > Implicit Skips may give us false positives. Maybe I want to test Mutlus, but my cluster is misconfigured. I want it to fail loudly.
> > What about returning an explicit error saying "Can't run suite X because of Y is missing. Please install Y or skip this suite with `flags or variables that can be set to skip this suite`".
> 
> I see your point. I believe that everyone is looking at skipped tests. So they would see that Multus is skipped but they didn't want that.... I am fine with explicit error as it improves the experience for newcomers.

It will be for sure annoying for people coming to KubeVirt to see it fail first. But hey, they will learn how to skip! I'm not worried about local-use, but CI. What if our CNAO lane breaks down and we silently pass with all tests skipped. Unless you tell me that `Skip` marks the lane red.

xpivarc:
> > > Implicit Skips may give us false positives. Maybe I want to test Mutlus, but my cluster is misconfigured. I want it to fail loudly.
> > > What about returning an explicit error saying "Can't run suite X because of Y is missing. Please install Y or skip this suite with `flags or variables that can be set to skip this suite`".
> > 
> > 
> > I see your point. I believe that everyone is looking at skipped tests. So they would see that Multus is skipped but they didn't want that.... I am fine with explicit error as it improves the experience for newcomers.
> 
> It will be for sure annoying for people coming to KubeVirt to see it fail first. But hey, they will learn how to skip! I'm not worried about local-use, but CI. What if our CNAO lane breaks down and we silently pass with all tests skipped. Unless you tell me that `Skip` marks the lane red.

One improvement we can make here is to check on CNAO lane if the setup is correct so these tests wouldn't be skipped. I think the test suite shouldn't be checking if the setup is correct as there are too many possible environments.

But with this error message the CNAO lane will still fail and newcomers will have more context.

rmohr:
/override pull-kubevirt-e2e-k8s-1.16

failed on the faulty operator test which should work again now.

stu-gott:
/hold

This approach isn't consistent with what the test suite has done so far. Let's come to an agreement on Fail vs Skip before merging.

xpivarc:
Closing this PR as we had consensus on a community meeting:
"Consensus: We should start to move in the direction of explicitly requesting some features -- and Fail if conditions aren’t met."

PR #3469, opened 2020-05-26
Minimize the scope of KUBEVIRT_POSTINBOUND chain

AlonaKaplan:
**What this PR does / why we need it**:\
Currently, in case the vm has a list of supported ports, the source
address of the incoming traffic to the vm is substituted by the address of
k6t-eth0 bridge (relevant for both ipv4 and ipv6).

The only reason for this behaviour I could find is for connecting to the vm
using localhost address as the destination (from an app inside the
virt-launcher pod).
In this case the source address would probably be also localhost.
The vm will try to reply to local host but of course won't succeed since the reply will stay
inside the vm and won't reach the default route.
 So in this specific case, the substitution of the source address (localhost) to the bridge address is indeed needed.

However, couldn't find any other use case for this substitution.
More over, the behaviour may be harmful since the vm thinks all its
clients have the same ip address.

The PR minimizes the rules to do the substitution only in case the source address is localhost.

Fixes #

A follow-up PR will update the [documentation](https://github.com/kubevirt/kubevirt/pull/3395/commits/1d2295fcd9804db0f4965c3900d41934259bec84#diff-971521d7f1fea6eeaa462181948819a9R304) 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/network/podinterface.go
+++ b/pkg/virt-launcher/virtwrap/network/podinterface.go
@@ -794,6 +794,7 @@ func (p *MasqueradePodInterface) createNatRulesUsingIptables(protocol iptables.P
 			strings.ToLower(port.Protocol),
 			"--dport",
 			strconv.Itoa(int(port.Port)),
+			"--source", getLoopbackAdrress(protocol),
 			"-j",
 			"SNAT",
 			"--to-source", p.getGatewayByProtocol(protocol))
@@ -896,6 +897,7 @@ func (p *MasqueradePodInterface) createNatRulesUsingNftables(proto iptables.Prot
 			strings.ToLower(port.Protocol),
 			"dport",
 			strconv.Itoa(int(port.Port)),
+			Handler.GetNFTIPString(proto), "saddr", getLoopbackAdrress(proto),
 			"counter", "snat", "to", p.getGatewayByProtocol(proto))
 		if err != nil {
 			return err
--- a/pkg/virt-launcher/virtwrap/network/podinterface_test.go
+++ b/pkg/virt-launcher/virtwrap/network/podinterface_test.go
@@ -397,7 +397,9 @@ var _ = Describe("Pod Network", func() {
 						"-p",
 						"tcp",
 						"--dport",
-						"80", "-j", "SNAT", "--to-source", GetMasqueradeGwIp(proto)).Return(nil).AnyTimes()
+						"80",
+						"--source", getLoopbackAdrress(proto),
+						"-j", "SNAT", "--to-source", GetMasqueradeGwIp(proto)).Return(nil).AnyTimes()
 					mockNetwork.EXPECT().IptablesAppendRule(proto, "nat",
 						"KUBEVIRT_PREINBOUND",
 						"-p",
@@ -445,6 +447,7 @@ var _ = Describe("Pod Network", func() {
 						"tcp",
 						"dport",
 						"80",
+						GetNFTIPString(proto), "saddr", getLoopbackAdrress(proto),
 						"counter", "snat", "to", GetMasqueradeGwIp(proto)).Return(nil).AnyTimes()
 					mockNetwork.EXPECT().NftablesAppendRule(proto, "nat",
 						"KUBEVIRT_PREINBOUND",


EdDev:
Based on http://linux-ip.net/html/routing-saddr-selection.html , for host generated traffic, it is enough to dnat the packets, the source will be determined by the exgress interface IP. So no snat is needed.

EdDev:
> Based on http://linux-ip.net/html/routing-saddr-selection.html , for host generated traffic, it is enough to dnat the packets, the source will be determined by the eggress interface IP. So no snat is needed.

I was wrong, the egress is determined before the OUTPUT chain, after which it needs to be overwritten using SNAT. Sorry for the noise.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -794,6 +794,7 @@ func (p *MasqueradePodInterface) createNatRulesUsingIptables(protocol iptables.P
 			strings.ToLower(port.Protocol),
 			"--dport",
 			strconv.Itoa(int(port.Port)),
+			"--source", getLoopbackAdrress(protocol),
```
EdDev, line 797: I think it is enough to specify here the keyword `localhost`.

pkg/virt-launcher/virtwrap/network/podinterface.go
```diff
@@ -896,6 +897,7 @@ func (p *MasqueradePodInterface) createNatRulesUsingNftables(proto iptables.Prot
 			strings.ToLower(port.Protocol),
 			"dport",
 			strconv.Itoa(int(port.Port)),
+			Handler.GetNFTIPString(proto), "saddr", getLoopbackAdrress(proto),
```
EdDev, line 900: I did not try nftables, but would it also support the keyword `localhost`?

pkg/virt-launcher/virtwrap/network/podinterface.go
AlonaKaplan: Thanks, I will try in on a different PR.

pkg/virt-launcher/virtwrap/network/podinterface.go
AlonaKaplan: same

EdDev:
https://travis-ci.org/github/kubevirt/kubevirt/builds/691442496#L2158

rmohr:
> The only reason for this behaviour I could find is for connecting to the vm
> using localhost address as the destination (from an app inside the
> virt-launcher pod).
> In this case the source address would probably be also localhost.
> The vm will try to reply to local host but of course won't succeed since the reply will stay
> inside the vm and won't reach the default route.
> So in this specific case, the substitution of the source address (localhost) to the bridge address is indeed needed.

This may be related to istio. @phoracek @SchSeba do you remember the details here?

AlonaKaplan:
> > The only reason for this behaviour I could find is for connecting to the vm
> > using localhost address as the destination (from an app inside the
> > virt-launcher pod).
> > In this case the source address would probably be also localhost.
> > The vm will try to reply to local host but of course won't succeed since the reply will stay
> > inside the vm and won't reach the default route.
> > So in this specific case, the substitution of the source address (localhost) to the bridge address is indeed needed.
> 
> This may be related to istio. @phoracek @SchSeba do you remember the details here?

Yes, indeed for istio (or any other side car/virt-launcher internal proccess communicating with the vm) this rule may be relevant.
The 'minimized rule' is enough for those cases.

Couldn't find any reason for having the rule for traffic with source address other than the localhost.

phoracek:
> > > The only reason for this behaviour I could find is for connecting to the vm
> > > using localhost address as the destination (from an app inside the
> > > virt-launcher pod).
> > > In this case the source address would probably be also localhost.
> > > The vm will try to reply to local host but of course won't succeed since the reply will stay
> > > inside the vm and won't reach the default route.
> > > So in this specific case, the substitution of the source address (localhost) to the bridge address is indeed needed.
> > 
> > 
> > This may be related to istio. @phoracek @SchSeba do you remember the details here?

Only that istio has some expectations. I don't know what those are and we don't have tests for istio, so it would be only blind assumption anyway.

I'm ok with this change. Once we start integrating (testing) istio, we can revisit this.

PR #3478, opened 2020-05-27
Set proper architecture label in generated container images

### LINKED ISSUE #3065: Wrong architecture label on container image built for Power (ppc64le)
/kind bug

**What happened**:
The container images built in Power are showing up with the wrong architecture (amd64) instead of ppc64le in dockerhub.

For example some of the images in: https://hub.docker.com/u/murilofv

They work properly on a Power platform without any issues despite the wrong label, but it is confusing for a user and would cause issues with maintaining images for both amd64 and ppc64le in the same repo.

Once we have this issue sorted out we can work on creating manifests list that could be used as a pointer for both architectures on a release.

Container images should be labeled correctly for the architecture they are built in.

`make bazel-build-images`
`make bazel-push-images`

This issue is a follow up item on the enablement of the ppc64le architecture in kubevirt done in https://github.com/kubevirt/kubevirt/pull/2944

murilofv:
With the 0.14.2 release of rules_docker a new attribute called
"architecture" has been exposed in the container_image rule. It allows
setting a custom architecture label, previously the images were always
labeled as amd64. So moving to this release and  adding a select in
order to properly label the architecture of the container images pushed
by Kubevirt.

Signed-off-by: Murilo Fossa Vicentini <muvic@linux.ibm.com>

**What this PR does / why we need it**: This PR allows the container images built by Kubevirt to be properly labeled for the architecture they are built in

Fixes #3065

A nice follow up enhancement would be to add support for a manifest list that could be used as a pointer in a release for multi architecture support.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -48,11 +48,10 @@ http_archive(
 
 http_archive(
     name = "io_bazel_rules_docker",
-    sha256 = "dc97fccceacd4c6be14e800b2a00693d5e8d07f69ee187babfd04a80a9f8e250",
-    strip_prefix = "rules_docker-0.14.1",
+    sha256 = "3efbd23e195727a67f87b2a04fb4388cc7a11a0c0c2cf33eec225fb8ffbb27ea",
+    strip_prefix = "rules_docker-0.14.2",
     urls = [
-        "https://github.com/bazelbuild/rules_docker/releases/download/v0.14.1/rules_docker-v0.14.1.tar.gz",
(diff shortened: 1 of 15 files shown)


rmohr:
@murilofv nice enhancement. Thanks!

murilofv:
I am not sure what is going on with the Travis CI, it is marking as pending but when I go in the details it has already finished successfully two days ago, is there a way to manually trigger it again?

PR #3480, opened 2020-05-28
Fix for checking if an environment has SRIOV enabled before executing the SRIOV functional tests. 

shwetaap:
SRIOV functional tests are run in all environments today, even those that do not support SRIOV, this fix checks if the environment has SRIOV enabled and only then allows the tests to be run otherwise they are skipped. 
Fixes #

**Release note**:
```NONE

```

### DIFF
--- a/tests/vmi_multus_test.go
+++ b/tests/vmi_multus_test.go
@@ -580,6 +580,12 @@ var _ = Describe("SRIOV", func() {
 
 	tests.BeforeAll(func() {
 		tests.BeforeTestCleanup()
+		// Check if the hardware supports SRIOV
+		sriovcheck := checkSriovEnabled(virtClient, sriovResourceName)
+		if !sriovcheck {
+			Skip("Sriov is not enabled in this environment. Skip these tests using - export FUNC_TEST_ARGS='--ginkgo.skip=SRIOV'")
+		}
+
 		// Create two sriov networks referring to the same resource name
 		result := virtClient.RestClient().
 			Post().
@@ -1047,4 +1053,21 @@ func configureNodeNetwork(virtClient kubecli.KubevirtClient) {
 		daemonSet := getNetworkConfigDaemonSet()
 		return int(daemonSet.Status.NumberAvailable)
 	}, time.Minute, time.Second).Should(Equal(len(nodes.Items)))
+}
+
+func checkSriovEnabled(virtClient kubecli.KubevirtClient, sriovResourceName string) bool {
+	nodes := tests.GetAllSchedulableNodes(virtClient)
+	Expect(nodes.Items).ToNot(BeEmpty(), "There should be some compute node")
+
+	for _, node := range nodes.Items {
+		resourceList := node.Status.Allocatable
+		for k, v := range resourceList {
+			if string(k) == sriovResourceName {
+				if v.Value() > 0 {
+					return true
+				}
+			}
+		}
+	}
+	return false
 }


tests/vmi_multus_test.go
```diff
@@ -580,6 +580,37 @@ var _ = Describe("SRIOV", func() {
 
 	tests.BeforeAll(func() {
 		tests.BeforeTestCleanup()
+		// Check if the hardware supports SRIOV
+		nodes, err := virtClient.CoreV1().Nodes().List(metav1.ListOptions{})
```
ksimon1, line 584: I would rather use nodes := tests.GetAllSchedulableNodes(virtClient)  (https://github.com/kubevirt/kubevirt/blob/master/tests/vmi_lifecycle_test.go#L865)

stu-gott: +1

This is a very important distinction. If a node is not schedulable, it should not be considered a candidate for running tests on. It sounds somewhat unlikely in the case of SRIOV, but this pattern has come up before. e.g. CPU count.

shwetaap: Made the change as suggested

tests/vmi_multus_test.go
```diff
@@ -580,6 +580,36 @@ var _ = Describe("SRIOV", func() {
...
+			for k, v := range capacityList {
+				if strings.HasPrefix(string(k), sriovResourceName) {
+					if v.Value() > 0 {
+						sriovcapacity = true
+						break
+					}
+				}
+			}
+			if (!sriovallocate || !sriovcapacity){
+				Skip("SRIOV is not enabled, test skipped")
```
phoracek, line 609: I would vote to use an early `Fail` instead. SR-IOV is enabled in tests, it is just not available on the cluster. As a user of these tests I would like to be clearly notified (the skip can be easily lost in the output). On the other hand, I understand that it is frustrating to get fail from the test-suite if you don't have specific hardware.

Would you mind using `Fail` instead and explain in the message why it failed and how can users skip the suite?

This PR would be a suggestion to skip all tests without sufficient cluster environment. I think it is dangerous. Failing early and with clear explanation would be a better service.

tests/vmi_multus_test.go
```diff
@@ -580,6 +580,36 @@ var _ = Describe("SRIOV", func() {
...
+		// Check if the hardware supports SRIOV
+		nodes := tests.GetAllSchedulableNodes(virtClient)
+		Expect(nodes.Items).ToNot(BeEmpty(), "There should be some compute node")
+
+		for _, node := range nodes.Items {
+			var sriovallocate bool = false
+			var sriovcapacity bool = false
+			resourceList := node.Status.Allocatable
+			for k, v := range resourceList {
+				if strings.HasPrefix(string(k), sriovResourceName) {
```
phoracek, line 592: I think we are looking for perfect match. If the `sriovResourceName` is set for `example.com/net`, but the cluster is configured only with `example.com/net1`, this check would pass but test would fail anyway.

tests/vmi_multus_test.go
```diff
@@ -580,6 +580,36 @@ var _ = Describe("SRIOV", func() {
 
 	tests.BeforeAll(func() {
 		tests.BeforeTestCleanup()
+		// Check if the hardware supports SRIOV
```
phoracek, line 583: This should be in a helper function to make it more readable. It is a lot of code occupying BeforeAll.

tests/vmi_multus_test.go
```diff
@@ -580,6 +580,36 @@ var _ = Describe("SRIOV", func() {
...
+			for k, v := range resourceList {
+				if strings.HasPrefix(string(k), sriovResourceName) {
+					if v.Value() > 0 {
+						sriovallocate = true
+						break
+					}
+				}
+			}
+			capacityList := node.Status.Capacity
+			for k, v := range capacityList {
```
phoracek, line 600: If you store the whole check in a function, you can perform an early return if you don't find even the resource. Then you can avoid this second iteration.

vladikr:
perhaps it's better to squash these commits, at least those with the same name. also please provide a meaningful name to the commit. It will just look weird in history.

tests/vmi_multus_test.go
vladikr: +1

tests/vmi_multus_test.go
```diff
@@ -1048,3 +1054,38 @@ func configureNodeNetwork(virtClient kubecli.KubevirtClient) {
...
+					break
+				}
+			}
+		}
+		if !sriovcapacity{
+			return false
+		}
+	}
+	return true
+	}
```
phoracek, line 1091: Seems that his file was not formated with goimports/gofmt

tests/vmi_multus_test.go
```diff
@@ -580,6 +580,12 @@ var _ = Describe("SRIOV", func() {
 
 	tests.BeforeAll(func() {
 		tests.BeforeTestCleanup()
+		// Check if the hardware supports SRIOV
+		var sriovcheck = checkSriovEnabled(virtClient, sriovResourceName)
```
phoracek, line 584: We usually use the shorthand:

```go
sriovcheck := checkSriovEnabled(virtClient, sriovResourceName)
```

tests/vmi_multus_test.go
```diff
@@ -580,6 +580,12 @@ var _ = Describe("SRIOV", func() {
...
+		if !sriovcheck {
+			Fail("Sriov is not enabled in this environment. Skip these tests using - export FUNC_TEST_ARGS='--ginkgo.skip=SRIOV'")
```
phoracek, line 586: Thanks!

tests/vmi_multus_test.go
```diff
@@ -1048,3 +1054,38 @@ func configureNodeNetwork(virtClient kubecli.KubevirtClient) {
 		return int(daemonSet.Status.NumberAvailable)
 	}, time.Minute, time.Second).Should(Equal(len(nodes.Items)))
 }
+
+func checkSriovEnabled(virtClient kubecli.KubevirtClient, sriovResourceName string) bool {
+	nodes := tests.GetAllSchedulableNodes(virtClient)
+	Expect(nodes.Items).ToNot(BeEmpty(), "There should be some compute node")
+
+	for _, node := range nodes.Items {
+		var sriovallocate bool = false
```
phoracek, line 1063: ditto `:=`

tests/vmi_multus_test.go
```diff
@@ -1048,3 +1054,38 @@ func configureNodeNetwork(virtClient kubecli.KubevirtClient) {
...
+		var sriovcapacity bool = false
```
phoracek, line 1064: ditto

tests/vmi_multus_test.go
```diff
@@ -1048,3 +1054,38 @@ func configureNodeNetwork(virtClient kubecli.KubevirtClient) {
...
+				if v.Value() > 0 {
+					sriovallocate = true
+					break
+				}
+			}
+		}
+		if !sriovallocate{
+			return false
+		}
+		capacityList := node.Status.Capacity
```
phoracek, line 1077: I think it is enough to check only for allocate. It should never be lower than allocatable

tests/vmi_multus_test.go
```diff
@@ -1048,3 +1054,25 @@ func configureNodeNetwork(virtClient kubecli.KubevirtClient) {
 		return int(daemonSet.Status.NumberAvailable)
 	}, time.Minute, time.Second).Should(Equal(len(nodes.Items)))
 }
+
+func checkSriovEnabled(virtClient kubecli.KubevirtClient, sriovResourceName string) bool {
+	nodes := tests.GetAllSchedulableNodes(virtClient)
+	Expect(nodes.Items).ToNot(BeEmpty(), "There should be some compute node")
+
+	for _, node := range nodes.Items {
+		sriovallocate := false
```
phoracek, line 1063: We can drop this variable

tests/vmi_multus_test.go
```diff
@@ -1048,3 +1054,25 @@ func configureNodeNetwork(virtClient kubecli.KubevirtClient) {
...
+		resourceList := node.Status.Allocatable
+		for k, v := range resourceList {
+			if string(k) == sriovResourceName {
+				if v.Value() > 0 {
+					sriovallocate = true
```
phoracek, line 1068: return true here

tests/vmi_multus_test.go
```diff
@@ -1048,3 +1054,25 @@ func configureNodeNetwork(virtClient kubecli.KubevirtClient) {
...
+					break
+				}
+			}
+		}
+		if !sriovallocate {
+			return false
+		}
+	}
+	return true
```
phoracek, line 1077: return false here

shwetaap:
> /lgtm
> 
> Thanks! This will be indeed quite helpful. Even with this check, SR-IOV tests were executed https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3480/pull-kubevirt-e2e-kind-k8s-sriov-1.17.0/1268638843146342400/build-log.txt?Expires=1591340789&GoogleAccessId=prow-46%40kubevirt-240711.iam.gserviceaccount.com&Signature=dYzxpwsJXsxXZnsF%2BUboaSzI%2Bw4dtYRgJt4QiqV0HKZcObHkVfudM5IBUoCFZ%2FekGGIWevlfzNaeebMHKFiGHFlZvpgNZnABKzWPMRhe1uRQPNOjWqQtuOrAdU2A%2BYrnAoSdm0ZBn4l8cur3W0OVhpfgsL%2Fc4R84nw8%2FEYtAk7nVgYtqQ8tZEjJ5sA%2FfPzHNYpRJV6lOW2qwvH0kkZ9VxCiJorScHyVFvV4ZjshmeyJYszdY5ihY3dmhW37tBetcwLal7C%2B4galeyRwMhvoyXVF2Cwx8%2Fmw5JgtVT5bDJ4BfIEwBpwhyk5U4twYGNf%2BHOZyC7fKQ0i%2BypgGtU9TdFg%3D%3D.
> 
> Have you verified that the check is skipping locally?

Yes.
This is the output

• Failure in Spec Setup (BeforeEach) [4.387 seconds]
SRIOV
/root/go/src/kubevirt.io/kubevirt/tests/vmi_multus_test.go:568
  VirtualMachineInstance with sriov plugin interface [BeforeEach]
  /root/go/src/kubevirt.io/kubevirt/tests/vmi_multus_test.go:623
    [test_id:3957]should connect to another machine with sriov interface over IPv6
    /root/go/src/kubevirt.io/kubevirt/tests/vmi_multus_test.go:874

    Sriov is not enabled in this environment. Skip these tests using - export FUNC_TEST_ARGS='--ginkgo.skip=SRIOV'

    /root/go/src/kubevirt.io/kubevirt/tests/vmi_multus_test.go:586
failed to dump entity named [sriovnetworknodestates]: the server could not find the requested resource
failed to dump entity named [sriovnetworknodepolicies]: the server could not find the requested resource
failed to dump entity named [sriovnetworks]: the server could not find the requested resource
failed to dump entity named [sriovoperatorconfigs]: the server could not find the requested resource

stu-gott:
/hold

I'm not sure this Fail is a good pattern. At least we're skipping directly elsewhere in the tests. I see no reason why SR-IOV would be different than e.g. not having enough CPUs. Let's come to an agreement on the best approach in general.

phoracek:
Similar PR solving similar issue: https://github.com/kubevirt/kubevirt/pull/3464

I am in favor of failing since it does not leave place for false positives. But you have a good point that if we extend this logic everywhere, it may become hard for users to skip everything they don't support.

stu-gott:
Hrm. yeah, having to add explicit skip flags as new features get added can escalate quickly.

If an environment isn't valid for a test to be applied, that's literally the definition of why Skip exists. The test didn't fail. The test was never run.

I think the best approach is to just be explicit in why a test was skipped in the message.

EdDev:
> If an environment isn't valid for a test to be applied, that's literally the definition of why Skip exists.

IMO this is a miss-use of skip.
It may skip because one has a bug to detect the hw or because the detection is invalid after some hw/driver/platform changes.
The runner should just not run the tests.
Tests for sriov should not be called at all on hw that does not support it.
I see skipping as an edge case for cases where there is a bug and the test not running is indeed a problem (xfail) or when you have a few tests that have a problem running but it is impractical to not run the at all.

stu-gott:
@EdDev we don't have "a few" tests that fall into this category. We have a significant amount of them. There are more than 100 explicit Skip calls in the functional tests--each of which is due to there being some env condition that makes the test invalid in that setup. At this point it's a very well established pattern.

I fundamentally disagree that tests requiring specific conditions (in this case, rare specialized hardware) should cause the entire test suite to fail. Taking this approach means *every* instance of the test suite *including* developer setups will automatically fail unless they remember to add the ill-documented FUNC_TEST_ARGS env var. If we enforced a fail for all 112 Skips in the test suite, we'd end up with a bewildering maze of frustration.

I get your point about the pedantic use of skip/fail, but the end result here is likely to be opposite of what you desire. It'll just become tribal knowledge that "you need to add a skip for SRIOV, masquerade, IPv6, migration, data volumes, CPU manager, etc"--and everybody will do that automatically. We need to be pragmatic here. default-to-fail is going to do more harm than good in the vast majority of cases.

vladikr:
> Similar PR solving similar issue: #3464
This is not similar to me, as you've requested a resource but it doesn't have network-attachment-definitions -> so tests should fail.
Here is a different situation. These tests shouldn't run at all in an environment that doesn't have sriov.
> 
> I am in favor of failing since it does not leave place for false positives. But you have a good point that if we extend this logic everywhere, it may become hard for users to skip everything they don't support.
What false positive do you envision with this approach?
I think it's pretty straight forward. If the environment is lacking resources to run certain tests, these tests should be skipped.

However, the real question to me is how do we make this model scalable. Right now were are testing SRIOV, but in the future, there will be multiple resources to test.
I would imagine that each environment could express some ow which features must be tested.

However, right now skipping these tests on an environment that doesn't have proper resources looks like the correct approach.

EdDev:
I guess I am indeed questioning the existing pattern of skipping every time a test does not fit the env requirements. This approach adds logic in the tests themselves and require them to "know" about the env before running. If you have several such conditions mixing into the tests you start to have a mess of skip conditions and with time a real problem to track how correct it is (e.g. do we skip the tests for good reason or is this a logic mistake).

Running only needed tests can be taken out of the tests and still be automated. Usually it is an external checked that validates the env and decides what to run. It can report it before or after the tests to make it clear what it detected so the user can check if it makes sense.
In some sense, this is like the difference between compilation and runtime where in the compilation you can decide what to include and what not.

I understand this may be an overkill topic for this PR, but continuing adding such skips all over makes it harder to maintain the existing tests. I do hope you all will consider this for the long run though.

stu-gott:
> I guess I am indeed questioning the existing pattern of skipping every time a test does not fit the env requirements. This approach adds logic in the tests themselves and require them to "know" about the env before running. If you have several such conditions mixing into the tests you start to have a mess of skip conditions and with time a real problem to track how correct it is (e.g. do we skip the tests for good reason or is this a logic mistake).
 
Re-architecting the overall approach would work, but the approach in this PR takes us down an even worse path. This PR as written enforces mandatory test failures in almost every environment. If the test must fail if invalid then the test **MUST** be opt in--not opt out.

> Running only needed tests can be taken out of the tests and still be automated. Usually it is an external checker that validates the env and decides what to run. It can report it before or after the tests to make it clear what it detected so the user can check if it makes sense.

You're not wrong, but we don't have that infrastructure in place. Skip() is the most reasonable compromise in the meantime.

Personally I just don't see real impactful value-add in changing our approach. What we have works as designed, even if people don't like it. There's a considerable time and effort in re-architecting this and what we get in return is something that works...

> I understand this may be an overkill topic for this PR, but continuing adding such skips all over makes it harder to maintain the existing tests. I do hope you all will consider this for the long run though.

I disagree with the notion that Skip() makes the tests harder to maintain. It adds to technical debt--assuming we all agree reversing this is even necessary or sensible, but it does not make the tests harder to maintain.

This conversation is indeed well beyond the scope or purpose of this PR. If the community agrees we should eradicate skips, then that's something we can work towards. But we should prioritize. Our test suite isn't really something I'd call stable at the moment--and that's a far greater concern.

We should not block this PR while the subject of skip vs fail tests is sorted out.

phoracek:
> What false positive do you envision with this approach?

The suite expects that SR-IOV network is available under the name specified as `SRIOV_RESOURCE_NAME` https://github.com/kubevirt/kubevirt/blob/master/tests/vmi_multus_test.go#L575. If somebody sets this value and makes a typo (or we change the name of the resource available in cluster or the namespace prefix `openshift.io` gets changed), our tests would be skipped. If `SRIOV_RESOURCE_NAME` is not available in the cluster, it should loudly fail. That is what happens now. This PR just makes sure we fail earlier and with a clearer message.

tests/vmi_multus_test.go
```diff
@@ -580,6 +580,12 @@ var _ = Describe("SRIOV", func() {
 
 	tests.BeforeAll(func() {
 		tests.BeforeTestCleanup()
+		// Check if the hardware supports SRIOV
+		sriovcheck := checkSriovEnabled(virtClient, sriovResourceName)
+		if !sriovcheck {
+			Fail("Sriov is not enabled in this environment. Skip these tests using - export FUNC_TEST_ARGS='--ginkgo.skip=SRIOV'")
```
stu-gott, line 586: Please change this Fail() to a Skip().

Because checkSriovEnabled can't check if SR-IOV is *supposed* to be enabled, this ends up being way too aggressive. The only way a Fail() is acceptable here is if the user explicitly asks for SR-IOV tests to be executed. We don't currently have a framework for such a thing.

tests/vmi_multus_test.go
shwetaap: Done

tests/vmi_multus_test.go
shwetaap: Done

tests/vmi_multus_test.go
shwetaap: Done

stu-gott:
/override pull-kubevirt-e2e-k8s-1.15-ceph

Failing tests are known to be flakey

PR #3490, opened 2020-05-29
Revert "travis-ci: Fail on first error in script"

rmohr:
Reverts kubevirt/kubevirt#3413


EdDev:
I would prefer to extract to an external file: #3491

PR #3497, opened 2020-06-01
Collect infra logs from infra containers to debug CI failures

danielBelenky:

```release-note
NONE
```

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -170,6 +170,16 @@ fi
 
 kubectl() { cluster-up/kubectl.sh "$@"; }
 
+collect_debug_logs() {
+    local containers
+
+    containers=( $(docker ps -a --format '{{ .Names }}') )
+    for container in "${containers[@]}"; do
+        echo "======== $container ========"
+        docker logs "$container"
+    done
+}
+
 export NAMESPACE="${NAMESPACE:-kubevirt}"
 
 # Make sure that the VM is properly shut down on exit
@@ -200,7 +210,7 @@ done
 set -e
 make bazel-build-images
 
-trap '{ echo "Dump kubevirt state:"; make dump; }' ERR
+trap '{ collect_debug_logs; echo "Dump kubevirt state:"; make dump; }' ERR
 make cluster-up
 trap - ERR
 


automation/test.sh
```diff
@@ -201,7 +211,7 @@ set -e
 make bazel-build-images
 
 trap '{ echo "Dump kubevirt state:"; make dump; }' ERR
-make cluster-up
+make cluster-up || collect_debug_logs
```
EdDev, line 214: Why this has been moved here from the `trap`?

oshoval: also take into account that even with trap, here there is trap - ERR on line 215
best to trap globally i think, in any case of error ?

oshoval: line 214 suppress line 213 trap

automation/test.sh
```diff
@@ -201,7 +211,7 @@ set -e
 make bazel-build-images
 
 trap '{ echo "Dump kubevirt state:"; make dump; }' ERR
```
EdDev, line 213:
```suggestion
trap '{ collect_debug_logs; echo "Dump kubevirt state:"; make dump; }' ERR
```

automation/test.sh
EdDev: Or is right, on failure of `make cluster-up`, the trap will not run `make dump` (because of the OR chain).
See suggestion above.

automation/test.sh
```diff
@@ -170,6 +170,16 @@ fi
 
 kubectl() { cluster-up/kubectl.sh "$@"; }
 
+collect_debug_logs() {
+    local containers
+
+    containers=( $(docker ps --format '{{ .Names }}') )
```
oshoval, line 176: i believe you need -a
since the wanted container is probably dead
i will try to simulate it locally as well, injecting an error

automation/test.sh
danielBelenky: Good catch. I've updated the PR.

automation/test.sh
danielBelenky: Added the `-a` flag.

automation/test.sh
```diff
@@ -200,7 +210,7 @@ done
 set -e
 make bazel-build-images
 
-trap '{ echo "Dump kubevirt state:"; make dump; }' ERR
+trap '{ collect_debug_logs; echo "Dump kubevirt state:"; make dump; }' ERR
```
oshoval, line 213: would you add exit 1 in the end please?
because as we groomed, it suppress the -e
we can do it on other PR if needed, up to you
thanks

EdDev: I could not recreate this, the exit trap is executed after the error trap.
```
#!/bin/bash -e                 
                           
echo "Starting"                
                               
trap '{ echo Exit-trap; }' EXIT
                               
trap '{ echo Err-trap; }' ERR  
false                          
trap - ERR                     
                               
echo MormExit                  
```

oshoval: ah didnt see there is another trap EXIT before
thanks

EdDev:
Error on the 1.17 lane:
```
 Error response from daemon: cannot join network of a non running container: e1b4e148fa110d46fff93d15bbf36597faa54d530997be62666216f69391431e
Makefile:96: recipe for target 'cluster-up' failed
++ collect_debug_logs
++ local containers
++ containers=($(docker ps -a --format '{{ .Names }}'))
+++ docker ps -a --format '{{ .Names }}'
++ for container in "${containers[@]}"
++ echo '======== pull-kubevirt-e2e-k8s-1.17-registry ========'
======== pull-kubevirt-e2e-k8s-1.17-registry ========
++ docker logs pull-kubevirt-e2e-k8s-1.17-registry
++ for container in "${containers[@]}"
++ echo '======== pull-kubevirt-e2e-k8s-1.17-dnsmasq ========'
======== pull-kubevirt-e2e-k8s-1.17-dnsmasq ========
++ docker logs pull-kubevirt-e2e-k8s-1.17-dnsmasq
dnsmasq: failed to create inotify for /etc/resolv.conf: No space left on device 
```

PR #3506, opened 2020-06-03
Fix vmirs watcher, Controller can filter abnormal vmis related abnorm…

### LINKED ISSUE #3332: Node failover function of VM/VMIRS
 /kind enhancement

**What happened**: When the worker node status is not ready(kubelet down or Host down), the VM keep staying in Terminating state

**What you expected to happen**:  The VM terminate successfully and re-create on another node. (if i want to recreate on another node, some specific flag is help to recreate vm automatically)

 - Create VM/VMIRS on the worker node
 - Stop the kubelet (systemctl stop kubelet) or Shutdown the worker node
 - Checking your pods ( kubectl get pods ) -> pods is staying in terminating state

- In k8s, replicaset guarantees failover in the above situation.
- I think, VM is one of replicaset for vmi 
- Even if it's not, vmirs seems to have to be guaranteed 

**Environment**:
- KubeVirt version (use `virtctl version`):  v0.27.0
- Kubernetes version (use `kubectl version`):  v1.15.3
- VM or VMI specifications: vm & vmirs
- OS (e.g. from /etc/os-release): Ubuntu 18.04.4 LTS
- Kernel (e.g. `uname -a`): 4.15.0-88-generic #88-Ubuntu SMP Tue Feb 11 20:11:34 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
- Install tools:
- Others:

haemyung:
Signed-off-by: haemyung_yang <haemyung_yang@tmax.co.kr>

When node was failed and pod goes to deleting phase, vmi was not recreated by controller. Because controller not detected pod phase for related vmi,vmirs. So
I changed the filterActiveVmi method

Fixes #3332 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-controller/watch/replicaset.go
+++ b/pkg/virt-controller/watch/replicaset.go
@@ -351,7 +351,7 @@ func (c *VMIReplicaSet) scale(rs *virtv1.VirtualMachineInstanceReplicaSet, vmis 
 // filterActiveVMIs takes a list of VMIs and returns all VMIs which are not in a final state and not terminating
 func (c *VMIReplicaSet) filterActiveVMIs(vmis []*virtv1.VirtualMachineInstance) []*virtv1.VirtualMachineInstance {
 	return filter(vmis, func(vmi *virtv1.VirtualMachineInstance) bool {
-		return !vmi.IsFinal() && vmi.DeletionTimestamp == nil
+		return !vmi.IsFinal() && vmi.DeletionTimestamp == nil && !controller.NewVirtualMachineInstanceConditionManager().HasConditionWithStatus(vmi, virtv1.VirtualMachineInstanceConditionType(k8score.PodReady), k8score.ConditionFalse)
 	})
 }
 


pkg/virt-controller/watch/replicaset.go
```diff
@@ -351,7 +351,7 @@ func (c *VMIReplicaSet) scale(rs *virtv1.VirtualMachineInstanceReplicaSet, vmis
 // filterActiveVMIs takes a list of VMIs and returns all VMIs which are not in a final state and not terminating
 func (c *VMIReplicaSet) filterActiveVMIs(vmis []*virtv1.VirtualMachineInstance) []*virtv1.VirtualMachineInstance {
 	return filter(vmis, func(vmi *virtv1.VirtualMachineInstance) bool {
-		return !vmi.IsFinal() && vmi.DeletionTimestamp == nil
+		return !vmi.IsFinal() && vmi.DeletionTimestamp == nil && !controller.NewVirtualMachineInstanceConditionManager().HasConditionWithStatus(vmi, virtv1.VirtualMachineInstanceConditionType(k8score.PodReady), k8score.ConditionFalse)
```
rmohr, line 354: I think what is happening in the k8s case is that the NoExecutionTaintManger is deleting all pods on the unresponsive node. The pods are then still marked as ready (at least if there is no explicit readiness probe specified) but are in the terminating phase. I think that this is the edge case which we are not detecting  right now.

haemyung: Sorry for the late response to the answer. What you said is that the podReady part of vmi is the part that remains ready even if the pod is terminating. I think the podReady part needs to be changed. So I think it needs to be corrected for the part marked as ready. Does your idea match? If not, I think filterActiveVMIs should see the actual pod status and judge it.

haemyung: Which one do you think is the right direction for you?

rmohr:
/close 

fixed in #4398, @haemyung thanks for bringing this up.

PR #3511, opened 2020-06-04
revert cpu model / feature tests

ksimon1:
revert cpu model / feature tests

With labeller PR these tests were changed, but when labeller
was reverted, these tests were not. This PR return old tests back.

**Release note**:
```
NONE
```


rmohr:
Merging with some CI fixes.

PR #3513, opened 2020-06-04
fix test default smbios values

ashleyschuett:

This test does not currently test for default values, but rather the values that are set for the VM. This change clears the set values so that the default values are used instead. 

```release-note
NONE
```

### DIFF
--- a/tests/vmi_configuration_test.go
+++ b/tests/vmi_configuration_test.go
@@ -2324,6 +2324,11 @@ var _ = Describe("Configurations", func() {
 			Expect(err).ToNot(HaveOccurred())
 			tests.WaitForSuccessfulVMIStart(vmi)
 
+			test_smbios := &cmdv1.SMBios{Family: "", Product: "", Manufacturer: ""}
+			smbiosJson, err := json.Marshal(test_smbios)
+			Expect(err).ToNot(HaveOccurred())
+			tests.UpdateClusterConfigValueAndWait(virtconfig.SmbiosConfigKey, string(smbiosJson))
+
 			By("Check values in domain XML")
 			domXml, err := tests.GetRunningVirtualMachineInstanceDomainXML(virtClient, vmi)
 			Expect(err).ToNot(HaveOccurred())


tests/vmi_configuration_test.go
```diff
@@ -2306,6 +2306,12 @@ var _ = Describe("Configurations", func() {
 			Expect(err).ToNot(HaveOccurred())
 			tests.WaitForSuccessfulVMIStart(vmi)
 
+			test_smbios := &cmdv1.SMBios{Family: "", Product: "", Manufacturer: ""}
```
rmohr, line 2309: Do you want to do this as a pre-caution? I did not find the place in the code where they would have been set.

tests/vmi_configuration_test.go
```diff
@@ -2306,6 +2306,12 @@ var _ = Describe("Configurations", func() {
...
+			smbiosJson, err := json.Marshal(test_smbios)
+			Expect(err).ToNot(HaveOccurred())
+			tests.UpdateClusterConfigValueAndWait(virtconfig.SmbiosConfigKey, string(smbiosJson))
+			defer tests.ClearKubeVirtConfigMap(virtconfig.SmbiosConfigKey)
```
stu-gott, line 2313: We can't blindly delete the SMBios settings. before we set them to something, we need to retrieve and store the original values. Once the test has concluded, we should restore that. It might make sense to do that in an AfterEach block.

It's possible that there wasn't anything set for SMBios, in which case Clearing the value is the correct response. However on clusters that did have this set before the test suite was run, this test would incur a side effect of deleting it.

tests/vmi_configuration_test.go
stu-gott: These values are already present on some clusters. (e.g. those deployed via HCO)

tests/vmi_configuration_test.go
ashleyschuett: hmmm, is that needed for this test as well then? https://github.com/kubevirt/kubevirt/pull/3513/files#diff-d47bf036797965d742615e64adc3b777L2337-L2341

ashleyschuett: I have updated this to use an `AfterEach` block to restore the ConfigMap SMBios settings to the originals at the beginning of the test

tests/vmi_configuration_test.go
```diff
@@ -2294,18 +2294,34 @@ var _ = Describe("Configurations", func() {
...
+		if val, ok := cfgMap.Data[virtconfig.SmbiosConfigKey]; ok {
+			originalSmbios = val
+		}
 
 		BeforeEach(func() {
 			vmi = tests.NewRandomFedoraVMIWithDmidecode()
 		})
 
+		AfterEach(func() {
+			tests.UpdateClusterConfigValueAndWait(virtconfig.SmbiosConfigKey, originalSmbios)
```
stu-gott, line 2310: Getting closer!

I think we should move this save/restore to the top-level describe for the whole module. a lot of tests are currently violating this restoration principle--we can fix all of them at once.

tests/operator_test.go has a really elegant trick worth incorporating here. That code tests like this:

`curKubeVirtConfig.ResourceVersion != originalKubeVirtConfig.ResourceVersion`

and if the configs don't match, it simply restores the entire configMap data. Any change anywhere in the configMap will be detected and corrected by this. Check the AfterEach on or around line 556.

Doing the above sidesteps another subtle issue: what about clusters where the smbios code wasn't present? Setting `SmbiosConfigKey` like this line does means that if the key wasn't set prior to the test, it will transition from not present at all to the empty string when this test is run. resetting the entire `data` stanza doesn't have that problem.

ashleyschuett:
I think this PR is meant to solve this same issue in a broader way https://github.com/kubevirt/kubevirt/pull/3542 and we can close this one?

tests/vmi_configuration_test.go
vatsalparekh: @stu-gott @ashleyschuett we need similar reset of value for machinetype as well

vatsalparekh: @stu-gott @ashleyschuett we need similar reset of value for machinetype as well

ashleyschuett: In this test or?

vatsalparekh: https://github.com/kubevirt/kubevirt/blob/2b35c0cf0f7657b9ad8b1c7a50ed642ca31bfc20/tests/vmi_configuration_test.go#L1538

Here also we need a similar mechanism to restore values to previous state

rmohr:
> @stu-gott @ashleyschuett we need similar reset of value for machinetype as well

Maybe we should resolve that with #3542 which will always restore the original config after each test.

ashleyschuett:
@rmohr I agree. I think this PR should wait until #3542 is merged. Then I will update this PR to only clear the smbios values and remove all the configuration manipulation for restoring the correct config.

tests/vmi_configuration_test.go
stu-gott: @vatsalparekh is the default machine type fix for *this* test or a related one in this module?

PR #3524, opened 2020-06-08
 Refactor Travis CI test script 

oshoval:
Separate bash script to functions.
Add prints to assist finding what is the actual failing command without
the need to scroll back.

Support providing list of functions via arguments,
in order to enable running parallel automation of the script.

```release-note
None
```

### DIFF
--- a/automation/travisci-test.sh
+++ b/automation/travisci-test.sh
@@ -1,8 +1,23 @@
 #!/bin/bash -e
 
 
-make generate-verify
-make bazel-build-verify
+make generate
+if [[ -n "$(git status --porcelain)" ]] ; then
+    echo "It seems like you need to run 'make generate'. Please run it and commit the changes"
+    git status --porcelain; false
+fi
+
+if diff <(git grep -c '') <(git grep -cI '') | egrep -v -e 'docs/.*\.png|swagger-ui' -e 'vendor/*' -e 'assets/*' | grep '^<'; then
+    echo "Binary files are present in git repostory."; false
+fi
+
+make
+
+if [[ -n "$(git status --porcelain)" ]] ; then
+    echo "It seems like you need to run 'make'. Please run it and commit the changes"; git status --porcelain; false
+fi
+
+make build-verify # verify that we set version on the packages built by bazel
 
 # The make bazel-test might take longer then the current timeout for a command in Travis-CI of 10 min, so adding a keep alive loop while it runs
 while sleep 9m; do echo "Long running job - keep alive"; done & LOOP_PID=$!
--- a/automation/travisci-test.sh
+++ b/automation/travisci-test.sh
@@ -1,41 +1,110 @@
 #!/bin/bash -e
+#
+# This file is part of the KubeVirt project
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2020 Red Hat, Inc.
+#
 
+last_running_func=""
+trap '{ [[ $last_running_func != "" ]] && echo "Last running function was:" $last_running_func; }' EXIT ERR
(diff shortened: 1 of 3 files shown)


EdDev:
Generaly speaking, improvements in this area are discouraged at this time because the plan is to migrate it to prow as multiple jobs. 

> Separate bash script to functions.

+1 to structure.

> Add prints to assist finding what is the actual failing command without
> the need to scroll back.

Most operations are of ‘make’ which print on what target they failed.

> 
> Add sanity check that all expected functions are executed.
> This will allow developers to change order, or run a subset
> of the functions during development, but still guard
> that at the end the debug change is reverted before Travis
> allows to merge.

This is an overkill. If you have functions, you can easily choose what to run.

—-

I feedback on specifics for completeness. As the plan is to move to prow and split into multiple jobs, effort should be focused on bugs and not enhancement at this stage.

oshoval:
> Generaly speaking, improvements in this area are discouraged at this time because the plan is to migrate it to prow as multiple jobs.
> 
> > Separate bash script to functions.
> 
> +1 to structure.
> 
> > Add prints to assist finding what is the actual failing command without
> > the need to scroll back.
> 
> Most operations are of ‘make’ which print on what target they failed.
some doesnt , and then at the end you must scroll back, now you dont

> 
> > Add sanity check that all expected functions are executed.
> > This will allow developers to change order, or run a subset
> > of the functions during development, but still guard
> > that at the end the debug change is reverted before Travis
> > allows to merge.
> 
> This is an overkill. If you have functions, you can easily choose what to run.
> 
was groomed with Dan, to do it as i did for the lanes, it allow people to easily remove functions, and have fast answer, without forgetting to revert it.
> —-
> 
> I feedback on specifics for completeness. As the plan is to move to prow and split into multiple jobs, effort should be focused on bugs and not enhancement at this stage.

oshoval:
synced with @danielBelenky 
Its now supports providing function list, and allow easier automation.
Rebased and fixed build-verify usage.

automation/travisci-test.sh
```diff
@@ -1,41 +1,121 @@
...
+    test_generate 
+    test_binaries
+    test_make 
+    test_goveralls_bazel_test 
+    test_apidocs 
+    test_client_python 
+    test_manifests 
+    test_olm_verify 
+    test_prom_rules_verify    
+)
```
EdDev, line 36: I'm sorry, but IMO this is unmaintainable.

In no language, especially not in a scripting one like bash, I expect one to list the functions of executions unless we have in our hand a pluggable program (and that I expect to see in a real programming language and not in a script).

If one wants to add a new functions, there is a need to add it to multiple places to make it kick. That is unreasonable IMO.

It also has little added value, as we expect the relevant action to fail and exit of first error. If you prefer to add `echo` prints at each function, I am fine with that.

oshoval: I will refactor it a bit, have an idea, it was for guarding that a user can run a specific sub group , but not merging the change
(as was discussed with Dan),

there is more value in this PR, for example the latest change that allow to use arg list of the desired functions to run,
allowing parallel run , as discussed with Daniel

oshoval: wanted to remove the guard, so it would be merged, even that this method was discussed with Dan,
and then saw https://github.com/kubevirt/kubevirt/commit/896cdec4f2775bd6e50c9358e5271f4330dcd462
i think this PR supplies a better readable and constructed method, wdyt ?
should we continue ?

oshoval: PTAL

automation/travisci-test.sh
```diff
@@ -1,26 +1,110 @@
...
+    fi
+}
 
-if [[ $TRAVIS_REPO_SLUG == "kubevirt/kubevirt" && $TRAVIS_CPU_ARCH == "amd64" ]]; then
-    make goveralls
-else
-    make bazel-test
-fi
+function test_binaries {
+    if diff <(git grep -c '') <(git grep -cI '') | egrep -v -e 'docs/.*\.png|swagger-ui' -e 'vendor/*' -e 'assets/*' | grep '^<'; then
```
oshoval, line 44: i can split to lines if needed as was done on
hack/check-for-binaries.sh

oshoval:
@danielBelenky 
Would you please review again ? 
did the requests you asked, reverted the change that moves logic to the Makefile, 
because this one keeps all the targets in one script, which is more maintainable and have common structure.
it support getting function name from command line as we spoke, so it can do the same as the one that was reverted.

oshoval:
Imo this is much cleaner than using snippets in makefile, but the CI task force decided otherwise.

PR #3527, opened 2020-06-08
wip: prefer qemu:///session over qemu:///system

victortoso:
Signed-off-by: Victor Toso <victortoso@redhat.com>

**What this PR does / why we need it**: This is a WIP to switch from [qemu:///system to qemu:///session](https://blog.wikichoon.com/2016/01/qemusystem-vs-qemusession.html) to verify the feasibility and what problems we can encounter.

This is a WIP/dicussion PR. I can still use examples/vmi-fedora with this change but likely we would remove the libvirtd from running at all.

**Release note**:

```release-note
NONE yet
```


victortoso:
This would need a lot more than I proposed here as moving away to use root user is a must.

PR #3537, opened 2020-06-09
Fix a race condition where we add a role and check for it immediately

stu-gott:
Fixes a race condition in a functional test.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/vm_test.go
+++ b/tests/vm_test.go
@@ -1525,7 +1525,7 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
 					Expect(err).ToNot(HaveOccurred(), "ERR: %s", stdOut+stdErr)
 
 					By("Ensuring user has the admin rights for the test namespace project")
-					// This simulates the ordinary user as an admin in his project
+					// This simulates the ordinary user as an admin in this project
 					stdOut, stdErr, err = tests.RunCommand(k8sClient, "adm", "policy", "add-role-to-user", "admin", fmt.Sprintf("system:serviceaccount:%s:%s", tests.NamespaceTestDefault, testUser))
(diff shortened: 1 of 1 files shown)


stu-gott:
/cc @jean-edouard 
Would you mind reviewing?

tests/vm_test.go
```diff
@@ -1546,9 +1546,11 @@ var _ = Describe("[rfe_id:1177][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
 
 					By("Checking VM creation permission using k8s client binary")
-					stdOut, _, err := tests.RunCommand(k8sClient, "auth", "can-i", "create", "vms", "--as", testUser)
-					Expect(err).ToNot(HaveOccurred())
-					Expect(strings.TrimSpace(stdOut)).To(Equal("yes"))
+					// It might take time for the role to propagate
+					Eventually(func() string {
+						stdOut, _, _ := tests.RunCommand(k8sClient, "auth", "can-i", "create", "vms", "--as", testUser)
+						return strings.TrimSpace(stdOut)
+					}, 120*time.Second, 10*time.Second).Should(Equal("yes"), fmt.Sprintf("test account '%s' was never granted permission to create a VM", testUser))
```
rmohr, line 1553: I would expect that we retry each second and wait up to 10 seconds. I am not so fluent in this area: What 3xactly need time to be propagated?

stu-gott: The BeforeEach of this test is essentially running "oc adm policy add-role-to-user admin testuser". This does not appear to take effect instantly.

Your approach to timing is probably more reasonable.

rmohr:
Will take the PR in once I am convinced by CI runs.

rmohr:
Test results are good enough. Merging to increase test stability.

PR #3540, opened 2020-06-10
Fix flaky test 1625 by removing outdated condition

oshoval:
This test, uses WaitForSuccessfulVMIStart.
The WaitForSuccessfulVMIStart helper
waits for a VMI to reach running phase.
    
Therefore, there is no need for another condition to wait
for the VMI to reach running.
The VMI condition manager status check is therefore removed.
    
The eventually clause in which the condition was examined is
also incorrect, and once it was checked, further retry on it did
not refreshed the data and just re-examined the same value.

Seen at
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3503/pull-kubevirt-e2e-k8s-1.16/1269547598943883264

```release-note
None
```

### DIFF
--- a/tests/vmi_lifecycle_test.go
+++ b/tests/vmi_lifecycle_test.go
@@ -38,7 +38,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/apimachinery/pkg/util/rand"
 
-	"kubevirt.io/kubevirt/pkg/controller"
 	device_manager "kubevirt.io/kubevirt/pkg/virt-handler/device-manager"
 
 	v1 "kubevirt.io/client-go/api/v1"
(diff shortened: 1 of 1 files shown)


oshoval:
Thanks
> BTW, the removal of the import required a `make generate`?
You meant i guess `make deps-update` which is the one that needed for deps updating, and no need for both (tried locally before, and if i am wrong travis would catch it)

oshoval:
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3540/pull-kubevirt-check-tests-for-flakes/1270631142336237570
Fails on other test, will open an issue about it, not related to this PR, which change only test 1625 itself.

Rerunning, i believe its a flaky test that one that fails there 1634
/test pull-kubevirt-check-tests-for-flakes

oshoval:
i just added a commit and revert it,
we are back to the same state as before
wanted to make sure pull-kubevirt-check-tests-for-flakes fails on master as well.

PTAL

rmohr:
Test failures are unrelated. Merging to improve test stability.

PR #3542, opened 2020-06-10
[tests] Wait for config changes to be propagated and reset them implicitly

rmohr:

Every kubevirt component now exposes a `/healthz` endpoint. This endpoint returns a in the json a field called `config-resource-version` and the value of it reflects the resource version of the config which the component currently uses.

The tests take advantage of this new information and query every kubevirt component for this version and wait until the resource-version matches the expected version whenever the KubeVirt config is updated.

After each test the value is reset to the default config, again with waiting until the config is propagated. This makes it unnecessary to reset the values as part of the test.

Finally let the tests make use of the new features and for all tests, except the networking tests which use `BeforeAll`, and drop code which tries to reset explicitly in the test to the original config version.

This should fix flakes in tests like 

```
```

as well as making testing with config changes much more robust.

Fixes #

**Release note**:

```release-note
NONE
```


rmohr:
/hold

Needs #3567 to be merged before this one, since two tests are now failing with my change which need #3567  to work again.

rmohr:
/hold 

Thanks for the reviews. I think that the tests which are failing here are really failing because the test propagation never worked for them and  my change reveals that. Have to go over the tests and fix them.

rmohr:
/hold

Needs #3649 to finally succeed.

rmohr:
/hold cancel
/retest
 #3649 is merged.

PR #3543, opened 2020-06-10
tests, vmi: Fix IPv6 configuration through the console

oshoval:
configureIPv6OnVMI didn't actually update the /etc/resolv.conf

Updating DNS for cirros was added as nice to have only.
The tests that actually need it, use fedora VM, which use
cloud-init for this purpose.
Hence its safe and better to remove it.
Motivation and discovery was due to flaky test 3184.

Reasons that resolv.conf wasnt updated:

 Missing sudo su for cirros.
 Using of backtick malformatted the command [1]

Adjust the expected characters when examining the echo $? response
such that it will catch the actual retcode and not the a character
which is echoed back (i.e is part of the sent command itself).

Example when match with its own command input:
```
{11 sleep 5\r\n$ sudo ip -6 route add default via fd10 [0]}
```
And how it is with the fix:
```
{11 sudo ip -6 route add default via fd10:0:2::1 src fd10:0:2::2\r\n$ echo $?\r\n0 [\n0]}
```

[1] https://golang.org/ref/spec#String_literals

---
Seen on

test_id 3184
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3526/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1269974170763005953

Same error test_id 1780
See https://github.com/kubevirt/kubevirt/issues/3619

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -2876,19 +2876,23 @@ func CheckForTextExpecter(vmi *v1.VirtualMachineInstance, expected []expect.Batc
 }
 
 func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter, virtClient kubecli.KubevirtClient, prompt string) error {
-	shouldConfigureIpv6 := func(vmi *v1.VirtualMachineInstance) bool {
-		shouldConfigureIpv6Batch := append([]expect.Batcher{
+	hasEth0Iface := func(vmi *v1.VirtualMachineInstance) bool {
+		hasNetEth0Batch := append([]expect.Batcher{
 			&expect.BSnd{S: "\n"},
 			&expect.BExp{R: prompt},
-			&expect.BSnd{S: "ip a | grep -q eth0\n"},
+			&expect.BSnd{S: "ip a | grep -q eth0; echo $?\n"},
+			&expect.BExp{R: retcode("0")}})
+		_, err := expecter.ExpectBatch(hasNetEth0Batch, 30*time.Second)
+		return err == nil
+	}
+
+	hasGlobalIPv6 := func(vmi *v1.VirtualMachineInstance) bool {
+		hasGlobalIPv6Batch := append([]expect.Batcher{
+			&expect.BSnd{S: "\n"},
 			&expect.BExp{R: prompt},
-			&expect.BSnd{S: "echo $?\n"},
-			&expect.BExp{R: "0"},
-			&expect.BSnd{S: "ip address show dev eth0 scope global | grep -q inet6\n"},
-			&expect.BExp{R: prompt},
-			&expect.BSnd{S: "echo $?\n"},
-			&expect.BExp{R: "1"}})
-		_, err := expecter.ExpectBatch(shouldConfigureIpv6Batch, 30*time.Second)
+			&expect.BSnd{S: "ip -6 address show dev eth0 scope global | grep -q inet6; echo $?\n"},
+			&expect.BExp{R: retcode("0")}})
+		_, err := expecter.ExpectBatch(hasGlobalIPv6Batch, 30*time.Second)
 		return err == nil
 	}
 
@@ -2902,34 +2906,35 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
 	if !netutils.IsIPv6String(dnsServerIP) ||
 		(vmi.Spec.Domain.Devices.Interfaces == nil || len(vmi.Spec.Domain.Devices.Interfaces) == 0 || vmi.Spec.Domain.Devices.Interfaces[0].InterfaceBindingMethod.Masquerade == nil) ||
 		(vmi.Spec.Domain.Devices.AutoattachPodInterface != nil && !*vmi.Spec.Domain.Devices.AutoattachPodInterface) ||
-		!shouldConfigureIpv6(vmi) {
+		(!hasEth0Iface(vmi) || hasGlobalIPv6(vmi)) {
 		return nil
 	}
 
-	ipv6Batch := append([]expect.Batcher{
+	addIPv6Address := append([]expect.Batcher{
 		&expect.BSnd{S: "\n"},
 		&expect.BExp{R: prompt},
-		&expect.BSnd{S: "sudo ip -6 addr add fd10:0:2::2/120 dev eth0\n"},
-		&expect.BExp{R: prompt},
-		&expect.BSnd{S: "echo $?\n"},
-		&expect.BExp{R: "0"},
-		&expect.BSnd{S: "sleep 5\n"},
-		&expect.BExp{R: prompt},
-		&expect.BSnd{S: "sudo ip -6 route add default via fd10:0:2::1 src fd10:0:2::2\n"},
-		&expect.BExp{R: prompt},
-		&expect.BSnd{S: "echo $?\n"},
-		&expect.BExp{R: "0"},
-		&expect.BSnd{S: fmt.Sprintf(`echo "nameserver %s" >> /etc/resolv.conf\n`, dnsServerIP)},
-		&expect.BExp{R: prompt},
-		&expect.BSnd{S: "echo $?\n"},
-		&expect.BExp{R: "0"}})
-	resp, err := expecter.ExpectBatch(ipv6Batch, 1*time.Minute)
-
+		&expect.BSnd{S: "sudo ip -6 addr add fd10:0:2::2/120 dev eth0; echo $?\n"},
+		&expect.BExp{R: retcode("0")}})
+	resp, err := expecter.ExpectBatch(addIPv6Address, 30*time.Second)
 	if err != nil {
-		log.DefaultLogger().Object(vmi).Infof("Configure ipv6: %v", resp)
+		log.DefaultLogger().Object(vmi).Infof("addIPv6Address failed: %v", resp)
 		expecter.Close()
 		return err
 	}
+
+	time.Sleep(5 * time.Second)
+	addIPv6DefaultRoute := append([]expect.Batcher{
+		&expect.BSnd{S: "\n"},
+		&expect.BExp{R: prompt},
+		&expect.BSnd{S: "sudo ip -6 route add default via fd10:0:2::1 src fd10:0:2::2; echo $?\n"},
+		&expect.BExp{R: retcode("0")}})
+	resp, err = expecter.ExpectBatch(addIPv6DefaultRoute, 30*time.Second)
+	if err != nil {
+		log.DefaultLogger().Object(vmi).Infof("addIPv6DefaultRoute failed: %v", resp)
+		expecter.Close()
+		return err
+	}
+
 	return nil
 }
 
@@ -4934,4 +4939,8 @@ func IsLauncherCapabilityValid(capability k8sv1.Capability) bool {
 		return true
 	}
 	return false
+}
+
+func retcode(retcode string) string {
+	return "\n" + retcode
 }


oshoval:
> There are a couple of things I don't understand:
> 
>     * if the issue is not being sudo, this would fail consistently. Why does it usually work and sometimes fails ?
> 
>     * if the issue is the trailing `\n` in the resolv.conf - which using double-quotes instead of back-ticks would prevent - I also think it should fail consistently. Why does it fail some times ?

It always fails on adding the nameserver,
the function was added just nice to have, not because we actually needed it,
because cirros didnt have cloud-init userdata, i dont know if its changed since then (does cirros support it right now),
The fedora already have the user-data in the cloud-init so it doesnt reach this code because of the condition above the code i changed.

i had to fix both the issues, in order to make it work, there might be other alternatives, but this is the right way imo, without it, commands get cutted (prints show it),
comand arent executed, or test reads its own output even if the sanity is added,
i mean that if i add the expected string it think its ok, but its not, and if i add a wrong string it shows its wrong.

tests/utils.go
```diff
@@ -2910,14 +2912,20 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
 		&expect.BExp{R: prompt},
 		&expect.BSnd{S: "echo $?\n"},
 		&expect.BExp{R: "0"},
-		&expect.BSnd{S: fmt.Sprintf(`echo "nameserver %s" >> /etc/resolv.conf\n`, dnsServerIP)},
```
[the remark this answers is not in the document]
oshoval: it fails always, just sometimes u do see it fails , its matter of race, i simulated it locally by adding a sanity check at the end , or by adding a seperated snippet for sanity (something like you want in the comment above) and saw it fails 10/10 times

tests/utils.go
```diff
@@ -2910,14 +2912,20 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
...
+		&expect.BSnd{S: "sudo su\n"},
```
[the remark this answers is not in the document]
oshoval: the code wont run on fedoras that already have the nameserver in their user-data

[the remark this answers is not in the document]
maiqueb: And on fedora VMs that don't ? I think having a function that has redundant stuff based on the input is even worse than having a function that does redundant stuff always.

tests/utils.go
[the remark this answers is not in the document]
maiqueb: Race ? Race between what ? 

On a separate note... I'm getting the impression that if we delete everything DNS related from this function we'll achieve the same result - since DNS is only required for the guestagent tests, where DNS is configured via cloud-init.

Instead of fixing, I would actually remove this from here - since it's not needed.

tests/utils.go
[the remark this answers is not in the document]
oshoval: the sandboxing here is needed thats my point, 
else we need a sleep (not deterministic) or something, because if you take the original code and add
only this:

```
cat /etc/resolv.conf | tail -1
expect the right thing
```
it will pass
but if you
```
expect wrong thing
```
it will fail

and it actually didnt change the file (and even shows permission denided).
so it uses its own strings, and the flow is broken because of the backquote

this is why i think the sandbox is important

tests/utils.go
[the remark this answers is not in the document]
oshoval: race between the expect of this own block or the following ones
see please the 2 reports in the commit message

indeed we might be able to remove it, i wanted to remove it for long time, (mentioned on couple PRs), but it need to be done safe if we want.
atm prow is broken, should we just remove it then and see that nothing fails, or we do want to have a function in case we need it in the future, and learn what i can from it ?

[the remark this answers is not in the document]
maiqueb: > race between the expect of this own block (see 3184) or the following block (see for example 3187)
> see please the 2 reports in the commit message

I don't understand what you're saying; please include links to kubevirt code instead of making me fish for (what I think are) lines of code in what I assume to be the same file. This is not needed since it appears we have an agreement on the course of action - remove DNS configuration via console).

> 
> indeed we might be able to remove it, i wanted to remove it for long time, (mentioned on couple PRs), but it need to be done safe if we want.
> atm prow is broken, should we just remove it then and see that nothing fails, or we do want to have a function in case we need it in the future, and learn what we can from it ?
> green light to remove ?

+1 to remove it.

tests/utils.go
```diff
@@ -2910,14 +2902,14 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
 		&expect.BExp{R: prompt},
 		&expect.BSnd{S: "echo $?\n"},
 		&expect.BExp{R: "0"},
-		&expect.BSnd{S: fmt.Sprintf(`echo "nameserver %s" >> /etc/resolv.conf\n`, dnsServerIP)},
-		&expect.BExp{R: prompt},
-		&expect.BSnd{S: "echo $?\n"},
-		&expect.BExp{R: "0"}})
+		&expect.BSnd{S: "ip -6 r | grep via | grep src\n"},
+		&expect.BExp{R: "default via fd10:0:2::1"},
+		&expect.BSnd{S: "sync\n"},
```
oshoval, line 2907:
```
{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"Configure ipv6 GOOD: [{1 \r\n$  [$ ]} {3 \r\n$  [$ ]} {5 sudo ip -6 addr add fd10 [0]} {7 :0:2::2/120 dev eth0\r\n$  [$ ]} {9 echo $?\r\n0\r\n$  [$ ]} {11 sleep 5\r\n$ sudo ip -6 route add default via fd10 [0]} {13 :0:2::1 src fd10:0:2::2\r\n$ echo $?\r\n0\r\n$ ip -6 r | grep via | grep src\r\ndefault via fd10:0:2::1  [default via fd10:0:2::1]} {15 dev eth0  src fd10:0:2::2  metric 1024 \r\n$  [$ ]}]","name":"testvmifkpxfvnmvrv7pd6lssnfwhxhxp6fxdvpw2b98m5qt8xfg5mp","namespace":"kubevirt-test-default","pos":"utils.go:2918","timestamp":"2020-06-11T14:13:00.468152Z","uid":"88bbb828-23ef-4607-8bb1-c4d0aef0ded4"}
```
FYI 
i added the sync so there wont be consumable characters left
i could wait for 
default via fd10:0:2::1 dev eth0  src fd10:0:2::2  metric 1024
but didnt want a static metric, if it will be changed one day it would fail

i could wait just for prompt, but wanted to be safer

oshoval: imo sudo su and exit would be the best, because it will sandbox all stdout and stderr

tests/utils.go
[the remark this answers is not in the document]
oshoval: fedora doesnt rerun it, because the beginning of the function checks if there is already ipv6, and then it doesnt do the 2nd block (ip + route + dns),
anyway that block always fails because the expect isnt formatted well, i added another commit (wip) that fix the expect and remove the dns, as we said that we dont need it actually, and it was added just nice to have, without actual proven need

tests/utils.go
```diff
@@ -2883,49 +2883,29 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
 		return err == nil
 	}
 
-	dnsServerIP, err := getClusterDnsServiceIP(virtClient)
-	if err != nil {
-		log.DefaultLogger().Object(vmi).Infof("Couldn't get DNS Service IP while configuring ipv6: %v", err)
-		expecter.Close()
-		return err
-	}
```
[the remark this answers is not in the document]
oshoval: as we said, we are removing the DNS from here, whoever needs the DNS will do it with cloud-init, 
the code that i am removing actually doesnt work (it doesnt set nameserver), and not needed, i also fix the problems on the expect there in the 2nd commit that i will squash at the end as we said, it still wip, it works, but i am checking one more thing.

the condition you speak on is in the same function. cirros doesnt have userdata so its the one that adding the dns, fedora ones that need dns added it via user-data
we dont have other that do need it (otherwise it would fail , because the backticks, not the sudo su)

tests/utils.go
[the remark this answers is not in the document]
maiqueb: sorry, I didn't mean to push this comment, since I've later realized you were pushing stuff on top to later squash. Couldn't find it afterwards since the conversation was marked as resolved (as as such, the comment tree was collapsed.) 

Since I couldn't find it, I thought I had deleted it.

Forget about this particular question.

oshoval:
> addDnsEntry

since we are removing this one, i wont refactor this but remove it and refactor the other 2 ok ?
adding it need couple things to be fixed (backticks + sudo su)
and there is another 3rd bug which is general, and that one i will purpose the fix for

tests/utils.go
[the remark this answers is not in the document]
maiqueb: I think I've expressed my self poorly: I am asking about **routes** and the **IP address** configuration, not DNS. I'm well aware we've agreed to remove DNS configuration via console, leaving cloud-init as the only option to do configure DNS in the guests.

We read the IP of the DNS service for 2 reasons: 
  1. [config DNS in the guest](https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L2913)
  2. [know if](https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L2893) we're on an IPv6 cluster

Here, I'm concerned about the second point; if you remove this condition, we will end up configuring an ipv6 address + route on guests created on ipv4 clusters, which, while *probably* safe, is **not** required. 

Keeping the code that retrieves the cluster DNS service IP, and also keeping it [in the condition](https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L2893) would be preferable, IMHO, since guests on IPv4 clusters do not need - at all - that IPv6 address. I find it preferable to stick to the minimal configuration that works, since the tests can be used as references in the future.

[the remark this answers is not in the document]
oshoval: right, will do thanks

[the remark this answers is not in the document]
oshoval: done

tests/utils.go
```diff
@@ -2871,14 +2871,10 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
...
-			&expect.BSnd{S: "ip a | grep -q eth0\n"},
-			&expect.BExp{R: prompt},
-			&expect.BSnd{S: "echo $?\n"},
-			&expect.BExp{R: "0"},
-			&expect.BSnd{S: "ip address show dev eth0 scope global | grep -q inet6\n"},
-			&expect.BExp{R: prompt},
-			&expect.BSnd{S: "echo $?\n"},
-			&expect.BExp{R: "1"}})
+			&expect.BSnd{S: "ip a | grep -q eth0; echo $?\n"},
+			&expect.BExp{R: "\n0"},
```
oshoval, line 2875: instead of \n0 
(?m)^0$
this would also work, if preferred ?

it means multi line mode (which is needed, else it looks at the current buffer that it already filled)
and then it searches for a line which has only 0 on it.

oshoval:
@phoracek 
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3543/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1272503364130705409
fyi
this one took 7h and killed due to that, started today
Thanks

tests/utils.go
EdDev: I think I understand now. Maybe adding "$" will be even more accurate. `"\n0$"`

oshoval: it means end of buffer if `?m` isnt used
end of buffer is reached easily, at the moment you have just those 2 characters "at hand", 
its not meaning its indeed come with end of line afterwards
this is why i suggested the above regex

tests/utils.go
```diff
@@ -2893,31 +2887,33 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
...
-		&expect.BSnd{S: "echo $?\n"},
-		&expect.BExp{R: "0"},
-		&expect.BSnd{S: fmt.Sprintf(`echo "nameserver %s" >> /etc/resolv.conf\n`, dnsServerIP)},
+		&expect.BSnd{S: "sudo ip -6 addr add fd10:0:2::2/120 dev eth0; echo $?\n"},
+		&expect.BExp{R: retcode("0")}})
+	resp, err := expecter.ExpectBatch(addIPv6Address, 30*time.Second)
+
+	if err != nil {
+		log.DefaultLogger().Object(vmi).Infof("addIPv6Address failed: %v", resp)
+		expecter.Close()
```
EdDev, line 2903: You should fix this in a following PR, this is unsafe to be used like this.
The expected needs to be released using a `defer` to make sure we do not forget it in one of the paths.

tests/utils.go
```diff
@@ -2893,31 +2887,33 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
...
+		return err
+	}
+
+	time.Sleep(5 * time.Second)
```
EdDev, line 2907: You touched it, so I think you should use a poll mode here.
We already do this already in another place using `for...`.

tests/utils.go
oshoval: you cant release the expecter, just if error occurs
this the whole logic in those functions  (and the callers)

tests/utils.go
oshoval: i dont think we need to do a for using expect, just to wait before and then to do the command
its safer and nicer to not use expect for it.

tests/utils.go
oshoval: will handle comments on sunday, thanks

oshoval:
/hold
the change here discover some other real problem in tests i think.

tests/utils.go
```diff
@@ -2870,58 +2870,63 @@ func CheckForTextExpecter(vmi *v1.VirtualMachineInstance, expected []expect.Batc
 }
 
 func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter, virtClient kubecli.KubevirtClient, prompt string) error {
-	shouldConfigureIpv6 := func(vmi *v1.VirtualMachineInstance) bool {
-		shouldConfigureIpv6Batch := append([]expect.Batcher{
+	hasEth0Iface := func(vmi *v1.VirtualMachineInstance) bool {
```
EdDev, line 2873: You have not reason on why you re-introduced this one.

Please mention significant changes between pushes which are not part of the last review (so the reviewers will identify what changed).

tests/utils.go
```diff
@@ -4918,3 +4930,7 @@ func DetectLatestUpstreamOfficialTag() (string, error) {
 	By(fmt.Sprintf("By detecting latest upstream official tag %s for current branch", tag))
 	return tag, nil
 }
+
+func Retcode(retcode string) string {
```
EdDev, line 4934: Why is this an exposed function now?

tests/utils.go
```diff
@@ -2870,58 +2870,63 @@ func CheckForTextExpecter(vmi *v1.VirtualMachineInstance, expected []expect.Batc
...
-			&expect.BExp{R: "1"}})
-		_, err := expecter.ExpectBatch(shouldConfigureIpv6Batch, 30*time.Second)
+			&expect.BSnd{S: "ip -6 address show dev eth0 scope global | grep -q inet6; echo $?\n"},
+			&expect.BExp{R: Retcode("0")}})
+		_, err := expecter.ExpectBatch(hasGlobalIPv6Batch, 30*time.Second)
 		return err == nil
 	}
 
 	dnsServerIP, err := getClusterDnsServiceIP(virtClient)
+	defer checkClose(&err, expecter)
```
EdDev, line 2894: When I commented on this, my intention and wish was that it will be closed on the same level of abstraction that it was created. I even mentioned that if I recall correctly.

I find this defer closure to be worse then the current because it asks the same question again.
If you do not intend to fix the expecter object life-cycle from the root, I prefer you leave the existing flow as is.

tests/utils.go
oshoval: as we spoke, test 1776 doesn't have eth0, i removed but readded back the code, so we are just like the original, refactored

tests/utils.go
oshoval: in case we would need to use the same logic outside of utils.go (some tests have echo $?)

tests/utils.go
EdDev: I understand that `eth0` may be missing (which is a problem by itself), but you need to explain why it is not enough to just use `hasGlobalIPv6`.

tests/utils.go
EdDev: I was very close to nack this PR only because you touched this huge endless file.
Please do not make me regret this overlook.

No one else is using it now, if someone will want to use it the can change it then.

tests/utils.go
oshoval: hasGlobalIPv6 might return false in case eth0 doesnt exists,
if eth0 doesnt exists, the function will try to config ipv6 on it and fail 
this is why we need to check that eth0 do exist and that it doesnt have ipv6 global, before we proceed configuration.

EdDev: OK, that is more clear now.

Can you paste here the test that did not have on the VM `eth0`?

oshoval: https://github.com/kubevirt/kubevirt/blob/82cbf4822c1708472e1dc2819705063125274a93/tests/vmi_networking_test.go#L584

test_id 1776

rmohr:
> Updating DNS for cirros was added as nice to have only.
> The tests that actually need it, use fedora VM, which use
> cloud-init for this purpose.
> Hence its safe and better to remove it.
> Motivation and discovery was due to flaky test 3184.

This message is not 100% clear to me. Do you mean that we have another test which tests the same thing?

AlonaKaplan:
> Please first (step 0) refactor this by extracting the expects so we can easily read what each is doing (e.g. addIpv6Address, addIPv6DefaultRoute, addDnsEntry).

The reason there was one big expect and it wasn't broken to parts is that for some reason, sometimes #expect 2 returned the `err` value of #expect 1 (if a sleep was added before #expect 2 the issue was solved). Please make sure this problem is not reproduced now you broke the expect into parts.

> After, do whatever changes are needed to fix the problem you identified.
> 
> It is already a mess, less not patch the mess.

oshoval:
> > Please first (step 0) refactor this by extracting the expects so we can easily read what each is doing (e.g. addIpv6Address, addIPv6DefaultRoute, addDnsEntry).
> 
> The reason there was one big expect and it wasn't broken to parts is that for some reason, sometimes #expect 2 returned the `err` value of #expect 1 (if a sleep was added before #expect 2 the issue was solved). Please make sure this problem is not reproduced now you broke the expect into parts.
> 
> > After, do whatever changes are needed to fix the problem you identified.
> > It is already a mess, less not patch the mess.

it doesn't reproduce, i found the root cause of it, (actually there were 3 problems that are fixed here)

oshoval:
> > Updating DNS for cirros was added as nice to have only.
> > The tests that actually need it, use fedora VM, which use
> > cloud-init for this purpose.
> > Hence its safe and better to remove it.
> > Motivation and discovery was due to flaky test 3184.
> 
> This message is not 100% clear to me. Do you mean that we have another test which tests the same thing?

Ye, adding DNS in this function was just addon without the need, 
the tests that do need DNS, do it via UserData (and we do have tests like that, and there its needed)

tests/utils.go
```diff
@@ -4918,3 +4923,7 @@ func DetectLatestUpstreamOfficialTag() (string, error) {
 	By(fmt.Sprintf("By detecting latest upstream official tag %s for current branch", tag))
 	return tag, nil
 }
+
+func retcode(retcode string) string {
+	return "\n" + retcode
```
AlonaKaplan, line 4928: Can you please explain why the "\n" is needed here?

oshoval: without it , if you write a command 
ip route add .... f10:....
and then do echo $?
and EXP for 0
the 0 of the command would enter the buffer, and it will match for the EXP

you can see the buffer and the matching if you use verbose or better imo by dumping the structs which say exactly which EXP it correspond to.

with \n , since we are not in multiline mode of re2, it will enforce that its on a new line and not the output of the command that we used (unless we have HEREDOC or something ,which we dont)

AlonaKaplan:
> configureIPv6OnVMI didn't actually update the /etc/resolv.conf
> 
> Updating DNS for cirros was added as nice to have only.
> The tests that actually need it, use fedora VM, which use
> cloud-init for this purpose.
> Hence its safe and better to remove it.
> Motivation and discovery was due to flaky test 3184.
> 
> Reasons that resolv.conf wasnt updated:
> 
> Missing sudo su for cirros.
> Using of backtick malformatted the command [1]
> 
> Adjust the expected characters when examining the echo $? response
> such that it will catch the response and not the sent command itself
> (which is echoed back).
> 
> [1] https://golang.org/ref/spec#String_literals
> 
> Seen on
> 
> test_id 3184
> https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3526/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1269974170763005953
> 
> Same error test_id 1780
> See #3619
> 
> Signed-off-by: Or Shoval [oshoval@redhat.com](mailto:oshoval@redhat.com)


Can you please explain what caused the test to be flaky? I don't understand it from the comment.

oshoval:
> > configureIPv6OnVMI didn't actually update the /etc/resolv.conf
> > Updating DNS for cirros was added as nice to have only.
> > The tests that actually need it, use fedora VM, which use
> > cloud-init for this purpose.
> > Hence its safe and better to remove it.
> > Motivation and discovery was due to flaky test 3184.
> > Reasons that resolv.conf wasnt updated:
> > Missing sudo su for cirros.
> > Using of backtick malformatted the command [1]
> > Adjust the expected characters when examining the echo $? response
> > such that it will catch the response and not the sent command itself
> > (which is echoed back).
> > [1] https://golang.org/ref/spec#String_literals
> > Seen on
> > test_id 3184
> > https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3526/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1269974170763005953
> > Same error test_id 1780
> > See #3619
> > Signed-off-by: Or Shoval [oshoval@redhat.com](mailto:oshoval@redhat.com)
> 
> Can you please explain what caused the test to be flaky? I don't understand it from the comment.

there were 3 bugs, the dns wasnt actually configured because 
backquote was used and malformatted the lines, 
sudo su was missing for echo nameserver >> /etc/resolv

the structure of the expect was bad because it didnt wait right for retcode 
all those results in a racey env that just once in a while, depend on the buffer that Expect look on , it decide if the EXP is met,
usually it did have the 0 or 1 it waited for, but sometimes it doesnt, even if it had, it was not the real one,
if you take the test and fix all the stuff related to EXP, u will see that adding the DNS always fails
(because sudo su and missing and the use of backquote)
and the inputs and outputs of the EXP are used wrongly to match the EXP, 
the error was silent usually, as it wasnt actually even needed in those tests,
just when the EXP buffer didnt have the needed chars (for example delay in the system), it failed.

tests/utils.go
AlonaKaplan: Sorry, I still don't understand.
Isn't it a good think the 0 return by echo $?\n will match the Exp?

oshoval: its not returned by the echo $? 
its the 0 that is part of the SND command itself
since its in the buffer that EXP look on

you can se here when it say {11 ... what it matches
```
----

the good

{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"Configure ipv6 GOOD: [{1 \r\n$  [$ ]} {3 \r\n$  [$ ]} {5 sudo ip -6 addr add fd10:0:2::2/120 dev eth0\r\n$ echo $?\r\n0 [\n0]} {7 \r\n$  [$ ]} {9 sleep 5\r\n$  [$ ]} {11 sudo ip -6 route add default via fd10:0:2::1 src fd10:0:2::2\r\n$ echo $?\r\n0 [\n0]}]","name":"testvmi2r97k92hd8mjxsdp84n4sszdhw65tzz2wn79krfrjwwk9vzz","namespace":"kubevirt-test-default","pos":"utils.go:2917","timestamp":"2020-06-15T06:59:56.111569Z","uid":"05a36d8f-91e4-4fa3-9ed6-52a432a2a626"}


the bad

{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"Configure ipv6 GOOD: [{1 \r\n$  [$ ]} {3 \r\n$  [$ ]} {5 sudo ip -6 addr add fd10 [0]} {7 :0:2::2/120 dev eth0\r\n$  [$ ]} {9 echo $?\r\n0\r\n$  [$ ]} {11 sleep 5\r\n$ sudo ip -6 route add default via fd10 [0]}]","name":"testvmi86vtkr99wdp7mlz7t7jjc4gk5jrds9z924n2l4p4mqzm58fr","namespace":"kubevirt-test-default","pos":"utils.go:2917","timestamp":"2020-06-15T07:01:26.190329Z","uid":"8c931971-b5db-4e11-9653-8996054547e6"}

---
```

AlonaKaplan: oh, it considers the 0 in the ip address as the EXP! Now I understand. Thanks for the explanation. Can you please change the PR message so it will be more clear.

oshoval: updated, hope its better now

AlonaKaplan:
@oshoval  Shouldn't we add the same fix to more places. For example - https://github.com/kubevirt/kubevirt/blob/a897c7bc904e42fcff92a1f9480acd580f3751db/tests/utils.go#L4256
Can you please do it on a separate PR?

oshoval:
> @oshoval Shouldn't we add the same fix to more places. For example -
> 
> https://github.com/kubevirt/kubevirt/blob/a897c7bc904e42fcff92a1f9480acd580f3751db/tests/utils.go#L4256
> 
> 
> Can you please do it on a separate PR?

We might need to, yes, planned on that, but every test that we change, i prefer first to dump the tuples and see its as expected.
i will take it when possible, will open a card so we don' forget.
thanks

tests/utils.go
AlonaKaplan: It's better but it is still unclear the EXP caught the '0' character of the ip address.

oshoval:
failures here aren't due to this PR, can someone review please ?
(lgtm is missing)

oshoval:
The failures are flaky tests, not due to this PR, can we override please?
This PR solves a lot of flakiness, thanks

rmohr:
One test failed on 

```
panic: send on closed channel

goroutine 896 [running]:
kubevirt.io/kubevirt/pkg/virt-handler/notify-server.(*Notify).HandleDomainEvent(0xc000410fc0, 0x375fa00, 0xc0014c7050, 0xc0008e3380, 0xc000410fc0, 0xc0014c7050, 0xc002531bd0)
	pkg/virt-handler/notify-server/server.go:82 +0x4b7
kubevirt.io/kubevirt/pkg/handler-launcher-com/notify/v1._Notify_HandleDomainEvent_Handler(0x1a9d460, 0xc000410fc0, 0x375fa00, 0xc0014c7050, 0xc0008e3320, 0x0, 0x375fa00, 0xc0014c7050, 0xc0021f4000, 0x1556)
	bazel-out/k8-fastbuild/bin/pkg/handler-launcher-com/notify/v1/linux_amd64_stripped/kubevirt_notify_go_proto%/kubevirt.io/kubevirt/pkg/handler-launcher-com/notify/v1/notify.pb.go:269 +0x23e
google.golang.org/grpc.(*Server).processUnaryRPC(0xc0027c8580, 0x378abc0, 0xc001c6a780, 0xc001cd2a00, 0xc000411260, 0x48d5020, 0x0, 0x0, 0x0)
	external/org_golang_google_grpc/server.go:995 +0x466
google.golang.org/grpc.(*Server).handleStream(0xc0027c8580, 0x378abc0, 0xc001c6a780, 0xc001cd2a00, 0x0)
	external/org_golang_google_grpc/server.go:1275 +0xda6
google.golang.org/grpc.(*Server).serveStreams.func1.1(0xc001fe03f0, 0xc0027c8580, 0x378abc0, 0xc001c6a780, 0xc001cd2a00)
	external/org_golang_google_grpc/server.go:710 +0x9f
created by google.golang.org/grpc.(*Server).serveStreams.func1
	external/org_golang_google_grpc/server.go:708 +0xa1
```

in virt-handler. Is this related to https://github.com/kubevirt/kubevirt/issues/3575#issuecomment-648019963 or something different wich looks similar?

oshoval:
> One test failed on
> 
> ```
> panic: send on closed channel
> 
> goroutine 896 [running]:
> kubevirt.io/kubevirt/pkg/virt-handler/notify-server.(*Notify).HandleDomainEvent(0xc000410fc0, 0x375fa00, 0xc0014c7050, 0xc0008e3380, 0xc000410fc0, 0xc0014c7050, 0xc002531bd0)
> 	pkg/virt-handler/notify-server/server.go:82 +0x4b7
> kubevirt.io/kubevirt/pkg/handler-launcher-com/notify/v1._Notify_HandleDomainEvent_Handler(0x1a9d460, 0xc000410fc0, 0x375fa00, 0xc0014c7050, 0xc0008e3320, 0x0, 0x375fa00, 0xc0014c7050, 0xc0021f4000, 0x1556)
> 	bazel-out/k8-fastbuild/bin/pkg/handler-launcher-com/notify/v1/linux_amd64_stripped/kubevirt_notify_go_proto%/kubevirt.io/kubevirt/pkg/handler-launcher-com/notify/v1/notify.pb.go:269 +0x23e
> google.golang.org/grpc.(*Server).processUnaryRPC(0xc0027c8580, 0x378abc0, 0xc001c6a780, 0xc001cd2a00, 0xc000411260, 0x48d5020, 0x0, 0x0, 0x0)
> 	external/org_golang_google_grpc/server.go:995 +0x466
> google.golang.org/grpc.(*Server).handleStream(0xc0027c8580, 0x378abc0, 0xc001c6a780, 0xc001cd2a00, 0x0)
> 	external/org_golang_google_grpc/server.go:1275 +0xda6
> google.golang.org/grpc.(*Server).serveStreams.func1.1(0xc001fe03f0, 0xc0027c8580, 0x378abc0, 0xc001c6a780, 0xc001cd2a00)
> 	external/org_golang_google_grpc/server.go:710 +0x9f
> created by google.golang.org/grpc.(*Server).serveStreams.func1
> 	external/org_golang_google_grpc/server.go:708 +0xa1
> ```
> 
> in virt-handler. Is this related to [#3575 (comment)](https://github.com/kubevirt/kubevirt/issues/3575#issuecomment-648019963) or something different wich looks similar?

totally #3575

oshoval:
> @rmohr: Overrode contexts on behalf of rmohr: pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6

typo Overrode

oshoval:
can we override pull-kubevirt-e2e-k8s-1.17 as well please?
the flake there is known as well

PR #3544, opened 2020-06-10
Allow libvirt to relay audit logs (SELinux)

jean-edouard:
The permission is originally dontaudit-ed (silently deny), and doesn't seem to cause any problem.
However, the blocked logs might be useful, and the denial triggers errors like:
`Failed to send audit message virt=kvm vm=\"kubevirt-test-default_testvmi5w79tl6bg28svmx92fkp76r2bttsk96ls7dnrhmkkt6vrnd9\`
Adding the permission will silence these errors by allowing libvirt/qemu to:
"Send user space audit messages to the kernel audit system."

Fixes #

**Release note**:

```release-note
NONE
```


danielBelenky:
/approve
/hold

Unhold when you're ready

rmohr:
Great patch. We have from time to time issues with audit logs. See https://www.redhat.com/archives/libvirt-users/2019-October/msg00031.html. However I don't think this is directly related since that issue happens on clusters without selinux too.

PR #3546, opened 2020-06-10
Fix cascading deletion race condition in some tests

stu-gott:
Fix a race condition in some tests. We delete a VM, which should cause k8s to delete a VMI and DV associated with this VM. In an abundance of caution, we also delete that same VMI and DV--only we didn't account for the fact that k8s could beat us to it.

**Release note**:
```release-note
NONE
```


rmohr:
Test failures are unrelated. Merging to improve test stability.

PR #3551, opened 2020-06-11
Use SriovNetwork API instead of NAD hack

ormergi:
Signed-off-by: Or Mergi <ormergi@redhat.com>

This PR vendors SriovNetwork API and make use of it in sriov lane tests.
Now we can create NAD for sriov network by creating SriovNetwork CR, then the sriov-operator 
reconcile it and create correspond NAD.
This PR is part of stabilizing sriov lane effort.

Fixes #

This PR is currently blocked until kubevirt/builder container go version is bumped to 1.13
which required by SriovNetwork dependency.
Bump go version is addressed at #3420 and #3945.
**Release note**:

```release-note
NONE
```


alonSadan:
Hi 
Not a review yet, but shouldn't there be also an API type for `sriovNetworkNodeState` and for `SriovNetworkNodePolicy`  as well? it might be useful.
what do you think?

phoracek:
@ormergi CI:

```
It seems like you need to run 'make generate'. Please run it and commit the changes
 M tests/BUILD.bazel
 M tests/vmi_multus_test.go
```

tests/vmi_multus_test.go
```diff
@@ -592,6 +591,58 @@ var _ = Describe("SRIOV", func() {
 		sriovResourceName = "openshift.io/sriov_net"
 	}
 
+	createSriovNetwork := func(name, namespace, resourceName, networkNamespace  string, vlan int, trust, spoofCheck, linkState bool) error {
```
alonSadan: maybe you can add a comment about the possibility to pass `nil` value for `vlan` (vlan as an example, maybe there are other attributes that can have `nil` value). because it doesn't have to get value (will be given `0` value by the operator)

tests/vmi_multus_test.go
```diff
@@ -592,6 +591,58 @@ var _ = Describe("SRIOV", func() {
...
+					Trust:            trustString(trust),
+					LinkState:        linkStateString(linkState),
+					MinTxRate:        nil,
+					MaxTxRate:        nil,
+				},
+			}
+
+		err := runtimeClient.Create(context.TODO(), &sriovNetwork)
+		if err != nil {
+
```
alonSadan, line 659: same comment about the blank as before

tests/vmi_multus_test.go
```diff
@@ -131,27 +162,15 @@ var _ = Describe("Multus", func() {
 
 		configureNodeNetwork(virtClient)
 
-		result := virtClient.RestClient().
-			Post().
-			RequestURI(fmt.Sprintf(postUrl, tests.NamespaceTestDefault, "linux-bridge-net-vlan100")).
-			Body([]byte(fmt.Sprintf(linuxBridgeConfCRD, "linux-bridge-net-vlan100", tests.NamespaceTestDefault))).
-			Do()
-		Expect(result.Error()).NotTo(HaveOccurred())
+		err := createNetworkAttachmentDefinition("linux-bridge-net-vlan100", tests.NamespaceTestDefault, linuxBridgeNADConfig)
```
alonSadan, line 165: maybe use const for the name. then you can use it also in the error message one row below

tests/vmi_multus_test.go
```diff
@@ -131,27 +162,15 @@ var _ = Describe("Multus", func() {
...
-			Do()
-		Expect(result.Error()).NotTo(HaveOccurred())
-
-		result = virtClient.RestClient().
-			Post().
-			RequestURI(fmt.Sprintf(postUrl, tests.NamespaceTestAlternative, "ptp-conf-2")).
-			Body([]byte(fmt.Sprintf(ptpConfCRD, "ptp-conf-2", tests.NamespaceTestAlternative))).
-			Do()
-		Expect(result.Error()).NotTo(HaveOccurred())
+		err = createNetworkAttachmentDefinition("ptp-conf-1", tests.NamespaceTestDefault, ptpNADConfig)
```
alonSadan, line 169: ditto

tests/vmi_multus_test.go
```diff
@@ -131,27 +162,15 @@ var _ = Describe("Multus", func() {
...
+		Expect(err).NotTo(HaveOccurred(), "should create 'ptp-conf-1' linux bridge network-attachment-definition")
+
+		err = createNetworkAttachmentDefinition("ptp-conf-2", tests.NamespaceTestAlternative, ptpNADConfig)
```
alonSadan, line 172: ditto

tests/vmi_multus_test.go
```diff
@@ -591,49 +610,49 @@ var _ = Describe("SRIOV", func() {
 		sriovResourceName = "openshift.io/sriov_net"
 	}
 
-	createSriovNetwork := func(name, namespace, resourceName, networkNamespace  string, vlan int, trust, spoofCheck, linkState bool) error {
-		spoofCheckString := func (trust bool) string {
-			if ! spoofCheck {
+	createSriovNetwork := func(name, namespace, resourceName, networkNamespace string, vlan int, trust, spoofCheck, linkState bool) error {
```
alonSadan, line 613: I think this change relates to this https://github.com/kubevirt/kubevirt/pull/3551/commits/ea26c7c19d29a90bd32d860f4f3ad668fc273084
commit

tests/vmi_multus_test.go
```diff
@@ -591,49 +610,49 @@ var _ = Describe("SRIOV", func() {
...
+		spoofCheckString := func(trust bool) string {
+			if !spoofCheck {
 				return "off"
 			}
 
 			return "on"
 		}
 
-		trustString := func (spoofCheck bool) string {
+		trustString := func(spoofCheck bool) string {
```
alonSadan, line 622: ditto

tests/vmi_multus_test.go
```diff
@@ -591,49 +610,49 @@ var _ = Describe("SRIOV", func() {
...
 			if trust {
 				return "on"
 			}
 
 			return "off"
 		}
 
-		linkStateString := func (linkState bool) string {
+		linkStateString := func(linkState bool) string {
```
alonSadan, line 630: ditto

tests/vmi_multus_test.go
```diff
@@ -591,49 +610,49 @@ var _ = Describe("SRIOV", func() {
...
+				IPAM:             "{ \"type\": \"host-local\", \"subnet\": \"10.1.1.0/24\" }",
+				Vlan:             vlan,
+				VlanQoS:          0,
+				SpoofChk:         spoofCheckString(spoofCheck),
+				Trust:            trustString(trust),
+				LinkState:        linkStateString(linkState),
+				MinTxRate:        nil,
+				MaxTxRate:        nil,
+			},
+		}
```
alonSadan, line 655: ditto

tests/vmi_multus_test.go
```diff
@@ -652,14 +671,14 @@ var _ = Describe("SRIOV", func() {
 		}
 
 		// Create two sriov networks referring to the same resource name
-		err := createSriovNetwork("sriov", tests.NamespaceTestDefault,"sriov_net", tests.NamespaceTestDefault, 0, false, true, false)
-		Expect(err).NotTo(HaveOccurred(),"should create SriovNetwork sriov")
+		err := createSriovNetwork("sriov", tests.NamespaceTestDefault, "sriov_net", tests.NamespaceTestDefault, 0, false, true, false)
+		Expect(err).NotTo(HaveOccurred(), "should create SriovNetwork sriov")
```
alonSadan, line 675: ditto

alonSadan:
/assign @alonSadan 

sorry for the noise. accidentally removed my self

tests/vmi_multus_test.go
```diff
@@ -65,6 +67,16 @@ var _ = Describe("Multus", func() {
 	virtClient, err := kubecli.GetKubevirtClient()
 	tests.PanicOnError(err)
 
+	useExistingCluster := true
```
alonSadan, line 70: please consider move this to https://github.com/kubevirt/kubevirt/pull/3551/commits/b02341089320843590a7ea40b968bcccd9926ae8
maybe it is more related to that one

ormergi:
> In order to review this change, I suggest the following generic points to be addressed:
> 
> * Commit messages: Each commit should give a small background, a **why** and then a high-level description of what was done. If there are special and nontrivial behaviors, explain them as well.
>   Make sure to use `topic: subject` on the first line, where the topic gives the context.
> * This PR should only treat SRIOV stuff, if there are commits that changes other things, I suggest you post it over this PR (rebase over it) and mention the dependency in the description.
>   
>   * Upgrades for examples, should not be part of this PR. The last commit seems to do just that.
> * A commit that does one thing is great, but the code it represent needs to be used to be eligible for a stand-alone commit. E.g. defining `foo()` in one commit and using it in the next commit is not a proper separation because it does not show how it is used and the reviewer cannot properly understand if it is fine or not.
>   note: This rule has exceptions, but they are rare and need good reasoning.
> * Function: Please do not use lambda functions when not needed, it is harder to read and cause confusion. When there is a need to use variables from an outer context to build a function, that merits a lambda.
> * String that repeat (e.g. "ok") should be constants.
> 
> In addition, consider moving out SRIOV tests from the `multus` module as the first commit. I am not sure if it is easy or not, but it will help reviewers to focus on SRIOV and not other stuff.

Hi Edy thanks for the feedback !
I pushed new commits I tried to address most of your suggestions,
about spliting the file vmi_multus_test.go, I prefer doing that in another PR
with converting those networkAttachmentDefinitions json strings.

tests/vmi_multus_test.go
```diff
@@ -592,6 +591,58 @@ var _ = Describe("SRIOV", func() {
 		sriovResourceName = "openshift.io/sriov_net"
 	}
 
+	createSriovNetwork := func(name, namespace, resourceName, networkNamespace  string, vlan int, trust, spoofCheck, linkState bool) error {
+		spoofCheckString := func (trust bool) string {
+			if ! spoofCheck {
+				return "off"
```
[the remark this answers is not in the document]
ormergi: done

tests/vmi_multus_test.go
```diff
@@ -592,6 +591,58 @@ var _ = Describe("SRIOV", func() {
...
+		sriovNetwork :=	sriovnetv1.SriovNetwork{
+				ObjectMeta: v13.ObjectMeta{
+					Name: name,
+					Namespace: namespace,
+				},
+				Spec:       sriovnetv1.SriovNetworkSpec{
+					NetworkNamespace: networkNamespace,
+					ResourceName:     resourceName,
+					Capabilities:     "",
+					IPAM:             "{ \"type\": \"host-local\", \"subnet\": \"10.1.1.0/24\" }",
```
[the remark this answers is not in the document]
ormergi: SriovNetworkSpec.PAM is string type so I had to do it that way.

tests/vmi_multus_test.go
ormergi: In go you cannot pass `nil` for `int` value, nor pass only few parameters so I had to do it that way I think it is good enough for this test :)

tests/vmi_multus_test.go
ormergi: done

tests/vmi_multus_test.go
ormergi: I pushed new commits to address that, now this changes is in the right place.

tests/vmi_multus_test.go
ormergi: I addressed this in a new commit

tests/vmi_multus_test.go
```diff
@@ -592,6 +591,58 @@ var _ = Describe("SRIOV", func() {
...
+
+		trustString := func (spoofCheck bool) string {
+			if trust {
+				return "on"
+			}
+
+			return "off"
+		}
+
+		linkStateString := func (linkState bool) string {
```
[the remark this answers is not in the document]
ormergi: SriovIPAM is string type,

> Everything in this [39ff905](https://github.com/kubevirt/kubevirt/commit/39ff9054c0ddda3dcb717f651ec06adedfc5d9a6) commit is auto-generated?
> if not, can you point to the files that have been changed by you?
> 
> can you place all this information in the commit message?

I pushed new commits and now all generated changes are in a separate commit

ormergi:
It is not possible to vendor and use SriovNetwork API until #3420  is merged

EdDev:
> I pushed new commits I tried to address most of your suggestions

Thank you.
The commit message still needs a bit more convincing. Try to explain what this library you added is, where it comes from and why it is better than the direct json approach.
It does have one noticeable disadvantage of depending on a large number of dependencies.

> about spliting the file vmi_multus_test.go, I prefer doing that in another PR
> with converting those networkAttachmentDefinitions json strings.

I suggest you do that then (create a PR and move it out) and then rebase this one over it.

> It is not possible to vendor and use SriovNetwork API until #3420 is merged

Why?
Please mention this as a special note to the reviewer so it will be clear that this is blocked by that PR.

ormergi:
/hold
Until go version in builder container is 1.13+ to align with SriovNetwork dependecies

tests/vmi_multus_test.go
```diff
@@ -587,24 +645,15 @@ var _ = Describe("SRIOV", func() {
...
-			Post().
-			RequestURI(fmt.Sprintf(postUrl, tests.NamespaceTestDefault, "sriov-link-enabled")).
-			Body([]byte(fmt.Sprintf(sriovLinkEnableConfCRD, "sriov-link-enabled", tests.NamespaceTestDefault, sriovResourceName))).
-			Do()
-		Expect(result.Error()).NotTo(HaveOccurred())
+		ipam := "{ \"type\": \"host-local\", \"subnet\": \"10.1.1.0/24\" }"
+		err := createSriovNetwork("sriov", tests.NamespaceTestDefault, "sriov_net", tests.NamespaceTestDefault, ipam, 0, false, true, false)
+		Expect(err).NotTo(HaveOccurred(), "should create SriovNetwork sriov")
+
+		err = createSriovNetwork("sriov2", "sriov_net", tests.NamespaceTestDefault, tests.NamespaceTestDefault, ipam, 0, false, true, false)
```
alonSadan, line 652: nit: 
maybe you can use constant bool (`const trusted  = true`). I think it will make it easier to understand what configurations
are given to the SRIOV-network.

tests/vmi_multus_test.go
```diff
@@ -587,24 +645,15 @@ var _ = Describe("SRIOV", func() {
...
+		Expect(err).NotTo(HaveOccurred(), "should create SriovNetwork sriov2")
+
+		err = createSriovNetwork("sriov-link-enabled", "sriov_net", tests.NamespaceTestAlternative, tests.NamespaceTestDefault, ipam, 0, false, true, true)
```
alonSadan, line 655: same...

PR #3557, opened 2020-06-14
Skip an extremely flakey migration test

rmohr:

Test
```
```
is blocking merges. See also the flakefinder history:

https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2020-06-14-168h.html#row0

It recently started to fail almost 100% of time, but was flaky before already. There may be a real issue in the config propagation code. https://github.com/kubevirt/kubevirt/pull/3542 tries to consistently wait until the config is distributed, but it still fails on the test.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -1053,7 +1053,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 			AfterEach(func() {
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
-			It("[test_id:2227]should abort a vmi migration without progress", func() {
+			PIt("[test_id:2227] [flaky] should abort a vmi migration without progress", func() {
 				tests.SkipStressTestIfRunnigOnKindInfra()
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
 				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse("1Gi")


rmohr:
@phoracek @danielBelenky I would suggest to merge this asap to unblock others. There is the config propagation PR where the test can later be enabled again.

EdDev:
> @phoracek @danielBelenky I would suggest to merge this asap to unblock others. There is the config propagation PR where the test can later be enabled again.

Can we at least specify/reference the issue/bug ID so we will be able to track this?
It seems that sometime `Skip` is used. Is there one way to xfail such tests?

maya-r:
using xfail is problematic here, it passes on some lanes

EdDev:
Can you please check https://github.com/kubevirt/kubevirt/pull/3567 ?

jean-edouard:
/lgtm
This test is flaky by nature, as it hopes that the migration of a VM with stressed memory will show 0 progress for 5 seconds.
A better test would be to drop packets from one virt-launcher to the other, guaranteeing 0 progresss. I will look at writing that.
In the meantime, we need to merge this PR asap to unblock others.

EdDev:
> /lgtm
> This test is flaky by nature, as it hopes that the migration of a VM with stressed memory will show 0 progress for 5 seconds.

It was never set to 5sec, it used the default (30s?) because the set was buggy.

> A better test would be to drop packets from one virt-launcher to the other, guaranteeing 0 progresss. I will look at writing that.
> In the meantime, we need to merge this PR asap to unblock others.

What is the difference between dropping packets and touching the memory at high rate so it will not reconcile fast? The reconcile of the deltas would have use a timeout as well, and the 5sec is much less than that.

The test itself was here for years, it was flaky at the default timeouts.

EdDev:
I find it odd that an alternative has not been examined before you take such an extreme action.

PR #3567, opened 2020-06-15
Stabilize functest vmi migration stress

### LINKED ISSUE #3323: Test 2227 , IPv6 lane failing, because the stress test isnt sufficent in order to failing the migration
/kind bug

**What happened**:
running test 2227 on ipv6 lane
the test tries to stress by dirty memory pages, 
and expect the migration to fail, which doesnt

i suspect its because its kind and not a provider like k8s with vm nodes.
tried to extend the stressing time to 60 seconds, and the 800M to 900M
it didnt help

migration should fail and test pass

run test 2227 on prow with all other ipv6 tests as well (on its own it passes)

**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

### LINKED ISSUE #3560: [flakey test] VM Live Migration Starting a VirtualMachineInstance migration monitor [test_id:2227]should abort a vmi migration without progress

/kind bug

**What happened**:

Test
```
```
is blocking merges. See also the flakefinder history:

https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2020-06-14-168h.html#row0

It recently started to fail almost 100% of time, but was flaky before already. There may be a real issue in the config propagation code. https://github.com/kubevirt/kubevirt/pull/3542 tries to consistently wait until the config is distributed, but it still fails on the test.

This test was marked as pending in https://github.com/kubevirt/kubevirt/pull/3557.

Issue of this test should be investigated and fixed.

Remove the "skip" and run the test.

EdDev:

Fixing flaky stress migration test.

tests, migration: Fix the cluster config update in the stress test

The integers used as values for the migration should have been strings
and not integers.
This change in now using json marshaling (to make sure there are no
problems there).

Error seen in logs:
```
{"component":"virt-controller","level":"error","msg":"Invalid cluster config using 'ConfigMap' resource version '37083', falling back to last good resource version '37042'","pos":"config-map.go:437","reason":"failed to parse migration config: json: invalid use of ,string struct tag, trying to unmarshal unquoted value into *int64","service":"http","timestamp":"2020-06-15T10:07:52.194163Z"}
```

---

tests, migration: Fix run stress expecter
tests, migration: Assert phase directly

Fixes #3560, fixes #3323 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -326,7 +326,10 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 	runStressTest := func(expecter expect.Expecter) {
 		By("Run a stress test to dirty some pages and slow down the migration")
 		_, err = expecter.ExpectBatch([]expect.Batcher{
-			&expect.BSnd{S: "stress --vm 1 --vm-bytes 800M --vm-keep --timeout 1600s&\n"},
+			&expect.BSnd{S: "\n"},
+			&expect.BExp{R: "\\#"},
+			&expect.BSnd{S: "stress --vm 1 --vm-bytes 800M --vm-keep --timeout 1600s&\n\n"},
+			&expect.BExp{R: "\\#"},
 		}, 15*time.Second)
 		Expect(err).ToNot(HaveOccurred(), "should run a stress test")
 		// give stress tool some time to trash more memory pages before returning control to next steps
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -309,19 +309,17 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 		By("Waiting until the Migration Completes")
 
 		uid := ""
-		Eventually(func() bool {
+		Eventually(func() v1.VirtualMachineInstanceMigrationPhase {
 			migration, err := virtClient.VirtualMachineInstanceMigration(migration.Namespace).Get(migration.Name, &metav1.GetOptions{})
 			Expect(err).ToNot(HaveOccurred())
 
-			Expect(migration.Status.Phase).NotTo(Equal(v1.MigrationSucceeded))
+			phase := migration.Status.Phase
+			Expect(phase).NotTo(Equal(v1.MigrationSucceeded))
 
 			uid = string(migration.UID)
-			if migration.Status.Phase == v1.MigrationFailed {
-				return true
-			}
-			return false
+			return phase
 
-		}, timeout, 1*time.Second).Should(BeTrue())
+		}, timeout, 1*time.Second).Should(Equal(v1.MigrationFailed))
 		return uid
 	}
 
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -1062,7 +1062,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
 			PIt("[test_id:2227] should abort a vmi migration without progress", func() {
-				tests.SkipStressTestIfRunnigOnKindInfra()
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
 				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse("1Gi")
 
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -1049,12 +1049,19 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				cfgMap, err = virtClient.CoreV1().ConfigMaps(tests.KubeVirtInstallNamespace).Get(virtconfig.ConfigMapName, options)
 				Expect(err).ToNot(HaveOccurred())
 				originalMigrationConfig = cfgMap.Data["migrations"]
-				tests.UpdateClusterConfigValueAndWait("migrations", `{"progressTimeout" : 5, "completionTimeoutPerGiB": 5}`)
+
+				data := map[string]string{
+					"progressTimeout":         "5",
+					"completionTimeoutPerGiB": "5",
+				}
+				migrationData, err := json.Marshal(data)
+				Expect(err).ToNot(HaveOccurred())
+				tests.UpdateClusterConfigValueAndWait("migrations", string(migrationData))
 			})
 			AfterEach(func() {
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
-			PIt("[test_id:2227] [flaky] should abort a vmi migration without progress", func() {
+			PIt("[test_id:2227] should abort a vmi migration without progress", func() {
 				tests.SkipStressTestIfRunnigOnKindInfra()
 				vmi := tests.NewRandomFedoraVMIWitGuestAgent()
 				vmi.Spec.Domain.Resources.Requests[k8sv1.ResourceMemory] = resource.MustParse("1Gi")
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -1053,6 +1053,7 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				data := map[string]string{
 					"progressTimeout":         "5",
 					"completionTimeoutPerGiB": "5",
+					"bandwidthPerMigration":   "1Mi",
 				}
 				migrationData, err := json.Marshal(data)
 				Expect(err).ToNot(HaveOccurred())
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -4747,12 +4747,6 @@ func IsRunningOnKindInfraIPv6() bool {
 	return strings.HasPrefix(provider, "kind-k8s-1.17.0-ipv6")
 }
 
-func SkipStressTestIfRunnigOnKindInfra() {
-	if IsRunningOnKindInfra() {
-		Skip("Skip stress test till issue https://github.com/kubevirt/kubevirt/issues/3323 is fixed")
-	}
-}
-
 func SkipPVCTestIfRunnigOnKindInfra() {
 	if IsRunningOnKindInfra() {
 		Skip("Skip PVC tests till PR https://github.com/kubevirt/kubevirt/pull/3171 is merged")


EdDev:
Should fix #3560

tests/migration_test.go
```diff
@@ -309,26 +309,27 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
...
 
 		uid := ""
-		Eventually(func() bool {
+		Eventually(func() v1.VirtualMachineInstanceMigrationPhase {
 			migration, err := virtClient.VirtualMachineInstanceMigration(migration.Namespace).Get(migration.Name, &metav1.GetOptions{})
 			Expect(err).ToNot(HaveOccurred())
 
-			Expect(migration.Status.Phase).NotTo(Equal(v1.MigrationSucceeded))
+			phase := migration.Status.Phase
+			Expect(phase).NotTo(Equal(v1.MigrationSucceeded))
```
oshoval, line 317: why do we need this , if the Eventually already checks its equal to Failed ?

tests/migration_test.go
```diff
@@ -309,26 +309,27 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
...
 		return uid
 	}
 
 	runStressTest := func(expecter expect.Expecter) {
 		By("Run a stress test to dirty some pages and slow down the migration")
 		_, err = expecter.ExpectBatch([]expect.Batcher{
-			&expect.BSnd{S: "stress --vm 1 --vm-bytes 800M --vm-keep --timeout 1600s&\n"},
+			&expect.BSnd{S: "\n"},
+			&expect.BExp{R: "\\#"},
+			&expect.BSnd{S: "stress --vm 1 --vm-bytes 800M --vm-keep --timeout 1600s&\n\n"},
```
oshoval, line 331: why double \n ?

tests/migration_test.go
EdDev: Nothing changed from the original.
It is used to fail earlier (and not retry) in case the migration succeeds.

tests/migration_test.go
EdDev: Because it is running in the background, try it locally and see one is not enough.

oshoval:
Nice, thanks
/lgtm

EdDev:
> /hold this PR doesn't really address the root cause of the flakiness of this test. Calling "stress" doesn't actually guarantee that the migration will stall.

Sorry, I do not understand your point. The cluster config update was not set correctly, therefore the default migration timeout took affect and not the low values.
The call to "stress" causes the migration to take longer, that was the original point I guess.

vladikr:
Overall it looks fine to me. The main issue is that it used to work in the past even without these changes. 
I think what has changed is that we moved to a new Fedora version.




> /hold this PR doesn't really address the root cause of the flakiness of this test. Calling "stress" doesn't actually guarantee that the migration will stall.

@stu-gott it should.

EdDev:
> Overall it looks fine to me. The main issue is that it used to work in the past even without these changes.
> I think what has changed is that we moved to a new Fedora version.

Probably it worked because with the default timeouts it still had hard time to pass. When set to 5sec, the chances for it to pass are slim.
I think there is an upper-limit to the time the deltas are left to converge, afterward it just pauses and migrates. Maybe the two default timeouts are close, that is why it is flaky.

BTW, the memory size was increased in the passes, and it helped for some time.

vladikr:
> > 
> 
> Probably it worked because with the default timeouts it still had hard time to pass. When set to 5sec, the chances for it to pass are slim.
> I think there is an upper-limit to the time the deltas are left to converge, afterward it just pauses and migrates. Maybe the two default timeouts are close, that is why it is flaky.
Without stress, it will easily pass, that's why I originally added this.

> 
> BTW, the memory size was increased in the past, and it helped for some time.
Yes, but in some cases, stress will get killed by oom.

I would suggest reducing the bandwidth as well to as low as possible. {"bandwidthPerMigration" : "1Mi"}
As @jean-edouard suggested on Friday.

vladikr:
Just to add. Even if we will fix this right now, we will need #3542 to address the flaky nature of this test.

jean-edouard:
> > > 
> > 
> > 
> > Probably it worked because with the default timeouts it still had hard time to pass. When set to 5sec, the chances for it to pass are slim.
> > I think there is an upper-limit to the time the deltas are left to converge, afterward it just pauses and migrates. Maybe the two default timeouts are close, that is why it is flaky.
> > Without stress, it will easily pass, that's why I originally added this.
> 
> > BTW, the memory size was increased in the past, and it helped for some time.
> > Yes, but in some cases, stress will get killed by oom.
> 
> I would suggest reducing the bandwidth as well to as low as possible. {"bandwidthPerMigration" : "1Mi"}
> As @jean-edouard suggested on Friday.

A lower bandwidth would more or less guarantee a migration failure, but that's not exactly what the test is about.
I bet the reason the test used to usually succeed (i.e. the migration failed) wasn't because of "no progress" as suggested by the name, but rather because there wasn't enough time overall for the whole migration to finish.
To properly test this, we need to actually guarantee a complete stall the migration process so that it gets killed accordingly, even if there's still time on the clock for it to finish.
That, in my opinion, required completely re-designing the test, which is something I'll be looking at.

EdDev:
> I would suggest reducing the bandwidth as well to as low as possible. {"bandwidthPerMigration" : "1Mi"}

Sounds right to me. It cannot hurt for sure.

> Just to add. Even if we will fix this right now, we will need #3542 to address the flaky nature of this test.

I agree that confirming about the config distribution is a right action to take, much better than waiting using a `sleep`. IMO this change is enough to fix 99% of the problem and should have been considered over taking out the test completely.

vladikr:
No harm is done. I agree that we need to skip this for now to unblock other important PRs.
Let's prepare this fix and get it in.

stu-gott:
/hold cancel

sounds like we have an argeement on a path forward

EdDev:
I find it hard to decode the plan here.
One wants to redesign the test so it will not fail on timeout but rather on it being stall (sounds the same to me), one suggests to just add another limitation on the bandwidth rate and another PR wants to confirm and block on the configuration distribution.

I could update this change to include the bw limit and change the test name to reflect it tests a migration timeout. Let me know if this is what you are looking for here.

jean-edouard:
> I find it hard to decode the plan here.

+1

> One wants to redesign the test so it will not fail on timeout but rather on it being stall (sounds the same to me), one suggests to just add another limitation on the bandwidth rate and another PR wants to confirm and block on the configuration distribution.

Test timeout and stall are very different.
It is my understanding that we have a system that stops the migration when it sees 0 progress for a given amount of time (`progressTimeout`).
So a migration with a timeout of 10 minutes and a "progress timeout" of 10 seconds will be killed after 10 seconds of stall, not after the whole 10 minutes of doing nothing.
Going by the title of the test and discussions with the author, the stall is what `[test_id:2227]should abort a vmi migration without progress` verifies.

So there really should be 2 tests there.

> I could update this change to include the bw limit and change the test name to reflect it tests a migration timeout. Let me know if this is what you are looking for here.

I think it would make more sense to create a new test instead of changing the title of the broken one, just so we don't forget to fix it later.
Also a test_id has already been established...
Anyway, +1 for a migration timeout test!

EdDev:
The `progressTimeout` implementation does not check that 0 bytes have been migrated for the specified time, it checks between intervals that the remaining data to migrate is less than what was calculated before (i.e. it is monotonic and converges). But the `stress` utility is touching memory and I suspect the needed data to migrate may increase in mid migration which eventually can fail the test.

The current test has a problem though, it is unclear which timeout causes the migration to fail.

rmohr:
> I find it hard to decode the plan here.

Looks to me like we potentially have to address multiple points to really stabilize the things here.
 1. What @EdDev fixes here: Obviously the config is not picked up by the tests, due to the fact that the test passes an invalid config. It is unclear if this alone fixes the test, but it needs to be fixed. So let's get this one in.
 2. Like in #3542, we need to wait for the config to be propagated. This resolves the issue that a component may not  see the right config as well as the issue that the config may be invalid, since the test would then fail on the invalid config in this PR. My PR already shows at least a second test where we pass in an invalid config
 3. We need to check what @jean-edouard is talking about, that it is not 100% clear for us if we trigger the stall-detection always by just setting this config. 

So @EdDev solves (1), I can solve (2) and @jean-edouard can solve or verify (3).

Sounds good?

tests/migration_test.go
rmohr: I can't confirm that this is needed. This may depend on the shell in use. I am not sure I saw the test ever failing there, but worst case it makes the test more robust.

tests/migration_test.go
```diff
@@ -1048,7 +1049,14 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 				cfgMap, err = virtClient.CoreV1().ConfigMaps(tests.KubeVirtInstallNamespace).Get(virtconfig.ConfigMapName, options)
 				Expect(err).ToNot(HaveOccurred())
 				originalMigrationConfig = cfgMap.Data["migrations"]
-				tests.UpdateClusterConfigValueAndWait("migrations", `{"progressTimeout" : 5, "completionTimeoutPerGiB": 5}`)
+
+				data := map[string]string{
```
rmohr, line 1053: This fixes also config propagation issues in 

```
[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system] VM Live Migration Starting a VirtualMachineInstance with auto converge enabled [test_id:3237]should complete a migration
```

:+1:

vladikr:
> The `progressTimeout` implementation does not check that 0 bytes have been migrated for the specified time, it checks between intervals that the remaining data to migrate is less than what was calculated before (i.e. it is monotonic and converges). But the `stress` utility is touching memory and I suspect the needed data to migrate may increase in mid migration which eventually can fail the test.

`stress` would also increase the cpu consumption in the guest and together with a high rate of memory updates it would stall the progress.

> 
> The current test has a problem though, it is unclear which timeout causes the migration to fail.

Right, this part needs to be improved. I thought we are hitting the `progressTimeout` but at this point, the goal is to demonstrate that the migration monitor cancels the migration when it's not making a progress or didn't complete in a given time.

I simply wanted you to add ` {"bandwidthPerMigration" : "1Mi"}` to the config.

vladikr:
I'm fine with merging this as it is if it improves things.

EdDev:
> I simply wanted you to add ` {"bandwidthPerMigration" : "1Mi"}` to the config.

I already added that one.
I need to push an updated version, it also removed the skip from the kind lane as it is working there as well now.

EdDev:
I changed the test name, not sure if it is better or not.
I can take it out if you prefer the old one.

vladikr:
> I already added that one.
> I need to push an updated version, it also removed the skip from the kind lane as it is working there as well now.

Great! Thank you!

tests/migration_test.go
```diff
@@ -1048,13 +1049,20 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
...
+					"progressTimeout":         "5",
```
rmohr, line 1054: I think that this is a regression which happend due the flakiness of the test already before. Look for instance at https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3468/pull-kubevirt-e2e-k8s-1.17/1267910401970933760. The virt-handler logs don't say anything about config errors.

tests/migration_test.go
```diff
@@ -1048,13 +1049,20 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
...
+					"completionTimeoutPerGiB": "5",
+					"bandwidthPerMigration":   "1Mi",
```
rmohr, line 1056: I tried that before the test was still flaky.

EdDev:
> I don't think that we can re-enable the test right now. See my comments. My suspicion is that the test would only be less flaky again. Can we merge this without enabling the test again?

Ack.
Then I hope you can re-enable it with the other change.

> 
> Just to illustrate what I had to do to make the tests locally fail in the past where the config could still be loaded correctly:
> 
>     * Setting the test to FMeasure where I repeat it 30 times in a row
> 
>     * Maybe even do that a second time.
> 
> 
> I did never manage to run it locally 60 times in a row. Due do more resource competition, the flake appears more often on CI.

Fair enough, the need to wait for the config to be propagated is real.

tests/migration_test.go
```diff
@@ -1048,13 +1049,20 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
...
+				migrationData, err := json.Marshal(data)
+				Expect(err).ToNot(HaveOccurred())
+				tests.UpdateClusterConfigValueAndWait("migrations", string(migrationData))
 			})
 			AfterEach(func() {
 				tests.UpdateClusterConfigValueAndWait("migrations", originalMigrationConfig)
 			})
-			PIt("[test_id:2227] [flaky] should abort a vmi migration without progress", func() {
-				tests.SkipStressTestIfRunnigOnKindInfra()
+			PIt("[test_id:2227] should abort a vmi migration on timeout", func() {
```
rmohr, line 1065: Not sure the new name is better. I mean yes we are waiting for a timeout, but the timeout is not a migration-lifetime timeout. The timeout is related to the seeing progress on the migration.

oshoval: we still have PIt ?

EdDev: I am fine dropping the name change (I asked about it when posting it).
Please confirm.

rmohr: Yes. As described above we should enable it with #3542, since even before we had the invalid-config issue it was very flaky. With the issue which @EdDev is fixing it will go back to that state. So let's fix the flake before re-enabling it.

rmohr: > I am fine dropping the name change (I asked about it when posting it).
> Please confirm.

Since you have to rebase anyway, please. It is slightly less confusing.

EdDev:
> @EdDev could you rebase? Will take this in manually once I see a green lane afterwards to make progress on the test stabilization in general.

Done

tests/migration_test.go
EdDev: ack

rmohr:
Have good enough test results now for merging.

rmohr:
/cherry-pick release-0.30

Needed as part of this backport: #3761.

PR #3568, opened 2020-06-15
Expose guest swap metrics

xpivarc:
Currently, we have `kubevirt_vmi_memory_swap_traffic_bytes_total` metric but we actually never collect this metric and therefore never expose. This metric is helpful to identify VMs which are swapping and are good candidates for increasing resources(or identify performance problems,...).

This PR is enabling memballoon device by default(Note, we don't enable ballooning). Pooling interval defaults to 10(inspired by https://specs.openstack.org/openstack/nova-specs/specs/juno/implemented/enabled-qemu-memballoon-stats.html) and model to virtio. Virt-handler will now propagate guest swap metrics of VMs through Prometheus endpoint

Fixes #

**Release note**:

```release-note
Guest swap metrics available
```


vladikr:
@xpivarc Perhaps you can add more context to the description. Just briefly explain what are we trying to achieve here.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -588,6 +588,22 @@ func Convert_v1_FeatureHyperv_To_api_FeatureHyperv(source *v1.FeatureHyperv, hyp
 	return nil
 }
 
+// If Period is lower or equal 0 than we don't add ballooning device. Defaulting to 10 seconds pooling.
```
vladikr, line 591: This point isn't clear to me. Specifically, why do we want to silently drop the configuration instead of rejecting the spec?
I think, creating a validating-webhook and rejecting the spec with an error message will make this less unexpected for the users.

Also, please add a few words wrt why values <10 aren't accepted.

pkg/virt-launcher/virtwrap/api/schema.go
```diff
@@ -759,8 +759,14 @@ type Address struct {
 
 //END Video -------------------
 
+type Stats struct {
+	Period int `xml:"period,attr"`
+}
+
 type Ballooning struct {
```
vladikr, line 766: Perhaps this is an opportunity to change the device name to a MemBalloon, but I don't mind. :)  What do you think @rmohr ?

pkg/virt-launcher/virtwrap/api/schema.go
```diff
@@ -759,8 +759,14 @@ type Address struct {
 
 //END Video -------------------
 
+type Stats struct {
+	Period int `xml:"period,attr"`
```
rmohr, line 763: I am not sure if we want to make this configurable by users.
@andreabolognani do you know about sources which wound indicate reasonable values for it? Would be great if you could share some insights.

andreabolognani: Sorry, I don't have any specific insights into what period could be a good default. The obvious answer would be to look at what OpenStack and oVirt are doing, though I'm not sure you'd find that a very satisfying one :)

I think I share your opinion that this is not necessarily something that you need to expose to users, at least not right away: if the goal is to have ballooning stats available, then picking a reasonable, static default might the the way to go.

pkg/virt-launcher/virtwrap/api/converter.go
andreabolognani: I'm not too familiar with KubeVirt's API design so apologies if this is out of line, but conflating "period is zero" with "no balloon should be present" doesn't sound like a good idea to me.

I can see "period is zero" mapping quite naturally to "disable stats", but disabling the balloon device altogether should look more along the lines of

```yaml
memBalloon:
  present: false
```

in my opinion.

pkg/virt-launcher/virtwrap/api/schema.go
andreabolognani: Since these structures are IIUC intended to map pretty much 1:1 to libvirt XML elements, and libvirt uses `<memballon>` for this device, I think renaming it would make a lot of sense.

pkg/virt-launcher/virtwrap/api/converter.go
xpivarc: This PR is just enabling the collection of stats and by default, we are disabling the balloon device, right now. Therefore 0 means disabling device... I agree  `present: false` looks nicer to me.

@vladikr I suggest to not reject it as user might don't want to have memballon device. Values <10 are accepted, just <0 values means users don't want to collect the metrics.

pkg/virt-launcher/virtwrap/api/schema.go
xpivarc: I will include this in my PR.

pkg/virt-launcher/virtwrap/api/schema.go
xpivarc: Just note we are exposing this https://github.com/kubevirt/kubevirt/pull/3568/files#diff-3bfc6e1b38439eeb33078da0a69ef239R378. I take period=10 as an example from OpenStack. (Couldn't find better...) In my understanding, if VM will be swapping for 9sec and we collect it on the 10th we will not capture this. Therefore I suggested to expose it as users can find a suitable value for their workloads.

Also, note that we collect metrics only when the Prometheus registry does... (so you can miss swapping entirely) but I might be wrong here...

xpivarc:
> Looks like a solid PR! Thanks @xpivarc
> I have a few questions in line.
> I'm also wondering how do we use these stats?
> Do we provide it via the guest agent? If yes, we should probably provide a functional test to demonstrate this.

We provide these metrics through Prometheus. I didn't see that we have tests on metrics but I can certainly add one.

pkg/virt-launcher/virtwrap/api/converter.go
vladikr: @xpivarc ok, this is reasonable, but then I wonder why do we want to expose these low level settings to the user at all? Can we just do the right thing? Alternatively, use a cluster wide setting?

ArthurSens:
Hey, let me intrude in this PR with a question...
I was trying to expose some new memory metrics [here](https://github.com/ArthurSens/kubevirt/commit/4a9f3190a49f82a841cfde3c7c92130eadda4126) and I had the same problem you describe in this PR, none of them were ever exposed.

After I read this PR, I googled what was "memory ballooning", and now I'm wondering if that's why the metrics I added aren't being exposed

Do you guys think that the lack of memory ballooning is indeed the issue I'm facing?

xpivarc:
> Hey, let me intrude in this PR with a question...
> I was trying to expose some new memory metrics [here](https://github.com/ArthurSens/kubevirt/commit/4a9f3190a49f82a841cfde3c7c92130eadda4126) and I had the same problem you describe in this PR, none of them were ever exposed.
> 
> After I read this PR, I googled what was "memory ballooning", and now I'm wondering if that's why the metrics I added aren't being exposed
> 
> Do you guys think that the lack of memory ballooning is indeed the issue I'm facing?

Yes, these stats are reported only when memBalloon device is available. I am not sure about NR.

pkg/virt-launcher/virtwrap/api/converter.go
xpivarc: https://github.com/kubevirt/kubevirt/pull/3568#discussion_r442108913

vladikr: Ok, I see. I probably wouldn't expose such a low level tunable to on the API, but I'm not against it. On the other hand, we shouldn't make the device appear or disappear unexpectedly.
Let's just reject values <0 and introduce an explicit `present: false` as @andreabolognani  suggested.

fabiand: actually, we have examples of how to not add a device:
`VMI.spec.domain.devices.autoAttachGraphics: false` for example will opt-out of adding a graphics adapter (I'm not 100% sure on the field name). We have additional fields opting out of other features as well - see autoAttach* in https://kubevirt.io/api-reference/master/definitions.html#_v1_devices

Thus here we could add `VMI.spec.domain.devices.auttoattachMemoryBallooning: false` to opt out.
WDYT?

andreabolognani: I was not aware there was prior art for this kind of behavior... I agree that something like `autoattachMemoryBalloon` makes the most sense.

xpivarc: Thanks, Fabian for suggestion! I revisited how we actually expose metrics. We collect metrics from virt-launcher only when /metrics endpoint of virt-handler is being scraped(which is cluster-wide setting in Prometheus). Therefore I am going with cluster-wide setting of period in virt-config. Also will add `autoattachMemoryBalloon` as it now makes the most sense(we don't need to have MemBalloon struct on API).

pkg/virt-config/config-map.go
```diff
@@ -196,6 +197,7 @@ func defaultClusterConfig() *v1.KubeVirtConfiguration {
 		SELinuxLauncherType:         DefaultSELinuxLauncherType,
 		SupportedGuestAgentVersions: supportedQEMUGuestAgentVersions,
 		OVMFPath:                    DefaultOVMFPath,
+		MemBalloonPeriod:            DefaultMemBallonPeriod,
```
andreabolognani, line 200: I think this should be called `MemBalloonStatsPeriod` or something like that, because the period doesn't apply to the balloon itself but just to stats collection.

Additionally, shouldn't there be a way to document the configuration knob? For example its intended use, its default value, and the acceptable range, including the fact that a value of zero turns off stats collection.

pkg/virt-launcher/virtwrap/api/converter_test.go
```diff
@@ -565,7 +565,21 @@ var _ = Describe("Converter", func() {
...
   <iothreads>3</iothreads>
-</domain>`, domainType)
+</domain>`, domainType, "%s")
+		var convertedDomainWith5Period = fmt.Sprintf(convertedDomain,
+			`<memballoon model="virtio">
+      <stats period="5"></stats>
+    </memballoon>`)
+		var convertedDomainWith0Period = fmt.Sprintf(convertedDomain,
+			`<memballoon model="virtio">
+      <stats period="0"></stats>
```
andreabolognani, line 575: According to the [schema for `<memballoon>`](https://gitlab.com/libvirt/libvirt/-/blob/master/docs/schemas/domaincommon.rng#L4460), the period attribute takes a positiveInteger, that is, a value which is strictly greater than 0: disabling stats collection is achieved by *not having* the `<stats>` element at all.

libvirt will still accept a period attribute equal to zero and behave as if the `<stats>` element were not present in that case, but it would be better if it was not there in the first place.

staging/src/kubevirt.io/client-go/api/v1/schema.go
```diff
@@ -355,6 +355,11 @@ type Devices struct {
 	// Whether to attach the default serial console or not.
 	// Serial console access will not be available if set to false. Defaults to true.
 	AutoattachSerialConsole *bool `json:"autoattachSerialConsole,omitempty"`
+	// Whether to attach the Memory balloon device with default 10s period.
+	// Period can be also set in virt-config.
```
andreabolognani, line 359: I would not mention the default period here, but I would instead include the exact name of the ConfigMap attribute that can be used to tweak it.

pkg/virt-launcher/virtwrap/api/converter_test.go
xpivarc: Right, I will align that.

pkg/virt-config/config-map.go
xpivarc: I think it should. I will document this in user-guide repo.
We could even do better by declining changes to virt-config while the user is editing it.

andreabolognani: You're still using `MemBalloonPeriod` instead of `MemBalloonStatsPeriod` or something along those lines, but it's not clear whether you disagree with my suggestion or you simply missed it...

xpivarc: I missed it(was too late for me).

andreabolognani:
Looks good to me now, at least on the surface/API level. I leave an actual review up to someone with more KubeVirt expertise :)

pkg/handler-launcher-com/cmd/v1/cmd.pb.go
```diff
@@ -911,50 +919,52 @@ var _Cmd_serviceDesc = grpc.ServiceDesc{
...
-	0x06, 0x1d, 0x9a, 0xe6, 0xa0, 0x41, 0x01, 0x74, 0x07, 0xd6, 0xda, 0x54, 0xa6, 0xb9, 0xea, 0x5a,
-	0x5e, 0x30, 0xa7, 0xc8, 0x70, 0x5b, 0x30, 0xd7, 0xa4, 0x32, 0xb6, 0x54, 0xb4, 0x9e, 0xa9, 0xbc,
-	0xfb, 0xcf, 0x61, 0x6d, 0x23, 0x93, 0x4e, 0x7b, 0xbd, 0x3a, 0xab, 0x85, 0x01, 0x4e, 0x19, 0xe8,
-	0x38, 0xe6, 0xa3, 0x11, 0xcc, 0x94, 0xbd, 0xe3, 0x12, 0x6a, 0x43, 0xad, 0x49, 0xe5, 0xc0, 0x8a,
-	0xc7, 0x61, 0x71, 0x26, 0x9d, 0x71, 0x71, 0x05, 0xad, 0x36, 0xa9, 0xb2, 0xbc, 0xb1, 0x3a, 0xb7,
-	0xf2, 0x81, 0x19, 0xbb, 0x2c, 0xa1, 0xcf, 0x6a, 0x04, 0x77, 0xac, 0x6b, 0x1c, 0x7a, 0x3b, 0x1f,
-	0x9d, 0x63, 0x7e, 0xb8, 0x84, 0x0e, 0x60, 0xfa, 0x9c, 0x71, 0x6b, 0x1c, 0xf3, 0x6f, 0x67, 0x7e,
-	0x30, 0xfd, 0x69, 0x2a, 0xdc, 0xbb, 0x9a, 0x51, 0x3f, 0xfa, 0x9e, 0xfe, 0x09, 0x00, 0x00, 0xff,
-	0xff, 0x67, 0xbf, 0xe6, 0x8c, 0x21, 0x0a, 0x00, 0x00,
+	// 738 bytes of a gzipped FileDescriptorProto
```
jean-edouard, line 922: Could you please explain what this blob is and how it is generated?
Sorry I know you didn't introduce it, but the comment is really non-descriptive...

tests/vmi_configuration_test.go
```diff
@@ -486,7 +486,7 @@ var _ = Describe("Configurations", func() {
 
 				res, err := expecter.ExpectBatch([]expect.Batcher{
 					&expect.BSnd{S: "free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2\n"},
-					&expect.BExp{R: "105"},
+					&expect.BExp{R: "104"},
```
jean-edouard, line 489: This is just looking for trouble... What is this even testing?
Sorry, again, not your doing, but I don't believe RAM usage in Linux is deterministic.
Could we maybe switch to a reasonable range to limit potential flakyness?

pkg/handler-launcher-com/cmd/v1/cmd.pb.go
xpivarc: This is auto-generated from cmd.proto.

xpivarc: This is auto-generated from cmd.proto. Nothing we can do about.

tests/vmi_configuration_test.go
xpivarc: It `grep`s total memory. That's deterministic or? The reason why this dropped is memballoon which is added by default.

jean-edouard: My bad, assumed it was used memory. Thanks!

xpivarc: Comment on top of the command wouldn't hurt me either :)

pkg/virt-config/config-map.go
```diff
@@ -196,6 +197,7 @@ func defaultClusterConfig() *v1.KubeVirtConfiguration {
 		SELinuxLauncherType:         DefaultSELinuxLauncherType,
 		SupportedGuestAgentVersions: supportedQEMUGuestAgentVersions,
 		OVMFPath:                    DefaultOVMFPath,
+		MemBalloonStatsPeriod:       DefaultMemBallonStatsPeriod,
```
kbidarkar, line 200: Do we want, DefaultMemBallo**o**nStatsPeriod   here?

andreabolognani: Yup, nice catch :)

xpivarc: @kbidarkar @andreabolognani can you elaborate?

andreabolognani: @xpivarc There's a typo in `DefaultMemBallonStatsPeriod`, it should be `DefaultMemBalloonStatsPeriod` (note the two "o"s).

PR #3581, opened 2020-06-16
Ignore 404's on VMI deletion when cleaning namespace

stu-gott:
if a VMI has an associated VM, the OwnerReference will cause a cascading deletion when that VM is deleted. We should account for this when examining errors caused by VMI deletion.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -1655,7 +1655,8 @@ func cleanNamespaces() {
 		PanicOnError(virtCli.RestClient().Delete().Namespace(namespace).Resource("virtualmachineinstancereplicasets").Do().Error())
 
 		// Remove all VMIs
-		PanicOnError(virtCli.RestClient().Delete().Namespace(namespace).Resource("virtualmachineinstances").Do().Error())
+		// VMIs with OwnerReferences can be deleted when their VM is deleted. Ignore not found errors to avoid this race
+		PanicOnErrorIgnoreNotFound(virtCli.RestClient().Delete().Namespace(namespace).Resource("virtualmachineinstances").Do().Error())
 		vmis, err := virtCli.VirtualMachineInstance(namespace).List(&metav1.ListOptions{})
 		PanicOnError(err)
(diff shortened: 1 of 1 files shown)


tests/utils.go
```diff
@@ -1655,7 +1655,8 @@ func cleanNamespaces() {
 		PanicOnError(virtCli.RestClient().Delete().Namespace(namespace).Resource("virtualmachineinstancereplicasets").Do().Error())
 
 		// Remove all VMIs
-		PanicOnError(virtCli.RestClient().Delete().Namespace(namespace).Resource("virtualmachineinstances").Do().Error())
+		// VMIs with OwnerReferences can be deleted when their VM is deleted. Ignore not found errors to avoid this race
+		PanicOnErrorIgnoreNotFound(virtCli.RestClient().Delete().Namespace(namespace).Resource("virtualmachineinstances").Do().Error())
```
rmohr, line 1659: I think that if you get a 404 here, then you have issues talking to the apiserver. We run this code since forever before each test. It never fails. Even if there are no VMIs present.

stu-gott: I used the wrong words, which confused the issue. the error message was this:

Panic: the server could not find the requested resource (delete virtualmachines.kubevirt.io)

But upon closer inspection, I mis-read which line panicked:

```
// Remove all VirtualMachines
PanicOnError(virtCli.RestClient().Delete().Namespace(namespace).Resource("virtualmachines").Do().Error())
```

So I disagree with the assertion that this code never fails: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3534/pull-kubevirt-e2e-k8s-1.17/1270527327452794880

stu-gott: But you are correct. This PR doesn't fix the problem. if we're getting a "resource of type VM is undefined" error, then that sounds more like a race condition in restoring the operator.

Issue #3561, opened 2020-06-15
[flakey test] Creating a VirtualMachineInstance when virt-handler is not responsive [test_id:1634]the node controller should react


phoracek:
**Is this a BUG REPORT or FEATURE REQUEST?**:

 /kind bug

**What happened**:

This test fails with:

```
/root/go/src/kubevirt.io/kubevirt/tests/vmi_lifecycle_test.go:609
Timed out after 180.000s.
Expected
    <v1.VirtualMachineInstancePhase>: Running
to equal
    <v1.VirtualMachineInstancePhase>: Failed
/root/go/src/kubevirt.io/kubevirt/tests/vmi_lifecycle_test.go:637
```

It was observed 5 times in last week: https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2020-06-14-168h.html#row3

It happened on k8s 1.17 and 1.18 lanes.

Example failed run: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3540/pull-kubevirt-e2e-k8s-1.18/1271432801274564608

**What you expected to happen**:

It should pass.

dhiller:
Two of the failures observed from flakefinder look different than the others, i.e. they are not a timeout after 180 seconds, [one is a panic](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3534/pull-kubevirt-e2e-k8s-1.17/1270527327452794880), the other is waiting on [condition "Virthandler should be ready to work" to be true](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3546/pull-kubevirt-e2e-k8s-1.17/1271647692564467712).

So the flake count goes down to one on k8s-1.17 and 2 on k8s-1.18.

stu-gott:
The panic test run is caused by an oddball race condition that we're doing to ourselves. We have OwnerReferences for things like a VMI which has a VM as a parent. if the VM is deleted, then k8s will cascade that deletion to the VMI and possibly to container disks and the like... the next thing we do is attempt to delete the same VMI -- and then we panic on an "isNotFound" error. https://github.com/kubevirt/kubevirt/pull/3581

rmohr:
What I see is that the virt-launcher pods get actually deleted (see the deletion timestamp):

```
[...]

PR #3582, opened 2020-06-16
Increase timeout for virt-handler to return to ready state

stu-gott:
Flakey test observed in the wild. 120s isn't always enough.

Addresses at least one failure mentioned in:

https://github.com/kubevirt/kubevirt/issues/3561

**Release note**:
```release-note
NONE
```


stu-gott:
@EdDev The issue linked in the description has context for this.

Let's not overthink the timeout increase. This extra time *only* applies to the (very) few cases that exceed the previous timeout. The point here is to pick something realistic and move on.

rmohr:
Merging since the tide batch run looked good. I don't understand why the batch did not get merged.

PR #3595, opened 2020-06-18
[flakey test] mark parent vmi as failed if associated pod is being deleted

### LINKED ISSUE #3561: [flakey test] Creating a VirtualMachineInstance when virt-handler is not responsive [test_id:1634]the node controller should react

ashleyschuett:

This PR fixes and edge case of a `virt-launcher` pod going into a Pending state when it is being deleted. 

Fixes #3561 

When the virt-handler is not present on a node and the `virt-launcher` pod gets deleted is deleted the node controller needs to update the VMI status as the `virt-handler` is not longer monitoring it. The node-controller however filters out the VMI and does not mark it as `Failed` since there is an "alive" pod for the VMI and it is in the correct namespace. This change checkes to see if the pod is trying to terminate and marks the VMI as failed instead of allowing it to continue to have `Status.Phase: Running` 

example pod spec:
```
"metadata": {
                "name": "virt-launcher-testvmi2wwfkkvt5rq6qpr7jtr4t7jkhl6cvhkqjbbxznk4tn",
                "deletionTimestamp": "2020-06-12T14:33:39Z",
                 ....
}, 
... 
"status": {
                "phase": "Pending",
[...]
                "containerStatuses": [
                    {
                        "name": "compute",
                        "state": {
                            "waiting": {
                                "reason": "PodInitializing"
                            }
                        },
                        "lastState": {},
                        "ready": false,
                        "restartCount": 0,
                        "image": "registry-proxy.engineering.redhat.com/rh-osbs/container-native-virtualization-virt-launcher@sha256:03a96f4bbd4f2ea1ce089f59e47a382a7383265b0a0d04f5fb2075c0a21fcc5c",
                        "imageID": "",
                        "started": false
                    },
                    {
                        "name": "volumedisk0",
                        "state": {
                            "waiting": {
                                "reason": "PodInitializing"
                            }
                        },
                        "lastState": {},
                        "ready": false,
                        "restartCount": 0,
                        "image": "kubevirt/cirros-container-disk-demo:v0.30.0",
                        "imageID": "",
                        "started": false
                    }

```

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/vmi_lifecycle_test.go
+++ b/tests/vmi_lifecycle_test.go
@@ -619,7 +619,10 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
 				})
 				Expect(err).ToNot(HaveOccurred(), "Should list pods successfully")
 				Expect(pods.Items).To(HaveLen(1), "There should be only one VMI pod")
-				Expect(virtClient.CoreV1().Pods(vmi.Namespace).Delete(pods.Items[0].Name, &metav1.DeleteOptions{})).To(Succeed(), "The vmi pod should be deleted successfully")
+				var gracePeriod int64 = 0
+				Expect(virtClient.CoreV1().Pods(vmi.Namespace).Delete(pods.Items[0].Name, &metav1.DeleteOptions{
+					GracePeriodSeconds: &gracePeriod,
+				})).To(Succeed(), "The vmi pod should be deleted successfully")
 
 				// it will take at least 45 seconds until the vmi is gone, check the schedulable state in the meantime
 				By("marking the node as not schedulable")


pkg/virt-controller/watch/node.go
```diff
@@ -282,7 +282,7 @@ func filterStuckVirtualMachinesWithoutPods(vmis []*virtv1.VirtualMachineInstance
 	for _, vmi := range vmis {
 		if vmi.IsScheduled() || vmi.IsRunning() {
 			if podsForVMI, exists := podsPerNamespace[vmi.Namespace]; exists {
-				if _, exists := podsForVMI[string(vmi.UID)]; exists {
+				if pod, exists := podsForVMI[string(vmi.UID)]; exists && pod.ObjectMeta.DeletionTimestamp == nil {
```
rmohr, line 285: I am not sure that this is the right angle to the fix. The test which fails is simulating loosing the connection at least partially to the node. To be on the save side, we say that for as long as we find alive pods for a VMI, we don't know the real state of the VMI. Assuming that the state of the pod which was last reported is the current state can be dangerous regarding to e.g. currupting disks.

We try to follow the same behavour like k8s itself has when it looses the connection to nodes. The pods pretty much stay in a stuck state and can't be deleted until an admin or a fencing agent force-deletes them. That is what we basically have to wait for, if the pod is not already in a terminated state.

So we explicitly don't want to react on still existing pods which are not already finished.

Does that make sense?

rmohr: What I still don't understand is why a VMI can appear as running although the pod claims to not even have been started.

ashleyschuett: what I am seeing as the time long of events is as follows: 

1. `virt-launcher` pod starts 
```
"reason": "Started",
"message": "Started container compute",
"source": {
    "component": "kubelet",
    "host": "node01"
},
"firstTimestamp": "2020-06-12T14:33:02Z",
```
```
"reason": "Started",
"message": "VirtualMachineInstance started.",
"source": {
    "component": "virt-handler",
    "host": "node01"
},
"firstTimestamp": "2020-06-12T14:33:07Z",
```
```
"reason": "Created",
"message": "VirtualMachineInstance defined.",
"source": {
    "component": "virt-handler",
    "host": "node01"
},
"firstTimestamp": "2020-06-12T14:33:07Z",
```
2. `virt-handler` is removed from node
```
"reason": "SuccessfulDelete",
"message": "Deleted pod: virt-handler-hq5zn",
"source": {
    "component": "daemonset-controller"
},
"firstTimestamp": "2020-06-12T14:33:07Z",
```
```
"reason": "Killing",
"message": "Stopping container virt-handler",
"source": {
    "component": "kubelet",
    "host": "node01"
},
"firstTimestamp": "2020-06-12T14:33:07Z",
```

3. `virt-launcher` pod gets deletions signal 
```
"reason": "Killing",
"message": "Stopping container compute",
"source": {
    "component": "kubelet",
    "host": "node01"
},
"firstTimestamp": "2020-06-12T14:33:09Z",
```
```
"reason": "NodeUnresponsive",
"message": "virt-handler is not responsive, marking node as unresponsive",
"source": {
    "component": "node-controller"
},
"firstTimestamp": "2020-06-12T14:33:09Z",
```
```
            "involvedObject": {
                "kind": "Pod",
                "namespace": "kubevirt-test-default",
                "name": "virt-launcher-testvmi2wwfkkvt5rq6qpr7jtr4t7jkhl6cvhkqjbbxznk4tn",
                "uid": "05e75973-a9e9-49a6-b66b-de4cc807f1ca",
                "apiVersion": "v1",
                "resourceVersion": "24863",
                "fieldPath": "spec.containers{volumedisk0}"
            },
            "reason": "Unhealthy",
            "message": "Readiness probe errored: rpc error: code = Unknown desc = container not running (5e72d6a682446d71b8cc04174c907335386159959f8c5518aa1838a301d76ca7)",
            "source": {
                "component": "kubelet",
                "host": "node01"
            },
            "firstTimestamp": "2020-06-12T14:33:09Z",
```


My thought is that on delete of `virt-launcher` pod is not deleted in time for the readiness probe which then fails and then the pod thinks it is in a pending state instead of deletion. 

If that makes sense?

rmohr: hm, this is totally reasonable. So this seems to be a kubelet bug. How about doing a force delete in our test? I am not sure if we should code in such a workaround :/

PR #3598, opened 2020-06-18
virt-handler: Remove duplicate socket close

oshoval:
Its done already in the defer function,
and isnt needed to be done twice in case the server was closed.

---

Update logs of grpc server

Log after the grpc server stop is finished,
in order to record only if it was stopped fully.

```release-note
None
```

### DIFF
--- a/pkg/virt-handler/notify-server/server.go
+++ b/pkg/virt-handler/notify-server/server.go
@@ -156,8 +156,6 @@ func RunServer(virtShareDir string, stopChan chan struct{}, c chan watch.Event, 
 	case <-stopChan:
 		log.Log.Info("stopping notify server")
 		grpcServer.Stop()
-		sock.Close()
-		os.Remove(sockFile)
 	}
 
 	return nil
--- a/pkg/virt-handler/notify-server/server.go
+++ b/pkg/virt-handler/notify-server/server.go
@@ -154,8 +154,8 @@ func RunServer(virtShareDir string, stopChan chan struct{}, c chan watch.Event, 
 	case <-done:
 		log.Log.Info("notify server done")
 	case <-stopChan:
-		log.Log.Info("stopping notify server")
 		grpcServer.Stop()
+		log.Log.Info("notify server stopped")
 	}
 
 	return nil


pkg/virt-handler/notify-server/server.go
```diff
@@ -156,8 +156,7 @@ func RunServer(virtShareDir string, stopChan chan struct{}, c chan watch.Event,
 	case <-stopChan:
 		log.Log.Info("stopping notify server")
 		grpcServer.Stop()
-		sock.Close()
-		os.Remove(sockFile)
+		log.Log.Info("notify server stopped")
```
oshoval, line 159: i would suggest to add a print also before line 149,
(i know it wont be atomic)
the reason is that the server is a singleton per virt-handler, (and is recreated if its dies).
We have a race that a go routine of the server isnt closed (or is using it before its ready),
so it would help to know generally what is the flow upon server creation and closing.

AlonaKaplan:
@oshoval added the hold label. Since it's virt code, @enp0s3 can you please take a look? If you're ok with the change, please remove the hold.

pkg/virt-handler/notify-server/server.go
EdDev, line 159: Why this log is needed? Stopping the server is already mentioned.

oshoval: The panic that i investigate now, is due to a child of GRPC server which exists after the stop, i want the log to show the flow, and the actual stopping of the server which finished and left a zombie (serveStreams).
Meanwhile i am investigating and trying to simulate the panic itself locally.

EdDev: When you see "stopping notify server" it means it is closed, if you prefer to see when it is already stopped, then remove the previous one. As it does not return an error, this can only panic if it fails.

If you hope to catch entries in between these actions (i.e. during the stop processing) this will not help and is wrong (because other logs may come in before the stop is started or after it finished, so you can never know if it was during).

EdDev: It is also not related to this change.
(and yes, I do not like logs, especially when they do not add real benefit on regular runs)

oshoval: its important because we have a real bug, and can be removed later, also it doesn't add that much noise.
i can remove the first one, but if i can keep it, it would be even better please, i do want to see the full flow when it actually started to Stop and when it finished to stop, since its important for this panic which can affect the virt core.
(we are using here external library go-grpc, and it would assist in researching and deciding what are the next step, for example use GracefulStop)
we can see it as a refactor so it does connected to this change (whichever way we choose) unless you insist on 2nd PR.
wdyt?

enp0s3: > When you see "stopping notify server" it means it is closed, if you prefer to see when it is already stopped, then remove the previous one. As it does not return an error, this can only panic if it fails.
> 
> If you hope to catch entries in between these actions (i.e. during the stop processing) this will not help and is wrong (because other logs may come in before the stop is started or after it finished, so you can never know if it was during).

@EdDev The logs have timestamps even if they aren't chronologically arranged. Why do you think that logging won't help?

EdDev: > its important because we have a real bug, and can be removed later, also it doesn't add that much noise.
> i can remove the first one, but if i can keep it, it would be even better please, i do want to see the full flow when it actually started to Stop and when it finished to stop, since its important for this panic which can affect the virt core.

I explained why you cannot catch problems this way in my previous comment.
I also do not think that a debugging needs to be done this way, please add that in your own PR or locally and test that way.

To be clear, logging the flow is fine, but here the flow is clear and the operation is not suppose to be blocking. So if it was logged before or after the operation is enough.

> (we are using here external library go-grpc, and it would assist in researching and deciding what are the next step, for example use GracefulStop)

If everyone will endorse this pattern of adding logs it will both convolute the code and the logs.
I see no beneficial outcome of adding this log and I explained why.

> we can see it as a refactor so it does connected to this change (whichever way we choose) unless you insist on 2nd PR.
> wdyt?

You could have piggy-back it in on the PR through a separate commit.

oshoval: > > its important because we have a real bug, and can be removed later, also it doesn't add that much noise.
> > i can remove the first one, but if i can keep it, it would be even better please, i do want to see the full flow when it actually started to Stop and when it finished to stop, since its important for this panic which can affect the virt core.
> 
> I explained why you cannot catch problems this way in my previous comment.
> I also do not think that a debugging needs to be done this way, please add that in your own PR or locally and test that way.

I want that the log will show that the "Stop" actually started, this is why i need both.
i cant put on my own PR, because it happens just 1 or 2 a week.

> 
> To be clear, logging the flow is fine, but here the flow is clear and the operation is not suppose to be blocking. So if it was logged before or after the operation is enough.
> 
> > (we are using here external library go-grpc, and it would assist in researching and deciding what are the next step, for example use GracefulStop)
> 
> If everyone will endorse this pattern of adding logs it will both convolute the code and the logs.
> I see no beneficial outcome of adding this log and I explained why.
> 
> > we can see it as a refactor so it does connected to this change (whichever way we choose) unless you insist on 2nd PR.
> > wdyt?
> 
> You could have piggy-back it in on the PR through a separate commit.

EdDev: > @EdDev The logs have timestamps even if they aren't chronologically arranged. Why do you think that logging won't help?

```
// t1
log() // t2
// t3
do() // t4
// t5
log() //t6
// t7
```
You can have things happening at t3 and t5 and you will think it is happening during t4.

oshoval: > > @EdDev The logs have timestamps even if they aren't chronologically arranged. Why do you think that logging won't help?
> 
> ```
> // t1
> log() // t2
> // t3
> do() // t4
> // t5
> log() //t6
> // t7
> ```
> 
> You can have things happening at t3 and t5 and you will think it is happening during t4.

The "Stop" is the one that trigger the flow, which eventually trigger the close of event channel (on other function that wait for this function to release 3rd function). even to explain it in a sentence isnt easy, the flow has lots of blocks.
and this simple prints would encapsulate the critical operation and show much better and without doubt that the Stop, even that its protected is related to our problem.

enp0s3: > > > @EdDev The logs have timestamps even if they aren't chronologically arranged. Why do you think that logging won't help?
> > 
> > 
> > ```
> > // t1
> > log() // t2
> > // t3
> > do() // t4
> > // t5
> > log() //t6
> > // t7
> > ```
> > 
> > 
> > You can have things happening at t3 and t5 and you will think it is happening during t4.
> 
> The "Stop" is the one that trigger the flow, which eventually trigger the close of event channel (on other function that wait for this function to release 3rd function). even to explain it in a sentence isnt easy, the flow has lots of blocks.
> and this simple prints would encapsulate the critical operation and show much better and without doubt that the Stop, even that its protected is related to our problem.

@oshoval Following @EdDev example, we may think that the domain ERROR event came in t4 (during the Close()) while actually the event could arrive before the Close() or after it, so we don't have a clear proof whether the ERROR event was received by an orphant GRPC child or not.

oshoval: The "Stop" is the one that trigger the flow, which eventually trigger the close of event channel as by product in the flow.
We need to know
1. It started, so it will know the channel was closed due to this and not other reason.
2. when the panic came , is it before the end or after the end. because it will say where to look in the flow.
are there are zombies, or the flow wasn't finish, but already closed the channel.
3. it ended, so if the panic comes after, it means its a zombie.

enp0s3: @oshoval I think there is no other way to debug it but from the GRPC code itself, somehow put a debug prints that will indicate abnormal Close() during Message handling.

EdDev: > The "Stop" is the one that trigger the flow, which eventually trigger the close of event channel as by product in the flow.
> We need to know
> 
>     1. It started, so it will know the channel was closed due to this and not other reason.
If the channel that got closed is related to this operation, then you can assume that it happened due to this operation. That is logged in the existing log already. 
 
>     2. when the panic came , is it before the end or after the end. because it will say where to look in the flow.
>        are there are zombies, or the flow wasn't finish, but already closed the channel.

If it comes after the first log and the channel it blow up on is related to this operation, you know already.

>     3. it ended, so if the panic comes after, it means its a zombie.

It means nothing, it may blow up one nano seconds after the log was dispatched.

PR #3599, opened 2020-06-18
Add tests related to qemu processes

kbidarkar:
**What this PR does / why we need it**: `Add tests related to qemu processes`

Fixes #

**Special notes for your reviewer**: `Add tests related to qemu processes`

**Release note**:

```
NONE
```

### DIFF
--- a/tests/security_features_test.go
+++ b/tests/security_features_test.go
@@ -20,6 +20,8 @@
 package tests_test
 
 import (
+	"strings"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	k8sv1 "k8s.io/api/core/v1"
@@ -42,6 +44,9 @@ var _ = Describe("SecurityFeatures", func() {
 	})
 
 	Context("Check virt-launcher securityContext", func() {
+		tests.BeforeAll(func() {
+			tests.SkipSELinuxTestIfRunnigOnKindInfra()
+		})
 
 		var container k8sv1.Container
 		var vmi *v1.VirtualMachineInstance
@@ -81,18 +86,48 @@ var _ = Describe("SecurityFeatures", func() {
 				}
 				Expect(*container.SecurityContext.Privileged).To(BeFalse())
 			})
+
+			It("[test_id:4297]Make sure qemu processes are MCS constrained", func() {
+
+				By("Starting a VirtualMachineInstance")
+				vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+				Expect(err).ToNot(HaveOccurred())
+				tests.WaitForSuccessfulVMIStart(vmi)
+
+				pod := tests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
+
+				qemuProcessSelinuxContext, err := tests.ExecuteCommandOnPod(
+					virtClient,
+					pod,
+					"compute",
+					[]string{"/usr/bin/bash", "-c", "ps -efZ | grep [/]usr/libexec/qemu-kvm | awk '{print $1}'"},
+				)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Checking that qemu-kvm process is of the SELinux type container_t")
+				Expect(strings.Split(qemuProcessSelinuxContext, ":")[2]).To(Equal("container_t"))
+
+				By("Checking that qemu-kvm process has SELinux category_set")
+				Expect(len(strings.Split(qemuProcessSelinuxContext, ":"))).To(Equal(5))
+
+				err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Delete(vmi.Name, &metav1.DeleteOptions{})
+				Expect(err).ToNot(HaveOccurred())
+			})
 		})
 
-		Context("With selinuxLauncherType defined", func() {
-			var superPrivilegedType string
-
-			BeforeEach(func() {
-				superPrivilegedType = "spc_t"
-				tests.UpdateClusterConfigValueAndWait(virtconfig.SELinuxLauncherTypeKey, superPrivilegedType)
-				vmi = tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskAlpine))
-			})
+		Context("With selinuxLauncherType defined as spc_t", func() {
 
 			It("[test_id:3787]Should honor custom SELinux type for virt-launcher", func() {
+
+				superPrivilegedType := "spc_t"
+				kubeVirtConfig, err := virtClient.CoreV1().ConfigMaps(tests.KubeVirtInstallNamespace).Get("kubevirt-config", metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				if kubeVirtConfig.Data[virtconfig.SELinuxLauncherTypeKey] != superPrivilegedType {
+					tests.UpdateClusterConfigValueAndWait(virtconfig.SELinuxLauncherTypeKey, superPrivilegedType)
+				}
+
+				vmi = tests.NewRandomVMIWithEphemeralDisk(tests.ContainerDiskFor(tests.ContainerDiskAlpine))
+
 				By("Starting a New VMI")
 				vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
 				Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


tests/security_features_test.go
```diff
@@ -141,6 +142,30 @@ var _ = Describe("SecurityFeatures", func() {
...
+					"compute",
+					[]string{"/usr/bin/bash", "-c", "ps -efZ | grep [q]emu-kvm | awk '{print $1}'"},
+				)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Checking that qemu-kvm process is of the SELinux type container_t")
+				Expect(strings.Split(qemuProcess, ":")[2]).To(Equal("container_t"))
+
+				By("Checking that qemu-kvm process has SELinux category_set")
+				Expect(len(strings.Split(qemuProcess, ":"))).To(Equal(5))
```
kbidarkar, line 167: The category_set is random and so not sure, what to compare it with.
Hence decided to check for it's length, if category_set exists the length would be 5, ifnot then 4.

tests/security_features_test.go
```diff
@@ -141,6 +142,30 @@ var _ = Describe("SecurityFeatures", func() {
...
+
+				qemuProcess, err := tests.ExecuteCommandOnPod(
+					virtClient,
+					pod,
+					"compute",
+					[]string{"/usr/bin/bash", "-c", "ps -efZ | grep [q]emu-kvm | awk '{print $1}'"},
+				)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Checking that qemu-kvm process is of the SELinux type container_t")
```
jean-edouard, line 163: Qemu processes should run as container_t only if `selinuxLauncherType` is not set in the kubevirt-config configmap.
Otherwise, the domain on qemu processes should be equal to the value of `selinuxLauncherType`.

tests/security_features_test.go
jean-edouard: :+1:

tests/security_features_test.go
```diff
@@ -141,6 +142,30 @@ var _ = Describe("SecurityFeatures", func() {
...
+			It("[test_id:5297]Make sure qemu processes are MCS constrained", func() {
+
+				By("Starting a VirtualMachineInstance")
+				vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+				Expect(err).ToNot(HaveOccurred())
+				tests.WaitForSuccessfulVMIStart(vmi)
+
+				pod := tests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)
+
+				qemuProcess, err := tests.ExecuteCommandOnPod(
```
jean-edouard, line 155: Nit? The variable `qemuProcess` contains the SELinux context of the qemu process, its name should probably reflect that.

tests/security_features_test.go
```diff
@@ -141,6 +142,30 @@ var _ = Describe("SecurityFeatures", func() {
...
+					virtClient,
+					pod,
+					"compute",
+					[]string{"/usr/bin/bash", "-c", "ps -efZ | grep [q]emu-kvm | awk '{print $1}'"},
```
jean-edouard, line 159: Is the `q` in brackets to avoid grepping the `grep` process? Cool trick!
Nit: we should probably `Expect()` the output of this to be 1 line only. To make sure that's true we could also grep for something bigger (like `/usr/libexec/[q]emu-kvm`).

tests/security_features_test.go
kbidarkar: yeah, agreed. hence this test-case is under `Context("With selinuxLauncherType undefined"`,  in kubevirt-config CM.

tests/security_features_test.go
kbidarkar: Agreed, updated to `qemuProcessSelinuxContext`

tests/security_features_test.go
kbidarkar: Updated this line to `"ps -efZ | grep [/]usr/libexec/qemu-kvm | awk '{print $1}'`

kbidarkar: yes, `q` in brackets to avoid grepping the grep process.

tests/security_features_test.go
jean-edouard: Ah, missed that, thanks!

tests/security_features_test.go
```diff
@@ -141,40 +142,55 @@ var _ = Describe("SecurityFeatures", func() {
 				}
 				Expect(*container.SecurityContext.Privileged).To(BeFalse())
 			})
-		})
 
-		Context("With selinuxLauncherType defined", func() {
```
kbidarkar, line 146: Updated this as `Context("With selinuxLauncherType defined as spc_t", func() {`

jean-edouard:
This PR now does more than just "Add [a] test to make sure qemu processes are MCS constrained".
Could you please describe what else is happening, maybe in the commit description?

kbidarkar:
> This PR now does more than just "Add [a] test to make sure qemu processes are MCS constrained".
> Could you please describe what else is happening, maybe in the commit description?

I updated it now to reflect the changes done, I pushed new pending changes yesterday. Thought of raising a new PR actually, but decided against as then it would in total be 4 PR's, instead of 2, if we also count in the release-30 backports.

tests/utils.go
```diff
@@ -4910,6 +4910,28 @@ func GetKubeVirtConfigMap() (*k8sv1.ConfigMap, error) {
 	return cfgMap, err
 }
 
+func ReplaceKubeVirtConfigMapDataValue(key string, value string) error {
```
jean-edouard, line 4913: This might not be needed anymore since https://github.com/kubevirt/kubevirt/pull/3542, @kbidarkar please have a look.

kbidarkar: Thanks @jean-edouard  , I guess I need to just use test.UpdateClusterConfigValueAndWait

ok, will check and update it.

kbidarkar: My bad, my function was in fact never needed. I now just used the existing one.

rmohr:
@kbidarkar you can use

/cherrypick release-0.30

to create cherrypicks in many cases.

kbidarkar:
Looked at the PR  History, 
the test passes for all the other lanes except on `ipv6 lane`, with below error:

Would need help, as me not entirely sure, how different the ipv6 lane setup is here.

```
• Failure [16.345 seconds]
SecurityFeatures
/root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:41
  Check virt-launcher securityContext
  /root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:88
    With selinuxLauncherType undefined
    /root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:93
      [test_id:4297]Make sure qemu processes are MCS constrained [It]
      /root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:146

      Expected
          <string>: spc_t
      to equal
          <string>: container_t

      /root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:164
```

```
• Failure [14.820 seconds]
SecurityFeatures
/root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:41
  Check virt-launcher securityContext
  /root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:88
    With selinuxLauncherType defined as virt_launcher.process
    /root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:207
      [test_id:4298]qemu process type is virt_launcher.process, when selinuxLauncherType is virt_launcher.process [It]
      /root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:209

      Expected
          <string>: spc_t
      to equal
          <string>: virt_launcher.process

      /root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:240
```

jean-edouard:
Could be that dockerd is running without "--enable-selinux", in which case everything is spc_t.

rmohr:
> Could be that dockerd is running without "--enable-selinux", in which case everything is spc_t.

I have seen this issue a few times now. Maybe it is bound to some nodes, because this issue most of the time does not happen.

rmohr:
> I have seen this issue a few times now. Maybe it is bound to some nodes, because this issue most of the time does not happen.

The last three ones showed a specific node. Checking.

rmohr:
> The last three ones showed a specific node. Checking.
The node is fine. I think the tests are written in a flaky way. They don't wait until config changes are propagated. Will push a fix.

kbidarkar:
> > The last three ones showed a specific node. Checking.
> 
> The node is fine. I think the tests are written in a flaky way. They don't wait until config changes are propagated.

At-least for `4298` testcase, I make use of `tests.UpdateClusterConfigValueAndWait `  which as per below comment, it waits.
https://github.com/kubevirt/kubevirt/blob/master/tests/utils.go#L4415

But for some reason, the value `selinuxLauncherType` does not get updated.

tests/security_features_test.go
```diff
@@ -81,18 +84,48 @@ var _ = Describe("SecurityFeatures", func() {
...
+
+				By("Checking that qemu-kvm process has SELinux category_set")
+				Expect(len(strings.Split(qemuProcessSelinuxContext, ":"))).To(Equal(5))
+
+				err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Delete(vmi.Name, &metav1.DeleteOptions{})
+				Expect(err).ToNot(HaveOccurred())
 			})
+		})
+
+		Context("With selinuxLauncherType defined as spc_t", func() {
```
jean-edouard, line 116: Nit: spc_t tests happen to pass on Kind because everything is spc_t there... But it's really not testing anything so I would recommend skipping this as well on Kind.

kbidarkar: ok, let me skip that as well.

PR #3609, opened 2020-06-19
Add simple profiling for test automation

dhiller:

Simple execution time output is now rendered when `KUBEVIRT_PROFILE_MAKE=1`
is used. This is set in `automation/test.sh` by default.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/profile.sh
+++ b/hack/profile.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e
+
+# use time to record the elapsed time for the step
+shift # get rid of the '-c' supplied by make.
+
+if [[ -n ${KUBEVIRT_PROFILE_MAKE} ]]; then
+    date
+fi
+/bin/bash -c "$*"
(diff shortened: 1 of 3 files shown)


hack/profile.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -e
+
+# use time to record the elapsed time for the step
+shift # get rid of the '-c' supplied by make.
+
+if [[ -z ${KUBEVIRT_PROFILE_MAKE} ]]; then
+    /bin/bash -c "$*"
+else
+    time /bin/bash -c "$*"
```
rmohr, line 11: Would it make sens to just use `date`? I think that would make it easier to detect where we are stalling.

dhiller: Done.

dankenigsberg:
I assume the purpose of this PR is to ease debugging. Am I correct @dhiller?

dhiller:
> I assume the purpose of this PR is to ease debugging. Am I correct @dhiller?

Yes, ATM we have problems with slow CI builds and want to know where the most time is consumed.

dankenigsberg:
Thanks, @dhiller. Please include This level of motivation in the primary message of PRs, so that laypersons like me would have a chance to differential a helper PR like this from a PR that that solves something.

dhiller:
Closing this in favor of kubevirt/kubevirt#3609

PR #3613, opened 2020-06-19
Add Kernel Boot feature to boot VMs

### LINKED ISSUE #2741: Add kernel provision through command line
FEATURE REQUEST

/kind enhancement

We are testing kubevirt and it seems to be great, however, we miss the feature of pass a kernel parameter to provide by command line a different kernel than it has in the image.

As an example in the case of kvm the command line will be:

qemu-system-x86_64 -kernel "custom_kernel_provided".

Thank you very much.
Xavi.

hritvi:
This is a draft PR.
This PR enables the VM to boot directly from a kernel and initrd stored in the host OS, allowing command line arguments to be passed directly to the installer.

Fixes #2741 

This is a draft PR. 
Proposed design document can be found [here ](https://docs.google.com/document/d/1RHIh8l906WIno-zhhpzgGhL-795uZUr_YM6_noLEtvw/edit#)

**Release note**:
```release-note
Allow users to boot a VM directly from kernel, initrd and command line arguments.
```


pkg/virt-launcher/virtwrap/api/converter_test.go
```diff
@@ -2202,6 +2202,135 @@ var _ = Describe("Converter", func() {
...
+			It("should configure normal boot", func() {
+				vmi.Spec.Domain.Firmware = &v1.Firmware{}
+				domainSpec := vmiToDomainXMLToDomainSpec(vmi, c)
+				Expect(domainSpec.OS.Kernel).To(BeNil())
+				Expect(domainSpec.OS.Initrd).To(BeNil())
+				Expect(domainSpec.OS.Cmdline).To(BeNil())
+			})
+		})
+
+		Context("when kernel boot is set and kernel is provided", func() {
```
dhiller, line 2230: Maybe you can refactor this by using Table driven tests, this could greatly reduce redundancy, have a look at that section in [this document](https://onsi.github.io/ginkgo/). Get back to me if you have questions.

dhiller: It's located under Extensions on the page mentioned.

examples/vmi-kernel-boot.yaml
```diff
@@ -0,0 +1,29 @@
...
+    devices:
+      disks:
+      - disk:
+          bus: virtio
+        name: containerdisk
+    firmware:
+      kernelBoot:
+        cmdline: console=ttyS0
+        kernelPath: kernelImage/
+        name: containerdisk
```
rmohr, line 19: I think it makes sense to inline something  like the `containerDisk` struct here. It just is not a volume. There are also fields on the containerDisk (like disk path) which have no meaning in this context. Let's create a dedicated struct.
Also let's think about the possibility that different sources may appear later on. A possible solution would be something like this:


```
kernelBoot:
   cmdline: <something>
   container:
     image: <image>
     imagePullSecret #optional
     imagePullPolicy #optional
     kernelPath: #optional
    <...>
```

If we want to provide it via hostPath (I don't think we should but just as an example) in the future, it would then for instance look like this:

```
kernelBoot:
   cmdline: <something>
   hostPath: # optional
     relativeKernelPath: #optional
    <...>
```

Note that I am also calling the subsection `container` and not `containerDisk`, to make clear that it is different.
Feel free to propose a file-structure which would be auto-discovered.

rmohr: We could also consider adding ad `cmdlinePath` where we would have a file which could provide a "default" commandline if none is extra specified.

examples/vmi-kernel-boot.yaml
```diff
@@ -0,0 +1,29 @@
...
+spec:
+  domain:
+    devices:
+      disks:
+      - disk:
+          bus: virtio
+        name: containerdisk
+    firmware:
+      kernelBoot:
+        cmdline: console=ttyS0
```
rmohr, line 17: It is fine to use the full name `commandline`

examples/vmi-kernel-boot.yaml
rmohr: Regarding to the implementation, here a possible implementation flow:

1) virt-controller would in template.go check if a container is specified there and add it to the pod template
2) virt-handler would have in pkg/virt-handler/container-disk/mount.go another Mount function (something like MountKernelArtifacts and call this function in `pkg/virt-handler/vm.go` before or after `Mount` for the containerdisks is called, if a container is specified.
3) virt-launcher would like for the mounted containerDiskPath translate the VMI spec to the path of the artifacts in the converter (if we go with default paths, it would probably check which files are mounted and pass them in via the converter context).

rmohr (changes requested):
Added some thoughts on how I think it would make sense to proceed with the PR. Happy to answer any questions if I am unclear or if you have other ideas.

examples/vmi-kernel-boot.yaml
vladikr: What if we will just reference a volume here? In this case, we will not need to use a specific struct here and it will be easy to extend the support in the future. We could limit the types that we can support right now.

rmohr: My suggestion was to use not a volume, since we see already on the containerDisk that it contains fields which are tailored to the need of exposing a disk.

vladikr: We had an offline conversation and I agree with Roman, it's best to reference the source as suggested.

gabrielecerami:
@hritvi: PR needs rebase.

<details>

</details>

EdDev (approved):
Thank you!

It still needs to be rebased after https://github.com/kubevirt/kubevirt/pull/4156 is merged, so adding a hold.

/hold

EDIT: Sorry, wrong PR :/

EdDev (changes requested):
Empty comment to remove the ack. Sorry and disregard it.

vladikr:
There are still changes that need to be done as per [1]
@hritvi are you still planning to work on this or would you prefer someone else to complete this work?


[1] https://github.com/kubevirt/kubevirt/pull/3613#pullrequestreview-470359733

daghaian:
@vladikr If it doesn't appear like effort is being made on this I'm more than happy to offer up whatever assistance I can to see this feature merged. I need the same functionality for some of our current usecases as well.

PR #3618, opened 2020-06-21
Add timestamp wrapper to our automation scripts

danielBelenky:
Sometimes, we see automation tasks that are taking longer than expected
and we hit our defined timeouts. By adding timestamps to each line of
our executed scripts, we will be able to detect where our bottleneck is.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/timestamps.sh
+++ b/hack/timestamps.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+function timestamps::time_wrapper() {
+    while IFS= read -r line; do
+        printf "%(%T)T: %s\n" "-1" "$line"
+    done
+}
+
+if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
+    /bin/sh "$@" 2>&1 | timestamps::time_wrapper
+fi
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 export GO15VENDOREXPERIMENT := 1
 
+SHELL = ./hack/timestamps.sh
 
 all:
 	hack/dockerized "DOCKER_PREFIX=${DOCKER_PREFIX} DOCKER_TAG=${DOCKER_TAG} IMAGE_PULL_POLICY=${IMAGE_PULL_POLICY} VERBOSITY=${VERBOSITY} ./hack/build-manifests.sh && \


danielBelenky:
> How about just using `ts` or `gawk`?
> 
> `command | gawk '{ print strftime("[%Y-%m-%d %H:%M:%S]"), $0 }'`
> or
> `command | ts` This one may need `moreutils` package.

What will be the benefit of using one of this tools?
`date` is a much more common tool and is more likely to exist in the environment while `ts` and `gawk` are extra tools that require a custom image.

EdDev:
> What will be the benefit of using one of this tools?
> `date` is a much more common tool and is more likely to exist in the environment while `ts` and `gawk` are extra tools that require a custom image.

The current solution spawns a process (date) for each line, which seems an overkill. And we do print a lot so this may slow down things.

If you still prefer the current per line injection, then maybe instead of `date` you can use bash `%(datefmt)T`.

danielBelenky:
> > What will be the benefit of using one of this tools?
> > `date` is a much more common tool and is more likely to exist in the environment while `ts` and `gawk` are extra tools that require a custom image.
> 
> The current solution spawns a process (date) for each line, which seems an overkill. And we do print a lot so this may slow down things.
> 
> If you still prefer the current per line injection, then maybe instead of `date` you can use bash `%(datefmt)T`.

And `gawk` or `ts` are not processes?

I think that we need per-line log for 2 reasons:
1. Monitor the progress.
2. As you mentioned, we generate a lot of output and I'm not sure that buffering it in memory will be a good idea.

But I do agree that calling `date` when not necessary is a bad idea. I forgot about `printf`'s built-in date formatting, thanks for mentioning it. I've updated the PR.

EdDev:
> And `gawk` or `ts` are not processes?

One process per command/script instead one per line. One is spawned and wraps a script, printing the timestamp for each line, the other is spawned per each line.

At least, this is how I understand it.

> 
> I think that we need per-line log for 2 reasons:
> 
>     1. Monitor the progress.
> 
>     2. As you mentioned, we generate a lot of output and I'm not sure that buffering it in memory will be a good idea.

Yes, buffering a complete script may be a problem. I guess using no-buffers is also an option, but it may cause other issues. Not sure.

danielBelenky:
> enough

Updated the print.

stu-gott:
Sorry for the confusion. We just backed this change out. The timestamps were causing prow jobs to erroneously succeed!

Example: https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3635/pull-kubevirt-e2e-k8s-1.16/1275470747132760064

hack/timestamps.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+function timestamps::time_wrapper() {
+    while IFS= read -r line; do
+        printf "%(%T)T: %s\n" "-1" "$line"
+    done
+}
+
+if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
+    /bin/sh "$@" 2>&1 | timestamps::time_wrapper
```
jean-edouard, line 10: `/bin/sh "$@" > >(timestamps::time_wrapper) 2> >(timestamps::time_wrapper)` would avoid hiding the return value and doesn't merge stdout and stderr.
Nit: why the weird function name?

danielBelenky: Nice, I didn't think think of this way. I'll update my fix.

About the function name, it ensures that if someone sources this script to use this function, the context from where this function came from remains.

PR #3635, opened 2020-06-23
Add a feature gate to enable hostDisks

jean-edouard:
This PR adds the feature gate "HostDisk" that has to be enabled to use hostDisks.

Fixes #

**Release note**:

```release-note
The "HostDisk" feature gate has to be enabled to use hostDisks
```

### DIFF
--- a/pkg/virt-config/feature-gates.go
+++ b/pkg/virt-config/feature-gates.go
@@ -34,6 +34,7 @@ const (
 	SnapshotGate          = "Snapshot"
 	DataVolumesGate       = "DataVolumes"
 	SRIOVGate             = "SRIOVGate"
+	HostDiskGate          = "HostDisk"
 )
 
 func (c *ClusterConfig) isFeatureGateEnabled(featureGate string) bool {
@@ -83,4 +84,8 @@ func (config *ClusterConfig) DataVolumesEnabled() bool {
 
 func (config *ClusterConfig) SriovEnabled() bool {
 	return config.isFeatureGateEnabled(SRIOVGate)
+}
+
+func (config *ClusterConfig) HostDiskEnabled() bool {
+	return config.isFeatureGateEnabled(HostDiskGate)
 }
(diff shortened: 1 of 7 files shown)


tests/storage_test.go
```diff
@@ -402,6 +410,23 @@ var _ = Describe("Storage", func() {
...
+
+					It("Should fail to start a VMI", func() {
+						diskName := "disk-" + uuid.NewRandom().String() + ".img"
+						diskPath := filepath.Join(hostDiskDir, diskName)
+						vmi = tests.NewRandomVMIWithHostDisk(diskPath, v1.HostDiskExistsOrCreate, "")
+						tests.RunVMIAndExpectLaunch(vmi, 30)
+						virtClient, err := kubecli.GetKubevirtClient()
+						Expect(err).ToNot(HaveOccurred())
+						_, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+						Expect(err).To(HaveOccurred())
```
rmohr, line 426: Can we check if the expected cause is included in the error?

pkg/virt-config/feature-gates.go
```diff
@@ -34,6 +34,7 @@ const (
 	SnapshotGate          = "Snapshot"
 	DataVolumesGate       = "DataVolumes"
 	SRIOVGate             = "SRIOVGate"
+	HostDiskGate          = "HostDiskVMIAccess"
```
rmohr, line 37: Should we just call it `HostDisk`?

tests/storage_test.go
jean-edouard: Good point, done.

pkg/virt-config/feature-gates.go
jean-edouard: Renamed.

rmohr:
/lgtm cancel

Removing the PR from the merge pool since it just entered it due to the timewrapper PR.

jean-edouard:
Depends on https://github.com/kubevirt/kubevirt/pull/3542 for stability

rmohr:
@jean-edouard I think the failing tests are failing because they also use hostDisk and the feature gate is not enabled for them.

jean-edouard:
@rmohr good catch, missed one, force-pushed a fix

PR #3500, opened 2020-06-01
Remove domain label from VMI metrics

ArthurSens:
**What this PR does / why we need it**:
This PR removes domain labels from VMI metrics. They are being removed because they originate from the original metrics implementation where the more useful labels like `name` and `namespace` did not exist and now can be removed

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3477 

**Special notes for your reviewer**:

**Release note**:

```release-note
This change will possibly break dashboards that use the domain label to identify VMIs. 

action required
Use name and namespace labels instead
```

Issue #3477, opened 2020-05-27
Remove `domain` label from Metrics


rmohr:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>

/kind bug

> /kind enhancement

**What happened**:

We still expose the `domain` label in our metrics. This originates from our original metrics implementation where the more useful labels like `name` and `namespace` did not exist. We can remove it now.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- KubeVirt version (use `virtctl version`):
- Kubernetes version (use `kubectl version`):
- VM or VMI specifications:
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #3636, opened 2020-06-23
Add k8s metadata labels to VMI metrics

### LINKED ISSUE #2108: virt-handler metrics do not integrate any k8s VMI properties (labels, names, etc)

FEATURE REQUEST

>
> /kind bug

/kind enhancement

**What happened**:

We're currently working on exposing the VMI metrics (cpu, mem, net, disk) that come from the virt-handler metrics endpoint. Right now these metrics come directly from the virt-handler pods and the only kubernetes identifiable information is "domain=$name-space_$pod-name" which is the libvirt domain information. We were hoping that these metrics would have identifiable information from the kubernetes metrics api, such as pod_labels, etc. 

I spoke with @rmohr  and he mentioned the intention of kubevirt metrics to integrate nicely with the k8s metrics and asked me to open an issue.

I had hoped that VMI identifying information might be passed down into the kubevirt_* metrics, such as the VMI name, VMI namespace, VMI labels, and other relevant k8s spec information. 

Access prometheus on a server running kubevirt with the kubevirt serviceMonitor. Query for and kubevirt_* metrics such as; kubevirt_vm_memory_resident_bytes

The output will be similar to the following;

kubevirt_vm_memory_resident_bytes{domain="golden-images_vm-centos-1",endpoint="metrics",instance="10.3.2.1:8443",job="kubevirt-prometheus-metrics",namespace="kubevirt",pod="virt-handler-c5m4b",service="kubevirt-prometheus-metrics"}

**Environment**:

Server Version: version.Info{GitVersion:"v0.15.0-alpha.0", GitCommit:"31d08ef5b6c0170b21f5823d351f43902168e2d5", GitTreeState:"clean", BuildDate:"2019-02-21T11:00:06Z", GoVersion:"go1.11.4", Compiler:"gc", Platform:"linux/amd64"}

Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.0", GitCommit:"ddf47ac13c1a9483ea035a79cd7c10005ff21a6d", GitTreeState:"clean", BuildDate:"2018-12-03T20:56:12Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"}

Baremetal

NAME="CentOS Linux"
VERSION="7 (Core)"

- Kernel (e.g. `uname -a`):

4.13.10-1.el7.elrepo.x86_64

- Install tools:
- Others:

ArthurSens:
Signed-off-by: arthursens <arthursens2005@gmail.com>

This PR adds K8s metadata labels to VMI metrics.

This facilitates users to identify VMIs and aggregate metrics as they prefer and to implement their own monitoring solutions that best suit their needs

#### Labels
As said at the [Kubernetes Labels and Selectors concept documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/)
> Labels enable users to map their own organizational structures onto system objects in a loosely coupled fashion, without requiring clients to store these mappings.
> Example labels:
>
> * "release" : "stable", "release" : "canary"
> * "environment" : "dev", "environment" : "qa", "environment" : "production"
> * "tier" : "frontend", "tier" : "backend", "tier" : "cache"
>*  "partition" : "customerA", "partition" : "customers"
> * "track" : "daily", "track" : "weekly"

Fixes #2108

To test this I'm using the [vmi-fedora example](https://github.com/kubevirt/kubevirt/blob/master/examples/vmi-fedora.yaml).

After deploying the VMI, I get the following metadata when describing the resource:
```
./cluster-up/kubectl.sh describe virtualmachineinstances/vmi-fedora

Name:         vmi-fedora
Namespace:    default
Labels:       kubevirt.io/nodeName=node01
              special=vmi-fedora
```
Which is then translated into metric labels on every VMI metric(Note that it doesn't occur on metric not related to VMIs):
```
curl -k https://localhost:8443/metrics | grep kubevirt
# TYPE kubevirt_info gauge
kubevirt_info{goversion="go1.12.8",kubeversion="{gitVersion}"} 1
# HELP kubevirt_vmi_memory_resident_bytes resident set size of the process running the domain.
# TYPE kubevirt_vmi_memory_resident_bytes gauge
kubevirt_vmi_memory_resident_bytes{domain="default_vmi-fedora",kubernetes_vmi_label_kubevirt_io_nodeName="node01",kubernetes_vmi_label_special="vmi-fedora",name="vmi-fedora",namespace="default",node="node01"} 4.34688e+08
# HELP kubevirt_vmi_network_errors_total network errors.
# TYPE kubevirt_vmi_network_errors_total counter
kubevirt_vmi_network_errors_total{domain="default_vmi-fedora",interface="vnet0",kubernetes_vmi_label_kubevirt_io_nodeName="node01",kubernetes_vmi_label_special="vmi-fedora",name="vmi-fedora",namespace="default",node="node01",type="rx"} 0
kubevirt_vmi_network_errors_total{domain="default_vmi-fedora",interface="vnet0",kubernetes_vmi_label_kubevirt_io_nodeName="node01",kubernetes_vmi_label_special="vmi-fedora",name="vmi-fedora",namespace="default",node="node01",type="tx"} 0
# HELP kubevirt_vmi_network_traffic_bytes_total network traffic.
# TYPE kubevirt_vmi_network_traffic_bytes_total counter
kubevirt_vmi_network_traffic_bytes_total{domain="default_vmi-fedora",interface="vnet0",kubernetes_vmi_label_kubevirt_io_nodeName="node01",kubernetes_vmi_label_special="vmi-fedora",name="vmi-fedora",namespace="default",node="node01",type="rx"} 13386
kubevirt_vmi_network_traffic_bytes_total{domain="default_vmi-fedora",interface="vnet0",kubernetes_vmi_label_kubevirt_io_nodeName="node01",kubernetes_vmi_label_special="vmi-fedora",name="vmi-fedora",namespace="default",node="node01",type="tx"} 5070
# HELP kubevirt_vmi_network_traffic_packets_total network traffic.
# TYPE kubevirt_vmi_network_traffic_packets_total counter
kubevirt_vmi_network_traffic_packets_total{domain="default_vmi-fedora",interface="vnet0",kubernetes_vmi_label_kubevirt_io_nodeName="node01",kubernetes_vmi_label_special="vmi-fedora",name="vmi-fedora",namespace="default",node="node01",type="rx"} 99
kubevirt_vmi_network_traffic_packets_total{domain="default_vmi-fedora",interface="vnet0",kubernetes_vmi_label_kubevirt_io_nodeName="node01",kubernetes_vmi_label_special="vmi-fedora",name="vmi-fedora",namespace="default",node="node01",type="tx"} 58

.
.
.
```

**Release note**:

```release-note
Adds kubernetes metadata.labels as VMI metrics' label
```

### DIFF
--- a/pkg/monitoring/vms/prometheus/prometheus.go
+++ b/pkg/monitoring/vms/prometheus/prometheus.go
@@ -17,703 +17,703 @@
  *
  */
 
- package prometheus
+package prometheus
 
- import (
-	 "fmt"
-	 "net/http"
-	 "strings"
-	 "time"
- 
-	 "github.com/prometheus/client_golang/prometheus"
-	 "github.com/prometheus/client_golang/prometheus/promhttp"
- 
-	 k6tv1 "kubevirt.io/client-go/api/v1"
-	 "kubevirt.io/client-go/kubecli"
-	 "kubevirt.io/client-go/log"
-	 "kubevirt.io/client-go/version"
-	 "kubevirt.io/kubevirt/pkg/util/lookup"
-	 cmdclient "kubevirt.io/kubevirt/pkg/virt-handler/cmd-client"
-	 "kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/stats"
- )
- 
- const statsMaxAge time.Duration = collectionTimeout + 2*time.Second // "a bit more" than timeout, heuristic again
- 
- var (
-	 // Formatter used to sanitize k8s metadata into metric labels
-	 labelFormatter = strings.NewReplacer(".", "_", "/", "_", "-", "_")
- 
-	 // see https://www.robustperception.io/exposing-the-software-version-to-prometheus
-	 versionDesc = prometheus.NewDesc(
-		 "kubevirt_info",
-		 "Version information",
-		 []string{"goversion", "kubeversion"},
-		 nil,
-	 )
- 
-	 // higher-level, telemetry-friendly metrics
-	 vmiCountDesc = prometheus.NewDesc(
-		 "kubevirt_vmi_phase_count",
-		 "VMI phase.",
-		 []string{
-			 "node", "phase",
-		 },
-		 nil,
-	 )
- 
-	 // lower level metrics
-	 storageIopsLabels = []string{"node", "namespace", "name", "drive", "type"}
-	 storageIopsDesc = prometheus.NewDesc(
-		 "kubevirt_vmi_storage_iops_total",
-		 "I/O operation performed.",
-		 storageIopsLabels,
-		 nil,
-	 )
- 
-	 storageTrafficLabels = []string{"node", "namespace", "name", "drive", "type"}
-	 storageTrafficDesc = prometheus.NewDesc(
-		 "kubevirt_vmi_storage_traffic_bytes_total",
-		 "storage traffic.",
-		 storageTrafficLabels,
-		 nil,
-	 )
- 
-	 storageTimesLabels = []string{"node", "namespace", "name", "drive", "type"}
-	 storageTimesDesc = prometheus.NewDesc(
-		 "kubevirt_vmi_storage_times_ms_total",
-		 "storage operation time.",
-		 storageTimesLabels,
-		 nil,
-	 )
- 
-	 vcpuUsageLabels = []string{"node", "namespace", "name", "id", "state"}
-	 vcpuUsageDesc = prometheus.NewDesc(
-		 "kubevirt_vmi_vcpu_seconds",
-		 "Vcpu elapsed time.",
-		 vcpuUsageLabels,
-		 nil,
-	 )
- 
-	 networkTrafficBytesLabels = []string{"node", "namespace", "name", "interface", "type"}
-	 networkTrafficBytesDesc = prometheus.NewDesc(
-		 "kubevirt_vmi_network_traffic_bytes_total",
-		 "network traffic.",
-		 networkTrafficBytesLabels,
-		 nil,
-	 )
- 
-	 networkTrafficPktsLabels = []string{"node", "namespace", "name", "interface", "type"}
-	 networkTrafficPktsDesc = prometheus.NewDesc(
-		 "kubevirt_vmi_network_traffic_packets_total",
-		 "network traffic.",
-		 networkTrafficPktsLabels,
-		 nil,
-	 )
- 
-	 networkErrorsLabels = []string{"node", "namespace", "name", "interface", "type"}
-	 networkErrorsDesc = prometheus.NewDesc(
-		 "kubevirt_vmi_network_errors_total",
-		 "network errors.",
-		 networkErrorsLabels,
-		 nil,
-	 )
- 
-	 memoryAvailableLabels = []string{"node", "namespace", "name"}
-	 memoryAvailableDesc = prometheus.NewDesc(
-		 "kubevirt_vmi_memory_available_bytes",
-		 "amount of usable memory as seen by the domain.",
-		 memoryAvailableLabels,
-		 nil,
-	 )
- 
-	 memoryResidentLabels = []string{"node", "namespace", "name"}
-	 memoryResidentDesc = prometheus.NewDesc(
-		 "kubevirt_vmi_memory_resident_bytes",
-		 "resident set size of the process running the domain",
-		 memoryResidentLabels,
-		 nil,
-	 )
- 
-	 swapTrafficLabels = []string{"node", "namespace", "name", "type"}
-	 swapTrafficDesc = prometheus.NewDesc(
-		 "kubevirt_vmi_memory_swap_traffic_bytes_total",
-		 "swap memory traffic.",
-		 swapTrafficLabels,
-		 nil,
-	 )
- )
- 
- func tryToPushMetric(desc *prometheus.Desc, mv prometheus.Metric, err error, ch chan<- prometheus.Metric) {
-	 if err != nil {
-		 log.Log.V(4).Warningf("Error creating the new const metric for %s: %s", memoryAvailableDesc, err)
-		 return
-	 }
-	 ch <- mv
- }
- 
- func updateMemory(vmi *k6tv1.VirtualMachineInstance, vmStats *stats.DomainStats, ch chan<- prometheus.Metric) {
-	 memoryResidentLabels = []string{"node", "namespace", "name"}
-	 memoryAvailableLabels = []string{"node", "namespace", "name"}
-	 swapTrafficLabels = []string{"node", "namespace", "name", "type"}
-	 var memoryResidentLabelValues = []string{vmi.Status.NodeName, vmi.Namespace, vmi.Name}
-	 var memoryAvailableLabelsValues = []string{vmi.Status.NodeName, vmi.Namespace, vmi.Name}
-	 var swapTrafficInLabelsValues = []string{vmi.Status.NodeName, vmi.Namespace, vmi.Name, "in"}
-	 var swapTrafficOutLabelsValues = []string{vmi.Status.NodeName, vmi.Namespace, vmi.Name, "out"}
- 
-	 // Add k8s metadata.Labels as metric labels
-	 labelPreffix := "vmi_k8s_label_"
-	 for label, val := range vmi.Labels {
-		 memoryResidentLabels = append(memoryResidentLabels, labelPreffix+labelFormatter.Replace(label))
-		 memoryResidentLabelValues = append(memoryResidentLabelValues, val)
- 
-		 memoryAvailableLabels = append(memoryAvailableLabels, labelPreffix+labelFormatter.Replace(label))
-		 memoryAvailableLabelsValues = append(memoryAvailableLabelsValues, val)
- 
-		 swapTrafficLabels = append(swapTrafficLabels, labelPreffix+labelFormatter.Replace(label))
-		 swapTrafficInLabelsValues = append(swapTrafficInLabelsValues, val)
-		 swapTrafficOutLabelsValues = append(swapTrafficOutLabelsValues, val)
-	 }
- 
-	 // Add k8s metadata.Annotations as metric labels
-	 annotationPreffix := "vmi_k8s_annotation_"
-	 for annotation, val := range vmi.Annotations {
-		 memoryResidentLabels = append(memoryResidentLabels, annotationPreffix+labelFormatter.Replace(annotation))
-		 memoryResidentLabelValues = append(memoryResidentLabelValues, val)
- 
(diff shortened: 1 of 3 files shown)


ArthurSens:
Currently working on splitting k8s_labels into multiple labels, e.g.:

VMI has the following labels:
```yaml
kubevirt.io/nodeName: node01
special: vmi-fedora
```
current label:
`k8s_labels="kubevir.io/nodeName=node01,secial=vmi-fedora"`

desired labels:
`k8s_label_kubevirt_io_nodeName="node01", k8s_label_special="vmi-fedora"`

ArthurSens:
Currently getting the following error on virt-handler logs

```json
{
	"component": "virt-handler",
	"level": "warning",
	"msg": "Error creating the new const metric for Desc{fqName: \"kubevirt_vmi_memory_available_bytes\", help: \"amount of usable memory as seen by the domain.\", constLabels: {}, variableLabels: [node namespace name vmi_k8s_label_kubevirt_io_nodeName vmi_k8s_label_special vmi_k8s_annotation_kubevirt_io_latest_observed_api_version vmi_k8s_annotation_kubevirt_io_storage_observed_api_version]}: duplicate label names",
	"pos": "prometheus.go:148",
	"timestamp": "2020-06-23T19:04:06.091264Z"
}
```

Running kubevirt and the [fedora example](https://github.com/kubevirt/kubevirt/blob/master/examples/vmi-fedora.yaml)

ArthurSens:
Getting the following error at virt-handler logs at the moment

```json
{
	"component": "virt-handler",
	"level": "warning",
	"msg": "Error creating the new const metric for Desc{fqName: \"kubevirt_vmi_memory_available_bytes\", help: \"amount of usable memory as seen by the domain.\", constLabels: {}, variableLabels: [node namespace name vmi_k8s_label_kubevirt_io_nodeName vmi_k8s_label_special vmi_k8s_annotation_kubevirt_io_latest_observed_api_version vmi_k8s_annotation_kubevirt_io_storage_observed_api_version]}: inconsistent label cardinality: expected 9 label values but got 10 in []string{\"node01\", \"default\", \"vmi-fedora\", \"vdb\", \"read\", \"node01\", \"vmi-fedora\", \"v1alpha3\", \"v1alpha3\", \"v1alpha3\"}",
	"pos": "prometheus.go:148",
	"timestamp": "2020-06-23T19:30:37.315607Z"
}
```

Somehow passing `\"vdb\", \"read\"` as label values for available_memory metric... still couldn't find out where exactly

pkg/monitoring/vms/prometheus/prometheus.go
```diff
@@ -161,34 +152,91 @@ func tryToPushMetric(desc *prometheus.Desc, mv prometheus.Metric, err error, ch
...
+	memoryResidentLabels = []string{"node", "namespace", "name"}
+	memoryAvailableLabels = []string{"node", "namespace", "name"}
+	swapTrafficLabels = []string{"node", "namespace", "name", "type"}
+	var memoryResidentLabelValues = []string{vmi.Status.NodeName, vmi.Namespace, vmi.Name}
+	var memoryAvailableLabelsValues = []string{vmi.Status.NodeName, vmi.Namespace, vmi.Name}
+	var swapTrafficInLabelsValues = []string{vmi.Status.NodeName, vmi.Namespace, vmi.Name, "in"}
+	var swapTrafficOutLabelsValues = []string{vmi.Status.NodeName, vmi.Namespace, vmi.Name, "out"}
+
+	// Add k8s metadata.Labels as metric labels
+	labelPreffix := "vmi_k8s_label_"
```
rmohr, line 164: Does this follow the pattern which we see on pod metrics in k8s which are exposed?

ArthurSens: Particularly, I've never seen this use-case anywhere before... Adding pod metadata to be exposed by another one.

So I can't say what is the best practice here

rmohr: Did you check the kubelet metrics?

ArthurSens: Are they supposed to be like this example?

image

`__meta_kubernetes_vmi_annotation_...` for annotations and `__meta_kubernetes_vmi_label_...` for labels

ArthurSens: or maybe 
`__meta_kubevirt_vmi`

ArthurSens: After some failed implementation and some research, I've noticed that it's not possible to add `__meta` labels to Prometheus metrics.

These sorts of labels are [reserved to Prometheus service-discoveries](https://github.com/prometheus/prometheus/tree/master/discovery#mapping-from-sd-to-prometheus), therefore I think the best approach would be `kubernetes_vmi_label_...` and `kubernetes_vmi_annotation...`

Maybe in the future, we could propose a Kubevirt service-discovery to Prometheus

ArthurSens:
@rmohr I didn't like declaring metrics descriptors at the beginning and then overriding everything at the update functions... but I couldn't think of a better solution since the `Describe` function needs all those descriptors as well

Any tips about that? Should I leave it as it is?

fabiand:
Two notes:
- `kubernetes_vmi_label_...` should it be `kubevirt_vmi_label_...`
- What about backwards compaibitility?

ArthurSens:
> * `kubernetes_vmi_label_...` should it be `kubevirt_vmi_label_...`?

My thought was that `metadata.label` and `metadata.annotation` is a Kubernetes thing, so that's why I named it that way, but I have no problems to change it if you guys think that `kubevirt_...` is better

> * What about backwards compaibitility?

This PR actually removes the `domain` label, as requested at #3477 and is also removed at the PR #3500 

But as we discussed at today's meeting, we shouldn't remove that label without a sending deprecation note before.

I think we should close #3500, I should re-add the domain label to this PR and then remove it in a future PR after the deprecation note, is that correct?

dhiller:
@rmohr any objections here? Otherwise I'm going to approve it tomorrow.

pkg/monitoring/vms/prometheus/prometheus.go
```diff
@@ -40,6 +40,18 @@ import (
...
 
 var (
+	// Formatter used to sanitize k8s metadata into metric labels
+	labelFormatter = strings.NewReplacer(".", "_", "/", "_", "-", "_")
+
+	// Preffixes used when transforming K8s metadata into metric labels
+	labelPreffix      = "kubernetes_vmi_label_"
+	annotationPreffix = "kubernetes_vmi_annotation_"
+
+	k8sLabels           = []string{}
```
xpivarc, line 50: Doesn't this cause a data race?

pkg/monitoring/vms/prometheus/prometheus.go
```diff
@@ -59,136 +71,105 @@ var (
...
-	networkErrorsDesc = prometheus.NewDesc(
-		"kubevirt_vmi_network_errors_total",
-		"network errors.",
-		[]string{
-			"node", "namespace", "name",
-			"domain", "interface", "type",
-		},
+
+	// Metrics descriptors used at the Describe function
+	storageIopsDesc         = prometheus.NewDesc("kubevirt_vmi_storage_iops_total", "", nil, nil)
```
xpivarc, line 76: Also here. As we spawn goroutine per vmi on node.

pkg/monitoring/vms/prometheus/prometheus.go
ArthurSens: Could you take a look and point if this race condition was removed on this last commit?

pkg/monitoring/vms/prometheus/prometheus.go
ArthurSens: I see the race condition here, but I'm not sure how to solve this :confused: 

Any suggestions?

ArthurSens: I know they shouldn't be global variables... but how can I use them at the `Describe` function and also update them at the `update...` functions?

ArthurSens:
Since the last commit done 2 weeks ago, I've been studying more about kubernetes, prometheus and I also tried to get involved with [SIG-Obeservability](https://github.com/cncf/sig-observability).

I've opened [this issue](https://github.com/cncf/sig-observability/issues/18) on their repository to ask for guidelines about some design problems concerning new metrics and labels.

About the usage of labels:

> All your examples and questions are very fair, but the rule of thumb is quite easy here: Just think if you would be grouping the aggregations across those labels. If yes, put it in the label, if no then the metric name is fine. Remember that a new label means a bigger index, so some overhead and cost again.

When adding annotations as new labels, I thought of a use case like this one below. One would add an annotation with contact information in case of an alert, which in this case `dbas@myorg.com` would be notified for database alerts
```yaml
---
kind: VirtualMachineInstance
metadata:
  annotations:
    resposibleEmailGroup: dbas@myorg.com
  labels:
    type: database
    environment: production
  name: vmi-ephemeral
spec:
.
.
.
```
But this can be easily configured in AlertManager with only the `type: database` label, thus, that annotation will only create an extra cost without a real benefit.

Also, a lot of annotations are often created when working with third-party tools, even Kubernetes itself add annotations to pods when using Deployments and Replicasets for example, which will also cost more capacity at the metrics store.

With that being said, and thinking about "what would be used to aggregate metric values",  I'll remove `metadata.annotations` from this PR, so only `metadata.labels` will be added as VMI metrics' labels.

pkg/monitoring/vms/prometheus/prometheus.go
rmohr: Why do you need them to be globally defined if you override them later on anyway per metric which you push?

ArthurSens:
@xpivarc I think the race conditions were solved and a functional test was added. Do you mind taking a look again?

pkg/monitoring/vms/prometheus/prometheus.go
```diff
@@ -384,6 +444,23 @@ func updateVersion(ch chan<- prometheus.Metric) {
 	)
 }
 
+type vmiMetrics struct {
```
xpivarc, line 447: I think you actually don't need this and all these Desc can be local. Please ignore this if I missed some cases.

Otherwise, it seems good to me. Please just squash commits and ping me :+1:

pkg/monitoring/vms/prometheus/prometheus.go
```diff
@@ -40,6 +40,13 @@ import (
 const statsMaxAge time.Duration = collectionTimeout + 2*time.Second // "a bit more" than timeout, heuristic again
 
 var (
+
+	// Formatter used to sanitize k8s metadata into metric labels
+	labelFormatter = strings.NewReplacer(".", "_", "/", "_", "-", "_")
+
+	// Preffixes used when transforming K8s metadata into metric labels
+	labelPreffix = "kubernetes_vmi_label_"
```
dhiller, line 48: nit: I think you mean 'prefix'

PR #3640, opened 2020-06-24
Fix automation timestamps

danielBelenky:
The previous iteration of this code was missing the pipefail option. It
ensures that the exit status of the pipe that passes stdout/stderr to
the timestamps function is propogated instead of being hidden behind the
timestamps function that can never fail.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/timestamps.sh
+++ b/hack/timestamps.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+function timestamps::time_wrapper() {
+    while IFS= read -r line; do
+        printf "%(%T)T: %s\n" "-1" "$line"
+    done
+}
+
+if [[ "${BASH_SOURCE[0]}" -ef "$0" ]]; then
+    set -o pipefail
+    /bin/sh "$@" 2>&1 | timestamps::time_wrapper
+fi
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 export GO15VENDOREXPERIMENT := 1
 
+SHELL = ./hack/timestamps.sh
 
 all:
 	hack/dockerized "DOCKER_PREFIX=${DOCKER_PREFIX} DOCKER_TAG=${DOCKER_TAG} IMAGE_PULL_POLICY=${IMAGE_PULL_POLICY} VERBOSITY=${VERBOSITY} ./hack/build-manifests.sh && \


hack/timestamps.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+function timestamps::time_wrapper() {
+    while IFS= read -r line; do
+        printf "%(%T)T: %s\n" "-1" "$line"
+    done
+}
+
+if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
+    set -o pipefail
```
EdDev, line 10: We will have to monitor how this behaves, I am unclear how it will affect the existing scripts in general.

jean-edouard:
@danielBelenky looks like the timestamp breaks travis-ci.
However, thanks to a couple flaky tests, we can see the lanes fail properly :)

EdDev:
> @danielBelenky looks like the timestamp breaks travis-ci.
> However, thanks to a couple flaky tests, we can see the lanes fail properly :)

@jean-edouard any idea why? Can you help with this?

hack/timestamps.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+function timestamps::time_wrapper() {
+    while IFS= read -r line; do
+        printf "%(%T)T: %s\n" "-1" "$line"
+    done
+}
+
+if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
+    /bin/sh "$@" > >(timestamps::time_wrapper) 2> >(timestamps::time_wrapper)
```
dollierp, line 10: stderr redirection should be improved to keep stdout and stderr separated:

`2> >(timestamps::time_wrapper >&2)`

danielBelenky: Nice catch! fixed, thanks.

danielBelenky:
@jean-edouard I had to go back to the pipe approach because the global redirection of stdout/stderr caused scripts that consume positional arguments from the cli to fail because the commands were first redirected to the timestamps function and then passed back to the tools.

hack/timestamps.sh
oshoval: return ${PIPESTATUS[0]} worth trying , it should preserve the original retcode, no need to change to pipefail if so, which might change the behavior as you said

PR #3642, opened 2020-06-24
Update VMI status with latest Pod network IP

vatsalparekh:

This adds a way to update VMI Status with latest Pod network IP for Masquerade bindings

Fixes #
https://bugzilla.redhat.com/show_bug.cgi?id=1686208

Discussion for the same: https://groups.google.com/forum/#!topic/kubevirt-dev/rUPRsG4CqqU
Old PR which had to be reverted with the same changes as this one: #3488 
**Release note**:

```release-note
Add a way to update VMI Status with latest Pod IP for Masquerade bindings
```


AlonaKaplan:
@vatsalparekh Maybe the `podInterfaceCache` is done too early. The cleanup should be done only when `'vmi.Status.NodeName != "" && vmi.Status.NodeName != d.host'` is true. Otherwise is can cause the `updateVMIStatus` on the source vm to stop working too early.

pkg/virt-handler/vm.go
```diff
@@ -448,6 +450,10 @@ func (d *VirtualMachineController) updateVMIStatus(vmi *v1.VirtualMachineInstanc
 	} else if vmi.Status.NodeName != "" && vmi.Status.NodeName != d.host {
 		// Only update the VMI's phase if this node owns the VMI.
 		// not owned by this host, likely the result of a migration
+
+		// Clear the pod network cache files now because we're sure VMI doesn't exist here anymore
```
AlonaKaplan, line 454: You sould verify `isPreMigrationTarget` is not true. So you won't cleanup the prepared by the target during a migration.

pkg/virt-handler/vm.go
```diff
@@ -334,9 +340,23 @@ func (d *VirtualMachineController) hasTargetDetectedDomain(vmi *v1.VirtualMachin
 
 func (d *VirtualMachineController) clearPodNetworkPhase1(vmi *v1.VirtualMachineInstance) {
```
enp0s3, line 341: @vatsalparekh looks like you don't need VMI object but only its UID, you can change the function arguments

pkg/virt-handler/vm.go
```diff
@@ -1299,6 +1385,12 @@ func (d *VirtualMachineController) processVmCleanup(vmi *v1.VirtualMachineInstan
 
 	vmiId := string(vmi.UID)
 
+	// Don't process if the vmi uid is nil
```
enp0s3, line 1388: @vatsalparekh If we exit here, vmGracefulShutdownTriggerClear won't be called but it doesn't need UID, maybe we should first do the graceful shutdown cleanup and then check for UID validity.

AlonaKaplan: Copying from the chat - 
IIUC, we can get to the situation there is no uid on the vmi only if the vmi and the domain don't exist on the virt-handler informers and also not in the GhostRecordCache. The vmi is removed from the GhostRecordCache during the cleanup. So it means cleanup was already done. So I think it is safe to return from the [execute](https://github.com/kubevirt/kubevirt/blob/ac98c6b523cd88223173c89c2e8dbf244d3744f4/pkg/virt-handler/vm.go#L1240).

@enp0s3 IIUC, if we get to here with no uid, it means the cleanup was already invoked.

pkg/virt-handler/vm.go
```diff
@@ -389,11 +411,49 @@ func domainMigrated(domain *api.Domain) bool {
 	return false
 }
 
+func (d *VirtualMachineController) getPodInterfacefromFileCache(vmi *v1.VirtualMachineInstance, ifaceName string) (*network.PodCacheInterface, error) {
```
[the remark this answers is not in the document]
vatsalparekh: Changed

pkg/virt-handler/vm.go
```diff
@@ -1235,9 +1321,13 @@ func (d *VirtualMachineController) execute(key string) error {
 			if uid != "" {
 				log.Log.Object(vmi).V(3).Infof("watchdog file provided %s as UID", uid)
 				vmi.UID = types.UID(uid)
+			} else {
+				// If UID is still not found, cleanup would already have been done
+				return nil
```
davidvossel, line 1326: I'm hesitant to change this here. There could be unintentional side effects. It's possible this is the right thing, but i'm not sure it's worth tackling how to handle VMI's with no UID in this PR. 

in general you shouldn't have to worry about this. If the UID doesn't exist, it's because the VMI has already been deleted and cleaned up. The only code paths at this point involve clean up.  Just make sure the cleanup logic is resilient to UID possibly not existing. 

basically, no UID means that part of the cleanup is a no-op. Take ```err = d.containerDiskMounter.Unmount(vmi)
``` for example, that does nothing if the UID is empty.

**Comment by AlonaKaplan** [comment]
> the `should move VirtualMachineInstance to Failed if configuring the networks on the virt-launcher fails` never actually worked. What actually made it pass wasn't the fact that the `network.SetupPodNetworkPhase1` function returned an error, it was because we detected an "unresponsive cmd client" `isUnresponsive, err := d.isLauncherClientUnresponsive(vmi) `
> 
> that test was unfortunately flawed. network.SetupPodNetworkPhase1 isn't actually called, and I don't fully understand why. @AlonaKaplan might need to help you debug this.

@davidvossel I will debug it.  @vatsalparekh, I don't want it to block your PR. Thanks for bringing it to my attention, I will fix it in a separate PR.

File: pkg/virt-handler/vm.go (hunk 5 above)
Comment (vatsalparekh) [reply pkg/virt-handler/vm.go]: uid check in Cleanup also look okay to me
Changed it now

File: pkg/virt-handler/vm_test.go [hunk 6]
```diff
@@ -804,21 +813,24 @@ var _ = Describe("VirtualMachineInstance", func() {
...
-			vmiFeeder.Add(vmi)
-
-			network.SetupPodNetworkPhase1 = func(vm *v1.VirtualMachineInstance, pid int) error { return fmt.Errorf("SetupPodNetworkPhase1 error") }
-
-			vmiInterface.EXPECT().Update(gomock.Any()).Do(func(vmi *v1.VirtualMachineInstance) {
-				Expect(vmi.Status.Phase).To(Equal(v1.Failed))
-			})
-			controller.Execute()
-		})
+		// TODO: Fix the test to work as expected; Assigning network.SetupPodNetworkPhase1 doesn't seem functional right now
```
vatsalparekh, line 816: @davidvossel Commented this test as we found that this never worked as expected and blocks this PR without reason.
@AlonaKaplan agreed she would fix this in a separate PR

pkg/virt-handler/vm.go
```diff
@@ -367,11 +373,25 @@ func (d *VirtualMachineController) hasTargetDetectedDomain(vmi *v1.VirtualMachin
 	return false, timeLeft
 }
 
-func (d *VirtualMachineController) clearPodNetworkPhase1(vmi *v1.VirtualMachineInstance) {
+func (d *VirtualMachineController) clearPodNetworkPhase1(uid types.UID) {
```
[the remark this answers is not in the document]
vatsalparekh: This was added to the processCleanup func from where this func was called
https://github.com/kubevirt/kubevirt/pull/3642/files#diff-27207346391eaa19190c76760eb8851fR1432

But don't mind moving it to just this function

pkg/virt-handler/vm.go
```diff
@@ -367,11 +373,25 @@ func (d *VirtualMachineController) hasTargetDetectedDomain(vmi *v1.VirtualMachin
...
+	for key, _ := range d.podInterfaceCache {
+		if strings.Contains(key, string(uid)) {
+			delete(d.podInterfaceCache, key)
+		}
+	}
+	d.podInterfaceCacheLock.Unlock()
 
-	delete(d.phase1NetworkSetupCache, vmi.UID)
+	vmiIfaceDir := fmt.Sprintf(virtutil.VMIInterfaceDir, uid)
+	err := os.RemoveAll(vmiIfaceDir)
```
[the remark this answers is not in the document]
vatsalparekh: Same as above comment, this was handled in the function calling this func https://github.com/kubevirt/kubevirt/pull/3642/files#diff-27207346391eaa19190c76760eb8851fR1432

But now moved to starting of this func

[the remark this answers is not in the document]
davidvossel: yay, that's a little safer I think, since we might move this logic around some day.

PR #3646, opened 2020-06-24
label tests of connectivity over masquerade

phoracek:

When KubeVirt is build on an base image that uses nftables but is
deployed on a cluster where some of the nodes have older iptables,
masquerade binding does not work (it can't call to older iptables on
hosts from newer nftables).

With this patch, we label tests that exercise connectivity over
masquerade binding so they can be skipped on environments where nftables
and iptables are mixed. For these environments, only the bridge binding
is relevant.

These tests can be skipped by passing an environment variable to the
test run:

`FUNC_TEST_ARGS="--ginkgo.skip=\[label:masquerade_binding_connectivity\]" make functest`

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/expose_test.go
+++ b/tests/expose_test.go
@@ -63,7 +63,7 @@ var _ = Describe("[rfe_id:253][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 		Context("Expose ClusterIP service", func() {
 			const servicePort = "27017"
 			const serviceName = "cluster-ip-vmi"
-			It("[test_id:1531]Should expose a Cluster IP service on a VMI and connect to it", func() {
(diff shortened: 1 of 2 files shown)


AlonaKaplan:
@phoracek most of our tests are using masquerade binding, since the [NewRandomVMIWithNS](https://github.com/kubevirt/kubevirt/blob/a897c7bc904e42fcff92a1f9480acd580f3751db/tests/utils.go#L1868 since) which is the base of most of the test vms is using masquerade. Are you sure the tests in the PR are the only failing (actually need network connectivity) tests?

phoracek:
@AlonaKaplan these are the ones that were reported to fail on a specific setup with nftables in containers and iptables on hosts. We do use masquerade in other places, but these tests seems to be the ones that check connectivity over it.

PR #3681, opened 2020-06-25
Rename chroot to virt-chroot

jfehlig:
During various experiments building the kubevirt components, a scenario
was encountered where a post-build file checker complained about
cmd/chroot not behaving like the system chroot. Since the interface is
quite different than the system chroot, rename it to virt-chroot.

Signed-off-by: Jim Fehlig <jfehlig@suse.com>

It renames cmd/chroot to cmd/virt-chroot to make it clear that it is different from the system chroot.

**Release note**:

```release-note
NONE
```


rmohr:
/lgtm

@jfehlig thanks. Let's see if we catched all.

PR #3692, opened 2020-06-27
Get QEMU logs from virtlogd

### LINKED ISSUE #2953: Evacuate QEMU logs from virt-launcher pod

>
> /kind bug
/kind enhancement

**What happened**:
A virt-launcher pod is being deleted because the QEMU crashed. This happens fast and if "spec.running" is True it happens in loops. Currently there is no option to treat the QEMU logs.

Print QEMU logs to the STDOUT and evacuate them to virt-handler for bug tracking.

Run Windows 2019 VM and enable kernel debugging via Network
In order to do so run the following commands as administrator from the command prompt:
bcdedit /debug on
bcdedit /dbgsettings net hostip:192.168.0.12 port:50000 key:this.key.isnt.secure

victortoso:
* It reads the logs from /var/log/libvirt/qemu/domain.log (which is handled by virtlogd), structuralizes it and log them with  kubevirt.io/client-go/log so it is reachable outside the pod

Fixes #2953 

This is a WIP, I want to double check that this is the right direction. I'm just reading from the file created by virtlogd for now.

**Release note**:

```release-note
QEMU logs can now be fetched from outside the pod
```

### DIFF
--- a/pkg/virt-launcher/virtwrap/util/libvirt_helper.go
+++ b/pkg/virt-launcher/virtwrap/util/libvirt_helper.go
@@ -230,7 +230,7 @@ func StartLibvirt(stopChan chan struct{}) {
 	}()
 }
 
-func StartVirtlog(stopChan chan struct{}) {
+func StartVirtlog(stopChan chan struct{}, domainName string) {
 	go func() {
 		for {
 			var args []string
@@ -245,6 +245,36 @@ func StartVirtlog(stopChan chan struct{}) {
 				log.Log.Reason(err).Error("failed to start virtlogd")
 				panic(err)
 			}
+
+			go func() {
+				logfile := fmt.Sprintf("/var/log/libvirt/qemu/%s.log", domainName)
+
+				// It can take a few seconds to the log file to be created
+				for {
+					_, err = os.Stat(logfile)
+					if !os.IsNotExist(err) {
+						break
+					}
+					time.Sleep(time.Second)
+				}
+
+				file, err := os.Open(logfile)
+				if err != nil {
+					log.Log.Reason(err).Error("failed to catch virtlogd logs")
+					return
+				}
+				defer file.Close()
+
+				scanner := bufio.NewScanner(file)
+				scanner.Buffer(make([]byte, 1024), 512*1024)
+				for scanner.Scan() {
+					log.LogQemuLogLine(log.Log, scanner.Text())
+				}
+
+				if err := scanner.Err(); err != nil {
+					log.Log.Reason(err).Error("failed to read virtlogd logs")
+				}
+			}()
 
 			go func() {
 				defer close(exitChan)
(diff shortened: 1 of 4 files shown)


victortoso:
CC @enp0s3

pkg/virt-launcher/virtwrap/util/libvirt_helper.go
```diff
@@ -242,10 +242,50 @@ func StartVirtlog(stopChan chan struct{}) {
 
 			err := cmd.Start()
 			if err != nil {
-				log.Log.Reason(err).Error("failed to start libvirtd")
+				log.Log.Reason(err).Error("failed to start virtlogd")
 				panic(err)
 			}
 
+			go func() {
+				logfile := fmt.Sprintf("/var/log/libvirt/qemu/%s.log", domainName)
```
rmohr, line 250: I think that it is possible to configure virtlogd to directly print to stdout. Would it be possible to read them directly from virtlogd?

enp0s3: @rmohr @victortoso Following my conversation with Daniel Berrange, the virtlogd can log to stderr only about it's own activity. Qemu logs will continue to be written to /var/log/libvirt/qemu. Even though we configure stdio_handler=logd in qemu.conf this pipe is only between qemu and virtlogd. As per my understanding from Daniel we cannot make virtlogd to redirect the incoming QEMU logging to something other than a file in /var/log/libvirt/qemu

rmohr: Fine by me then. That explains why my configuration attempts in the past had no influence.

pkg/virt-launcher/virtwrap/util/libvirt_helper.go
```diff
@@ -242,10 +242,50 @@ func StartVirtlog(stopChan chan struct{}) {
 
 			err := cmd.Start()
 			if err != nil {
-				log.Log.Reason(err).Error("failed to start libvirtd")
+				log.Log.Reason(err).Error("failed to start virtlogd")
 				panic(err)
 			}
 
+			go func() {
```
rmohr, line 249: Can we add some unit tests for this to `libvirt_helper_test.go` which transforms typical qemu logs to kubevirt logs?

pkg/virt-launcher/virtwrap/util/libvirt_helper.go
victortoso: Yep, same with me. I was about to send an email to libvirt folks to see what I was doing wrong. Thanks @enp0s3 for the quick check with Daniel.

pkg/virt-launcher/virtwrap/util/libvirt_helper.go
victortoso: Sure, I'll be adding that shortly. Thanks!

victortoso:
Besides adding a unit test, I changed the subcomponent to "qemu" as suggested by @enp0s3 and renamed the log function to reflect that.

rmohr:
Let's see what the tests say about the change. @victortoso thanks.

PR #3694, opened 2020-06-28
virt-launcher: remove redundant condition check

enp0s3:
A small fix that removes redundant condition check in virt-launcher main. 
err variable remains the same before and after call to util.StartLibvirt since
the function didn't return any value.

Fixes #

```release-note
None
```


rmohr:
@enp0s3 the PR got merged, but it ended up in an inconsistent state on the github side due to the ongoing degraded availability of github. Verified that the commits are on master and closed.

PR #3706, opened 2020-06-29
virt-controller: fix null pointer dereference in Pod Disruption Budget controller

enp0s3:
There was an attempt to read PDB name after it's deletion.
Instead of using it's name, use PDB key for recording the deletion event.

Prevent virt-controller process from a crash when recording a deletion event.

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/virt-controller/watch/drain/disruptionbudget/disruptionbudget.go
+++ b/pkg/virt-controller/watch/drain/disruptionbudget/disruptionbudget.go
@@ -362,7 +362,7 @@ func (c *DisruptionBudgetController) sync(key string, vmi *virtv1.VirtualMachine
 			c.recorder.Eventf(vmi, v12.EventTypeWarning, FailedDeletePodDisruptionBudgetReason, "Error deleting the PodDisruptionBudget %s: %v", pdb.Name, err)
 			return err
 		}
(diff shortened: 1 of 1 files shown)


pkg/virt-controller/watch/drain/disruptionbudget/disruptionbudget.go
```diff
@@ -362,7 +362,7 @@ func (c *DisruptionBudgetController) sync(key string, vmi *virtv1.VirtualMachine
 			c.recorder.Eventf(vmi, v12.EventTypeWarning, FailedDeletePodDisruptionBudgetReason, "Error deleting the PodDisruptionBudget %s: %v", pdb.Name, err)
 			return err
 		}
-		c.recorder.Eventf(vmi, v12.EventTypeNormal, SuccessfulDeletePodDisruptionBudgetReason, "Deleted PodDisruptionBudget %s", pdb.Name)
+		c.recorder.Eventf(vmi, v12.EventTypeNormal, SuccessfulDeletePodDisruptionBudgetReason, "Deleted PodDisruptionBudget %s", pdbKey)
```
rmohr, line 365: Could you add a unit test?

PR #3719, opened 2020-06-30
Moved cleanup of test from AfterEach() hook to test body

ashleyschuett:
When this test fails the AfterEach block is ran before the reporter
calls Dump(). This means we are unable to get the logs for the NFS
and the virt-launcher pods.

Signed-off-by: Ashley Schuett <ashleyns1992@gmail.com>

This allows us to get more logs for debuging if this test fails in the future 

Fixes #

It looks like the VMI clean up at one point was happening in the body of the test and was then moved to the `AfterEach()` block. https://github.com/kubevirt/kubevirt/commit/8646e5b851b62ebcd76f2e3833fa5ef2e9750132 

I'm unable to figure out the reason for this change however and it seems to be the standard in other test cases to clean up the VMI in the test body. 

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/migration_test.go
+++ b/tests/migration_test.go
@@ -768,7 +768,6 @@ var _ = Describe("[rfe_id:393][crit:high][vendor:cnv-qe@redhat.com][level:system
 
 				By("Waiting for VMI to disappear")
 				tests.WaitForVirtualMachineToDisappearWithTimeout(vmi, 120)
-
 			})
 		})
 		Context("with an Cirros shared ISCSI PVC", func() {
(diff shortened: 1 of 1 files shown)


rmohr:
> When this test fails the AfterEach block is ran before the reporter
> calls Dump(). This means we are unable to get the logs for the NFS
> and the virt-launcher pods.

I wonder if we should move the `Dump()` code to a [JustAfterEach](https://onsi.github.io/ginkgo/#separating-creation-and-configuration-justbeforeeach) and out of the reporter. It would make it possible again for people to use `AfterEach` more freely. What do you think?

ashleyschuett:
I think that makes sense. The only problem I see with that is if something fails during clean up then it could be hard for us to debug that

rmohr:
> I think that makes sense. The only problem I see with that is if something fails during clean up then it could be hard for us to debug that

How about wrapping the current FailHandler and do our dump in-between: https://onsi.github.io/ginkgo/#bootstrapping-a-suite ?

ashleyschuett:
If we wrap the Fail handler, and use the `RegisterFailHandler` function we then lose access to the spec summary from the test completion. In our case this means losing access to the length of the test that was ran which is how we are parsing to get the relevant portion of logs. https://github.com/kubevirt/kubevirt/blob/master/tests/reporter/kubernetes.go#L94

In order to create the kubernetes archives we are using a custom report and only really have access to the hooks specified here https://onsi.github.io/ginkgo/#writing-custom-reporters if we want to be able to leverage using the report as we are.

ashleyschuett:
@rmohr what are your thoughts on my assessment, am I missing something and it's actually easier than I'm understanding?

rmohr:
> @rmohr what are your thoughts on my assessment, am I missing something and it's actually easier than I'm understanding?

I think that with a combination of the reporter and  the fail handler, we could over come that (the reporter is called before the `It` is started, so one could collect a timestamp there and use this timestamp to collect the logs. Any way, the change is independent of that fine. Let's look into that later again.

PR #3729, opened 2020-07-01
[release-0.30] tests, vmi: Fix IPv6 configuration through the console

oshoval:
This is a manual cherry-pick of #3543 

configureIPv6OnVMI didn't actually update the /etc/resolv.conf

Updating DNS for cirros was added as nice to have only.
The tests that actually need it, use fedora VM, which use
cloud-init for this purpose.
Hence its safe and better to remove it.
Motivation and discovery was due to flaky test 3184.

Reasons that resolv.conf wasnt updated:

 Missing sudo su for cirros.
 Using of backtick malformatted the command [1]

Adjust the expected characters when examining the echo $? response
such that it will catch the actual retcode and not the a character
which is echoed back (i.e is part of the sent command itself).

[1] https://golang.org/ref/spec#String_literals

```release-note
None
```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -2807,19 +2807,23 @@ func CheckForTextExpecter(vmi *v1.VirtualMachineInstance, expected []expect.Batc
 }
 
 func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter, virtClient kubecli.KubevirtClient, prompt string) error {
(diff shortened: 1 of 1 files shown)


oshoval:
https://github.com/kubevirt/kubevirt/pull/3727
would fix the failure of 
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3729/pull-kubevirt-e2e-k8s-cnao-1.17/1278329543114887168

PR #3730, opened 2020-07-01
notify-client: Improve domain event error information

oshoval:
In case of a domain event error,
propagate the domain name and an application layer error,
which will be presented on the virt-handler side.
    
See #3806 which logs the error instead of resetting the
gRPC server.
    
```release-note
None
```

### DIFF
--- a/pkg/virt-launcher/notify-client/client.go
+++ b/pkg/virt-launcher/notify-client/client.go
@@ -200,8 +200,10 @@ func eventCallback(c cli.Connection, domain *api.Domain, libvirtEvent libvirtEve
 	d, err := c.LookupDomainByName(util.DomainFromNamespaceName(domain.ObjectMeta.Namespace, domain.ObjectMeta.Name))
 	if err != nil {
 		if !domainerrors.IsNotFound(err) {
-			log.Log.Reason(err).Error("Could not fetch the Domain.")
-			client.SendDomainEvent(newWatchEventError(err))
+			errMsg := "Could not fetch the Domain"
+			log.Log.Reason(err).Error(errMsg)
+			canNotFetchDomainError := fmt.Errorf("%s, %s, reason %v", errMsg, domain.Spec.Name, err)
+			client.SendDomainEvent(newWatchEventError(canNotFetchDomainError))
 			return
 		}
 		domain.SetState(api.NoState, api.ReasonNonExistent)
@@ -214,8 +216,10 @@ func eventCallback(c cli.Connection, domain *api.Domain, libvirtEvent libvirtEve
 		status, reason, err := d.GetState()
 		if err != nil {
 			if !domainerrors.IsNotFound(err) {
-				log.Log.Reason(err).Error("Could not fetch the Domain state.")
-				client.SendDomainEvent(newWatchEventError(err))
+				errMsg := "Could not fetch the Domain state"
+				log.Log.Reason(err).Error(errMsg)
+				canNotFetchStateError := fmt.Errorf("%s, %s, reason %v", errMsg, domain.Spec.Name, err)
+				client.SendDomainEvent(newWatchEventError(canNotFetchStateError))
 				return
 			}
 			domain.SetState(api.NoState, api.ReasonNonExistent)
@@ -227,8 +231,10 @@ func eventCallback(c cli.Connection, domain *api.Domain, libvirtEvent libvirtEve
 		if err != nil {
 			// NOTE: Getting domain metadata for a live-migrating VM isn't allowed
 			if !domainerrors.IsNotFound(err) && !domainerrors.IsInvalidOperation(err) {
-				log.Log.Reason(err).Error("Could not fetch the Domain specification.")
-				client.SendDomainEvent(newWatchEventError(err))
+				errMsg := "Could not fetch the Domain specification"
+				log.Log.Reason(err).Error(errMsg)
+				canNotFetchSpecError := fmt.Errorf("%s, %s, reason %v", errMsg, domain.Spec.Name, err)
+				client.SendDomainEvent(newWatchEventError(canNotFetchSpecError))
 				return
 			}
 		} else {
@@ -274,7 +280,7 @@ func eventCallback(c cli.Connection, domain *api.Domain, libvirtEvent libvirtEve
 		}
 		err := client.SendDomainEvent(watch.Event{Type: watch.Modified, Object: domain})
 		if err != nil {
-			log.Log.Reason(err).Error("Could not send domain notify event.")
+			log.Log.Reason(err).Errorf("Could not send %s domain notify event.", domain.Spec.Name)
 		}
 	}
 }
@@ -337,7 +343,7 @@ func (n *Notifier) StartDomainNotifier(
 				eventCallback(domainConn, domainCache, libvirtEvent{}, n, deleteNotificationSent,
 					interfaceStatuses, guestOsInfo)
 			case <-reconnectChan:
-				n.SendDomainEvent(newWatchEventError(fmt.Errorf("Libvirt reconnect")))
+				n.SendDomainEvent(newWatchEventError(fmt.Errorf("Libvirt reconnect, domain %s", domainName)))
 				return
 			}
 		}


pkg/virt-launcher/notify-client/client.go
```diff
@@ -196,12 +196,13 @@ func newWatchEventError(err error) watch.Event {
...
 
 func eventCallback(c cli.Connection, domain *api.Domain, libvirtEvent libvirtEvent, client *Notifier, events chan watch.Event,
-	interfaceStatus []api.InterfaceStatus, osInfo *api.GuestOSInfo) {
+	interfaceStatus []api.InterfaceStatus, osInfo *api.GuestOSInfo, domainName string) {
 	d, err := c.LookupDomainByName(util.DomainFromNamespaceName(domain.ObjectMeta.Namespace, domain.ObjectMeta.Name))
 	if err != nil {
 		if !domainerrors.IsNotFound(err) {
 			log.Log.Reason(err).Error("Could not fetch the Domain.")
-			client.SendDomainEvent(newWatchEventError(err))
+			errMessage := fmt.Errorf(fmt.Sprintf("Could not fetch the Domain %s, reason %s", domainName, err.Error()))
```
EdDev, line 204: `Errorf` is supposed to be similar to `Printf`, so I do not think you need `Sprintf`.
Also, I saw errors being formatted using %v and without `Error()` (e.g. `fmt.Errorf("%v", err)`)

The name is also not representing what this is. It is not a message, it is an error. You could call it something like `canNotFetchDomainError`.

Same for the others below.

pkg/virt-launcher/notify-client/client.go
```diff
@@ -196,12 +196,13 @@ func newWatchEventError(err error) watch.Event {
 }
 
 func eventCallback(c cli.Connection, domain *api.Domain, libvirtEvent libvirtEvent, client *Notifier, events chan watch.Event,
-	interfaceStatus []api.InterfaceStatus, osInfo *api.GuestOSInfo) {
+	interfaceStatus []api.InterfaceStatus, osInfo *api.GuestOSInfo, domainName string) {
```
EdDev, line 199: I think that with so many arguments, this merits a struct already. Everyone is adding "just" one and it becomes too big to manage/maintain/read.

Maybe you can add a small refactor commit to fix it.

pkg/virt-launcher/notify-client/client.go
```diff
@@ -274,7 +277,7 @@ func eventCallback(c cli.Connection, domain *api.Domain, libvirtEvent libvirtEve
 		}
 		err := client.SendDomainEvent(watch.Event{Type: watch.Modified, Object: domain})
 		if err != nil {
-			log.Log.Reason(err).Error("Could not send domain notify event.")
+			log.Log.Reason(err).Errorf("Could not send notify event of domain %s", domainName)
```
EdDev, line 280: How about "Could not send %s domain notify event." ?

pkg/virt-launcher/notify-client/notify_test.go
```diff
@@ -104,7 +104,9 @@ var _ = Describe("Notify", func() {
 				mockDomain.EXPECT().GetXMLDesc(gomock.Eq(libvirt.DomainXMLFlags(0))).Return(string(x), nil)
 				mockDomain.EXPECT().GetMetadata(libvirt.DOMAIN_METADATA_ELEMENT, "http://kubevirt.io", libvirt.DOMAIN_AFFECT_CONFIG).Return(`<kubevirt></kubevirt>`, nil)
 
-				eventCallback(mockCon, util.NewDomainFromName("test", "1234"), libvirtEvent{Event: &libvirt.DomainEventLifecycle{Event: event}}, client, deleteNotificationSent, nil, nil)
+				domainCache := util.NewDomainFromName("test", "1234")
```
EdDev, line 107: What is a "domainCache"? Where this name came from?

pkg/virt-launcher/notify-client/notify_test.go
```diff
@@ -104,7 +104,9 @@ var _ = Describe("Notify", func() {
...
+				eventCallback(mockCon, domainCache, libvirtEvent{Event: &libvirt.DomainEventLifecycle{Event: event}},
```
EdDev, line 108: I do not understand how the line got split here.
I suggest you either put each argument on a separate line or all in one line.
These are the options that usually fit:
```
foo(a, b, c)
foo(
  a, b, c,
)
foo(
  a,
  b,
  c,
)
```
I am surprised that `gofmt` is not taking care of this.
```

File: pkg/virt-launcher/notify-client/notify_test.go (hunk 4 above)
Comment (oshoval) [reply pkg/virt-launcher/notify-client/notify_test.go:107]: as the client does
https://github.com/kubevirt/kubevirt/blob/2d22cb6a9706e0ef0e1a9cc7b0b7fd1d57480730/pkg/virt-launcher/notify-client/client.go#L320

but in some cases it might be null, so i cant just use it in the last parameter.

Comment (EdDev) [reply pkg/virt-launcher/notify-client/notify_test.go:107]: I just see it returns a `domain`, I do not understand what makes it a `cache`.

File: pkg/virt-launcher/notify-client/client.go (hunk 2 above)
Comment (enp0s3) [reply pkg/virt-launcher/notify-client/client.go:199]: @oshoval Since *api.Domain already passed to the function I think you can invoke domain name from api.Domain.Spec

Comment (oshoval) [reply pkg/virt-launcher/notify-client/client.go:199]: @enp0s3 
Thanks but i am afraid it can be nil
see please
https://github.com/kubevirt/kubevirt/blob/2d22cb6a9706e0ef0e1a9cc7b0b7fd1d57480730/pkg/virt-launcher/notify-client/client.go#L333

Comment (enp0s3) [reply pkg/virt-launcher/notify-client/client.go:199]: @oshoval How come? here the domain cache is being created: https://github.com/kubevirt/kubevirt/blob/2d22cb6a9706e0ef0e1a9cc7b0b7fd1d57480730/pkg/virt-launcher/notify-client/client.go#L320
If it should be thread-safe lets protect it with lock, otherwise how does it become nil?

Comment (oshoval) [reply pkg/virt-launcher/notify-client/client.go:199]: its not about thread safe, 
its about 
https://github.com/kubevirt/kubevirt/blob/2d22cb6a9706e0ef0e1a9cc7b0b7fd1d57480730/pkg/virt-launcher/notify-client/client.go#L333

there are three select cases there in one infinite loop, the first channel that will happen will execute (in case two at the same time one is taken randomly, this is the select logic of go), and see please that this case even check its not nil

**Comment by oshoval** [comment]
As we spoke, i would separate this PR to 2 PRs
1. this one would be about the logs improving (since most of the comments here is about it)
2. PR that would update just server side, no just log instead of propagating the error further, eliminating the server restart upon client application layer errors.

File: pkg/virt-launcher/notify-client/client.go [hunk 6]
```diff
@@ -318,13 +328,19 @@ func (n *Notifier) StartDomainNotifier(
 
 	// Run the event process logic in a separate go-routine to not block libvirt
 	go func() {
-		var interfaceStatuses []api.InterfaceStatus
-		var guestOsInfo *api.GuestOSInfo
+		data := domainEventData{
```
EdDev, line 331: I suggest you build it in one shot and not in stages. Stages are prone to errors.
Redefine the data at the call/actual usuage.

pkg/virt-launcher/notify-client/client.go
```diff
@@ -195,18 +206,17 @@ func newWatchEventError(err error) watch.Event {
 	return watch.Event{Type: watch.Error, Object: &metav1.Status{Status: metav1.StatusFailure, Message: err.Error()}}
 }
 
-func eventCallback(c cli.Connection, domain *api.Domain, libvirtEvent libvirtEvent, client *Notifier, events chan watch.Event,
-	interfaceStatus []api.InterfaceStatus, osInfo *api.GuestOSInfo, domainName string) {
-	d, err := c.LookupDomainByName(util.DomainFromNamespaceName(domain.ObjectMeta.Namespace, domain.ObjectMeta.Name))
+func eventCallback(data *domainEventData) {
```
EdDev, line 209: Why do you use a pointer instead of a regular value like it was before?

pkg/virt-launcher/notify-client/client.go
oshoval: you dont have all the parameter in the first place, and some of the values needs to be saved from one stage to another, but i can try to rebuilt it in the two needed places from scratch, saving the shared values between the two occurences,
lemme suggest the diff,
indeed at the moment there is a bug there.

pkg/virt-launcher/notify-client/client.go
oshoval: in order to save data copy, i can use "by value" if needed

oshoval:
As we spoke, due to https://github.com/kubevirt/kubevirt/pull/3814
removed the new variable and the refactor commit (which i might open new PR for).

oshoval:
due to the changes at 
https://github.com/kubevirt/kubevirt/pull/3806
its look like this one might become not needed, and be part of 3806 commits.

PR #3734, opened 2020-07-02
Tests suite cleanup to always create artifacts dir

danielBelenky:
We used to create the artifacts directory at the very end of the suite
cleanup. The problem with this approach is that if something failed or
paniced along the way, we never actaully create the artifacts directory
and then, other Ginkgo reporters such as the JUnit reporter fail to
write the report.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -26,6 +26,7 @@ import (
 	"testing"
 
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 
 	"kubevirt.io/kubevirt/tests/reporter"
 
@@ -50,7 +51,12 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = AfterSuite(func() {
-	tests.AfterTestSuitCleanup()
+	// We have to create the artifacts directory before running any cleanup code
+	// because in case of a failure in one of the cleanup steps, we still want to
+	// be able to store artifacts.
+	artifactsDir, err := tests.CreateDirectory(os.Getenv("ARTIFACTS"))
+	Expect(err).ToNot(HaveOccurred())
+	tests.AfterTestSuitCleanup(artifactsDir)
 })
 
 func getMaxFailsFromEnv() int {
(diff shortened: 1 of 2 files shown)


tests/tests_suite_test.go
```diff
@@ -50,7 +51,12 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = AfterSuite(func() {
-	tests.AfterTestSuitCleanup()
+	// We have to create the artifacts directory before running any cleanup code
+	// because in case of a failure in one of the cleanup steps, we still want to
+	// be able to store artifacts.
+	artifactsDir, err := tests.CreateDirectory(os.Getenv("ARTIFACTS"))
```
ormergi, line 57: Artifacts directory created before starting the test suite, if we want to do it from within the test suite 
we should remove this line:
https://github.com/kubevirt/kubevirt/blob/master/automation/test.sh#L286

tests/tests_suite_test.go
```diff
@@ -50,7 +51,12 @@ var _ = BeforeSuite(func() {
...
+	Expect(err).ToNot(HaveOccurred())
```
ormergi, line 58: Could you please add error message somthing like `should successfully create artifacts directory..`

tests/utils.go
```diff
@@ -578,7 +566,21 @@ func GenerateRESTReport() error {
 	return nil
 }
 
-func AfterTestSuitCleanup() {
+// CreateDirectory creates the a directory on the given path if it doesn't exist.
+func CreateDirectory(artifactsDir string) (string, error) {
```
ormergi, line 570: I think this functuion name should be `CreateArtifactsDirectory`

danielBelenky: Why? It can create an arbitrary directory and it is in "utils.go"

rmohr:
> We used to create the artifacts directory at the very end of the suite
> cleanup.

We have https://github.com/kubevirt/kubevirt/blob/master/automation/test.sh#L286 which creates the artifacts dir before ginkgo. Are you sure that we fix something here?

tests/utils.go
xpivarc: Maybe change `artifactsDir` to `path`

danielBelenky:
The error is fixed by this PR: https://github.com/kubevirt/kubevirt/pull/3779

PR #3735, opened 2020-07-02
Exclude test suites from fetching changed tests

dhiller:

Related to the previous grep exit code 123 problem there exists another corner case
when a suite file has been changed, that actually does not contain tests
but was counted as a test that changed. This is now excluded.

Also minor refactorings:
* reusing the existing NEW_TESTS instead of again finding all tests files
* replacing the word count with a count in grep

Fixes #

/cc @danielBelenky FYI, your PR fails the lane check-tests-for-flakes due to the changes in the test suite file.

**Release note**:

```release-note
NONE
```


stu-gott:
/approve

failing tests are flaky and unrelated.

PR #3736, opened 2020-07-02
[release-0.30] Enable Secure Boot by default when EFI is enabled

jean-edouard:
Cherry-pick of https://github.com/kubevirt/kubevirt/pull/3529

```release-note
Enabling EFI will also enable Secure Boot, which requires SMM to be enabled.
To use EFI without Secure Boot, secureBoot must be set to false, and SMM is then optional.
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -4821,7 +4821,7 @@
     "type": "object",
     "properties": {
      "secureBoot": {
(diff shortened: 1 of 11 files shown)


stu-gott:
@jean-edouard the failure in 1.17 ... can you double check if that's related to this PR or not?

jean-edouard:
@stu-gott thanks for the heads-up. However it looks like it's just flakyness...
/retest

PR #3738, opened 2020-07-02
virt-api: validate labels with kubevirt.io prefix upon VMI creation

enp0s3:
While creating a VMI, some of the kubevirt.io prefixed labels should be restricted for creation.
These labels are meant to be used internally by the kubevirt components. 

To prevent situations that may lead to VMI state inconsistency. For example if it will be allowed to craft
kubevirt.io/nodename label with incorrect value, it will lead to inconsistent VMI state between the controller and the handler.

Fixes #
https://bugzilla.redhat.com/show_bug.cgi?id=1843456

```release-note
Restrict creation of VMI if it has labels that are used internally by Kubevirt components.
```


enp0s3:
> What about `kubevirt.io/ignitiondata`?

@xpivarc That's an annotation. The suggested restrictions are on labels.

xpivarc:
kubevirt.io/created-by

> > What about `kubevirt.io/ignitiondata`?
> 
> @xpivarc That's an annotation. The suggested restrictions are on labels.

Sorry,  I often take them as equals. Is there a reason to not use a blacklist instead of a whitelist?(As only one label can do damage now, AFAIK) We have `kubevirt.io/` labels in examples on official pages and this would break them. I would generally be more inclined to be less restrictive. WDYT?

Note: tests must be corrected because they use `kubevirt.io/` labels.

enp0s3:
> kubevirt.io/created-by
> 
> > > What about `kubevirt.io/ignitiondata`?
> > 
> > 
> > @xpivarc That's an annotation. The suggested restrictions are on labels.
> 
> Sorry, I often take them as equals. Is there a reason to not use a blacklist instead of a whitelist?(As only one label can do damage now, AFAIK) We have `kubevirt.io/` labels in examples on official pages and this would break them. I would generally be more inclined to be less restrictive. WDYT?
> 
> Note: tests must be corrected because they use `kubevirt.io/` labels.

@xpivarc Thanks for the notice about the tests. Indeed, I need to update the VMI preset test suite. @rmohr @stu-gott @fabiand What do you think? Should we use a blacklist or a whitelist?

enp0s3:
@xpivarc kubevirt.io/created-by is a label that should not be created by users, as mentioned [here](https://github.com/kubevirt/kubevirt/blob/5665b11cbd249edfd03da626860b653c1af29b69/staging/src/kubevirt.io/client-go/api/v1/types.go#L451) 
In general we need to decide whether we restrict the whole kubevirt.io domain or we can use subdomain like internal.kubevirt.io for internal use-cases

danielBelenky:
I think that we probably should separate the kubevirt labels that are used by kubevirt components and labels that are used by users. Just like the `kubernetes.io/` and `k8s.io/` prefixes are reserved for K8s core components. It makes this change more complicated and we should also consider a deprecation notice first, before we actually enforce this. We will need to support another type of label for user configuration. Or, maybe add a subname to the label? `kubevirt.io/core` for the core components and then the user can use anything else for configurations?

enp0s3:
> I think that we probably should separate the kubevirt labels that are used by kubevirt components and labels that are used by users. Just like the `kubernetes.io/` and `k8s.io/` prefixes are reserved for K8s core components. It makes this change more complicated and we should also consider a deprecation notice first, before we actually enforce this. We will need to support another type of label for user configuration. Maybe even consider moving our user-configured labels to proper CR fields?

@danielBelenky Looks like a temporary approach will be to do a blacklist for internal labels, and bring up this subject to community discussion for a long term solution, WDYT?

danielBelenky:
> > I think that we probably should separate the kubevirt labels that are used by kubevirt components and labels that are used by users. Just like the `kubernetes.io/` and `k8s.io/` prefixes are reserved for K8s core components. It makes this change more complicated and we should also consider a deprecation notice first, before we actually enforce this. We will need to support another type of label for user configuration. Maybe even consider moving our user-configured labels to proper CR fields?
> 
> @danielBelenky Looks like a temporary approach will be a blacklist for internal labels, and bring up this subject to community discussion for a long term solution, WDYT?

Let's block labels that we know that cause an issue for now. About the other `kubevirt.io/` labels, since it's a user facing API, we should be more careful with it. I think that once we decide on a plan (i.e: replacing `kubevirt.io/` with some other label, moving those settings to a fields on the CR, ...) we can share it with the community and gather feedback.

enp0s3:
@danielBelenky @xpivarc I've updated the PR with the new relaxed approach, now we allow creation unless we hit the list of internal labels.

pkg/util/util.go
```diff
@@ -1,7 +1,13 @@
 package util
 
 import (
+	"fmt"
+	"strings"
+
 	v1 "kubevirt.io/client-go/api/v1"
+	"kubevirt.io/client-go/log"
+	clientutil "kubevirt.io/client-go/util"
+	"kubevirt.io/kubevirt/pkg/virt-operator/creation/rbac"
```
rmohr, line 10: Could you leave the rbac related helper function in the webhooks packages? This here creates a binding across the whole project to the operator otherwise. Can make changes difficult.

pkg/util/util.go
```diff
@@ -30,3 +36,34 @@ func IsGPUVMI(vmi *v1.VirtualMachineInstance) bool {
 	}
 	return false
 }
+func FilterKubevirtLabels(labels map[string]string) map[string]string {
```
rmohr, line 39: This seems to be only relevant for the webhooks. Best to leave it close to it. Also, you already go over the labels. Can't you just apply your filter check here right now and only return labels which we care about?

rmohr:
> I think that we probably should separate the kubevirt labels that are used by kubevirt components and labels that are used by users. Just like the `kubernetes.io/` and `k8s.io/` prefixes are reserved for K8s core components. It makes this change more complicated and we should also consider a deprecation notice first, before we actually enforce this. We will need to support another type of label for user configuration. Maybe even consider moving our user-configured labels to proper CR fields?

Let's tackle it like k8s does. k8s says you should not use labels with such a prefix since it is reserved for k8s itself but it is not blocking you. I think we should block internally used labels to avoid surprises, but let's just ignore the rest. We should probably still write to kubevirt-dev and update the docs that people should not use them, even if we don't block it. Ideally our components would also be able to handle weird entities and not crash at least.

@enp0s3 @danielBelenky what do you think?

pkg/util/util.go
enp0s3: @rmohr Done

pkg/util/util.go
enp0s3: @rmohr because of it's usage in VMI update admitter as well, FilterKubevirtLabels should return all the kubevirt.io labels. But I've added a boolean to indicate whether we have restricted labels among all the kubevirt.io related.

enp0s3:
> > I think that we probably should separate the kubevirt labels that are used by kubevirt components and labels that are used by users. Just like the `kubernetes.io/` and `k8s.io/` prefixes are reserved for K8s core components. It makes this change more complicated and we should also consider a deprecation notice first, before we actually enforce this. We will need to support another type of label for user configuration. Maybe even consider moving our user-configured labels to proper CR fields?
> 
> Let's tackle it like k8s does. k8s says you should not use labels with such a prefix since it is reserved for k8s itself but it is not blocking you. I think we should block internally used labels to avoid surprises, but let's just ignore the rest. We should probably still write to kubevirt-dev and update the docs that people should not use them, even if we don't block it. Ideally our components would also be able to handle weird entities and not crash at least.
> 
> @enp0s3 @danielBelenky what do you think?

@rmohr @danielBelenky sounds legit to me.

pkg/util/util.go
rmohr: I don't think that it makes sense to forbid only a subset of labels on object creation but later on update block updating labels which were not blocked in the first place. What do you think?

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-update-admitter.go
```diff
@@ -124,3 +111,23 @@ func getAllowedServiceAccounts() map[string]struct{} {
 		fmt.Sprintf("%s:%s", prefix, rbac.ControllerServiceAccountName): {},
 	}
 }
+
+func FilterKubevirtLabels(labels map[string]string) (map[string]string, bool) {
```
danielBelenky, line 115: Should it be exported?

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
```diff
@@ -64,6 +64,15 @@ var validInterfaceModels = map[string]*struct{}{"e1000": nil, "e1000e": nil, "ne
 var validIOThreadsPolicies = []v1.IOThreadsPolicy{v1.IOThreadsPolicyShared, v1.IOThreadsPolicyAuto}
 var validCPUFeaturePolicies = map[string]*struct{}{"": nil, "force": nil, "require": nil, "optional": nil, "disable": nil, "forbid": nil}
 
+var filteredVmiKubevirtLabels = map[string]bool{
```
danielBelenky, line 67: Maybe call it `restriectedVmiLabels` instead?

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
```diff
@@ -950,9 +960,25 @@ func ValidateVirtualMachineInstanceMandatoryFields(field *k8sfield.Path, spec *v
...
+	filteredLabels, hasRestrictedLabels := FilterKubevirtLabels(labels)
+
+	// Validate kubevirt.io labels presence. Restricted labels allowed
+	// to be created only by known service accounts
+	if len(filteredLabels) > 0 && hasRestrictedLabels {
+		allowed := GetAllowedServiceAccounts()
+		if _, ok := allowed[accountName]; !ok {
+			causes = append(causes, metav1.StatusCause{
+				Type:    metav1.CauseTypeFieldValueNotSupported,
+				Message: "creation of kubevirt.io/ labels on a VMI object is restricted",
```
danielBelenky, line 977: We don't restrict all labels... just a subset of `kubevirt.io/` for now.
Also, I think that it is worth telling the user exactly which label he added which is restricted. Otherwise, it can be a frustrating experience.

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-update-admitter.go
```diff
@@ -93,21 +94,7 @@ func admitVMILabelsUpdate(
...
-	for label, value := range vmi.Labels {
-		if strings.HasPrefix(label, "kubevirt.io") {
-			m[label] = value
-		}
-	}
-	return m
-}
-
-func getAllowedServiceAccounts() map[string]struct{} {
+func GetAllowedServiceAccounts() map[string]struct{} {
```
danielBelenky, line 97: Why you exported this function?

fabiand:
Igor, would you mid adding a user-guide documentation about these specific
white listed labes? Or should we remove them on the long run?

stu-gott:
> Igor, would you mid adding a user-guide documentation about these specific white listed labes? Or should we remove them on the long run?

I think it makes sense to add some documentation regarding this. I concur with @rmohr's idea: we should be advertising that nothing with a "kubevirt.io/" prefix should be used--even if we don't specifically respond to or block it.

stu-gott:
Code looks good overall. @danielBelenky's requested changes make sense.

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
xpivarc: This can be a slice. As Map doesn't add additional value IMO

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
```diff
@@ -950,9 +960,25 @@ func ValidateVirtualMachineInstanceMandatoryFields(field *k8sfield.Path, spec *v
 	return causes
 }
 
-func ValidateVirtualMachineInstanceMetadata(field *k8sfield.Path, metadata *metav1.ObjectMeta, config *virtconfig.ClusterConfig) []metav1.StatusCause {
+func ValidateVirtualMachineInstanceMetadata(field *k8sfield.Path, metadata *metav1.ObjectMeta, config *virtconfig.ClusterConfig, accountName string) []metav1.StatusCause {
+
 	var causes []metav1.StatusCause
 	annotations := metadata.Annotations
+	labels := metadata.Labels
```
xpivarc, line 967: I would write this as follow: 
1. Check if the user is allowed to create restricted kubevirt.io labels 
2. If not, check if did...

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
xpivarc: +1 for including the label we restrict.

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-update-admitter.go
```diff
@@ -73,13 +74,13 @@ func admitVMILabelsUpdate(
...
 	if _, ok := allowed[ar.Request.UserInfo.Username]; ok {
 		return nil
 	}
 
-	oldLabels := filterKubevirtLabels(oldVMI)
-	newLabels := filterKubevirtLabels(newVMI)
+	oldLabels, _ := FilterKubevirtLabels(oldVMI.ObjectMeta.Labels)
+	newLabels, _ := FilterKubevirtLabels(newVMI.ObjectMeta.Labels)
 
 	if !reflect.DeepEqual(oldLabels, newLabels) {
```
xpivarc, line 85: What happens here if the order is changed?

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter_test.go
```diff
@@ -332,15 +335,66 @@ var _ = Describe("Validating VMICreate Admitter", func() {
 	)
 
 	Context("with VirtualMachineInstance metadata", func() {
-
+		table.DescribeTable(
+			"Should allow VMI creation with kubevirt.io/ labels only for kubevirt service accounts",
```
xpivarc, line 339: What about "Should allow VMI creation with restricted kubevirt.io/ labels only ...."?

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
enp0s3: @danielBelenky I've rnamed it. @xpivarc AFAIK map lookup is better then iterating over a slice

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
enp0s3: @danielBelenky @xpivarc Lets just document the restriction over the usage of system reserved labels. Filtering the exact ones that were changed by the user will take more effort.

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-update-admitter.go
enp0s3: @danielBelenky export reverted

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-update-admitter.go
enp0s3: @xpivarc Since it's a map the order should not affect the validation. Labels are the keys in these maps.

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter_test.go
enp0s3: @xpivarc Already included these tests, allow creation by service accounts

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-create-admitter.go
enp0s3: @xpivarc Done

enp0s3:
> Igor, would you mid adding a user-guide documentation about these specific white listed labes? Or should we remove them on the long run?

@fabiand Sure. I will add it in a separate PR

pkg/virt-api/webhooks/validating-webhook/admitters/vmi-update-admitter.go
enp0s3: @danielBelenky Done

stu-gott:
/lgtm

Nice job @enp0s3!

PR #3744, opened 2020-07-02
vendor: "Unreplace" prometheus-operator, update packages

andreabolognani:

The longer version is in the commit message, but the short version is that `vim-go` doesn't seem to work properly without these changes.

I find Go modules to be extremely confusing, so I might have gotten everything completely wrong.

**Release note**:

```release-note
NONE
```


rmohr:
/lgtm
/approve

The original reason to ping this dependency to align it with some other k8s dependencies. Since we are generating the client now from the api instead of using the shipped one, this issue is gone. All our generated clients work with the same set of k8s dependencies.

PR #3750, opened 2020-07-03
Ensure that selinux context tests wait for configpropagation

rmohr:

Let the selinux tests use the new config propagation mechanism. This fixes test flakes like this:

```
/root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:146
Expected
    <string>: spc_t
to equal
    <string>: container_t
/root/go/src/kubevirt.io/kubevirt/tests/security_features_test.go:164
```

like experienced in https://prow.apps.ovirt.org/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3599/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1278713572813180928.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/security_features_test.go
+++ b/tests/security_features_test.go
@@ -20,16 +20,10 @@
 package tests_test
 
 import (
-	"encoding/json"
-	"fmt"
-	"time"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	k8sv1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 
 	v1 "kubevirt.io/client-go/api/v1"
 	"kubevirt.io/client-go/kubecli"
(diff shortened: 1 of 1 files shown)


tests/security_features_test.go
```diff
@@ -40,48 +34,11 @@ import (
...
-
-			originalKubeVirtConfig, err = virtClient.CoreV1().ConfigMaps(tests.KubeVirtInstallNamespace).Patch("kubevirt-config", types.JSONPatchType, []byte(data))
-			Expect(err).ToNot(HaveOccurred())
-
-			// Allow time for virt-controller's ConfigMap cache to sync
-			time.Sleep(3 * time.Second)
-		}
-
+	BeforeEach(func() {
+		tests.BeforeTestCleanup()
```
kbidarkar, line 41: Wondering if this would create an empty kubevirt-config cfgmap if non exists? At-least in downstream 
I know of that it does exist always. Not sure if it is the same in upstream.

kbidarkar: Anyways all tests in all the lanes seem to have passed.

rmohr: We always create a config-map before we start the tests. Should be good.

PR #3755, opened 2020-07-05
tests, reporter: Collect journal logs from nodes

EdDev:

Collect journal logs from the nodes on test failure.
This should allow the debugging of node related operations, e.g. kubectl, NetworkManager.

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/reporter/kubernetes.go
+++ b/tests/reporter/kubernetes.go
@@ -91,7 +91,8 @@ func (r *KubernetesReporter) Dump(duration time.Duration) {
 		return
 	}
 
-	since := time.Now().Add(-duration).Add(-5 * time.Second)
+	duration += 5 * time.Second
+	since := time.Now().Add(-duration)
 
 	r.logEvents(virtCli, since)
 	r.logNodes(virtCli)
@@ -103,7 +104,8 @@ func (r *KubernetesReporter) Dump(duration time.Duration) {
 	r.logSecrets(virtCli)
 	r.logAuditLogs(virtCli, since)
 	r.logDMESG(virtCli, since)
-	r.logJournal(virtCli, duration+5*time.Second)
+	r.logJournal(virtCli, duration, "")
+	r.logJournal(virtCli, duration, "kubelet")
 	r.logVMs(virtCli)
 	r.logDomainXMLs(virtCli)
 	r.logLogs(virtCli, since)
@@ -303,9 +305,15 @@ func (r *KubernetesReporter) logAuditLogs(virtCli kubecli.KubevirtClient, since 
 	}
 }
 
-func (r *KubernetesReporter) logJournal(virtCli kubecli.KubevirtClient, duration time.Duration) {
+func (r *KubernetesReporter) logJournal(virtCli kubecli.KubevirtClient, duration time.Duration, unit string) {
 
-	const component = "journal"
+	var component string = "journal"
+	var unitCommandArgs []string
+
+	if unit != "" {
+		component += "_" + unit
+		unitCommandArgs = append(unitCommandArgs, "-u", unit)
+	}
 
 	logsdir := filepath.Join(r.artifactsDir, "nodes")
 	if err := os.MkdirAll(logsdir, 0777); err != nil {
@@ -334,6 +342,8 @@ func (r *KubernetesReporter) logJournal(virtCli kubecli.KubevirtClient, duration
 			"--since",
 			"-" + logDuration + "s",
 		}
+		commands = append(commands, unitCommandArgs...)
+
 		stdout, stderr, err := tests.ExecuteCommandOnPodV2(virtCli, pod, "virt-handler", commands)
 		if err != nil {
 			fmt.Fprintf(
--- a/tests/reporter/kubernetes.go
+++ b/tests/reporter/kubernetes.go
@@ -103,6 +103,7 @@ func (r *KubernetesReporter) Dump(duration time.Duration) {
 	r.logSecrets(virtCli)
 	r.logAuditLogs(virtCli, since)
 	r.logDMESG(virtCli, since)
+	r.logJournal(virtCli, duration+5*time.Second)
 	r.logVMs(virtCli)
 	r.logDomainXMLs(virtCli)
 	r.logLogs(virtCli, since)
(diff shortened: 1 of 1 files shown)


tests/reporter/kubernetes.go
```diff
@@ -302,6 +303,56 @@ func (r *KubernetesReporter) logAuditLogs(virtCli kubecli.KubevirtClient, since
...
+
+	logDuration := strconv.FormatInt(int64(duration/time.Second), 10)
+
+	nodes := getNodesWithVirtLauncher(virtCli)
+
+	for _, node := range nodes {
+		pod, err := kubecli.NewVirtHandlerClient(virtCli).Namespace(tests.KubeVirtInstallNamespace).ForNode(node).Pod()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to get virt-handler pod on node %s: %v", node, err)
+			return
```
oshoval, line 324: continue, same for the others, so node error wont stop them all

tests/reporter/kubernetes.go
```diff
@@ -302,6 +303,56 @@ func (r *KubernetesReporter) logAuditLogs(virtCli kubecli.KubevirtClient, since
...
+			"--",
+			"/usr/bin/journalctl",
+			"--since",
+			"-" + logDuration + "s",
+		}
+		stdout, stderr, err := tests.ExecuteCommandOnPodV2(virtCli, pod, "virt-handler", commands)
+		if err != nil {
+			fmt.Fprintf(
+				os.Stderr,
+				"failed to execute command: %s, stdout: %s, stderr: %s, error: %v",
```
oshoval, line 341: maybe add node name to the err message?

tests/reporter/kubernetes.go
```diff
@@ -302,6 +303,56 @@ func (r *KubernetesReporter) logAuditLogs(virtCli kubecli.KubevirtClient, since
...
+				commands, stdout, stderr, err,
+			)
+			return
+		}
+
+		fileName := fmt.Sprintf("%d_%s_%s.log", r.failureCount, component, node)
```
oshoval, line 347: i would suggest please to switch node and component, so in case of couple components, file list would be sorted according node

tests/reporter/kubernetes.go
```diff
@@ -612,3 +663,14 @@ func getNodesWithVirtLauncher(virtCli kubecli.KubevirtClient) []string {
 
 	return nodes
 }
+
+func writeStringToFile(filePath string, data string) error {
+	f, err := os.OpenFile(filePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		return fmt.Errorf("failed to open the file %s: %v", filePath, err)
```
oshoval, line 670: it would print the error twice, here and on `fmt.Fprintf(os.Stderr, "%v", err)`
is it ok ?

tests/reporter/kubernetes.go
EdDev: Done

tests/reporter/kubernetes.go
EdDev: Done

tests/reporter/kubernetes.go
EdDev: This is the current pattern used in all the other node level collections, it will be wrong to change this in only one place.

oshoval:
```
Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes #

Special notes for your reviewer:
```
Why we leave this in the PR when we dont use it ?
Ain't it better to be as clean as possible imo ?

EdDev:
> ```
> Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
> Fixes #
> 
> Special notes for your reviewer:
> ```
> 
> Why we leave this in the PR when we dont use it ?
> Ain't it better to be as clean as possible imo ?

I personally do not bother.
I have seen no where written that it is required to remove it. If there is some guidence, please link me to it or add one.

oshoval:
> > ```
> > Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
> > Fixes #
> > 
> > Special notes for your reviewer:
> > ```
> > 
> > 
> > Why we leave this in the PR when we dont use it ?
> > Ain't it better to be as clean as possible imo ?
> 
> I personally do not bother.
> I have seen no where written that it is required to remove it. If there is some guidence, please link me to it or add one.

Personal taste only so far, thanks

oshoval:
Nice, thanks
/lgtm

EdDev:
> I think that the kubelet deserves a dedicated log file.

Why would you prefer a dedicated one plus a general one?
I think it will be harder to debug that way as kubectl may interact with other units (e.g kernel) and having the whole picture is an advantage.
The log file itself may be filtered easily if one wants just the kubelet lines.

rmohr:
> Why would you prefer a dedicated one plus a general one?
> I think it will be harder to debug that way as kubectl may interact with other units (e.g kernel) and having the whole picture is an advantage.
> The log file itself may be filtered easily if one wants just the kubelet lines.

Keeping the logfiles small and focused allows at least me so far to diagnose most issues without having to download files.

rmohr:
> I think it will be harder to debug that way as kubectl may interact with other units (e.g kernel) and having the whole picture is an advantage.

And if that is your concern, I am not talking about filtering out the kubelet logs from the general journal. Just having them in addition.

EdDev:
> Keeping the logfiles small and focused allows at least me so far to diagnose most issues without having to download files.

Having two logs, one specific and one that includes all is a bit of a waste, but I have no problem to add another with the kubectl unit. On it.

tests/reporter/kubernetes.go
```diff
@@ -302,6 +305,63 @@ func (r *KubernetesReporter) logAuditLogs(virtCli kubecli.KubevirtClient, since
...
+			"/usr/bin/virt-chroot",
+			"--mount",
+			"/proc/1/ns/mnt",
+			"exec",
+			"--",
+			"/usr/bin/journalctl",
+			"--since",
+			"-" + logDuration + "s",
+		}
+		if unit != "" {
```
oshoval, line 342: instead of two if, maybe worth to set a command_postfix that would always be appended, and would be set for -u unit in or empty string, in the first if ?

EdDev: Done

oshoval:
Thanks
/lgtm

rmohr:
/override pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6
/override pull-kubevirt-e2e-k8s-1.18

One failed on the dev/null issue, the other one on a timeout in AfterSuite

PR #3761, opened 2020-07-06
[release-0.30] [tests] Wait for config changes to be propagated and reset them implicitly

rmohr:
This is a cherry-pick of #3542

```release-note
NONE
```


rmohr:
> This seems to bring bits of #3766, probably better to wait until that's merged.

:+1:

rmohr:
/hold cancel

Should be better now. One config entry had the wrong format.

Issue #3764, opened 2020-07-06
Find an alternative to parsing the logs in vmi lifecycle tests


danielBelenky:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
Currently, we are parsing the logs from virt-handler to decide if the communication between
virt-handler and virt-launcher works and check for various domain lifecycle events.
This is risky because parsing logs in unreliable: I can print a line to the log that announces
that I do something but fail to do so in my code and the test will still pass.

**What you expected to happen**:
We should define the overall behaviour that we want to check and assert on it. During e2e
tests we probably shouldn't care about "how" things happened but we want to know that
they happened in general.
I understand that in some cases this can be hard to do during e2e tests and this is what
the actual goal of this ticket: to see if we really need those tests and if we do, try to find
a solution on how to test the behaviour without relying on log parsing.

**Anything else we need to know?**:
Example: https://github.com/kubevirt/kubevirt/blob/master/tests/vmi_lifecycle_test.go#L1084

EdDev:
I would also expect the functional/e2e tests to interact with the API and not with logs. Logs cannot be trusted, they can change (edited, added or removed).

Could we give an example in this issue on what kind of test is expected here?
I could imagine a VMI lifecycle to include the different phases and states it is in, checking that we detect them all in the different paths they can take. Is this the case?

victortoso:
/reopen

It is still an issue.

victortoso:
@xpivarc would `help-wanted` label makes sense here? It isn't a high priority but not necessarily a `good-first-issue`, depending on the work involved... needs discussion.

PR #3765, opened 2020-07-06
Add assertion to domain start/stop test

### LINKED ISSUE #3573: [flakey test] should log libvirt start and stop lifecycle events of the domain [test_id:1641]
> /kind bug

**What happened**:

Today failing on K8s-1.18
https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2020-06-15-024h.html

last week saw it failing almost every time on the check for flakes lane (k8s-1.18)

locally it runs fun (checking in loops)

test should pass

I believe its matter of timeout, if i reach it, i will try to increase the timeout and run in loops on prow
checking loops of it locally, so far so good (without changes even)

**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

danielBelenky:
There are 2 paths that we support in which a domain can be removed in
case of a VMI deletion:

1. virt-controllers kicks in and registers a deletion request on the
   launcher and then K8s removes the pod before we get a chance to
   actually delete the domain.
2. virt-launcher/virt-handler sets the deletion timestamp on the domain
   and we get to properly remove it.

This test was failing sporadically because we used to check only the 2nd
path.

Since log parsing during tests can lead to various issues, I've opened
opened https://github.com/kubevirt/kubevirt/issues/3764 to see if
we can find a better way to address those tests.

Fixes #3573

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/vmi_lifecycle_test.go
+++ b/tests/vmi_lifecycle_test.go
@@ -1096,12 +1096,22 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
 
 				// Check if the stop event was logged
 				By("Checking that virt-handler logs VirtualMachineInstance deletion")
+				/*
+						Since we deleted the VMI object, there are two possible outcomes and both are expected:
+						1. virt-controller kicks in, registers a deletion request on the launcher pod and K8s deletes the pod
+					       before virt-handler had a chance to set or check the deletion timestamp on the domain.
+						2. virt-handler detects the deletion timestamp on the domain and removes it.
+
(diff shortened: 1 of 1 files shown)


rmohr:
/hold

Seems like we have an issue related to the change in the cnao lane.

danielBelenky:
/hold cancel

cancelling because I fixed the issue.

danielBelenky:
/override pull-kubevirt-e2e-kind-1.17-sriov
This laneis optional as it fails 100% of the time but there's no reason for the retest bot to keep retesting this.

PR #3720, opened 2020-06-30
Bump kubevirtci: sriov provider fixes

ormergi:
Signed-off-by: Or Mergi <ormergi@redhat.com>

**What this PR does / why we need it**:
This PR bumps kubevirtci in order to get changes from https://github.com/kubevirt/kubevirtci/pull/397:
1. Stop cluster-up when sriov operator webhooks certificate generation fails
2. Print informative error message where wait_for_taint return code 1

This PR is part of stabilizing sriov lane effort.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #3767, opened 2020-07-06
Fix kubevirtci bumping

oshoval:
Move sync-kubevirtci.sh after bazel build.
This way it won't generate unneeded BUILD.bazel files,
and the bumping result of make bump-kubevirtci would equal
the result of make generate (after kubevirtci commit hash updating).

```release-note
None
```


jean-edouard:
Pardon my ignorance, but what is the point of that script?
At first glance it looks like it will just revert any change that was made to `cluster-up/` since the last `make bump-kubevirtci`...

oshoval:
> Pardon my ignorance, but what is the point of that script?
> At first glance it looks like it will just revert any change that was made to `cluster-up/` since the last `make bump-kubevirtci`...

This PR just move the sync-kubevirtci.sh from `hack/generate.sh` to the end of `make generate`
There are 2 ways to bump (even before this PR)

1. Just run make bump-kubevirtci - usually for local use only and not for official bump PR
2. Run make generate after updating the kubevirtci hash, either manually, or just by running (1) before that, which will update the hash and bump-kubevirtci already, but, and this is what this PR tries to fix:

There is difference between bump result of (1) and (2)
because (2) creates a BUILD.bazel file in cluster-up folder, which doesn't exists in kubevirtci original cluster-up folder,
so people that want to create official bump PR, need to run the two commands usually, and finish with a file that is there but not needed (from what i understand).

Now, assuming the only change that is needed is bumping, it would be sufficient to run only (1).
and no need to run (2) anymore for bumping only (of course if there are more changes that make generate does, and we want them, we would need to).

Its done by moving the sync-kubevirtci.sh after the bazel command that generate the BUILD.bazel files,
so it would vendor it clean.

The reason sync-kubevirtci.sh is also part of hack/bump-kubevirtci.sh (result of make bump-kubevirtci)
is in order to give a faster method for bumping, which originally just bumped the hash, but it also actually vendor the folder, so if someone uses (1), he wouldnt need to run the whole make generate just to vendor the folder.

Hope it sounds fine ?

oshoval:
@jean-edouard 
Travis fails for example, because this PR would need to come with a fresh bump
```
It seems like you need to run 'make generate'. Please run it and commit the changes
 M cluster-up-sha.txt
 D cluster-up/cluster/kind-k8s-sriov-1.17.0/csrcreator/BUILD.bazel
The command "./automation/travisci-test.sh" exited with 1.
```
After a bump + this PR merged, it will be enough to run short and simple
`make bump-kubevirtci`, in order to only bump-kubevirtci (even PR ready, and not just quick and dirty locally, as its now).

jean-edouard:
Thank you for the detailed explanation, I've been using only `make bump-kubevirtci` for official bumps, I probably need to fix my workflow.
Please note that travis-ci is not happy about this.

oshoval:
> Thank you for the detailed explanation, I've been using only `make bump-kubevirtci` for official bumps, I probably need to fix my workflow.
> Please note that travis-ci is not happy about this.

Thanks, after 
https://github.com/kubevirt/kubevirt/pull/3720 is merged, i would fine tune this PR

qinqon:
@oshoval why do we dump kubevirtci too ?

oshoval:
> @oshoval why do we dump kubevirtci too ?

what do you mean ?
the PR is for fix bump of kubevirtci at kubevirt, there are some delicate issues, 
this PR makes the result of `make bump-kubevirtci `equal the result of changing the hash and running `make generate`
before it wasnt the situation.

hack/sync-kubevirtci.sh
```diff
@@ -10,7 +10,9 @@ version_file="cluster-up/version.txt"
 sha_file="cluster-up-sha.txt"
 download_cluster_up=true
 function getClusterUpShasum() {
-    find ${KUBEVIRT_DIR}/cluster-up -type f | sort | xargs sha1sum | sha1sum | awk '{print $1}'
+    pushd ${KUBEVIRT_DIR} >/dev/null
```
qinqon, line 13: you can do 
```bash
(
    cd $KUBIVRT_DIR
    find ...
)
```

Parentesis are creating a subshell and after the subshell you will be at the previous dir.

oshoval: done thx

oshoval:
/hold 
want to make sure 
pull-kubevirt-e2e-kind-1.17-sriov passes
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3767/pull-kubevirt-e2e-kind-1.17-sriov/1281220660827262976

oshoval:
/hold cancel
no regression about csrcreator in sriov lane, as seen above

oshoval:
@jean-edouard 
would you please take another look ?

PR #3779, opened 2020-07-07
K8s reporter to own a sub dir under $ARTIFACTS

danielBelenky:
The K8s reporter removes the artifacts directory whenver it starts the
test and it doesn't create it later if it failed to generate a client
for the cluster. Since other reporters also depend on the shared
artifacts dir, they fail as well.

This commit gives the K8s reporter it's own artifacts sub dir to own and
manage however it needs to.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -22,6 +22,7 @@ package tests_test
 import (
 	"fmt"
 	"os"
+	"path"
 	"strconv"
 	"testing"
 
@@ -35,7 +36,9 @@ import (
 
 func TestTests(t *testing.T) {
 	maxFails := getMaxFailsFromEnv()
-	reporters := []Reporter{reporter.NewKubernetesReporter(os.Getenv("ARTIFACTS"), maxFails)}
+	reporters := []Reporter{
+		reporter.NewKubernetesReporter(path.Join(os.Getenv("ARTIFACTS"), "k8s-reporter"), maxFails),
+	}
 	if ginkgo_reporters.Polarion.Run {
 		reporters = append(reporters, &ginkgo_reporters.Polarion)
 	}


jean-edouard:
@danielBelenky as a frequent local functest user I'm not a big fan of this...
Couldn't we just `mkdir -p` whatever directory sometimes fails to create? (or better yet, figure out why it sometimes fails to create)

danielBelenky:
> @danielBelenky as a frequent local functest user I'm not a big fan of this...
> Couldn't we just `mkdir -p` whatever directory sometimes fails to create? (or better yet, figure out why it sometimes fails to create)

@jean-edouard The failure happens when the tests are breaking the cluster or the kubevirt deployment and
we are not able to create a new client here: https://github.com/kubevirt/kubevirt/blob/8831332bda73b0dc7184a213d2697e5d1bcb7b99/tests/reporter/kubernetes.go#L83

The reporter shouldn't assume that it is the first/only user of the artifacts dir.
What if we add another reporter that runs before the K8s reporter?
Since Gingo's documentation doesn't mention anything about ensuring that
the reporters will be called in order, this cleanup code is a bug.

I am a frequent local functests user myself and I think that it is ok to request the user to explicitly
clean after himself if he decides to.

jean-edouard:
It's just that I rarely look at the artifacts and kinda forget about them, but when I actually need them I don't want to find a mix of old and new stuff.
However, I guess I can easily sort through stuff using file timestamps, so feel free to ignore my initial comment.

danielBelenky:
>  However, I guess I can easily sort through stuff using file timestamps, so feel free to ignore my initial comment.

Thanks for the quick reply.

If this actually will be a problem, and people will not want to use the `clean` target in our Makefile,
we will find where to add this removal to make it safe.

qinqon:
@danielBelenky what about `artifacts/[timestamp]/` ?

tests/reporter/kubernetes.go
```diff
@@ -52,7 +52,6 @@ func (r *KubernetesReporter) SpecSuiteWillBegin(config config.GinkgoConfigType,
 }
 
 func (r *KubernetesReporter) BeforeSuiteDidRun(setupSummary *types.SetupSummary) {
-	r.Cleanup()
```
qinqon, line 55: We can check `CI` environment variable here and do this only then ?

rmohr: I would prefer just a sub-directory like `k8s-reporter`.

rmohr:
/hold

due to the mentioned issues.

PR #3784, opened 2020-07-08
Add build verification targets to the Makefile

danielBelenky:
The purpose of those targets is to serve as an entrypoint for Prow jobs.
The idea is to eventually replace automation/travisci-test.sh and the
single Travis-CI job with multiple Prow jobs that will run in parallel.

I also modified the travis-ci testing script in this commit to get proper CI
coverage for this work. There should be no change in the functionality
of the targets as opposed to calling them and wrapping the calls with
some checks from a shell script.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,9 @@ bazel-generate:
 bazel-build:
 	hack/dockerized "hack/bazel-fmt.sh && hack/bazel-build.sh"
 
+bazel-build-verify: TARGET_TO_RUN='make'
+bazel-build-verify: bazel-build check-git-tree-state build-verify bazel-test
+
 bazel-build-images:
 	hack/dockerized "DOCKER_PREFIX=${DOCKER_PREFIX} DOCKER_TAG=${DOCKER_TAG} DOCKER_TAG_ALT=${DOCKER_TAG_ALT} IMAGE_PREFIX=${IMAGE_PREFIX} IMAGE_PREFIX_ALT=${IMAGE_PREFIX_ALT} ./hack/bazel-build-images.sh"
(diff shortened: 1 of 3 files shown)


automation/automation_entrypoint.sh
```diff
@@ -0,0 +1,81 @@
...
+    make manifests
+}
+
+function automation::verify_prometheus_rules() {
+    make prom-rules-verify
+}
+
+function automation::generate() {
+    make generate
+    if [[ -n "$(git status --porcelain)" ]] ; then
```
dhiller, line 59: I would like to have this in the makefile rather as a separate make target.

automation/automation_entrypoint.sh
```diff
@@ -0,0 +1,81 @@
...
+        echo "It seems like you need to run 'make generate'. Please run it and commit the changes"
+        git status --porcelain; false
+    fi
+
+    if diff <(git grep -c '') <(git grep -cI '') | grep -E -v -e 'docs/.*\.png|swagger-ui' -e 'vendor/*' -e 'assets/*' | grep '^<'; then
```
dhiller, line 64: Same here.

automation/automation_entrypoint.sh
```diff
@@ -0,0 +1,81 @@
...
+        echo "Binary files are present in git repostory."; false
+    fi
+}
+
+function automation::bazel_build_verify() {
+    make
+    if [[ -n "$(git status --porcelain)" ]] ; then
```
dhiller, line 71: Same here.

PR #3789, opened 2020-07-08
Print error message on test utils

alonSadan:

A part of the cleanup that is done after the test suit is to remove all the tests namespaces.
we give an initial delete instruction and then wait in an Eventually for the deletion to be complete,
by polling a delete and wait for an IsNotFound error,
which tells us that the delete performed successfully.
If another error occurred while waiting for the the deletion (IsNotFound returned false),
it would be helpful to know what error did returned upon waiting for the deletion to be completed.

This PR prints the error in case of timeout while waiting for the namespace deletion.

Signed-off-by: alonSadan <asadan@redhat.com>

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -555,7 +555,7 @@ func AfterTestSuitCleanup() {
 	if flags.DeployTestingInfrastructureFlag {
 		WipeTestingInfrastructure()
 	}
-	removeNamespaces()
+	DeleteAndWaitForNamespacesToDisappear()
 
 	CleanNodes()
 
@@ -1678,7 +1678,7 @@ func cleanNamespaces() {
 	}
 }
 
-func removeNamespaces() {
+func DeleteAndWaitForNamespacesToDisappear() {
 	virtCli, err := kubecli.GetKubevirtClient()
 	PanicOnError(err)
 
@@ -1691,11 +1691,13 @@ func removeNamespaces() {
 	}
 
 	// Wait until the namespaces are terminated
-	fmt.Println("")
 	for _, namespace := range testNamespaces {
 		fmt.Printf("Waiting for namespace %s to be removed, this can take a while ...\n", namespace)
-		EventuallyWithOffset(1, func() bool { return errors.IsNotFound(virtCli.CoreV1().Namespaces().Delete(namespace, nil)) }, 240*time.Second, 1*time.Second).
-			Should(BeTrue())
+
+		EventuallyWithOffset(1, func() error {
+			_, err = virtCli.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
+			return err
+		}, 240*time.Second, 1*time.Second).Should(SatisfyAll(HaveOccurred(), WithTransform(errors.IsNotFound, BeTrue())), "The namesapce should be gone within the given timeout")
 	}
 }
 


tests/utils.go
```diff
@@ -1737,8 +1737,11 @@ func removeNamespaces() {
 	fmt.Println("")
 	for _, namespace := range testNamespaces {
 		fmt.Printf("Waiting for namespace %s to be removed, this can take a while ...\n", namespace)
-		EventuallyWithOffset(1, func() bool { return errors.IsNotFound(virtCli.CoreV1().Namespaces().Delete(namespace, nil)) }, 240*time.Second, 1*time.Second).
-			Should(BeTrue())
+		EventuallyWithOffset(1, func() bool {
+			err = (virtCli.CoreV1().Namespaces().Delete(namespace, nil))
+			return errors.IsNotFound(err)
+		}, 240*time.Second, 1*time.Second).
+			Should(BeTrue(), "While deleting test namespace %s an unexpected error occurred : %v", namespace, err)
```
EdDev, line 1744: I think you should return the error and then evaluate it instead the boolean.
I would expect the error to include the namespace and other data about why it failed. Could you check how this assertion below outputs the failure? I would expect it to print the returned value (the error) so some of the additional message you added may not be needed.

So it should look like this:
```
EventuallyWithOffset(1, func() error {
	return virtCli.CoreV1().Namespaces().Delete(namespace, nil)
	return errors.IsNotFound(err)
}, 240*time.Second, 1*time.Second).
	Should(WithTransform(errors.IsNotFound, BeTrue()), "While deleting test namespace %s an unexpected error occurred : %v", namespace, err)
```

alonSadan: > I think you should return the error and then evaluate it instead the boolean.
> I would expect the error to include the namespace and other data about why it failed. Could you check how this assertion below outputs the failure? I would expect it to print the returned value (the error) so some of the additional message you added may not be needed.
> 
> So it should look like this:
> 
> ```
> EventuallyWithOffset(1, func() error {
> 	return virtCli.CoreV1().Namespaces().Delete(namespace, nil)
> 	return errors.IsNotFound(err)
I guess this line is mistake?
> }, 240*time.Second, 1*time.Second).
> 	Should(WithTransform(errors.IsNotFound, BeTrue()), "While deleting test namespace %s an unexpected error occurred : %v", namespace, err)
> ```

Thanks @EdDev. the error output includes the object type we tried to delete (namesapce) and the objects name. so I updated the additional message. 
the snippet as it is won't work in case of a `nil` error, so I added a condition to verify that.

alonSadan:
@EdDev  what do you think of the change?

tests/utils.go
```diff
@@ -1686,8 +1686,15 @@ func removeNamespaces() {
...
 		fmt.Printf("Waiting for namespace %s to be removed, this can take a while ...\n", namespace)
-		EventuallyWithOffset(1, func() bool { return errors.IsNotFound(virtCli.CoreV1().Namespaces().Delete(namespace, nil)) }, 240*time.Second, 1*time.Second).
-			Should(BeTrue())
+
+		EventuallyWithOffset(1, func() error {
+			err = virtCli.CoreV1().Namespaces().Delete(namespace, nil)
+			if err != nil {
+				return err
+			}
+			return fmt.Errorf("namespace %s still exists", namespace)
```
EdDev, line 1695: This is obviously wrong because a deletion with a nil error is a success delivery of the request to the API.
If we take a few steps back, I do not understand why a delete instruction is retried.

If we look how the VMI deletion was handled, the retry is on the "Get" while the "Delete" is checked only once.

alonSadan:
sorry @EdDev  for the late response (and for the bad code).
I hope now it is better. Also I fixed the typos in the commit message.

tests/utils.go
alonSadan: changed. I Think now it is better

alonSadan:
https://github.com/kubevirt/kubevirt/pull/3797/commits/655c4e2255341fbe2355938b17980258cd1bdb4b

PR #3791, opened 2020-07-08
check for readiness of device manager plugin

### LINKED ISSUE #3664: [flaky ci] [test_id:1635] VMIlifecycle Creating a VirtualMachineInstance with node tainted the vmi with  tolerations should be scheduled

>
kind bug
> /kind enhancement

**What happened**:

[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:component]VMIlifecycle   [rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:component]Creating  a VirtualMachineInstance with node tainted [test_id:1635]the vmi with  tolerations should be scheduled 

/root/go/src/kubevirt.io/kubevirt/tests/vmi_lifecycle_test.go:689 VMI testvmisl2862zxxqshrwtcj6g4z7glf445q486fhwww6cglsmfmt5c unexpectedly stopped. State: Failed Expected     <bool>: true to be false /root/go/src/kubevirt.io/kubevirt/tests/utils.go:2657

Link to the failing job : https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3502/pull-kubevirt-e2e-k8s-1.17/1275546119513837568

**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ashleyschuett:

Currently we have no way of knowing if the device manager is up and running when checking the status of the virt-handler. This change allows us to check that the `virt-handler` is running and the device manager has started before creating a VMI. 

If we create a VMI before the `virt-handler` is ready the `virt-controller` will create the `virt-launcher` pod which will then be unable to get the requested resources. Immediately the `virt-launcher` pod will be put into a state of `Failed` and [will not retry](https://github.com/kubernetes/kubernetes/issues/70010)

This change will allow checks such as [this](https://github.com/kubevirt/kubevirt/blob/master/tests/vmi_lifecycle_test.go#L659-L664) to be meaningful. 

The sequence of events looks like such:
```
            "reason": "Scheduled",
            "message": "Successfully assigned kubevirt-test-default/virt-launcher-testvmisl2862zxxqshrwtcj6g4z7glf445q486fhwwwr4tkv to node01",
            "source": {
                "component": "default-scheduler"
            },
            "firstTimestamp": "2020-06-24T00:12:25Z",
```

```
            "reason": "UnexpectedAdmissionError",
            "message": "Update plugin resources failed due to requested number of devices unavailable for devices.kubevirt.io/kvm. Requested: 1, Available: 0, which is unexpected.",
            "source": {
                "component": "kubelet",
                "host": "node01"
            },
            "firstTimestamp": "2020-06-24T00:12:25Z",
```

Fixes #3664

**Release note**:
```release-note
make node as kubevirt.io/schedulable=false on virt-handler restart 
```

### DIFF
--- a/cmd/virt-handler/virt-handler.go
+++ b/cmd/virt-handler/virt-handler.go
@@ -25,7 +25,9 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"os/signal"
 	"path/filepath"
+	"syscall"
 	"time"
 
 	"github.com/emicklei/go-restful"
@@ -34,6 +36,7 @@ import (
 	k8sv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	k8coresv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
@@ -125,6 +128,14 @@ func (app *virtHandlerApp) prepareCertManager() (err error) {
 	return
 }
 
+func (app *virtHandlerApp) markNodeAsUnschedulable(logger *log.FilteredLogger) {
+	data := []byte(fmt.Sprintf(`{"metadata": { "labels": {"%s": "false"}}}`, v1.NodeSchedulable))
+	_, err := app.virtCli.CoreV1().Nodes().Patch(app.HostOverride, types.StrategicMergePatchType, data)
+	if err != nil {
+		logger.V(1).Level(log.ERROR).Log("Unable to mark node as unschedulable", err.Error())
+	}
+}
+
 func (app *virtHandlerApp) Run() {
 	// HostOverride should default to os.Hostname(), to make sure we handle errors ensure it here.
 	if app.HostOverride == "" {
@@ -159,6 +170,20 @@ func (app *virtHandlerApp) Run() {
 	if err != nil {
 		panic(err)
 	}
+
+	app.markNodeAsUnschedulable(logger)
+
+	go func() {
+		sigint := make(chan os.Signal, 1)
+
+		signal.Notify(sigint, syscall.SIGTERM)
+
+		<-sigint
+
+		app.markNodeAsUnschedulable(logger)
+		os.Exit(0)
+	}()
+
 	broadcaster := record.NewBroadcaster()
 	broadcaster.StartRecordingToSink(&k8coresv1.EventSinkImpl{Interface: app.virtCli.CoreV1().Events(k8sv1.NamespaceAll)})
 	// Scheme is used to create an ObjectReference from an Object (e.g. VirtualMachineInstance) during Event creation
--- a/pkg/virt-handler/device-manager/device_controller.go
+++ b/pkg/virt-handler/device-manager/device_controller.go
@@ -100,3 +100,13 @@ func (c *DeviceController) Run(stop chan struct{}) error {
 	logger.Info("Shutting down device plugin controller")
 	return nil
 }
+
+func (c *DeviceController) Initialized() bool {
+	for _, dev := range c.devicePlugins {
+		if !dev.GetInitialized() {
+			return false
+		}
+	}
+
+	return true
+}
--- a/pkg/virt-handler/device-manager/generic_device.go
+++ b/pkg/virt-handler/device-manager/generic_device.go
@@ -46,33 +46,36 @@ type GenericDevice interface {
 	Start(chan struct{}) (err error)
 	GetDevicePath() string
 	GetDeviceName() string
+	GetInitialized() bool
 }
 
 type GenericDevicePlugin struct {
-	counter    int
-	devs       []*pluginapi.Device
-	server     *grpc.Server
-	socketPath string
-	stop       chan struct{}
-	health     chan string
-	devicePath string
-	deviceName string
-	done       chan struct{}
-	deviceRoot string
-	preOpen    bool
+	counter     int
+	devs        []*pluginapi.Device
+	server      *grpc.Server
+	socketPath  string
+	stop        chan struct{}
+	health      chan string
+	devicePath  string
+	deviceName  string
+	done        chan struct{}
+	deviceRoot  string
+	preOpen     bool
+	initialized bool
 }
 
 func NewGenericDevicePlugin(deviceName string, devicePath string, maxDevices int, preOpen bool) *GenericDevicePlugin {
 	serverSock := SocketPath(deviceName)
 	dpi := &GenericDevicePlugin{
-		counter:    0,
-		devs:       []*pluginapi.Device{},
-		socketPath: serverSock,
-		health:     make(chan string),
-		deviceName: deviceName,
-		devicePath: devicePath,
-		deviceRoot: util.HostRootMount,
-		preOpen:    preOpen,
+		counter:     0,
+		devs:        []*pluginapi.Device{},
+		socketPath:  serverSock,
+		health:      make(chan string),
+		deviceName:  deviceName,
+		devicePath:  devicePath,
+		deviceRoot:  util.HostRootMount,
+		preOpen:     preOpen,
+		initialized: false,
 	}
 	for i := 0; i < maxDevices; i++ {
 		dpi.addNewGenericDevice()
@@ -166,6 +169,7 @@ func (dpi *GenericDevicePlugin) Start(stop chan struct{}) (err error) {
 		errChan <- dpi.healthCheck()
 	}()
 
+	dpi.setInitialized(true)
 	logger.Infof("%s device plugin started", dpi.deviceName)
 	err = <-errChan
 
@@ -176,6 +180,7 @@ func (dpi *GenericDevicePlugin) Start(stop chan struct{}) (err error) {
 func (dpi *GenericDevicePlugin) Stop() error {
 	defer close(dpi.done)
 	dpi.server.Stop()
+	dpi.setInitialized(false)
 	return dpi.cleanup()
 }
 
@@ -339,4 +344,16 @@ func (dpi *GenericDevicePlugin) healthCheck() error {
 
 func SocketPath(deviceName string) string {
 	return filepath.Join(pluginapi.DevicePluginPath, fmt.Sprintf("kubevirt-%s.sock", deviceName))
+}
+
+func (dpi *GenericDevicePlugin) GetInitialized() bool {
+	dpi.lock.Lock()
+	defer dpi.lock.Unlock()
+	return dpi.initialized
+}
+
+func (dpi *GenericDevicePlugin) setInitialized(initialized bool) {
+	dpi.lock.Lock()
+	dpi.initialized = initialized
+	dpi.lock.Unlock()
 }
--- a/pkg/virt-handler/device-manager/generic_device.go
+++ b/pkg/virt-handler/device-manager/generic_device.go
@@ -26,6 +26,7 @@ import (
 	"path"
 	"path/filepath"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
@@ -62,6 +63,7 @@ type GenericDevicePlugin struct {
 	deviceRoot  string
 	preOpen     bool
 	initialized bool
+	lock        *sync.Mutex
 }
 
 func NewGenericDevicePlugin(deviceName string, devicePath string, maxDevices int, preOpen bool) *GenericDevicePlugin {
@@ -76,6 +78,7 @@ func NewGenericDevicePlugin(deviceName string, devicePath string, maxDevices int
 		deviceRoot:  util.HostRootMount,
 		preOpen:     preOpen,
 		initialized: false,
+		lock:        &sync.Mutex{},
 	}
 	for i := 0; i < maxDevices; i++ {
 		dpi.addNewGenericDevice()
--- a/pkg/virt-handler/vm.go
+++ b/pkg/virt-handler/vm.go
@@ -46,6 +46,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 
 	container_disk "kubevirt.io/kubevirt/pkg/virt-handler/container-disk"
+	device_manager "kubevirt.io/kubevirt/pkg/virt-handler/device-manager"
 
 	v1 "kubevirt.io/client-go/api/v1"
 	"kubevirt.io/client-go/kubecli"
@@ -60,7 +61,6 @@ import (
 	virtconfig "kubevirt.io/kubevirt/pkg/virt-config"
 	virtcache "kubevirt.io/kubevirt/pkg/virt-handler/cache"
 	cmdclient "kubevirt.io/kubevirt/pkg/virt-handler/cmd-client"
-	device_manager "kubevirt.io/kubevirt/pkg/virt-handler/device-manager"
 	"kubevirt.io/kubevirt/pkg/virt-handler/isolation"
 	migrationproxy "kubevirt.io/kubevirt/pkg/virt-handler/migration-proxy"
 	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
@@ -2207,7 +2207,13 @@ func (d *VirtualMachineController) heartBeat(interval time.Duration, stopCh chan
 				log.DefaultLogger().Reason(err).Errorf("Can't determine date")
 				return
 			}
-			data := []byte(fmt.Sprintf(`{"metadata": { "labels": {"%s": "true"}, "annotations": {"%s": %s}}}`, v1.NodeSchedulable, v1.VirtHandlerHeartbeat, string(now)))
+
+			kubevirtSchedulable := "true"
+			if !d.kvmController.Initialized() {
+				kubevirtSchedulable = "false"
+			}
+
+			data := []byte(fmt.Sprintf(`{"metadata": { "labels": {"%s": "%s"}, "annotations": {"%s": %s}}}`, v1.NodeSchedulable, kubevirtSchedulable, v1.VirtHandlerHeartbeat, string(now)))
 			_, err = d.clientset.CoreV1().Nodes().Patch(d.host, types.StrategicMergePatchType, data)
 			if err != nil {
 				log.DefaultLogger().Reason(err).Errorf("Can't patch node %s", d.host)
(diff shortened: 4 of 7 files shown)


pkg/virt-handler/device-manager/device_controller.go
```diff
@@ -95,6 +97,8 @@ func (c *DeviceController) Run(stop chan struct{}) error {
 		go c.startDevicePlugin(dev, stop)
 	}
 
+	util.MarkReady(ReadinessFile)
```
rmohr, line 100: Could you explain how this is working? When looking at the snippes provided in the PR description it looks to me like our device plugin already reported the resources and that virt-handler marked the node as schedulable too. How does a readiness probe on virt-handler help mitigating this situation?

ashleyschuett: What can happen is:  
   - virt-handler starts, it registers the device plugin 
   - virt-handler pods are deleted / restarted 
   -  virt-handler is ready but device plugin will not yet be re-registered 
This leaves the node to have the following state when the virt-hander is first running.. 
```  
allocatable:
    devices.kubevirt.io/kvm: "0"
    ... 
capacity:
    devices.kubevirt.io/kvm: "110"
    ... 
```

Since the pod currently has no readiness check we see it has available as soon as it is running but there is about 60 seconds from when it starts to the device manager running and register the plugins.

rmohr:
Thanks for  the details. I am still now sure how this would help. I think to minimize the "downtime window, you will have to mark the node as unschedulable by setting `kubevirt.io/schedulable=false` as soon as virt-handler starts. Have a look at https://github.com/kubevirt/kubevirt/blob/2d22cb6a9706e0ef0e1a9cc7b0b7fd1d57480730/pkg/virt-handler/vm.go#L1996.

I guess, in addition we could on clean shutdowns also explicitly try to set it to false. Something which we also don't do right now. We only have the node-controller which will do that for crasched virt-handlers after some timeout.

pkg/virt-handler/device-manager/device_controller.go
rmohr: What I don't understand: If there are no allocatable, why are pods scheduled there?

xpivarc: "Update plugin resources failed due to requested..." is produced by Allocate of device plugin handler in kubelet. So it suggests that device plugin(virt-handler) was running(because we have resources on node) but is unable to serve resources at that moment.  
Do we do this  https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubelet/pkg/apis/deviceplugin/v1alpha/api.proto#L43 ?

rmohr: What I mean:

```
allocatable:
    devices.kubevirt.io/kvm: "0"
    ... 
capacity:
    devices.kubevirt.io/kvm: "110"
    ... 
```

Looks correct. No pods should be scheduled to the node if it looks like this. Can it be that the kubelet only detects that right after one pod got there and requested it?

xpivarc: I think so. Scheduler have own copy of the state so it can be(right?). Does the allocatable stays at 0? Why did it crash in first place?

ashleyschuett: I've update this PR to instead update the `kubevirt.io/schedulable=false` label so that if the device plugin is not running the node is not schedulable. The event's in the case of a pod being created during a `virt-handler` pods restarting now look as follows

```
89s         Normal    SuccessfulCreate       virtualmachineinstance/testvmizpzf7f9fhhmcxwl9bj7bn5sl5clf9xg2csg9d4fztxc86jqh   Created virtual machine pod virt-launcher-testvmizpzf7f9fhhmcxwl9bj7bn5sl5clf9xg2csg9drz7ws
89s         Warning   FailedScheduling       pod/virt-launcher-testvmizpzf7f9fhhmcxwl9bj7bn5sl5clf9xg2csg9drz7ws              0/2 nodes are available: 2 node(s) didn't match node selector.
58s         Warning   FailedScheduling       pod/virt-launcher-testvmizpzf7f9fhhmcxwl9bj7bn5sl5clf9xg2csg9drz7ws              0/2 nodes are available: 1 Insufficient devices.kubevirt.io/kvm, 1 node(s) didn't match node selector.
58s         Warning   FailedScheduling       pod/virt-launcher-testvmizpzf7f9fhhmcxwl9bj7bn5sl5clf9xg2csg9drz7ws              0/2 nodes are available: 2 Insufficient devices.kubevirt.io/vhost-net.
53s         Warning   FailedScheduling       pod/virt-launcher-testvmizpzf7f9fhhmcxwl9bj7bn5sl5clf9xg2csg9drz7ws              0/2 nodes are available: 1 Insufficient devices.kubevirt.io/kvm, 1 Insufficient devices.kubevirt.io/tun.
56s         Warning   FailedScheduling       pod/virt-launcher-testvmizpzf7f9fhhmcxwl9bj7bn5sl5clf9xg2csg9drz7ws              0/2 nodes are available: 2 Insufficient devices.kubevirt.io/tun.
53s         Warning   FailedScheduling       pod/virt-launcher-testvmizpzf7f9fhhmcxwl9bj7bn5sl5clf9xg2csg9drz7ws              0/2 nodes are available: 1 Insufficient devices.kubevirt.io/kvm, 1 Insufficient devices.kubevirt.io/vhost-net.
```

The allocatable stays at 0 until the virt-handler re-registers the device manager, in this case the virt-handler was restarted because the test that runs before it updates the DS restarting the pods.

cmd/virt-handler/virt-handler.go
```diff
@@ -155,6 +166,19 @@ func (app *virtHandlerApp) Run() {
...
+	app.markNodeAsUnschedulable(logger)
+
+	go func() {
+		sigint := make(chan os.Signal, 1)
+
+		signal.Notify(sigint, syscall.SIGTERM)
+
+		<-sigint
+
+		app.markNodeAsUnschedulable(logger)
```
rmohr, line 179: Don't you have to exit here? I would expect that otherwise we would get a sigkill when the grace period is over.

ashleyschuett: Yeah, you are correct, i missed that. I have update it

pkg/virt-handler/device-manager/device_controller.go
xpivarc: Thanks, that looks better. Can you share which test? I feel like we can improve it so it doesn't disturb other tests.

pkg/virt-handler/vm.go
```diff
@@ -2010,7 +2010,14 @@ func (d *VirtualMachineController) heartBeat(interval time.Duration, stopCh chan
 				log.DefaultLogger().Reason(err).Errorf("Can't determine date")
 				return
 			}
-			data := []byte(fmt.Sprintf(`{"metadata": { "labels": {"%s": "true"}, "annotations": {"%s": %s}}}`, v1.NodeSchedulable, v1.VirtHandlerHeartbeat, string(now)))
+
+			kubevirtSchedulable := "true"
```
dhiller, line 2014: I see this as being "stringly typed". Although I don't have a real objection against this I find this looking odd.

rmohr: It is a label. Labels are of type `map[string]string`.

tests/vmi_lifecycle_test.go
```diff
@@ -93,6 +93,25 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
+			})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(pods.Items)).Should(BeNumerically(">", 0))
+
+			pod := pods.Items[0]
+			handlerNamespace := pod.GetNamespace()
+			err = virtClient.CoreV1().Pods(handlerNamespace).Delete(pod.Name, &metav1.DeleteOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+			n, err := virtClient.CoreV1().Nodes().Get(pod.Spec.NodeName, metav1.GetOptions{})
```
xpivarc, line 109: Just to be on safe side. I would use `Eventually` here.

ashleyschuett: I had originally thought the same thing, but the delete call shouldn't return until the go routine is ran and exits. By this point it should have been set, if it is not set the update failed. If we check using `Eventually` that would mean that is was marked during pod start-up.

tests/vmi_lifecycle_test.go
```diff
@@ -93,6 +93,25 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
 		tests.EnsureKVMPresent()
 	})
 
+	Context("when virt-launcher is deleted", func() {
+		It("should label the node with kubevirt.io/schedulable=false", func() {
+			pods, err := virtClient.CoreV1().Pods("").List(metav1.ListOptions{
+				LabelSelector: fmt.Sprintf("%s=%s", v1.AppLabel, "virt-handler"),
+			})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(pods.Items)).Should(BeNumerically(">", 0))
```
rmohr, line 102: `Expect(pods.Items).ToNot(BeEmpty())` should work too.

tests/vmi_lifecycle_test.go
rmohr: I think @xpivarc is right. You actually have to use an Eventually function here. I wonder why it works at all. Setting the DeletionTimestamp on a Pod is completely independent from the kubelet which will send the SIGTERM.

Have a look here on how to ensure that virt-handler pod gets evicted just from the target node and is not rescheduled until you are done with your check: https://github.com/kubevirt/kubevirt/blob/master/tests/vmi_lifecycle_test.go#L609.

ashleyschuett: Gottcha, makes sense. I have updated this to use `Eventually`

tests/vmi_lifecycle_test.go
```diff
@@ -93,6 +93,29 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
 		tests.EnsureKVMPresent()
 	})
 
+	Context("when virt-launcher is deleted", func() {
```
rmohr, line 96: "when virt-handler is deleted"

tests/vmi_lifecycle_test.go
```diff
@@ -93,6 +93,29 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
+		It("should label the node with kubevirt.io/schedulable=false", func() {
```
rmohr, line 97: We have also a test which looks like this:

```go
			It("[test_id:1634]the node controller should react", func() {

				// Update virt-handler heartbeat, to trigger a timeout
				data := []byte(fmt.Sprintf(`{"metadata": { "annotations": {"%s": "%s"}}}`, v1.VirtHandlerHeartbeat, nowAsJSONWithOffset(-10*time.Minute)))
				_, err = virtClient.CoreV1().Nodes().Patch(nodeName, types.StrategicMergePatchType, data)
				Expect(err).ToNot(HaveOccurred(), "Should patch node successfully")
```

I think that with your code-change it would not properly test this section anymore:


```go
				// it will take at least 45 seconds until the vmi is gone, check the schedulable state in the meantime
				By("marking the node as not schedulable")
				Eventually(func() string {
					node, err := virtClient.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
					Expect(err).ToNot(HaveOccurred(), "Should get node successfully")
					return node.Labels[v1.NodeSchedulable]
				}, 20*time.Second, 1*time.Second).Should(Equal("false"), "The node should not be schedulable")

```

I think that we need to tweak the patch in this test there, to explicitliy set the node to schedulable too, since now the eviction of virt-handler would already set it to false, while we want to test here the node-controller which should do that if virt-handler couldn't.

ashleyschuett: ahh yes, you are correct. Thank you catching that. I have updated it as you suggested to make sure this is testing node-controller and not the virt-handler ds

cmd/virt-handler/virt-handler.go
```diff
@@ -155,6 +166,20 @@ func (app *virtHandlerApp) Run() {
 	if err != nil {
 		panic(err)
 	}
+
+	app.markNodeAsUnschedulable(logger)
+
+	go func() {
+		sigint := make(chan os.Signal, 1)
+
+		signal.Notify(sigint, syscall.SIGTERM)
```
danielBelenky, line 175: We will hit the bug again if virt-handler will be killed.

Though this appraoch minimizes the chances for that to happen but we still can hit the bug.

Maybe we need to move the logic the marks nodes as un-schedulable to another component (virt-controller? :thinking: )
and do something like:

1. If there's a healthy virt-handler on that node, do nothing.
2. Otherwise, mark it as un-schedulable.

virt-handler will then be responsible to mark the node as schedulable instead of unschedulable.

tests/vmi_lifecycle_test.go
```diff
@@ -93,6 +93,29 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
...
+			err = virtClient.CoreV1().Pods(handlerNamespace).Delete(pod.Name, &metav1.DeleteOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+			Eventually(func() string {
+				n, err := virtClient.CoreV1().Nodes().Get(pod.Spec.NodeName, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+
+				return n.Labels[v1.NodeSchedulable]
+			}, 20*time.Second, 1*time.Second).Should(Equal("false"))
+
```
danielBelenky, line 115: I think that we need another test that checks what happens when there is no graceful shut-down of virt-handler.

cmd/virt-handler/virt-handler.go
rmohr: > We will hit the bug again if virt-handler will be killed.

Not sure we have a bug here. We can thing about faster recovery. But we will recover in the case that a bug or another serious issues in the cluster will cause a forced shutdown. We have a heartbeat right now from virt-handler. Once we are sure that virt-handler is not in control anymore, our node-controller (in the virt-controller binary already) will mark it as unschedulable.

> Maybe we need to move the logic the marks nodes as un-schedulable to another component (virt-controller?  )
> and do something like:
> 
> 1. If there's a healthy virt-handler on that node, do nothing.
> 2. Otherwise, mark it as un-schedulable.

On the one hand I like the idea of immediately marking the node as unschedulable if we see that no virt-handler is running on the node. There is then really no need to wait for the heartbeat to expire. On the other hand I like about the heartbeat check that it is a clear interface and that virt-controller does not interpret virt-handler. That makes things more complicated in case we re-architect parts of virt-handler.

tests/vmi_lifecycle_test.go
rmohr: This test exists already. Our node-controller in the virt-controller binary takes care of this. It watches the nodes and sees if virt-handlers heartbeat is gone.

tests/vmi_lifecycle_test.go
rmohr: @ashleyschuett that typo is probably worth fixing.

pkg/virt-handler/vm.go
```diff
@@ -2010,7 +2010,14 @@ func (d *VirtualMachineController) heartBeat(interval time.Duration, stopCh chan
...
+			sp := device_manager.SocketPath(device_manager.KVMName)
```
rmohr, line 2015: Sorry for the late change request, but I think this is a little bit too unclear. We should establish a clear check.

How about explicitly probing our internal device plugins if the successfully managed to register to the kubelet?

ashleyschuett: @rmohr would you mind relooking over this PR? I have updated it to have a clear check. If it seems like the correct way to go ill add tests for it.

cmd/virt-handler/virt-handler.go
danielBelenky: Alright, as you suggested - let's discuss it as part of future optimizations and see if the suggested approach here worth the work that it requires. Since we have the heartbeat mechanism, I'm not 100% sure that it worth it.

tests/vmi_lifecycle_test.go
danielBelenky: :+1: thanks for the info

pkg/virt-handler/device-manager/generic_device.go
```diff
@@ -340,3 +345,17 @@ func (dpi *GenericDevicePlugin) healthCheck() error {
...
 	return filepath.Join(pluginapi.DevicePluginPath, fmt.Sprintf("kubevirt-%s.sock", deviceName))
 }
+
+func (dpi *GenericDevicePlugin) Started() bool {
+	return dpi.started
+}
+
+func (dpi *GenericDevicePlugin) Healthy() bool {
+	for _, dev := range dpi.devs {
+		if dev.Health != pluginapi.Healthy {
```
rmohr, line 355: This seems to indicate that we would only ever set the node to schedulable if the kvm device is actually present. I think what we want is just a check that we could talk to the kubelet successfully.

pkg/virt-handler/device-manager/generic_device.go
```diff
@@ -340,3 +345,7 @@ func (dpi *GenericDevicePlugin) healthCheck() error {
 func SocketPath(deviceName string) string {
 	return filepath.Join(pluginapi.DevicePluginPath, fmt.Sprintf("kubevirt-%s.sock", deviceName))
 }
+
+func (dpi *GenericDevicePlugin) Initialized() bool {
+	return dpi.initialized
```
rmohr, line 350: Do you need a lock?

ashleyschuett: I didn't think I did? The `Start()` function where the variable is changed is being called from a single go routine and the other location where it is change in  and `Stop()` is called from `Start()` and is ran synchronously. Or am I thinking about this wrong? Since it's in a go routine and being called from code not in the same routine should it have a lock anyway?

rmohr: I think that you are reading the value in the health check. If you change it via `Stop()` in one go-routine you may still read the old value in the health-check.

rmohr:
/retest

Interesting. Seems like all jobs got terminated.

pkg/virt-handler/device-manager/generic_device.go
```diff
@@ -176,6 +183,7 @@ func (dpi *GenericDevicePlugin) Start(stop chan struct{}) (err error) {
 func (dpi *GenericDevicePlugin) Stop() error {
 	defer close(dpi.done)
 	dpi.server.Stop()
+	dpi.initialized = false
```
rmohr, line 186: This would also have to be written to by the lock. Maybe create helpers `setInitialized` and `getInitialized` which use a lock, just for clarity, always when you change or read it (even on Start).

rmohr:
I interrupted to 1.18 lane with a CI update.

PR #3795, opened 2020-07-08
[release-0.30] label tests of connectivity over masquerade

phoracek:

When KubeVirt is build on an base image that uses nftables but is
deployed on a cluster where some of the nodes have older iptables,
masquerade binding does not work (it can't call to older iptables on
hosts from newer nftables).

With this patch, we label tests that exercise connectivity over
masquerade binding so they can be skipped on environments where nftables
and iptables are mixed. For these environments, only the bridge binding
is relevant.

These tests can be skipped by passing an environment variable to the
test run:

```bash
FUNC_TEST_ARGS="--ginkgo.skip=\[label:masquerade_binding_connectivity\]" make functest
```

Fixes #

**Release note**:

```release-note
All tests exercising network connectivity over masquerade binding were label to make filtering easier
```

### DIFF
--- a/tests/expose_test.go
+++ b/tests/expose_test.go
@@ -63,7 +63,7 @@ var _ = Describe("[rfe_id:253][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 		Context("Expose ClusterIP service", func() {
 			const servicePort = "27017"
 			const serviceName = "cluster-ip-vmi"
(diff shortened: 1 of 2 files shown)


phoracek:
/assign @AlonaKaplan 

Could you please take a look?

phoracek:
/assign @stu-gott 

Would you please consider approving this for 0.30? I want to use this to skip masquerade tests on RHEL7 test runs.

rmohr:
@stu-gott ping.

PR #3796, opened 2020-07-08
Allow publish of an already published CSV

danielBelenky:

Currently, the manifest-templator fails if we try to publish a bundle
that was already published. This can happen for example if we try to run
the release and fail after we published the CSV.

**Release note**:

```release-note
NONE
```


rmohr:
/hold

Wait, I don't think that we can do that. We rebuild the images every time when we trigger the release job. If we don't replace the published CSV, we would publish an inconsistent state.

rmohr:
@danielBelenky maybe we should just replace the bundle, since we pushed at this stage already fresh images anyway.

danielBelenky:
Closing as we decided that this way introduces a bug in the release.

PR #3801, opened 2020-07-09
Fix some synchronization issues in the static tests

jean-edouard:
When running the go tests with the race detector on, some tests panic on synchronization issues.
This PR attempts to fix most of them (leaving only a tricky one related to grpc server).
Some of those fixes may not be appropriate, feedback is welcome!

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-handler/isolation/isolation_test.go
+++ b/pkg/virt-handler/isolation/isolation_test.go
@@ -24,6 +24,7 @@ var _ = Describe("Isolation", func() {
 		var tmpDir string
 		var podsDir string
 		var podUID string
+		var finished chan struct{} = nil
 
 		podUID = "pid-uid-1234"
 		vm := v1.NewMinimalVMIWithNS("default", "testvm")
@@ -48,10 +49,12 @@ var _ = Describe("Isolation", func() {
 			os.MkdirAll(filepath.Dir(socketFile), os.ModePerm)
 			socket, err = net.Listen("unix", socketFile)
 			Expect(err).ToNot(HaveOccurred())
+			finished = make(chan struct{})
 			go func() {
 				for {
 					conn, err := socket.Accept()
 					if err != nil {
+						close(finished)
 						// closes when socket listener is closed
 						return
 					}
@@ -100,6 +103,9 @@ var _ = Describe("Isolation", func() {
 			socket.Close()
 			os.RemoveAll(tmpDir)
 			os.RemoveAll(podsDir)
+			if finished != nil {
+				<-finished
+			}
 		})
 	})
 })
--- a/pkg/virt-launcher/monitor_test.go
+++ b/pkg/virt-launcher/monitor_test.go
@@ -47,6 +47,7 @@ var _ = Describe("VirtLauncher", func() {
 	var mon *monitor
 	var cmd *exec.Cmd
 	var cmdLock sync.Mutex
+	var gracefulShutdownChannel chan struct{}
 
 	uuid := "123-123-123-123"
 
@@ -56,8 +57,6 @@ var _ = Describe("VirtLauncher", func() {
 	dir = strings.TrimSuffix(dir, "pkg/virt-launcher")
 
 	processStarted := false
-
-	gracefulShutdownCallbackTriggered := false
 
 	StartProcess := func() {
 		cmdLock.Lock()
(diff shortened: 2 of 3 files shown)


pkg/virt-launcher/monitor_test.go
```diff
@@ -120,6 +121,8 @@ var _ = Describe("VirtLauncher", func() {
 			syscall.Kill(pid, syscall.SIGTERM)
 		}
 		gracefulShutdownCallback := func() {
+			gracefulShutdownCallbackTriggeredLock.Lock()
```
rmohr, line 124: Could we use a channel here too?

jean-edouard: Do you mean replacing the boolean `gracefulShutdownCallbackTriggered` with a channel?
How would I deal with `Expect(gracefulShutdownCallbackTriggered).To(BeFalse())`?

rmohr: Something like

```
Eventually(channel).Should(BeClosed())
```
Should work

jean-edouard: Right, that would replace the `BeTrue()` test, but what about `BeFalse()`? Just skip it?

rmohr: Did not try this but I would assume `Consistenlty(channel).Should(Not(BeClosed)))`

jean-edouard: Interesting, I wonder how that's checked without doing a read(), probably some golang magic. I'll try that, thank you!

jean-edouard: Seems to work just fine, force-pushed.

pkg/virt-launcher/monitor_test.go
```diff
@@ -206,11 +205,11 @@ var _ = Describe("VirtLauncher", func() {
 				}()
 
 				time.Sleep(time.Second)
-				Expect(gracefulShutdownCallbackTriggered).To(BeFalse())
+				Expect(gracefulShutdownChannel).To(Not(BeClosed()))
```
rmohr, line 208: Let's get rid of the Sleep and do instead

```
Eventually(gracefulShutdownChannel).To(Not(BeClosed()))
```

pkg/virt-launcher/monitor_test.go
```diff
@@ -206,11 +205,11 @@ var _ = Describe("VirtLauncher", func() {
...
 				close(stopChan)
 
 				time.Sleep(time.Second)
-				Expect(gracefulShutdownCallbackTriggered).To(BeTrue())
+				Expect(gracefulShutdownChannel).To(BeClosed())
```
rmohr, line 212: Let's use here

```
Eventually(gracefulShutdownChannel).To(BeClosed())
```

pkg/virt-launcher/monitor_test.go
```diff
@@ -205,12 +204,9 @@ var _ = Describe("VirtLauncher", func() {
 					done <- "exit"
 				}()
 
-				time.Sleep(time.Second)
-				Expect(gracefulShutdownCallbackTriggered).To(BeFalse())
+				Consistently(gracefulShutdownChannel).Should(Not(BeClosed()))
```
EdDev, line 207: This is using polling and changes the timeout from 1sec to 100msec.
How about using events/signals instead? E.g a timer with a select.

It may be an overkill, but I dislike the use of the polling too much to avoid commenting about it. :)

pkg/virt-handler/isolation/isolation_test.go
```diff
@@ -48,16 +49,18 @@ var _ = Describe("Isolation", func() {
 			os.MkdirAll(filepath.Dir(socketFile), os.ModePerm)
 			socket, err = net.Listen("unix", socketFile)
 			Expect(err).ToNot(HaveOccurred())
-			go func() {
+			finished = make(chan bool)
```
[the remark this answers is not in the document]
rmohr: Channels are preferred. The rest makes sense.

pkg/virt-launcher/monitor_test.go
rmohr: `Consistently` and `Eventually` are the way to go. `selects` would only help with `Eventually` and not with `Consistently`, so we would end up again with `Sleep` which does not mean much. The gomega language transports pretty clearly what one wants to test.

PR #3802, opened 2020-07-10
Add the KubeVirt resource to the published openapi spec

rmohr:

It is an oversight that the KubeVirt resource was never published.
Changing that.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -118,6 +118,509 @@
      }
     }
    },
+   "/apis/kubevirt.io/v1alpha3/kubevirt": {
+    "get": {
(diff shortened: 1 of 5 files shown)


api/openapi-spec/swagger.json
```diff
@@ -7525,6 +8513,26 @@
      }
     }
    },
+   "v1.SMBiosConfiguration": {
+    "type": "object",
+    "properties": {
+     "family": {
+      "type": "string"
```
rmohr, line 8520: We should definitely improve in a follow-up our very minimalistic description of all these fields.

PR #3814, opened 2020-07-13
virt-launcher: Remove unneeded condition from StartDomainNotifier

oshoval:
AgentUpdated event depends on agentPoller.Start(),
so the flow would be:
Receiving event on eventChan
Assigning domainCache
Start agent poller
    
And just then AgentUpdated event can occur,
and domainCache isnt nil in this case, as it was
set as seen above.

```release-note
None
```

### DIFF
--- a/pkg/virt-launcher/notify-client/client.go
+++ b/pkg/virt-launcher/notify-client/client.go
@@ -330,7 +330,7 @@ func (n *Notifier) StartDomainNotifier(
 			case agentUpdate := <-agentStore.AgentUpdated:
 				interfaceStatuses = agentUpdate.DomainInfo.Interfaces
 				guestOsInfo = agentUpdate.DomainInfo.OSInfo
-				if domainCache != nil && interfaceStatuses != nil {
+				if interfaceStatuses != nil {
 					interfaceStatuses = agentpoller.MergeAgentStatusesWithDomainData(domainCache.Spec.Devices.Interfaces, interfaceStatuses)
 				}
 


oshoval:
/cc @enp0s3 
Thanks for the research

enp0s3:
@petrkotas Hi. Can you please confirm this condition is redundant? Thanks!

enp0s3:
> Hi yes, it is like you said. @enp0s3 the condition was there from early phase. It can be removed at this point.

@petrkotas Thanks!

oshoval:
Travis has some issue, draft and undraft didnt help, will try to commit amend
`{"message":"service_job_id (707543227) must be unique for Travis Jobs not supplying a Coveralls Repo Token","error":true}`

oshoval:
@enp0s3 
https://github.com/kubevirt/kubevirt/pull/3814/files#diff-7175e08324a35a860b12d6518d396988L341
this return was removed yesterday, it will be shown when i rebase,
question is should this case Stop the Poller ?
does it now without the return do have a possible race where also the domainCache can be outdated,
if it is from the first round ?

AlonaKaplan:
> AgentUpdated will start only after agentPoller.Start()
> is called, and Start is called only after creating domainCache,
> so the check isnt needed and is misleading that eventCallback
> might be called with nil domainCache.
> 

It was very hard for me to follow this PR message.

I would suggest something like this.

The flow is as following -
create `domainCache` -> `agentPoller.Start()` -> `AgentUpdated`

So when an `AgentUpdated` event is received. the `domainCache` is initialized and there is no need to verify it.

oshoval:
> > AgentUpdated will start only after agentPoller.Start()
> > is called, and Start is called only after creating domainCache,
> > so the check isnt needed and is misleading that eventCallback
> > might be called with nil domainCache.
> 
> It was very hard for me to follow this PR message.
> 
> I would suggest something like this.
> 
> The flow is as following -
> create `domainCache` -> `agentPoller.Start()` -> `AgentUpdated`
> 
> So when an `AgentUpdated` event is received. the `domainCache` is initialized and there is no need to verify it.
> 
> Edit: I don't know if what I wrote here is correct, just tried to translate your comment.

done

oshoval:
@enp0s3 
please take another look, rebased + changed commit msg

oshoval:
lots of failues at
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3814/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1283745673329512449
but most of them i saw in the past, i dont believe its because of this PR, lets verify

/retest

PR #3825, opened 2020-07-14
virtctl: Print response body on error from upload proxy.

awels:
To help users debug issues.

Virtctl ignores any content from the body on an error during upload. This means we just report the status code. With https://github.com/kubevirt/containerized-data-importer/pull/1299 CDI now reports validation failures as part of the body response in case for instance the virtual disk image size > PVC size. This PR updates virtctl to print the error message from the response body.

Fixes #

**Release note**:

```release-note
Virtctl now prints error messages from the response body on upload errors.
```

### DIFF
--- a/pkg/virtctl/imageupload/imageupload.go
+++ b/pkg/virtctl/imageupload/imageupload.go
@@ -22,6 +22,7 @@ package imageupload
 import (
 	"crypto/tls"
 	"fmt"
+	"io/ioutil"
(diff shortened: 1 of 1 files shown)


jean-edouard:
/approve
Good catch on the close (from the http lib doc: `If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close.`)

PR #3837, opened 2020-07-16
libvirt: disable PXE rom on interfaces with no boot order

jean-edouard:
The API reference documentation specifies that boot will not be tried on interfaces with no boot order.
However, PXE boot does get attempted on VMs without disks.
This PR adds <rom bar="off"> to the libvirt xml of such interfaces to prevent that.

Fixes #

**Release note**:

```release-note
- VM interfaces with no `bootOrder` will no longer be candidates for boot when using the BIOS bootloader, as documented
- A new API option was introduced to log emulated BIOS messages to serial
```


rmohr:
> The API reference documentation specifies that boot will not be tried on interfaces with no boot order.
> However, PXE boot does get attempted on VMs without disks.
> This PR adds to the libvirt xml of such interfaces to prevent that.

Why do we want to disable PXE boot?

jean-edouard:
> Why do we want to disable PXE boot?

Apart from fixing the incoherence between docs and reality, I think it's good to have a way to disable PXE boot in VMs.
This effectively disables it by default. If that's too much, maybe we want to add a new interface option?

rmohr:
> Apart from fixing the incoherence between docs and reality, I think it's good to have a way to disable PXE boot in VMs.
> This effectively disables it by default.

Now I get it. Thanks.

jean-edouard:
/hold
Breaks rom-related tests.

rmohr:
/approve cancel

Let's clarify this a little more. What we try to solve is the case where the vmi has no disks, right? I am axtually not sure if we would break migrating existing vms with the general change, which were already started. It could work since the domain xml is transferred too, but let's explore this a little bit more.

jean-edouard:
@rmohr I'm not sure I understand your concern.
As you said, migration is not a problem, and it wouldn't be even if the xml wasn't transferred over.
The only potential concern is if users previously created non-network-bootable VMs and actually used network boot on them, but the release note should take care of that...

jean-edouard:
Added a functional test in a separate commit.
The test requires the emulated BIOS to log to serial, so an API option was added for that.
@rmohr PTAL!

tests/vmi_configuration_test.go
```diff
@@ -460,6 +460,62 @@ var _ = Describe("Configurations", func() {
...
+
+				By("Ensuring network boot is disabled on the network interface")
+				Expect(vmi.Spec.Domain.Devices.Interfaces[0].BootOrder).To(BeNil())
+
+				By("Starting a VirtualMachineInstance")
+				vmi, err = virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(vmi)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Expecting no bootable NIC")
+				_, err = tests.NetBootExpecter(vmi)
```
rmohr, line 486: The ntebootexpecter seems to be very specific. It may make sense to not place it in util.go, but that is just a nit.

rmohr (approved):
/lgtm
/approve

PR #3838, opened 2020-07-16
Update cloud images

jean-edouard:
Just updates Alpine and Cirros to the latest available versions.
Both now requires 128M of RAM.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -73,37 +73,37 @@ http_file(
 # Disk images
 http_file(
     name = "alpine_image",
-    sha256 = "5a4b2588afd32e7024dd61d9558b77b03a4f3189cb4c9fc05e9e944fb780acdd",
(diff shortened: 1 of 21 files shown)


danielBelenky (changes requested):
@jean-edouard can you please elaborate in the description why we want to update? does it fix any bug what we currently have?

jean-edouard:
@danielBelenky I did this locally when I was looking for a secureboot-signed image, hoping one of the latest images would be signed.
They don't happen to be, but it took a while to update everything to 128MB (which both Alpine and Cirros require in recent versions), so I thought I shouldn't let it go to waste, assuming we'll want to update the images one day or another anyway.

I'm not against closing this PR for now, I can always reopen it if/when we decide to update.

PR #3841, opened 2020-07-17
New metric for total memory available for a VMI

### LINKED ISSUE #3081: Add kubevirt_vmi_memory_total_bytes metric

>
> /kind bug

/kind enhancement

**What happened**:

virt-handler expose `kubevirt_vmi_memory_total_bytes `metric

**Environment**:
- VM or VMI specifications:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ArthurSens:
Signed-off-by: arthursens <arthursens2005@gmail.com>

This PR adds a new metric that represents the total available memory for a VMI, in bytes. This allows users to easily identify VMIs that are lacking or have too much memory and then resize them as needed.

Fixes #3081

**Release note**:

```release-note
Add metric kubevirt_vmi_memory_total_bytes
```

### DIFF
--- a/pkg/monitoring/vms/prometheus/prometheus.go
+++ b/pkg/monitoring/vms/prometheus/prometheus.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	k8sv1 "k8s.io/api/core/v1"
 
 	k6tv1 "kubevirt.io/client-go/api/v1"
 	"kubevirt.io/client-go/kubecli"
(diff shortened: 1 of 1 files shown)


ArthurSens:
I'm using `vmi.Spec.Domain.Resources.Limits` as value for this metric

But when the limit is not set, which is not uncommon, how is the total available memory set for a VMI? Currently, I'm using  `vmi.Spec.Domain.Resources.Requests` when limits == 0, but I think the VMI could use all memory available on the node it is running, is that correct?

xpivarc:
Do we need this if we have `kubevirt_vmi_memory_available_bytes
`?

ArthurSens:
> Do we need this if we have `kubevirt_vmi_memory_available_bytes `?

yeah... I've created this PR some time ago just because there was an issue, but not really thinking about if there was a need for it :facepalm: 

`kubevirt_vmi_memory_available_bytes ` and `kubevirt_vmi_memory_unused_bytes ` should be enough for any memory usage use-case

PR #3844, opened 2020-07-17
Remove unused feature gates

ashleyschuett:

Removing unused feature gates from the kubevirt configuration. 

Since these feature gate flags are not checked I think they should be removed. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/virt-config/feature-gates.go
+++ b/pkg/virt-config/feature-gates.go
@@ -32,7 +32,6 @@ const (
 	SidecarGate           = "Sidecar"
 	GPUGate               = "GPU"
 	SnapshotGate          = "Snapshot"
-	DataVolumesGate       = "DataVolumes"
 	SRIOVGate             = "SRIOVGate"
 	HostDiskGate          = "HostDisk"
 )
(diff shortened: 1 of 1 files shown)


pkg/virt-config/feature-gates.go
```diff
@@ -32,8 +32,6 @@ const (
 	SidecarGate           = "Sidecar"
 	GPUGate               = "GPU"
 	SnapshotGate          = "Snapshot"
-	DataVolumesGate       = "DataVolumes"
-	SRIOVGate             = "SRIOVGate"
```
rmohr, line 36: @vladikr I am surpised that SRIOV is not gated. I thought that it involves some extra privileges.

vladikr: @rmohr seems like you've approved it: https://github.com/kubevirt/kubevirt/pull/2546

vladikr: @rmohr, afaik, there are no additional privileges with sr-iov, but we will lock 1G of memory for VMIs with SR-IOV.

rmohr: > @rmohr seems like you've approved it: #2546

ups

> @rmohr, afaik, there are no additional privileges with sr-iov, but we will lock 1G of memory for VMIs with SR-IOV.

 perfect that was what I was unsure about.

PR #3849, opened 2020-07-20
Add a small helper script to pass git tokens in a secure way

rmohr:

Add a script which allows setting `GIT_ASKPASS` to securely provide a
github token to git. git expects a single "binary" without arguments which will print the token to stdout when invoked.

Needed to provide the git credentials in a secure way to the jobs in https://github.com/kubevirt/project-infra/pull/352

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/git-askpass.sh
+++ b/hack/git-askpass.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+cat $1


rmohr:
/close

I made up my mind. I think it is better to have this somewhere in project-infra and provide it in a general fashion to jobs, so that it works for different repos.

PR #3853, opened 2020-07-20
Support qcow2 image format for HostDisk & FS PVC

jcaamano:
Adds support for qcow2 images for HostDisk and file-system PVC configurations whereas currently only raw images are supported. This does not add support for creating qcow2 images on those configurations, only supports existing images.

**Release note**:

```release-note
Add support for exisiting qcow2 images on HostDisk and file-system PVC configurations.
```


pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -1009,9 +1009,17 @@ func (l *LibvirtDomainManager) preStartHook(vmi *v1.VirtualMachineInstance, doma
 		return domain, fmt.Errorf("creating service account disk failed: %v", err)
 	}
 
-	// set drivers cache mode
 	for i := range domain.Spec.Devices.Disks {
-		err := api.SetDriverCacheMode(&domain.Spec.Devices.Disks[i])
+		// Set missing image format. For some live images (host disk, fs pvc...)
+		// it is convenient to check image formats on the pre-start hook as it
+		// might fail if VM is already running and the image is already in use.
+		err := api.SetFileDriverType(&domain.Spec.Devices.Disks[i], GetImageInfo)
```
rmohr, line 1016: In general we have a converter context which allows passing in the disk type to the libvirt converter. Consider using that. You can find an example at https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-launcher/virtwrap/manager.go#L818 where we do this for the containerDisk. This allows us to keep all translations inside the converter.

pkg/virt-launcher/virtwrap/manager.go
```diff
@@ -1221,6 +1229,7 @@ func (l *LibvirtDomainManager) SyncVMI(vmi *v1.VirtualMachineInstance, useEmulat
 
 	// To make sure, that we set the right qemu wrapper arguments,
 	// we update the domain XML whenever a VirtualMachineInstance was already defined but not running
+	// FIXME: This does not work as we lose all the changes done to the domain spec in preStartHook
```
rmohr, line 1232: Do we already do that, or do you talk about your changes? Let's not change the domain spec in the preStartHook. The domain-spec from the kubevirt-perspective should be final and only hooks, in the sense of external hooks, should modify it at this stage.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -189,6 +189,29 @@ func SetDriverCacheMode(disk *Disk) error {
...
+	diskFileFormatUnset := disk.Type == "file" && disk.Source.File != "" && disk.Driver.Type == ""
+	if diskFileFormatUnset {
+		diskInfo, err := diskInfoGetter(disk.Source.File)
+		if err != nil {
+			return err
+		}
+		disk.Driver.Type = diskInfo.Format
+	}
+	// Check also for a backing file
+	backingStore := disk.BackingStore
```
rmohr, line 202: qcow2 is great, but it has this history of allowing DOS or security relevant attack angles if not properly sanitized. For containerDisk we explicitly disallow backing files and we ensure that we already do a validation-check on the qcow2 in a confined environment before it ever reaches virt-launcher. Have a look at https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-handler/container-disk/mount.go#L258 and https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-handler/isolation/validation.go#L18.

I would recommend that we do the same for any disk type which saves qcow2. If backing-files are needed, let's maybe analyze after this PR what we can do there, but let's start without them.

pkg/virt-launcher/virtwrap/api/converter.go
```diff
@@ -189,6 +189,29 @@ func SetDriverCacheMode(disk *Disk) error {
 	return nil
 }
 
+func SetFileDriverType(disk *Disk, diskInfoGetter func(string) (*containerdisk.DiskInfo, error)) error {
+	diskFileFormatUnset := disk.Type == "file" && disk.Source.File != "" && disk.Driver.Type == ""
+	if diskFileFormatUnset {
+		diskInfo, err := diskInfoGetter(disk.Source.File)
```
aglitke, line 195: We should always rely on a trusted source of information (other than the file itself) to determine if a file should be treated as a qcow2 image.  Several years ago there was a CVE where a malicious VM could write a fake qcow2 header and qcow2 metadata to the beginning of a raw disk image to trick the hypervisor into believing the file is qcow2.  This could lead to exposing sensitive data to the VM on subsequent reboots.  For this reason libvirt requires the format to be stated explicitly.  You will probably need to modify this PR so that the disk file format is exposed in the VM spec.  It's unfortunate that such a detail must be exposed to the end user but I do not see any way around it.

rmohr: > We should always rely on a trusted source of information (other than the file itself) to determine if a file should be treated as a qcow2 image. Several years ago there was a CVE where a malicious VM could write a fake qcow2 header and qcow2 metadata to the beginning of a raw disk image to trick the hypervisor into believing the file is qcow2. This could lead to exposing sensitive data to the VM on subsequent reboots. For this reason libvirt requires the format to be stated explicitly. You will probably need to modify this PR so that the disk file format is exposed in the VM spec. It's unfortunate that such a detail must be exposed to the end user but I do not see any way around it.

I think we should talk that through a little. It is out of question that qcow2 requires careful validation. But let's also keep in mind that we are now getting pretty close to the unprivileged pod scenario for running VMs. Even if one can escape it will not see anything which a user can't access anyway then.

I am not really convinced if it changes anything if a user specifies the disk type. I think this all boils down to no use questionable sources in general and it is less about the file format detection. Still i think that we can reach a pretty solid security barrier by properly validating the qcow and the encapsulation which pods give us. At the end I think that allowing people to pull and execute  container images per se is already as least as insecure as this scenario, where one would mostly just escape into the containers jail. vs. directly executing untrusted code in the pod.

pkg/virt-launcher/virtwrap/manager.go
jcaamano: Yes, I was aware of the converter context on purpose. `qemu-img` fails if the image is already in use as non backing store on every synchronization other than the first startup of the VM. This is the reason I used the pre-start hook, as it only runs on initial first startup. I try to explain this in the comment above, so maybe i need to re-write so that it is better understood.

pkg/virt-launcher/virtwrap/manager.go
jcaamano: This problem is already happening and is not related to my changes. Any changes done in the pre-start hook (for example, network bindings are resolved in the pre-start hook) will get lost if a defined non-running VM spec is re-synchronized afterwards. I have observed this many times on failure scenarios where the first startup will fail because something done on the pre-start hook causes problems, and on further attempts it will fail because the pre-start hook changes are not considered. Of course this is not observed most of the time because the normal state of a VM is either to be running or undefined on synchronizations.

pkg/virt-launcher/virtwrap/api/converter.go
jcaamano: On the backing files, this only applies for ephemeral disks backed with file-system PVC. These are created by kubevirt and are always raw images. There should be no issue. But I will see a way around to avoid it.

ACK on validation, I will see a way to validate the image on the virt-handler context as well.

pkg/virt-launcher/virtwrap/manager.go
rmohr: hm, that sounds like there are bugs in the retry logic then. Would be great to get a bug report next time you see it. Thanks for the details.

rmohr: @AlonaKaplan FYI, maybe we should look through our code parts there again and try to detect non-idempotent  steps.

pkg/virt-launcher/virtwrap/manager.go
rmohr: ok, I still think that we should know at this stage already what we are dealing with. We can extract and save the information somewhere else after a first successful extraction.

jcaamano:
@rmohr @aglitke 

Sorry it took me so long to get back to this. I added validation in the virt-handler context.

> ok, I still think that we should know at this stage already what we are dealing with. We can extract and save the information somewhere else after a first successful extraction.

I have not found a good place to save this other than the VMI spec itself. Maybe you had something more specific in mind or I am missing something?

> You will probably need to modify this PR so that the disk file format is exposed in the VM spec. It's unfortunate that such a detail must be exposed to the end user but I do not see any way around it.

If we do this, I am assuming it would only be used in the cases where the image already exist to verify the format (vs allowing to create an image of the specified format)? And that no empty value would be allowed? Or do you have a different opinion?

PR #3857, opened 2020-07-20
test connectivity over both bridge and masquerade

phoracek:

Currently, we extensively test the pod network connectivity only through
the bridge binding mechanism.

We want to use the same set of tests for the masquerade binding
mechanism which is another supported binding option that unlike the
bridge binding supports migration too.

This patch makes the current test more generic and exercises it with
both bindings set as defaults.

Signed-off-by: Petr Horacek <phoracek@redhat.com>

Fixes #

This PR is split to two commits to make it easier to review.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/vmi_networking_test.go
+++ b/tests/vmi_networking_test.go
@@ -118,300 +118,304 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 		tests.UpdateClusterConfigValueAndWait("permitBridgeInterfaceOnPodNetwork", fmt.Sprintf("%t", enable))
 	}
 
-	Describe("Multiple virtual machines connectivity using bridge binding interface", func() {
-		tests.BeforeAll(func() {
-			tests.BeforeTestCleanup()
+	testMultipleVirtualMachinesConnectivityUsingDefaultBindingMechanism := func() {
+		Describe("Multiple virtual machines connectivity using bridge binding interface", func() {
+			tests.BeforeAll(func() {
+				tests.BeforeTestCleanup()
 
-			// Prepare inbound and outbound VMI definitions
+				// Prepare inbound and outbound VMI definitions
 
-			// inboundVMI expects implicitly to be added to the pod network
-			inboundVMI = tests.NewRandomVMIWithEphemeralDiskAndUserdata(tests.ContainerDiskFor(tests.ContainerDiskCirros), "#!/bin/bash\necho 'hello'\n")
-			inboundVMI.Labels = map[string]string{"expose": "me"}
-			inboundVMI.Spec.Subdomain = "myvmi"
-			inboundVMI.Spec.Hostname = "my-subdomain"
-			// Remove the masquerade interface to use the default bridge one
-			inboundVMI.Spec.Domain.Devices.Interfaces = nil
-			inboundVMI.Spec.Networks = nil
+				// inboundVMI expects implicitly to be added to the pod network
+				inboundVMI = tests.NewRandomVMIWithEphemeralDiskAndUserdata(tests.ContainerDiskFor(tests.ContainerDiskCirros), "#!/bin/bash\necho 'hello'\n")
+				inboundVMI.Labels = map[string]string{"expose": "me"}
+				inboundVMI.Spec.Subdomain = "myvmi"
+				inboundVMI.Spec.Hostname = "my-subdomain"
+				// Remove the masquerade interface to use the default bridge one
+				inboundVMI.Spec.Domain.Devices.Interfaces = nil
+				inboundVMI.Spec.Networks = nil
 
-			// outboundVMI is used to connect to other vms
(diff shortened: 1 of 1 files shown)


tests/vmi_networking_test.go
```diff
@@ -118,43 +118,53 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
 				outboundVMI.Spec.Domain.Devices.Interfaces = nil
 				outboundVMI.Spec.Networks = nil
 
 				// inboudnVMIWithPodNetworkSet adds itself in an explicit fashion to the pod network
 				inboundVMIWithPodNetworkSet = tests.NewRandomVMIWithEphemeralDiskAndUserdata(tests.ContainerDiskFor(tests.ContainerDiskCirros), "#!/bin/bash\necho 'hello'\n")
-				// Remove the masquerade interface to use the default bridge one
-				inboundVMIWithPodNetworkSet.Spec.Domain.Devices.Interfaces = nil
-				inboundVMIWithPodNetworkSet.Spec.Networks = nil
-				v1.SetDefaults_NetworkInterface(inboundVMIWithPodNetworkSet)
-				Expect(inboundVMIWithPodNetworkSet.Spec.Domain.Devices.Interfaces).NotTo(BeEmpty())
```
RamLavi: in the original (left side) we set the `inboundVMIWithPodNetworkSet` with [SetDefaults_NetworkInterface](https://github.com/kubevirt/client-go/blob/e297527e7722c2a1aa3445dfc0b77bac29504ac5/api/v1/defaults.go#L145), then `Expect` it to not be empty, 
why did we remove this?

phoracek: I know it is not empty since it was explicitly set in a function defined few lines above. Using `Expect` felt redundant.

tests/vmi_networking_test.go
```diff
@@ -118,300 +118,319 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+				defer expecter.Close()
+
+				addrShow = "ip address show eth0\n"
+				resp, err := expecter.ExpectBatch([]expect.Batcher{
+					&expect.BSnd{S: "\n"},
+					&expect.BExp{R: "\\$ "},
+					&expect.BSnd{S: addrShow},
+					&expect.BExp{R: fmt.Sprintf(".*%s.*\n", expectedMtuString)},
+					&expect.BSnd{S: "echo $?\n"},
+					&expect.BExp{R: "0"},
```
oshoval, line 242: you might want to use \n0 as #3543 fixed,
Alona would create a general mechanisem, and not sure here here there is an issue, but just up to you meanwhile

same for the 2nd batch, which is bit more sensitive, i am not sure the error can happen there, but better be safe,
because we did see some lines which are prone for problems like

```
		&expect.BSnd{S: "sudo ip -6 route add default via fd10:0:2::1 src fd10:0:2::2\n"},
		&expect.BExp{R: prompt},
		&expect.BSnd{S: "echo $?\n"},
		&expect.BExp{R: "0"},
```
where the 0 of the first input was catched as the 2 EXP (in case that i am not wrong and this was the atomic block that caused it without other code around that was affecting it)

phoracek: Tha is not related to this PR. I am only using existing tests with different environment setup.

oshoval: > That is not related to this PR. I am only using existing tests with different environment setup.
> 
> Or do you suspect it is related to IPv6 fails?

not that i know of, i just say that if there is flakiness , or will be flakiness, this is one thing i would surely check

EdDev:
One thing I am unclear about: Why do we need to set a default binding? Isn't an explicit definition per VMI enough?

phoracek:
> The approach to generate the tests with multiple configurations is creative, unfortunately IMO it is a bit too creative to be easily understood and maintained.

I don't agree, it is quite simple to understand and the maintenance cost is not higher than if we had a duplicated tests for each binding or table-driven tests which are quite hard to read.

> 
> Have you considered grouping the setup and tests into functions and calling them using tables?

Yes. The issue with tables is that you cannot setup more `Contexts` inside them or use `It`. Due to that, we cannot simply run the whole test module twice with the only difference being the binding. We would need to parametrize each of the used Tables to run both bindings.

> The VMI/s configuration can be created with both bindings in advanced and the correct ones set as input to the table tests.

Yes. But I don't see how is it easier to read than the function.

> 
> Reusing the same tests would have made sense to me if the caller would have chosen (though a flag) the "default" binding type. But this will probably not work well at this time (as we do not run the tests multiple times).

> One thing I am unclear about: Why do we need to set a default binding? Isn't an explicit definition per VMI enough?

It is mostly to exercise the functionality with the default switched. It makes generalization of the test easier, but that is not a valid argument. We could make it so we test the default and then explicit masquerade and explicit bridge (and both of them with and without explicit MAC).

Transforming this into multiple tables parametrized would be a bigger effort. I understand why you would prefer that, but I currently only care about getting a single connectivity test with masquerade binding. So if you don't agree with the function-based approach, we can close this PR and I would add a new test focused on the basic connectivity. Your call.

PR #3859, opened 2020-07-21
tests: Create flags and containerdisk sub-package

EdDev:

Convert the flags and containerdisk files each to a sub-package in order to consume it in future
changes from other tests sub-packages (and avoid a dependency loop).

Fixes #

Depends on #3846 

**Release note**:
```release-note
NONE
```


EdDev:
> I guess you plan to create more packages and files. Would it make sense to create something like a `tests/util/` or `tests/infra` parent directory where you place all test-infra code in, to make the discovery easier?

I wanted to go in that path, creating a `libtest` folder, but there was some resistance because it looks like the current `utils` from the outside. As the cost to move everything into such a folder is low, I thought it will just become obvious when we will have enough such packages/folders under `/tests` (and it will bother us).

EdDev:
Just rebased to make sure the commit from the dependent PR (which got merged) is not messing up the history/log.

EdDev:
Rebase due to merge conflicts with master.

PR #3866, opened 2020-07-22
WIP check lane status with dual stack

oshoval:
```release-note
None
```


oshoval:
dual stack didnt cause regression at least now
(going to update calico, hope it still wont regression as well)

PR #3479, opened 2020-05-27
Test sriov over vlan

alonSadan:
**What this PR does / why we need it**:
we need to improve sriov testing.
so in this pr we:
1) Change pingVirtualMachine to return
a error and move the assertion outside of it's scope.
This is done so that we can assert that the ping failed as well.

2) warped the http requests for creating network function and deleting
network attachment definitions.  

3) Test  connectivity between two
VMs that share vlan over an SRIOV network.
A third vm without which do not share the
vlan should not get connected.
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:
1. second commit relies on the change in the first one, because as a part of the test we check that a ping to a vm which is not connected to vlan- fails.  
2. if this PR would be approved , we will add follow up PR to wrap all not pretty http requests in functions. I thought it was not related to this PR so I made it just in the SRIOV scope. 
**Release note**:

```release-note
NONE
```

PR #3871, opened 2020-07-22
tests, network: Test connectivity pre/post migration

EdDev:

Live migration is (partially) supported when using masquerade binding
on the primary network.

A new test is introduced, checking the VMI connectivity before and after
the migration.

Fixes #

Depends on #3883 , #3997

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/job.go
+++ b/tests/job.go
@@ -8,46 +8,72 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	k8sv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"kubevirt.io/client-go/kubecli"
-
-	. "github.com/onsi/gomega"
 )
 
-func WaitForJobToSucceed(virtClient *kubecli.KubevirtClient, job *batchv1.Job, timeoutSec time.Duration) {
-	EventuallyWithOffset(1, func() bool {
-		job, err := (*virtClient).BatchV1().Jobs(job.Namespace).Get(job.Name, metav1.GetOptions{})
-		ExpectWithOffset(2, err).ToNot(HaveOccurred())
-		for _, c := range job.Status.Conditions {
-			switch c.Type {
-			case batchv1.JobComplete:
-				if c.Status == k8sv1.ConditionTrue {
-					return true
-				}
-			case batchv1.JobFailed:
-				ExpectWithOffset(2, c.Status).NotTo(Equal(k8sv1.ConditionTrue), "Job should succeed")
-			}
-		}
-		return false
-	}, timeoutSec*time.Second, 1*time.Second).Should(BeTrue(), "Job should succeed")
+const (
+	toSucceed = true
+	toFail    = false
+)
+
+// WaitForJobToSucceed blocks until the given job finishes.
+// On success, it returns with a nil error, on failure or timeout it returns with an error.
+func WaitForJobToSucceed(job *batchv1.Job, timeout time.Duration) error {
+	return waitForJob(job, toSucceed, timeout)
 }
 
-func WaitForJobToFail(virtClient *kubecli.KubevirtClient, job *batchv1.Job, timeoutSec time.Duration) {
-	EventuallyWithOffset(1, func() bool {
-		job, err := (*virtClient).BatchV1().Jobs(job.Namespace).Get(job.Name, metav1.GetOptions{})
-		ExpectWithOffset(2, err).ToNot(HaveOccurred())
+// WaitForJobToFail blocks until the given job finishes.
+// On failure, it returns with a nil error, on success or timeout it returns with an error.
+func WaitForJobToFail(job *batchv1.Job, timeout time.Duration) error {
+	return waitForJob(job, toFail, timeout)
+}
+
+func waitForJob(job *batchv1.Job, toSucceed bool, timeout time.Duration) error {
+	virtClient, err := kubecli.GetKubevirtClient()
+	if err != nil {
+		return err
+	}
+
+	jobFailedError := func(job *batchv1.Job) error {
+		if toSucceed {
+			return fmt.Errorf("Job %s finished with failure, status: %+v", job.Name, job.Status)
+		}
+		return nil
+	}
+	jobCompleteError := func(job *batchv1.Job) error {
+		if toSucceed {
+			return nil
+		}
+		return fmt.Errorf("Job %s finished with success, status: %+v", job.Name, job.Status)
+	}
+
+	const finish = true
+	err = wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		job, err = virtClient.BatchV1().Jobs(job.Namespace).Get(job.Name, metav1.GetOptions{})
+		if err != nil {
+			return finish, err
+		}
 		for _, c := range job.Status.Conditions {
 			switch c.Type {
+			case batchv1.JobComplete:
+				if c.Status == k8sv1.ConditionTrue {
+					return finish, jobCompleteError(job)
+				}
 			case batchv1.JobFailed:
 				if c.Status == k8sv1.ConditionTrue {
-					return true
+					return finish, jobFailedError(job)
 				}
-			case batchv1.JobComplete:
-				ExpectWithOffset(2, c.Status).NotTo(Equal(k8sv1.ConditionTrue), "Job should fail")
 			}
 		}
-		return false
-	}, timeoutSec*time.Second, 1*time.Second).Should(BeTrue(), "Job should fail")
+		return !finish, nil
+	})
+
+	if err != nil {
+		return fmt.Errorf("Job %s timeout reached, status: %+v, err: %v", job.Name, job.Status, err)
+	}
+	return nil
 }
 
 // Default Job arguments to be used with NewJob.
--- a/tests/ping.go
+++ b/tests/ping.go
@@ -27,6 +27,7 @@ import (
 
 	expect "github.com/google/goexpect"
 	"google.golang.org/grpc/codes"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/net"
 
 	v1 "kubevirt.io/client-go/api/v1"
@@ -97,4 +98,37 @@ func vmiConsoleExpectBatch(vmi *v1.VirtualMachineInstance, expected []expect.Bat
 		log.DefaultLogger().Object(vmi).Infof("%v", resp)
 	}
 	return err
+}
+
+// PingAppJob performs a netcat check (tcp by default) using a pod.
+// Optional port and arguments for the netcat command may be provided.
+// Returns the job deletion/cleanup function and an error.
+// The caller is expected to use the returned function to cleanup the resources.
+func PingAppJob(host, port string, args ...string) (func() error, error) {
+	const (
+		netcat = "nc"
+
+		jobRetry   = 1
+		jobTimeout = 40
+	)
+
+	virtClient, err := kubecli.GetKubevirtClient()
+	if err != nil {
+		return nil, err
+	}
+
+	args = append([]string{netcat}, args...)
+	args = append(args, host, port)
+
+	args = []string{strings.Join(args, " ")}
+	job := NewJob("ping-application", []string{"/bin/bash", "-c"}, args, jobRetry, JobTTL, jobTimeout)
+	job, err = virtClient.BatchV1().Jobs(NamespaceTestDefault).Create(job)
+	if err != nil {
+		return nil, err
+	}
+
+	err = WaitForJobToSucceed(job, jobTimeout*time.Second)
+	return func() error {
+		return virtClient.BatchV1().Jobs(NamespaceTestDefault).Delete(job.Name, &metav1.DeleteOptions{})
+	}, err
 }
--- a/tests/vmi_networking_test.go
+++ b/tests/vmi_networking_test.go
@@ -47,6 +47,7 @@ import (
 	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
 	"kubevirt.io/kubevirt/tests"
 	cd "kubevirt.io/kubevirt/tests/containerdisk"
+	"kubevirt.io/kubevirt/tests/flags"
 )
 
 var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:component]Networking", func() {
@@ -762,6 +763,105 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 			table.Entry("without a specific port number", []v1.Port{}, defaultCIDR),
 			table.Entry("with custom CIDR", []v1.Port{}, customCIDR),
 		)
+
+		When("performing migration", func() {
+			var vmi *v1.VirtualMachineInstance
+			var virtHandlerIP string
+			var jobCleanup func() error
+
+			ping := func(ipAddr string) error {
+				return tests.PingFromVMConsole(vmi, ipAddr, "-c 1", "-w 2")
+			}
+
+			getVirtHandlerPod := func() (*k8sv1.Pod, error) {
+				node := vmi.Status.NodeName
+				pod, err := kubecli.NewVirtHandlerClient(virtClient).Namespace(flags.KubeVirtInstallNamespace).ForNode(node).Pod()
+				if err != nil {
+					return nil, fmt.Errorf("failed to get virt-handler pod on node %s: %v", node, err)
+				}
+				return pod, nil
+			}
+
+			runMigrationAndExpectCompletion := func(migration *v1.VirtualMachineInstanceMigration, timeout int) {
+				migration, err = virtClient.VirtualMachineInstanceMigration(migration.Namespace).Create(migration)
+				Expect(err).NotTo(HaveOccurred())
+
+				Eventually(func() error {
+					migration, err := virtClient.VirtualMachineInstanceMigration(migration.Namespace).Get(migration.Name, &v13.GetOptions{})
+					if err != nil {
+						return err
+					}
+
+					Expect(migration.Status.Phase).ToNot(Equal(v1.MigrationFailed))
+
+					if migration.Status.Phase == v1.MigrationSucceeded {
+						return nil
+					}
+					return fmt.Errorf("Migration is in phase %s", migration.Status.Phase)
+
+				}, timeout, time.Second).Should(Succeed(), fmt.Sprintf("migration should succeed after %d s", timeout))
+			}
+
+			BeforeEach(func() {
+				var err error
+
+				By("Create VMI")
+				vmi = masqueradeVMI([]v1.Port{}, "")
+				vmi, err = virtClient.VirtualMachineInstance(vmi.Namespace).Create(vmi)
+				Expect(err).ToNot(HaveOccurred())
+				tests.WaitUntilVMIReady(vmi, tests.LoggedInCirrosExpecter)
+
+				vmi, err = virtClient.VirtualMachineInstance(vmi.Namespace).Get(vmi.Name, &v13.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+
+				virtHandlerPod, err := getVirtHandlerPod()
+				Expect(err).ToNot(HaveOccurred())
+				virtHandlerIP = virtHandlerPod.Status.PodIPs[0].IP
+
+				By("Check connectivity")
+				Expect(ping(virtHandlerIP)).To(Succeed())
+
+				By("Execute migration")
+				migration := tests.NewRandomMigration(vmi.Name, vmi.Namespace)
+				runMigrationAndExpectCompletion(migration, migrationWaitTime)
+
+				vmi, err = virtClient.VirtualMachineInstance(vmi.Namespace).Get(vmi.Name, &v13.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(vmi.Status.Phase).To(Equal(v1.Running))
+			})
+
+			AfterEach(func() {
+				if vmi != nil {
+					By("Delete VMI")
+					Expect(virtClient.VirtualMachineInstance(vmi.Namespace).Delete(vmi.Name, &v13.DeleteOptions{})).To(Succeed())
+
+					Eventually(func() error {
+						_, err := virtClient.VirtualMachineInstance(vmi.Namespace).Get(vmi.Name, &v13.GetOptions{})
+						return err
+					}, time.Minute, time.Second).Should(
+						SatisfyAll(HaveOccurred(), WithTransform(errors.IsNotFound, BeTrue())),
+						"The VMI should be gone within the given timeout",
+					)
+				}
+			})
+
+			AfterEach(func() {
+				if jobCleanup != nil {
+					Expect(jobCleanup()).To(Succeed())
+				}
+			})
+
+			It("preserves connectivity", func() {
+				// Workaround a live-migration limitation:
+				// Post migration, the gateway mac points to the old pod bridge and not to the new one.
+				// Initiate ingress traffic in order to refresh the VM arp table.
+				vmiPodIP := vmi.Status.Interfaces[0].IPs[0]
+				jobCleanup, err = tests.PingAppJob(vmiPodIP, "8080")
+				Eventually(func() error {
+					return ping(virtHandlerIP)
+				}, 10*time.Second).Should(Succeed(), "application ping resulted with error: %v", err)
+			})
+		})
 	})
 
 	Context("VirtualMachineInstance with TX offload disabled", func() {
--- a/tests/vmi_networking_test.go
+++ b/tests/vmi_networking_test.go
@@ -278,7 +278,7 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 
 			job, err = virtClient.BatchV1().Jobs(inboundVMI.ObjectMeta.Namespace).Create(job)
 			Expect(err).ToNot(HaveOccurred())
-			tests.WaitForJobToSucceed(&virtClient, job, 90)
+			Expect(tests.WaitForJobToSucceed(job, 90*time.Second)).To(Succeed())
 		},
 			table.Entry("[test_id:1543]on the same node from Pod", v12.NodeSelectorOpIn, false),
 			table.Entry("[test_id:1544]on a different node from Pod", v12.NodeSelectorOpNotIn, false),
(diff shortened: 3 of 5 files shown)


tests/vmi_networking_test.go
```diff
@@ -789,6 +789,91 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+
+				By("Create VMI")
+				vmi = masqueradeVMI([]v1.Port{}, "")
+				vmi, err = virtClient.VirtualMachineInstance(vmi.Namespace).Create(vmi)
+				Expect(err).ToNot(HaveOccurred())
+				tests.WaitUntilVMIReady(vmi, tests.LoggedInCirrosExpecter)
+
+				vmi, err = virtClient.VirtualMachineInstance(vmi.Namespace).Get(vmi.Name, &v13.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				if netutils.IsIPv6String(vmi.Status.Interfaces[0].IP) {
```
AlonaKaplan, line 840: you're missing not

EdDev: Right, thanks!

tests/vmi_networking_test.go
```diff
@@ -789,6 +789,91 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 			table.Entry("without a specific port number", []v1.Port{}, defaultCIDR),
 			table.Entry("with custom CIDR", []v1.Port{}, customCIDR),
 		)
+
+		When("performing migration", func() {
+			var vmi *v1.VirtualMachineInstance
+			var ping func()
+
+			ping4 := func() {
+				pingVirtualMachine(vmi, gatewayIPFromCIDR(api.DefaultVMCIDR), "\\$ ")
```
EdDev, line 798: There is a need to use both the DG and an external address (e.g. 8.8.8.8).

EdDev:
Needs more work, the ping is not working after the migration.
/hold

tests/vmi_networking_test.go
phoracek: We should not test connectivity to remote IPs. That would prevent us from running these tests on disconnected environments.

And pinging to GW does not check the connectivity outside virt-launcher Pod.

We should do the same as in the tests above and create a client/server VMIs where we check connectivity within the cluster.

EdDev: I have change it to ping the virt-handler pod.

EdDev:
Changes:
- Rebased over #3883 
- The ping is targetting the virt-handler (instead of the GW, e.g. the bridge at the virt-launcher).

Issues:
Based on the test results, post migration there is no connectivity.
This has been identified as a real problem due to the arp entry of the GW which still points to the source pod bridge mac address.

EdDev:
> Issues:
> Based on the test results, after the migration, there is no connectivity.
> This has been identified as a real problem due to the arp entry of the GW which still points to the source pod bridge mac address.

In order to overcome this limitation, traffic is now generated towards the VM post migration (by the test).
This will need to be addressed by the production code as well at some point.

tests/ping.go
```diff
@@ -0,0 +1,50 @@
...
+	err := CheckForTextExpecter(vmi, []expect.Batcher{
+		&expect.BSnd{S: "\n"},
+		&expect.BExp{R: prompt},
+		&expect.BSnd{S: cmdCheck},
+		&expect.BExp{R: prompt},
+		&expect.BSnd{S: "echo $?\n"},
+		&expect.BExp{R: "0"},
+	}, 30)
+	if err != nil {
+		return fmt.Errorf("Failed to ping VMI %s, error: %v", vmi.Name, err)
```
alonSadan, line 77: [also replied to the same issue raised here : https://github.com/kubevirt/kubevirt/pull/3479#discussion_r458291020]

I am not sure here, but don't you think the test logs will be more readable if  we only return the error here, and then format it in the `Expect`? or maybe add a more high-level message in the `Expect` formatting on top of this message?

If a test fails, the whole error object is printed including the message. I find it less convenient reading the error message from there, and always looking for the high level logs that are a part of the test logs. if I want to get into more details I will read the error message. 

as it is now, instead if a stand-by-it's-own high level message , and another more low-level message in the error object, you will have a long error message within the error object. 
another thing is that the high-level message won't show up in the tests summery, which I find also convenient to look for a quick debugging.(about the last one:  I am not sure the high level message will apear as well, but anyway it will be more readable, I think...)

I hope this is clear ...
What do you think?

tests/vmi_networking_test.go
```diff
@@ -790,6 +790,101 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+			AfterEach(func() {
+				if vmi != nil {
+					By("Delete VMI")
+					err := virtClient.VirtualMachineInstance(vmi.Namespace).Delete(vmi.Name, &v13.DeleteOptions{})
+					Expect(err).ToNot(HaveOccurred())
+
+					Eventually(func() bool {
+						_, err = virtClient.VirtualMachineInstance(vmi.Namespace).Get(vmi.Name, &v13.GetOptions{})
+						return errors.IsNotFound(err)
+					}, 60*time.Second, 1*time.Second).Should(BeTrue(), "The VMI %s failed to be removed", vmi.Name)
```
alonSadan, line 858: maybe print the error here would help with the debugging. 
I mean, it could be that the error will be `nil`, and then the test will fail and that's OK, but it could be also that the error will be some other unexpected error. so consider change to something like: 

```
ventually(func() error {
						return virtClient.VirtualMachineInstance(vmi.Namespace).Get(vmi.Name, &v13.GetOptions{})
						
					}, 60*time.Second, 1*time.Second).Should(HaveOccurred(), "The VMI %s failed to be removed", vmi.Name)
if err != nil {
Expect(IsNotFound(err).To(BeTrue(), unexpected error occurred %v, err)
}
```

tests/ping.go
```diff
@@ -89,3 +91,28 @@ func expectPrompt() *expect.BCas {
 		},
 	}}
 }
+
+// PingAppFromVirtHandler performs a natcat tcp check through the provided virt-handler pod.
+// Optional port and arguments for the netcat command may be provided.
+// In port is not provided, port "8080" is used.
```
AlonaKaplan, line 97: ד.Iמ.Iכ

tests/vmi_networking_test.go
EdDev: Done

tests/ping.go
EdDev: Done

tests/ping.go
```diff
@@ -101,3 +102,28 @@ func vmiConsoleExpectBatch(vmi *v1.VirtualMachineInstance, expected []expect.Bat
 	}
 	return err
 }
+
+// PingAppFromVirtHandler performs a natcat tcp check through the provided virt-handler pod.
```
phoracek, line 106: Using virt-handler here feels like an abuse. Using the deployed application as a utility for testing, expecting it has nc available. It may cause issues on D/S in case virt-handler is built from a different base image that does not have this binary.

Could you spawn a VM for this which would be eventually replaced by a dedicated Pod? The new container image for testing should be added to https://github.com/kubevirt/kubevirt/tree/master/cmd but it may take quite some time to get it right.

tests/vmi_networking_test.go
```diff
@@ -789,6 +790,104 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+			AfterEach(func() {
+				if vmi != nil {
+					By("Delete VMI")
+					err := virtClient.VirtualMachineInstance(vmi.Namespace).Delete(vmi.Name, &v13.DeleteOptions{})
+					Expect(err).ToNot(HaveOccurred())
+
+					Eventually(func() error {
+						_, err = virtClient.VirtualMachineInstance(vmi.Namespace).Get(vmi.Name, &v13.GetOptions{})
+						return err
+					}, 60*time.Second, 1*time.Second).Should(
```
phoracek, line 858: `time.Minute, time.Second`

tests/ping.go
EdDev: virt-handler is used by the tests for much more than this, e.g. the reporting and accessing the node.
I think depending on it in other places seems fair enough.

I see no need to spawn a VM  from such a simple task.
I'm fine to replace it in the future with a dedicated pod/job that can be used for such tasks.

> It may cause issues on D/S in case virt-handler is built from a different base image that does not have this binary.

Originally I wanted to reach the node and do that from there, but it is the node that has different binaries (in this case the kind provider did not have it).
If we do not see a problem now, I think it is a good enough compromise to use it.

tests/ping.go
```diff
@@ -101,3 +102,28 @@ func vmiConsoleExpectBatch(vmi *v1.VirtualMachineInstance, expected []expect.Bat
...
+// Optional port and arguments for the netcat command may be provided.
+// If port is not provided, port "8080" is used.
+func PingAppFromVirtHandler(pod *k8sv1.Pod, ipAddr, port string, args ...string) error {
```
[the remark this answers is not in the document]
EdDev: The "ping" specifies here that it checks connectivity.

It is named `PingApp` and the "tcp" is an implementation detail here (it could be UDP with the correct arguments).
I prefer to leave it abstracted and generic. If we find this colliding with other things, we can change it.

tests/vmi_networking_test.go
```diff
@@ -789,6 +790,104 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 			table.Entry("without a specific port number", []v1.Port{}, defaultCIDR),
 			table.Entry("with custom CIDR", []v1.Port{}, customCIDR),
 		)
+
+		When("performing migration", func() {
```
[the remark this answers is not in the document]
EdDev: It is under "VirtualMachineInstance with masquerade binding mechanism", and this closure is indeed using such a binding for the test.

tests/vmi_networking_test.go
EdDev: Done

tests/vmi_networking_test.go
```diff
@@ -789,6 +790,104 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+					Eventually(func() error {
+						_, err = virtClient.VirtualMachineInstance(vmi.Namespace).Get(vmi.Name, &v13.GetOptions{})
+						return err
+					}, 60*time.Second, 1*time.Second).Should(
+						SatisfyAll(HaveOccurred(), WithTransform(errors.IsNotFound, BeTrue())),
+						"The VMI should be gone within the given timeout",
+					)
+				}
+			})
+			It("preserves connectivity", func() {
```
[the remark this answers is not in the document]
EdDev: I agree there is no sync between what it says and what it does, but the original intention is still correct and what needs adjusting is the implementation. I will move the migration outside the test itself as we are not testing the migration itself (that should be done on a dedicated migration test) but the connectivity (that it is preserved).

tests/ping.go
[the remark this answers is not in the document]
phoracek: Ack

tests/vmi_networking_test.go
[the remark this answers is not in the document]
phoracek: It does not use the var defined for the context. I close my eyes since this context is already spoiled by the DescribeTable. Once we start refactoring of this, we have to chose Contexts wiser.

tests/ping.go
phoracek: It must not use virt-handler. nc is not our dependency as is in the image only coincidentally. I can speak for our D/S where we use a minimal base image for virt-handler and nc is not available there. Please pick a different solution so these tests can be executed on an arbitrary cluster.

EdDev:
- It is flaky.
- Use [jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/) instead of pods (now that the helpers got in).

/hold

tests/job.go
```diff
@@ -10,37 +10,88 @@ import (
...
+		jobFailedError = func(job *batchv1.Job) error {
+			return fmt.Errorf("Job %s finished with failure, status: %+v", job.Name, job.Status)
+		}
+	} else {
+		jobCompleteError = func(job *batchv1.Job) error {
+			return fmt.Errorf("Job %s finished with success, status: %+v", job.Name, job.Status)
+		}
+	}
+
+	for seconds := 0.0; seconds < timeout.Seconds(); seconds++ {
```
xpivarc, line 50: This will not work as expected. It can happen that this will run for `timeout * 10` seconds.

tests/job.go
```diff
@@ -10,37 +10,88 @@ import (
...
-	}, timeoutSec*time.Second, 1*time.Second).Should(BeNumerically(">", 0), "Job should fail")
+func waitForJob(job *batchv1.Job, toSucceed bool, timeout time.Duration) error {
+	virtClient, err := kubecli.GetKubevirtClient()
+	if err != nil {
+		return err
+	}
+
+	jobCompleteError := func(job *batchv1.Job) error { return nil }
+	jobFailedError := func(job *batchv1.Job) error { return nil }
+	if toSucceed {
```
xpivarc, line 40: Can this be simplified to
 ```
jobFailedError = func(job *batchv1.Job) error {
                        if toSucceed{
			         return fmt.Errorf("Job %s finished with failure, status: %+v", job.Name, job.Status)
                         }
                         return nil
		}
```

File: tests/vmi_networking_test.go (hunk 10 above)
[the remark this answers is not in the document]
Comment (xpivarc) [reply tests/vmi_networking_test.go:849]: I would also prefer to change the context to "It preserves connectivity before&after migration" & "when using post-copy migration" as I had a hard time to understand what is the goal of this test.
I am not sure that using `BeforeEach` is a good choice even though you don't test migration... It should be used for setup and then you should perform the test(check -> migration -> check).

WDYT?

File: tests/vmi_networking_test.go [hunk 13]
```diff
@@ -762,6 +763,100 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+				// Workaround a live-migration limitation:
+				// Post migration, the gateway mac points to the old pod bridge and not to the new one.
+				// Initiate ingress traffic in order to refresh the VM arp table.
+				vmiPodIP := vmi.Status.Interfaces[0].IPs[0]
+				err := tests.PingAppJob(vmiPodIP, "")
```
xpivarc, line 854: You are ignoring this error intentionally?

tests/job.go
EdDev: Not sure if it is a simplification, but I am fine with your proposal.
(the cost of asking twice is negligible)

tests/job.go
EdDev: Sorry, I do not understand. Why did you multiply by 10?

- For timeout=0, it will not run.
- For timeout=1, it will run once.
- For timeout=2, it will run twice.
etc...

tests/vmi_networking_test.go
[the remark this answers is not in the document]
EdDev: We need to test that connectivity is preserved/still-exists after migration. That is the test.
The migration itself if a per-requisite of the test and if it fails, it is a pre-requisite failure and not what we are focusing in testing. The Before* and After* are fixtures, which setup (and teardown) the tests, from my perspective the migration operation is a setup (i.e. given a migrated VMI we can do the test).

The context is not supposed to say what it does, only in what env it is.
"Given a masquerade binding, when migration is performed, the connectivity is preserved."
I think the sentence is clear, more details are in the test itself is someone wants to dig into details.

Obviously the context can always be improved & changed per the commonality with other tests.

tests/vmi_networking_test.go
EdDev: Yes, we do not care if it fails or not, we just want traffic to be generated towards the VMI pod so the arp table is updated (see comment above). In case the "regular" ping fails later, we do report the error for debugging.

tests/job.go
xpivarc: But that does not mean it will run for 0,1,2 seconds... As call to Kubernetes API can block for some time...

phoracek, line 50: You can use apimachinery/wait module: https://godoc.org/k8s.io/apimachinery/pkg/util/wait#PollImmediate

tests/vmi_networking_test.go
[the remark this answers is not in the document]
xpivarc: I somehow thought it test before&after...

tests/job.go
EdDev: K8S API calls are asynchronous, they may take a few msec in the worse case, but not more than that.
Using a monotonic clock for such a inaccuracy is an overkill IMO, especially for the tests. I think we can live with a few msec drift here.

tests/vmi_networking_test.go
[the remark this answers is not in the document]
EdDev: You mean the connectivity check? It is checked before the migration to assure the connectivity is there, otherwise there is no point in continuing. We could assume it is fine, but the cost of checking is not high and may save us from doing more work. Consider this check like the existence of the VMI in a running state/phase.

tests/job.go
EdDev: Done

tests/job.go
EdDev: Done

tests/ping.go
```diff
@@ -98,3 +98,38 @@ func vmiConsoleExpectBatch(vmi *v1.VirtualMachineInstance, expected []expect.Bat
...
 	return err
 }
+
+// PingAppJob performs a natcat check (tcp by default) using a pod.
+// Optional port and arguments for the netcat command may be provided.
+// If port is not provided, port "8080" is used.
+func PingAppJob(host, port string, args ...string) error {
+	const (
+		netcat      = "nc"
+		defaultPort = "8080"
```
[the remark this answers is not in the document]
EdDev: The reason for the default is to express "I do not mind what port is used".
Which is the exact reason this came to be in the first place and a valid scenario IMO.

If we require the user to specify it even though he does not care about it, it may raise the obvious question: Why do you use this port and not something else? Something that is avoidable if we use "nothing".

tests/vmi_networking_test.go
```diff
@@ -762,6 +763,100 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+				tests.WaitUntilVMIReady(vmi, tests.LoggedInCirrosExpecter)
+
+				vmi, err = virtClient.VirtualMachineInstance(vmi.Namespace).Get(vmi.Name, &v13.GetOptions{})
+				Expect(err).ToNot(HaveOccurred())
+
+				virtHandlerPod, err := getVirtHandlerPod()
+				Expect(err).ToNot(HaveOccurred())
+				virtHandlerIP = virtHandlerPod.Status.PodIPs[0].IP
+
+				By("Check connectivity")
```
phoracek, line 822: We should not "Check" in before each. If you have this to just wait for the VM to become reachable, word it differently "Waiting for the VM to become reachable"

phoracek (changes requested):

tests/vmi_networking_test.go
```diff
@@ -762,6 +763,100 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+				node := vmi.Status.NodeName
+				pod, err := kubecli.NewVirtHandlerClient(virtClient).Namespace(flags.KubeVirtInstallNamespace).ForNode(node).Pod()
+				if err != nil {
+					return nil, fmt.Errorf("failed to get virt-handler pod on node %s: %v", node, err)
+				}
+				return pod, nil
+			}
+
+			runMigrationAndExpectCompletion := func(migration *v1.VirtualMachineInstanceMigration, timeout int) {
+				Eventually(func() error {
```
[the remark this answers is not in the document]
EdDev: This is the way it is currently used in utils, changing it is ignoring some pain points from the pasts which I cannot confirm if they are gone or not.
I do not like it as well, but unless we know all the details on why it was added, I prefer to keep the same logic other tests are taking.

tests/ping.go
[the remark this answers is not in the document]
phoracek: But this is directly connected with the port that is being exposed by a different function, starting the server. You have to care about the port unless there is a different mechanism making sure that we by default expose the port that is expected by this Ping function.

tests/vmi_networking_test.go
```diff
@@ -762,6 +763,100 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+					}
+
+					Expect(migration.Status.Phase).ToNot(Equal(v1.MigrationFailed))
+
+					if migration.Status.Phase == v1.MigrationSucceeded {
+						return nil
+					}
+					return fmt.Errorf("Migration is in phase %s", migration.Status.Phase)
+
+				}, timeout, time.Second).Should(Succeed(), fmt.Sprintf("migration should succeed after %d s", timeout))
```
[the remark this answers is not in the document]
EdDev: Not sure why you commented it here as this value is received as an argument.
`migrationWaitTime` is the value used for the migration tests, makes sense to use the same one in case it changes in the future. It is not this module global.

tests/vmi_networking_test.go
EdDev: This is a nit, playing with sentences for logging.
We are not waiting for connectivity, we just make sure it exists before we proceed on and spending more time on this test. The check itself is not a must, but an optimization and a debugging helper (helps confirming that the vm originally had connectivity in case after the migration it does not).

I see no problem with what is described here.

tests/vmi_networking_test.go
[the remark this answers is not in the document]
phoracek: That is a dangerous precedent for re factoring. If you can live with it, we can leave it here, up to your consideration.

Eventually more often hide bugs than helps with synchronisation of tests.

tests/vmi_networking_test.go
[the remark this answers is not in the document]
phoracek: Oh, the closure is so large I missed the fact that it is a param. All good then.

tests/vmi_networking_test.go
phoracek: Correct logging messages describing the intention would helps us with debugging.

tests/ping.go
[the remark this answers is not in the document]
EdDev: > But this is directly connected with the port that is being exposed by a different function, starting the server.

Yes, that is the common use case but not the only one.
If we want to express that we do not care, we use "". The usage in this test uses that as it does not care (because there is no server on the other side).

> You have to care about the port unless there is a different mechanism making sure that we by default expose the port that is expected by this Ping function.

All correct, but I fail to understand why this is bothering you that we support an empty value.

Again, if we explicitly use a port, one may ask why we used a specific one... and in the case we have here, there is no reason at all, it is because we do not care.

If I still did not managed to convince you, I'll change it for team peace.

tests/vmi_networking_test.go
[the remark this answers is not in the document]
EdDev: I'm removing it although it changes the current norm at the migration tests.
It is indeed improper to retry on something we expect to be perfectly stable.

tests/ping.go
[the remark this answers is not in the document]
phoracek: While reading the code, I would ask what service are we pinging and why are we expecting it to succeed. Seeing `testServicePort` passed to this function would help me finding that the same port was previously used in `BeforeEach`. Unless a single module has functions bound together where one sets up the service and another pings it, I'd like to see the port passed and connecting the dependencies.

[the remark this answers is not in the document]
EdDev: Done

tests/ping.go
```diff
@@ -98,3 +98,38 @@ func vmiConsoleExpectBatch(vmi *v1.VirtualMachineInstance, expected []expect.Bat
 	}
 	return err
 }
+
+// PingAppJob performs a natcat check (tcp by default) using a pod.
```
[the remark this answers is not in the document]
EdDev: Done.

phoracek (approved):
/lgtm

tests/ping.go
```diff
@@ -98,3 +98,38 @@ func vmiConsoleExpectBatch(vmi *v1.VirtualMachineInstance, expected []expect.Bat
...
+// PingAppJob performs a netcat check (tcp by default) using a pod.
+// Optional port and arguments for the netcat command may be provided.
+// If port is not provided, port "8080" is used.
+func PingAppJob(host, port string, args ...string) error {
+	const (
+		netcat      = "nc"
+		defaultPort = "8080"
+
+		jobRetry   = 1
+		jobTTL     = 10
```
xpivarc, line 111: Any reason to use only 10? You don't assert here so the caller of this function can assert after... I am just worrying to lose some valuable debug info...

EdDev: 10 seconds is a long of time and our usage of jobs in its current state is anyway problematic because a large number of them will be left behind, uncleaned. (we need to take care of that and add it to the namespace cleanup function and probably provide a delete API for jobs).

This helper will create the the job, then wait for it to finish and return. At the same time (more or less), the 10sec timer will start so the caller has now ~10sec to check something. If he decides to assert immediately based on the returned result, there is no problem.

If he will check something else and based on that result assert, then we may get into a problem of not collecting the job logs as you mentioned. In such a case, no number we will use here will really answer the issue because it depends on how long that other check will take.
One generic solution I can think of is to require a timeout value to be passed in to this helper, but I feel this will not make sense to the caller and will complicate the helper signature.

What I think we can do (in a follow up PR hopefully) is to provide a helper that collects the logs from job pods, letting the caller explicitly fetch the information in the scenario mentioned above.

In the meantime, I will increase this to 15 (as the caller is indeed checking another operation for 10sec max).
I will also document this on the doctext.

xpivarc: I just realized that this feature is still feature-gated. (Sorry, I am using it in my project...) I assume we don't have this enabled so indeed we need to have our own clean-ups so we don't dirty other tests. (follow up PR is good)

EdDev:
Wait for the dependent PR to get in.
/hold

tests/ping.go
EdDev: If this is the case (that it is feature gated), we have a problem. The TTL cannot be an API of the job abstraction nor for this ping one. An alternative is to launch a goroutine to do this instead, but I do not know if it is worth the trouble as other objects do not do the same at the moment.

phoracek:
/lgtm cancel

https://github.com/kubevirt/kubevirt/pull/3871#discussion_r475453121

tests/ping.go
xpivarc: I think it would be best to add the clean-up of jobs after each test.

EdDev:
change: Just a rebase.

EdDev:
change: Do not specify Job TTL (only use the default) in the ping-app and make sure the job is cleaned by the test.

tests/ping.go
```diff
@@ -98,3 +99,36 @@ func vmiConsoleExpectBatch(vmi *v1.VirtualMachineInstance, expected []expect.Bat
 	}
 	return err
 }
+
+// PingAppJob performs a netcat check (tcp by default) using a pod.
+// Optional port and arguments for the netcat command may be provided.
+// Returns the job deletion/cleanup function and an error.
+// The caller is expected to use the returned function to cleanup the resources.
+func PingAppJob(host, port string, args ...string) (func() error, error) {
```
maiqueb, line 107: I think having 'ping' in the name is misleading - I immediately think ICMP ping. This is a simple TCP client afaiu. Why not name it for what it is ?

EdDev: The ping verb [definition](https://www.dictionary.com/browse/ping) and my understanding of it talks about a connectivity check and adding "App" or "Application" should make it clear what  it does.. especially when we add a doctext and port to the arguments.

I managed to convinced this another reviewer which had the same comment as you do. Hope you'll get convinced as well.
If you do have another catchy name without starting to enter implementation details please suggest it. This was the closest I could find/think.

PR #3875, opened 2020-07-22
Add ProductName and Version labels to KubeVirt objects

stu-gott:
This PR sets the groundwork for applying commonly recognizes labels to all objects created within KubeVirt.

ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

**Release note**:
```release-note
Resources created by KubeVirt are now labelled more clearly in terms of relationship and role.
```

### DIFF
--- a/pkg/virt-operator/install-strategy/create.go
+++ b/pkg/virt-operator/install-strategy/create.go
@@ -2370,10 +2370,10 @@ func SyncAll(queue workqueue.RateLimitingInterface, kv *v1.KubeVirt, prevStrateg
 
 	// Avoid log spam by logging this issue once early instead of for once each object created
 	if !isValidLabel(kv.Spec.ProductVersion) {
-		log.Log.Errorf("invalid kubevirt.spec.productVersion: labels must be 63 characters or less, begin and end with alphanumeric characters, and contain only dot, hyphen or dash")
+		log.Log.Errorf("invalid kubevirt.spec.productVersion: labels must be 63 characters or less, begin and end with alphanumeric characters, and contain only dot, hyphen or underscore")
 	}
 	if !isValidLabel(kv.Spec.ProductName) {
-		log.Log.Errorf("invalid kubevirt.spec.productName: labels must be 63 characters or less, begin and end with alphanumeric characters, and contain only dot, hyphen or dash")
+		log.Log.Errorf("invalid kubevirt.spec.productName: labels must be 63 characters or less, begin and end with alphanumeric characters, and contain only dot, hyphen or underscore")
 	}
 
 	targetVersion := kv.Status.TargetKubeVirtVersion
--- a/pkg/virt-operator/install-strategy/create.go
+++ b/pkg/virt-operator/install-strategy/create.go
@@ -26,6 +26,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strings"
 	"time"
 
@@ -110,10 +111,26 @@ func controllerDeployments(strategy *InstallStrategy) []*appsv1.Deployment {
 	return deployments
 }
 
+func isValidLabel(label string) bool {
+	// First and last character must be alphanumeric
+	// middle chars can be alphanumeric, or dot hyphen or dash
+	// entire string must not exceed 63 chars
+	r := regexp.MustCompile(`^([a-z0-9A-Z]([a-z0-9A-Z\-\_\.]{0,61}[a-z0-9A-Z])?)?$`)
+	return r.Match([]byte(label))
+}
+
 func injectOperatorMetadata(kv *v1.KubeVirt, objectMeta *metav1.ObjectMeta, version string, imageRegistry string, id string) {
 	if objectMeta.Labels == nil {
 		objectMeta.Labels = make(map[string]string)
 	}
+	if kv.Spec.ProductVersion != "" && isValidLabel(kv.Spec.ProductVersion) {
+		objectMeta.Labels[v1.AppVersionLabel] = kv.Spec.ProductVersion
+	}
+	if kv.Spec.ProductName != "" && isValidLabel(kv.Spec.ProductName) {
+		objectMeta.Labels[v1.AppPartOfLabel] = kv.Spec.ProductName
+	}
+	objectMeta.Labels[v1.AppComponentLabel] = v1.AppComponent
+
 	objectMeta.Labels[v1.ManagedByLabel] = v1.ManagedByLabelOperatorValue
 
 	if objectMeta.Annotations == nil {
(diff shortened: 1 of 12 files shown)


stu-gott:
/hold

This PR still needs functional tests, but I wanted to get the code committed for initial review.

staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -467,8 +467,24 @@ const (
...
+	AppNameLabel = AppLabelPrefix + "/name"
+	// This label is commonly used by 3rd party management tools to identify
+	// an application's version.
+	AppVersionLabel = AppLabelPrefix + "/version"
+	// This label is commonly used by 3rd party management tools to identify
+	// a higher level application.
+	AppPartOfLabel = AppLabelPrefix + "/part-of"
+	// This label is commonly used by 3rd party management tools to identify
+	// the component this application is a part of.
+	AppComponentLabel = AppLabelPrefix + "/compnent"
```
ashleyschuett, line 483: "/component"

staging/src/kubevirt.io/client-go/api/v1/types.go
```diff
@@ -467,8 +467,24 @@ const (
...
+	AppLabelPrefix = "app.kubernetes.io"
+	// This label is commonly used by 3rd party management tools to identify
+	// an application's name.
+	AppNameLabel = AppLabelPrefix + "/name"
+	// This label is commonly used by 3rd party management tools to identify
+	// an application's version.
+	AppVersionLabel = AppLabelPrefix + "/version"
+	// This label is commonly used by 3rd party management tools to identify
+	// a higher level application.
+	AppPartOfLabel = AppLabelPrefix + "/part-of"
```
ashleyschuett, line 480: should this be `AppProductLabel = AppLabelPrefix + "/product"` I find it strange the terminology is different for this label than for rest of the code.

pkg/virt-operator/install-strategy/strategy.go
```diff
@@ -273,28 +273,42 @@ func GenerateCurrentInstallStrategy(config *operatorutil.KubeVirtDeploymentConfi
...
 	strategy.validatingWebhookConfigurations = append(strategy.validatingWebhookConfigurations, components.NewVirtAPIValidatingWebhookConfiguration(config.GetNamespace()))
 	strategy.mutatingWebhookConfigurations = append(strategy.mutatingWebhookConfigurations, components.NewVirtAPIMutatingWebhookConfiguration(config.GetNamespace()))
 
 	strategy.services = append(strategy.services, components.NewPrometheusService(config.GetNamespace()))
 	strategy.services = append(strategy.services, components.NewApiServerService(config.GetNamespace()))
 	strategy.services = append(strategy.services, components.NewOperatorWebhookService(operatorNamespace))
-	apiDeployment, err := components.NewApiServerDeployment(config.GetNamespace(), config.GetImageRegistry(), config.GetImagePrefix(), config.GetApiVersion(), config.GetImagePullPolicy(), config.GetVerbosity(), config.GetExtraEnv())
+	apiDeployment, err := components.NewApiServerDeployment(config.GetNamespace(), config.GetImageRegistry(), config.GetImagePrefix(), config.GetApiVersion(), productName, productVersion, config.GetImagePullPolicy(), config.GetVerbosity(), config.GetExtraEnv())
```
ashleyschuett, line 297: These method signatures are very long, is there a reason to not pass the whole config object?

pkg/virt-operator/install-strategy/strategy.go
```diff
@@ -273,28 +273,42 @@ func GenerateCurrentInstallStrategy(config *operatorutil.KubeVirtDeploymentConfi
 		}
 	}
 
+	var productName string
+	var productVersion string
+
+	if isValidLabel(config.AdditionalProperties[operatorutil.ProductNameKey]) {
```
ashleyschuett, line 279: It doesn't matters since `AdditionalProperties` and all the other fields are public/exported for config, but it seems the convention that is being used is to add helper methods to access each of it's fields. Maybe we should do the same for `GetProductName()` and `GetProductVersion()`. It does have the additional benefit of making them more discoverable since they are hidden under the additionalproperties map

staging/src/kubevirt.io/client-go/api/v1/types.go
stu-gott: Thank you!

staging/src/kubevirt.io/client-go/api/v1/types.go
stu-gott: The commonly used label is "app.kubernetes.io/part-of"

I guess it's all a matter of point of view. In this case our components are all from the kubevirt project, but pretend we shipped a postfix database for some reason. The question a user or admin would be asking when looking at a pod named "postfix-XXXXX" is "what is this? where did it come from?" -- so from that point of view the "part of" label makes sense.

As far as the decision to use the name "productName" on the kubevirt resource is the person making the decision at that level is thinking in terms of the bundle of applications they're trying to package together. "this is the ____ product offering".

So it just seemed to make sense. naming it the "AppPartOfLabel" didn't seem to convey as much intuitive meaning. names are hard. What do you think?

pkg/virt-operator/install-strategy/strategy.go
stu-gott: I'll double check, but I don't see any harm in using the entire object.

pkg/virt-operator/install-strategy/strategy.go
stu-gott: +1 thanks for the insight!

staging/src/kubevirt.io/client-go/api/v1/types.go
ashleyschuett: I didn't realize that the "part-of" label was a standard/community thing. In that case this makes sense and the naming does as well

pkg/virt-operator/install-strategy/strategy.go
stu-gott: So it turns out the unit tests here don't have a config object. They basically test the parameters directly. In the interests of not boiling the ocean I think I'm going to defer this.

pkg/virt-operator/install-strategy/create.go
```diff
@@ -110,10 +111,26 @@ func controllerDeployments(strategy *InstallStrategy) []*appsv1.Deployment {
 	return deployments
 }
 
+func isValidLabel(label string) bool {
+	// First and last character must be alphanumeric
+	// middle chars can be alphanumeric, or dot hyphen or dash
+	// entire string must not exceed 63 chars
+	r := regexp.MustCompile(`^([a-z0-9A-Z]([a-z0-9A-Z\-\_\.]{0,61}[a-z0-9A-Z])?)?$`)
```
jean-edouard, line 118: Nit: could be shortened with: `^([[:alnum:]]([\w\-\.]{0,61}[[:alnum:]])?)?$` (though `len("[a-z0-9A-Z]") == len("[[:alnum:]]")` and `\w` might be a bit obfuscating...)

pkg/virt-operator/install-strategy/create.go
```diff
@@ -2351,6 +2368,14 @@ func SyncAll(queue workqueue.RateLimitingInterface, kv *v1.KubeVirt, prevStrateg
 		GracePeriodSeconds: &gracePeriod,
 	}
 
+	// Avoid log spam by logging this issue once early instead of for once each object created
+	if !isValidLabel(kv.Spec.ProductVersion) {
+		log.Log.Errorf("invalid kubevirt.spec.productVersion: labels must be 63 characters or less, begin and end with alphanumeric characters, and contain only dot, hyphen or dash")
```
jean-edouard, line 2373: Nit: underscore is not a dash :)

stu-gott: ... oops. thanks!

PR #3693, opened 2020-06-28
Flaky functest - [test_id:1779] fixing race in the Expecter

AlonaKaplan:
**What this PR does / why we need it**:
Having multiple `expect.BExp` one after the other without an `expect.BSnd`
between them is raceful.

`expect.BExp` is waiting for the desired expression in the current buffer
and flushes the buffer once the expression is found.

For example -
```
&expect.BSnd{S: "a\nb"},
&expect.BExp{R: "a"},
&expect.BExp{R: "b"}
```

In the good scenario we will get -
```
Sent: "a\nb"
Match for RE: "a" found: ["a"] Buffer:
$a
Match for RE: "b" found: ["b"] Buffer:
b
```

In the bad scenario we will get -
```
Sent: "a\nb"
Match for RE: "a" found: ["a"] Buffer:
$a
b
```

The second match (looking for 'b') will never be met since 'b' was already
in the buffer when the first match was found, therefore it was flushed by
the first `expect.BExp`.

To avoid this we need to make sure the `expect.BExp` is only expecting text
that was added to the buffer after the previous expectation was met.

Possible race in [test_id:1779] that is fixed by this PR -

```
Sent: "cat /etc/resolv.conf\n"
Match for RE: "search example.com" found: ["search example.com"] Buffer: 
$ cat /etc/resolv.conf
search example.com
Match for RE: "\\$" found: ["$"] Buffer: 
nameserver 8.8.8.8
nameserver 4.2.2.1
$
Sent: "cat /etc/resolv.conf\n"
Match for RE: "nameserver 8.8.8.8" found: ["nameserver 8.8.8.8"] Buffer:  cat /etc/resolv.conf
search example.com
nameserver 8.8.8.8
nameserver 4.2.2.1
$ 
```

Since the '$' is part of the last buffer, the next `Exp` expecting for it won't ever be met.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3602

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

Issue #3619, opened 2020-06-21
[Flaky test][test_id:1780]should allow regular network connection with custom CIDR


oshoval:
/kind bug
/label build-officer

**What happened**:
Test 1780 has a flake
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3604/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1273898340039266304

Flake finder report
https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2020-06-19-024h.html

https://storage.googleapis.com/kubevirt-prow/reports/flakefinder/kubevirt/kubevirt/flakefinder-2020-06-20-024h.html

Lane pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6 

**What you expected to happen**:
Test should pass

oshoval:
Should be label cop
my PR #3543 should fix it

```
{"component":"tests","kind":"VirtualMachineInstance","level":"info","msg":"Configure ipv6: [{1 \r\n$ [$ ]} {3 \r\n$ [$ ]} {5 sudo ip -6 addr add fd10 [0]} {7 :0:2::2/120 dev eth0\r\n$ [$ ]} {9 echo $?\r\n0\r\n$ [$ ]} {11 sleep 5\r\n$ sudo ip -6 route add default via fd10 [0]} {13 :0:2::1 src fd10:0:2::2\r\n$ echo $?\r\n0\r\n$ echo \"nameserver fd00:10:96::a [$ ]} {15 \" \u003e\u003e /etc/resolv.conf\\necho $?\r\n-sh: can't create /etc/resolv.confnecho: Permission denied\r\n$ []}]","name":"testvmiqtn9xtpbq4stmd5w6z6gkv49l2s9pffkqm2ztbl97c4mv79j","namespace":"kubevirt-test-default","pos":"utils.go:2923","timestamp":"2020-06-19T12:06:35.264459Z","uid":"dc662e97-f6d4-415b-8a7d-61ffd1a0c15e"}
```

oshoval:
I saw several failures of test 1537 on other PRs / flake finder,
that the PR which fix this issue would fix as well

for example this is from flake finder
[...]

Issue #3602, opened 2020-06-18
[flakey test] Networking VirtualMachineInstance with custom dns [test_id:1779] should have custom resolv.conf


vladikr:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
```
[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:component]Networking VirtualMachineInstance with custom dns [test_id:1779]should have custom resolv.conf 

/root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:666
 Unexpected error:
     <expect.TimeoutError>: 15000000000
     expect: timer expired after 15 seconds
 occurred
 /root/go/src/kubevirt.io/kubevirt/tests/vmi_networking_test.go:691
```

**What you expected to happen**:
The k8s-cnao-1.17 lane often fails with the above error

AlonaKaplan:
links to failing jobs - 
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3550/pull-kubevirt-e2e-kind-k8s-1.17.0-ipv6/1274404687931510784
https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3406/pull-kubevirt-e2e-k8s-1.18/1274979489599721472

oshoval:
```
				&expect.BExp{R: "nameserver 4.2.2.1"},
				&expect.BExp{R: "\\$"},
```
i would suggest to add sending \n between those two, because i see in the logs the first one "steals" the prompt when the race occurs.

PR #3882, opened 2020-07-23
Safe expect batch

AlonaKaplan:
Wrong usage of the `expecter` may cause the tests to be flaky.
    
`ExpectBatchWithValidatedSend` adds the `expect.BSnd` command to the `expect.BExp`
 expression. It is done to make sure the found match is in the result of
 the `expect.BSnd` command and not in a leftover that wasn't removed from
 the buffer.
    
 Issue https://github.com/kubevirt/kubevirt/issues/3619 describes such
 scenario.
    
 `ExpectBatchWithValidatedSend` also prevents having two sequential `expect.BExp` which
 cause unpredictable behaviour as happened in -
 https://github.com/kubevirt/kubevirt/issues/3602 and explained in PR
 https://github.com/kubevirt/kubevirt/pull/3693.
 
 For simplicity, `ExpectBatchWithValidatedSend` also prevents two sequential `expect.BSnd`.

The PR is adding the `ExpectBatchWithValidatedSend` and to example it, uses it in some selected places. A follow-up PR is needed to go over all the uses of `expecter.ExpectBatch` and in case it makes sense/supported use the `ExpectBatchWithValidatedSend` instead.

The PR also introduces a `RetValue` method ("\n" + expectedValue + "\r\n" + ".*" + prompt)  which can be used on `BExp`s. Same as for `ExpectBatchWithValidatedSend`, a follow-up PR is needed to actually use this method where needed. The current PR doesn't aim to cover all the places.

Fixes #

For some reason `\n` appears in the expecter buffer as `\r\n`.
So when trying to find in the buffer a previously sent command that ended
with `\n` it fails.

To workaround this issue, the `\n` suffix is trimmed from the expected send command. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/config_test.go
+++ b/tests/config_test.go
@@ -105,7 +105,7 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 					// mount iso ConfigMap image
 					&expect.BSnd{S: "mount /dev/sda /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: tests.Retcode("0", "\\#")},
+					&expect.BExp{R: tests.RetValue("0", "\\#")},
 					&expect.BSnd{S: "cat /mnt/option1 /mnt/option2 /mnt/option3\n"},
 					&expect.BExp{R: expectedOutput},
 				}, 200*time.Second)
@@ -199,7 +199,7 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 					// mount iso Secret image
 					&expect.BSnd{S: "mount /dev/sda /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: tests.Retcode("0", "\\#")},
+					&expect.BExp{R: tests.RetValue("0", "\\#")},
 					&expect.BSnd{S: "cat /mnt/user /mnt/password\n"},
 					&expect.BExp{R: expectedOutput},
 				}, 200*time.Second)
@@ -282,7 +282,7 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 				// mount service account iso image
 				&expect.BSnd{S: "mount /dev/sda /mnt\n"},
 				&expect.BSnd{S: "echo $?\n"},
-				&expect.BExp{R: tests.Retcode("0", "\\#")},
+				&expect.BExp{R: tests.RetValue("0", "\\#")},
 				&expect.BSnd{S: "cat /mnt/namespace\n"},
 				&expect.BExp{R: tests.NamespaceTestDefault},
 				&expect.BSnd{S: "tail -c 20 /mnt/token\n"},
@@ -377,7 +377,7 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 					&expect.BExp{R: "#"},
 					&expect.BSnd{S: "mount /dev/vdc /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: tests.Retcode("0", "\\#")},
+					&expect.BExp{R: tests.RetValue("0", "\\#")},
 					&expect.BSnd{S: "cat /mnt/config1 /mnt/config2 /mnt/config3\n"},
 					&expect.BExp{R: expectedOutputCfgMap},
 				}, 200*time.Second)
@@ -403,7 +403,7 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 					// mount Secret image
 					&expect.BSnd{S: "mount /dev/vdd /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: tests.Retcode("0", "\\#")},
+					&expect.BExp{R: tests.RetValue("0", "\\#")},
 					&expect.BSnd{S: "cat /mnt/user /mnt/password\n"},
 					&expect.BExp{R: expectedOutputSecret},
 				}, 200*time.Second)
@@ -501,13 +501,13 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 					&expect.BExp{R: "\\#"},
 					&expect.BSnd{S: "mount /dev/sda /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: tests.Retcode("0", "\\#")},
+					&expect.BExp{R: tests.RetValue("0", "\\#")},
 					&expect.BSnd{S: "grep \"PRIVATE KEY\" /mnt/ssh-privatekey\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: tests.Retcode("0", "\\#")},
+					&expect.BExp{R: tests.RetValue("0", "\\#")},
 					&expect.BSnd{S: "grep ssh-rsa /mnt/ssh-publickey\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: tests.Retcode("0", "\\#")},
+					&expect.BExp{R: tests.RetValue("0", "\\#")},
 				}, 200*time.Second)
 				log.DefaultLogger().Object(vmi).Infof("%v", res)
 				Expect(err).ToNot(HaveOccurred())
--- a/tests/config_test.go
+++ b/tests/config_test.go
@@ -105,7 +105,7 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 					// mount iso ConfigMap image
 					&expect.BSnd{S: "mount /dev/sda /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: "0"},
+					&expect.BExp{R: tests.Retcode("0", "\\#")},
 					&expect.BSnd{S: "cat /mnt/option1 /mnt/option2 /mnt/option3\n"},
 					&expect.BExp{R: expectedOutput},
 				}, 200*time.Second)
@@ -199,7 +199,7 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 					// mount iso Secret image
 					&expect.BSnd{S: "mount /dev/sda /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: "0"},
+					&expect.BExp{R: tests.Retcode("0", "\\#")},
 					&expect.BSnd{S: "cat /mnt/user /mnt/password\n"},
 					&expect.BExp{R: expectedOutput},
 				}, 200*time.Second)
@@ -282,7 +282,7 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 				// mount service account iso image
 				&expect.BSnd{S: "mount /dev/sda /mnt\n"},
 				&expect.BSnd{S: "echo $?\n"},
-				&expect.BExp{R: "0"},
+				&expect.BExp{R: tests.Retcode("0", "\\#")},
 				&expect.BSnd{S: "cat /mnt/namespace\n"},
 				&expect.BExp{R: tests.NamespaceTestDefault},
 				&expect.BSnd{S: "tail -c 20 /mnt/token\n"},
@@ -377,7 +377,7 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 					&expect.BExp{R: "#"},
 					&expect.BSnd{S: "mount /dev/vdc /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: "0"},
+					&expect.BExp{R: tests.Retcode("0", "\\#")},
 					&expect.BSnd{S: "cat /mnt/config1 /mnt/config2 /mnt/config3\n"},
 					&expect.BExp{R: expectedOutputCfgMap},
 				}, 200*time.Second)
@@ -403,7 +403,7 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 					// mount Secret image
 					&expect.BSnd{S: "mount /dev/vdd /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: "0"},
+					&expect.BExp{R: tests.Retcode("0", "\\#")},
 					&expect.BSnd{S: "cat /mnt/user /mnt/password\n"},
 					&expect.BExp{R: expectedOutputSecret},
 				}, 200*time.Second)
@@ -498,16 +498,16 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 				res, err := expecter.ExpectBatch([]expect.Batcher{
 					// mount iso Secret image
 					&expect.BSnd{S: "sudo su -\n"},
-					&expect.BExp{R: "#"},
+					&expect.BExp{R: "\\#"},
 					&expect.BSnd{S: "mount /dev/sda /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: "0"},
+					&expect.BExp{R: tests.Retcode("0", "\\#")},
 					&expect.BSnd{S: "grep \"PRIVATE KEY\" /mnt/ssh-privatekey\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: "0"},
+					&expect.BExp{R: tests.Retcode("0", "\\#")},
 					&expect.BSnd{S: "grep ssh-rsa /mnt/ssh-publickey\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: "0"},
+					&expect.BExp{R: tests.Retcode("0", "\\#")},
 				}, 200*time.Second)
 				log.DefaultLogger().Object(vmi).Infof("%v", res)
 				Expect(err).ToNot(HaveOccurred())
--- a/tests/pausing_test.go
+++ b/tests/pausing_test.go
@@ -367,7 +367,7 @@ var _ = Describe("[rfe_id:3064][crit:medium][vendor:cnv-qe@redhat.com][level:com
 		grepSleepPid := func(expecter expect.Expecter) string {
 			res, err := tests.ExpectBatchWithValidatedSend(expecter, []expect.Batcher{
 				&expect.BSnd{S: `pgrep -f "sleep 5"` + "\n"},
-				&expect.BExp{R: tests.Retcode("[0-9]+", "\\# ")}, // pid
+				&expect.BExp{R: tests.RetValue("[0-9]+", "\\# ")}, // pid
 			}, 15*time.Second)
 			log.DefaultLogger().Infof("a:%+v\n", res)
 			Expect(err).ToNot(HaveOccurred())
--- a/tests/pausing_test.go
+++ b/tests/pausing_test.go
@@ -20,6 +20,7 @@
 package tests_test
 
 import (
+	"regexp"
 	"strings"
 	"time"
 
@@ -363,30 +364,32 @@ var _ = Describe("[rfe_id:3064][crit:medium][vendor:cnv-qe@redhat.com][level:com
 
 	Context("A long running process", func() {
 
-		startProcess := func(expecter expect.Expecter) string {
-			By("Start a long running process")
-			res, err := expecter.ExpectBatch([]expect.Batcher{
-				&expect.BSnd{S: "sleep 5&\n"},
-				&expect.BExp{R: "\\# "}, // prompt
+		grepSleepPid := func(expecter expect.Expecter) string {
+			res, err := tests.ExpectBatchWithValidatedSend(expecter, []expect.Batcher{
 				&expect.BSnd{S: `pgrep -f "sleep 5"` + "\n"},
-				&expect.BExp{R: "sleep 5"},    // command
-				&expect.BExp{R: "[0-9]+\r\n"}, // pid
+				&expect.BExp{R: tests.Retcode("[0-9]+", "\\# ")}, // pid
 			}, 15*time.Second)
 			log.DefaultLogger().Infof("a:%+v\n", res)
 			Expect(err).ToNot(HaveOccurred())
-			return strings.TrimSpace(res[2].Match[0])
+			re := regexp.MustCompile("\r\n[0-9]+\r\n")
+			return strings.TrimSpace(re.FindString(res[0].Match[0]))
+		}
+
+		startProcess := func(expecter expect.Expecter) string {
+			By("Start a long running process")
+			res, err := tests.ExpectBatchWithValidatedSend(expecter, []expect.Batcher{
+				&expect.BSnd{S: "sleep 5&\n"},
+				&expect.BExp{R: "\\# "}, // prompt
+			}, 15*time.Second)
+			log.DefaultLogger().Infof("a:%+v\n", res)
+			Expect(err).ToNot(HaveOccurred())
+
+			return grepSleepPid(expecter)
 		}
 
 		checkProcess := func(expecter expect.Expecter) string {
 			By("Checking the long running process")
-			res, err := expecter.ExpectBatch([]expect.Batcher{
-				&expect.BSnd{S: `pgrep -f "sleep 5"` + "\n"},
-				&expect.BExp{R: "sleep 5"},    // command
-				&expect.BExp{R: "[0-9]+\r\n"}, // pid
-			}, 15*time.Second)
-			log.DefaultLogger().Infof("a:%+v\n", res)
-			Expect(err).ToNot(HaveOccurred())
-			return strings.TrimSpace(res[1].Match[0])
+			return grepSleepPid(expecter)
 		}
 
 		It("[test_id:3090]should be continued after the VMI is unpaused", func() {
--- a/tests/storage_test.go
+++ b/tests/storage_test.go
@@ -201,7 +201,7 @@ var _ = Describe("Storage", func() {
 					&expect.BSnd{S: "sudo mkfs.ext4 /dev/vdc\n"},
 					&expect.BExp{R: "\\$ "},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: tests.Retcode("0", "\\$ ")},
+					&expect.BExp{R: tests.RetValue("0", "\\$ ")},
 				}, 20*time.Second)
 				log.DefaultLogger().Object(vmi).Infof("%v", res)
 				Expect(err).ToNot(HaveOccurred())
@@ -302,7 +302,7 @@ var _ = Describe("Storage", func() {
 					// Because "/" is mounted on tmpfs, we need something that normally persists writes - /dev/sda2 is the EFI partition formatted as vFAT.
 					&expect.BSnd{S: "mount /dev/sda2 /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: tests.Retcode("0", "\\#")},
+					&expect.BExp{R: tests.RetValue("0", "\\#")},
 					&expect.BSnd{S: "echo content > /mnt/checkpoint\n"},
 					// The QEMU process will be killed, therefore the write must be flushed to the disk.
 					&expect.BSnd{S: "sync\n"},
@@ -330,10 +330,10 @@ var _ = Describe("Storage", func() {
 					// Same story as when first starting the VirtualMachineInstance - the checkpoint, if persisted, is located at /dev/sda2.
 					&expect.BSnd{S: "mount /dev/sda2 /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: tests.Retcode("0", "\\#")},
+					&expect.BExp{R: tests.RetValue("0", "\\#")},
 					&expect.BSnd{S: "cat /mnt/checkpoint &> /dev/null\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: tests.Retcode("1", "\\#")},
+					&expect.BExp{R: tests.RetValue("1", "\\#")},
 				}, 200*time.Second)
 				Expect(err).ToNot(HaveOccurred())
 			})
--- a/tests/storage_test.go
+++ b/tests/storage_test.go
@@ -201,7 +201,7 @@ var _ = Describe("Storage", func() {
 					&expect.BSnd{S: "sudo mkfs.ext4 /dev/vdc\n"},
 					&expect.BExp{R: "\\$ "},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: "0"},
+					&expect.BExp{R: tests.Retcode("0", "\\$ ")},
 				}, 20*time.Second)
 				log.DefaultLogger().Object(vmi).Infof("%v", res)
 				Expect(err).ToNot(HaveOccurred())
@@ -302,7 +302,7 @@ var _ = Describe("Storage", func() {
 					// Because "/" is mounted on tmpfs, we need something that normally persists writes - /dev/sda2 is the EFI partition formatted as vFAT.
 					&expect.BSnd{S: "mount /dev/sda2 /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: "0"},
+					&expect.BExp{R: tests.Retcode("0", "\\#")},
 					&expect.BSnd{S: "echo content > /mnt/checkpoint\n"},
 					// The QEMU process will be killed, therefore the write must be flushed to the disk.
 					&expect.BSnd{S: "sync\n"},
@@ -330,10 +330,10 @@ var _ = Describe("Storage", func() {
 					// Same story as when first starting the VirtualMachineInstance - the checkpoint, if persisted, is located at /dev/sda2.
 					&expect.BSnd{S: "mount /dev/sda2 /mnt\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: "0"},
+					&expect.BExp{R: tests.Retcode("0", "\\#")},
 					&expect.BSnd{S: "cat /mnt/checkpoint &> /dev/null\n"},
 					&expect.BSnd{S: "echo $?\n"},
-					&expect.BExp{R: "1"},
+					&expect.BExp{R: tests.Retcode("1", "\\#")},
 				}, 200*time.Second)
 				Expect(err).ToNot(HaveOccurred())
 			})
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -2845,7 +2845,7 @@ func CheckForTextExpecter(vmi *v1.VirtualMachineInstance, expected []expect.Batc
 	}
 	defer expecter.Close()
 
-	resp, err := expecter.ExpectBatch(expected, time.Second*time.Duration(wait))
+	resp, err := ExpectBatchWithValidatedSend(expecter, expected, time.Second*time.Duration(wait))
 	if err != nil {
 		log.DefaultLogger().Object(vmi).Infof("%v", resp)
 	}
@@ -2880,7 +2880,7 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
 			&expect.BExp{R: prompt},
 			&expect.BSnd{S: "ip a | grep -q eth0; echo $?\n"},
 			&expect.BExp{R: retcode("0")}})
-		_, err := expecter.ExpectBatch(hasNetEth0Batch, 30*time.Second)
+		_, err := ExpectBatchWithValidatedSend(expecter, hasNetEth0Batch, 30*time.Second)
 		return err == nil
 	}
 
@@ -2890,7 +2890,7 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
 			&expect.BExp{R: prompt},
 			&expect.BSnd{S: "ip -6 address show dev eth0 scope global | grep -q inet6; echo $?\n"},
 			&expect.BExp{R: retcode("0")}})
-		_, err := expecter.ExpectBatch(hasGlobalIPv6Batch, 30*time.Second)
+		_, err := ExpectBatchWithValidatedSend(expecter, hasGlobalIPv6Batch, 30*time.Second)
 		return err == nil
 	}
 
@@ -2913,7 +2913,7 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
 		&expect.BExp{R: prompt},
 		&expect.BSnd{S: "sudo ip -6 addr add fd10:0:2::2/120 dev eth0; echo $?\n"},
 		&expect.BExp{R: retcode("0")}})
-	resp, err := expecter.ExpectBatch(addIPv6Address, 30*time.Second)
+	resp, err := ExpectBatchWithValidatedSend(expecter, addIPv6Address, 30*time.Second)
 	if err != nil {
 		log.DefaultLogger().Object(vmi).Infof("addIPv6Address failed: %v", resp)
 		expecter.Close()
@@ -2926,7 +2926,7 @@ func configureIPv6OnVMI(vmi *v1.VirtualMachineInstance, expecter expect.Expecter
 		&expect.BExp{R: prompt},
 		&expect.BSnd{S: "sudo ip -6 route add default via fd10:0:2::1 src fd10:0:2::2; echo $?\n"},
 		&expect.BExp{R: retcode("0")}})
-	resp, err = expecter.ExpectBatch(addIPv6DefaultRoute, 30*time.Second)
+	resp, err = ExpectBatchWithValidatedSend(expecter, addIPv6DefaultRoute, 30*time.Second)
 	if err != nil {
 		log.DefaultLogger().Object(vmi).Infof("addIPv6DefaultRoute failed: %v", resp)
 		expecter.Close()
@@ -4254,7 +4254,7 @@ func StartTCPServer(vmi *v1.VirtualMachineInstance, port int) {
 	Expect(err).ToNot(HaveOccurred())
 	defer expecter.Close()
 
-	resp, err := expecter.ExpectBatch([]expect.Batcher{
+	resp, err := ExpectBatchWithValidatedSend(expecter, []expect.Batcher{
 		&expect.BSnd{S: "\n"},
 		&expect.BExp{R: "\\$ "},
 		&expect.BSnd{S: fmt.Sprintf("screen -d -m nc -klp %d -e echo -e \"Hello World!\"\n", port)},
@@ -4285,7 +4285,7 @@ func StartHTTPServer(vmi *v1.VirtualMachineInstance, port int, isFedoraVM bool) 
 	}
 	defer expecter.Close()
 
-	resp, err := expecter.ExpectBatch([]expect.Batcher{
+	resp, err := ExpectBatchWithValidatedSend(expecter, []expect.Batcher{
 		&expect.BSnd{S: "\n"},
 		&expect.BExp{R: prompt},
 		&expect.BSnd{S: httpServerMaker},
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -2861,11 +2861,11 @@ func configureConsole(expecter expect.Expecter, prompt string, shouldSudo bool) 
 		&expect.BSnd{S: "stty columns 500\n"},
 		&expect.BExp{R: prompt},
 		&expect.BSnd{S: "echo $?\n"},
-		&expect.BExp{R: Retcode("0", prompt)},
+		&expect.BExp{R: RetValue("0", prompt)},
 		&expect.BSnd{S: fmt.Sprintf("%sdmesg -n 1\n", sudoString)},
 		&expect.BExp{R: prompt},
 		&expect.BSnd{S: "echo $?\n"},
-		&expect.BExp{R: Retcode("0", prompt)}})
+		&expect.BExp{R: RetValue("0", prompt)}})
 	resp, err := expecter.ExpectBatch(batch, 30*time.Second)
 	if err != nil {
 		log.DefaultLogger().Infof("%v", resp)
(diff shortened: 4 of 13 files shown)


tests/vmi_lifecycle_test.go
```diff
@@ -296,8 +296,6 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
 
 				By("Checking console text")
 				err = tests.CheckForTextExpecter(vmi, []expect.Batcher{
-					&expect.BSnd{S: "\n"},
-					&expect.BSnd{S: "\n"},
 					&expect.BExp{R: consoleText},
```
oshoval, line 299: not sure we want to eliminate both, at least one \r as most of the place do, 
well as long as each batch finishes in a prompt , it wont be needed, but since most still have \n (or \r) in the begining
i would stick to it

tests/vmi_networking_test.go
```diff
@@ -217,14 +218,14 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 			Expect(err).ToNot(HaveOccurred())
 			defer expecter.Close()
 
-			addrShow = "ip address show eth0\n"
+			addrShow = "ip address show eth0\r"
 			resp, err := expecter.ExpectBatch([]expect.Batcher{
```
oshoval, line 222: generally speaking, just taking this as an example
when there is a repeating prompt string i would suggest please to assign a var and use that var if possible

oshoval:
If you want, discuss with Dan, he suggested that the command + the retcode would be send to a function that would wrap both

tests/config_test.go
```diff
@@ -500,13 +500,13 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 					&expect.BExp{R: "#"},
```
oshoval, line 500: those should be \\# according Jed, maybe worth to handle on other PR?

tests/utils.go
```diff
@@ -2792,6 +2793,48 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
 	}, timeout, opts...)
 }
 
+// The method adds the expect.BSnd command to the exect.BExp expression.
+// It is done to make sure the match was found in the result of the expect.BSnd
+// command and not in a leftover the wasn't removed from the buffer.
+func ExpectBatchWithValidatedSend(expecter expect.Expecter, batch []expect.Batcher, timeout time.Duration) ([]expect.BatchRes, error) {
+	sendFlag := 0
```
oshoval, line 2800: i think it should enforce that batch start with SND?

tests/vmi_lifecycle_test.go
AlonaKaplan: Can you please explain what sending those `\n`s gives?
`&expect.BExp{R: consoleText}` is looking for output from the startup.

The new `ExpectBatchWuthSendValidation` will be flaky if we will keep those `\n` sends. Since it assumes the `expect.BExp` should be the result (and as so, appear after) the `expect.BSnd`. Since `consoleText` is not the result of `\n`, it has to be eliminated.

tests/storage_test.go
```diff
@@ -199,7 +199,7 @@ var _ = Describe("Storage", func() {
 					&expect.BSnd{S: "sudo mkfs.ext4 /dev/vdc\n"},
 					&expect.BExp{R: "\\$ "},
 					&expect.BSnd{S: "echo $?\n"},
```
oshoval, line 201: there are some places with Snd \n still
is it fine ?

for me it was fine when i worked on it, i guess something delicate was changed

tests/vmi_networking_test.go
AlonaKaplan: When you say "prompt string" do you mean the `\r`? Are you suggesting to keep the `\r` a constant in the utils file?

AlonaKaplan:
> If you want, discuss with Dan, he suggested that the command + the retcode would be send to a function that would wrap both

Can you please explain how this suggestion contradicts with the approach in this PR?

tests/vmi_lifecycle_test.go
oshoval: > Can you please explain what sending those `\n`s gives?

It make sure that this batch isnt depend on the previous one gracefulStop, up to you, we dont have it on some other places still, unless you changed them , if you didnt ,it means that if someone try to adapt them to ExpectBatchWithSendValidation, he will have problems, 

that being sad i would suggest if possible please to create a how to list with the known culprits and guides?

> `&expect.BExp{R: consoleText}` is looking for output from the startup.
> 
> The new `ExpectBatchWuthSendValidation` will be flaky if we will keep those `\n` sends. Since it assumes the `expect.BExp` should be the result (and as so, appear after) the `expect.BSnd`. Since `consoleText` is not the result of `\n`, it has to be eliminated.

Understood, 
just a nit i would suggest hat ExpectBatchWithSendValidation will enforce to start with Snd but not mandatory or maybe not good by design, if you want to keep the freedom still.

tests/vmi_networking_test.go
oshoval: i mean "\\$ " , not the \r
console prompt

oshoval:
> > If you want, discuss with Dan, he suggested that the command + the retcode would be send to a function that would wrap both
> 
> Can you please explain how this suggestion contradicts with the approach in this PR?

It might not contradict, just up to you if you decide to check out what he suggested and either do it on other PR or whatever you deicde

tests/config_test.go
AlonaKaplan: Yes, the string in `expect.BExp{R: str}` is actually a regular expression. Most of our tests don't keep it in mind. I make sure that at least the code that was added in this PR is aware to it. Thanks for the comment.

oshoval: `\\#` correction of course

tests/vmi_lifecycle_test.go
oshoval: i think i saw once that you do need to send \n
which indeed the result is prompt
(like doing enter)
else you might have a race that the buffer is empty and you will wait forever without it ?
even that the previous expecter was done gracefully

this is also what happened at least on most of the places
that it start by sending \n and expecting prompt

u even use still CheckForTextExpecter here dont you ?

tests/utils.go
```diff
@@ -2792,6 +2793,48 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
 	}, timeout, opts...)
 }
 
+// The method adds the expect.BSnd command to the exect.BExp expression.
+// It is done to make sure the match was found in the result of the expect.BSnd
+// command and not in a leftover the wasn't removed from the buffer.
+func ExpectBatchWithValidatedSend(expecter expect.Expecter, batch []expect.Batcher, timeout time.Duration) ([]expect.BatchRes, error) {
```
oshoval, line 2799: personally , i prefer that the fixes would be embedded in the current batches, 
i.e dont have double EXP, and wait for the right \n0  / prompt
or alternatively wrapping the command with the expected retcode (as discussed with Dan), and if there are special cases, handle them without generic method

as this approach might add some more nits that are need to be taken care of 
such as using \r instead of \n , maybe it was also before, but i didnt have this problem at least in #3543 

but its up to you, and the other reviewers
i am afraid that we will now have mixed approaches, and mixed use of \n or \r
batches that cant start with Snd \n (as you said but i am not sure we finished the discussion there)

tests/utils.go
AlonaKaplan: That was my approach when I started working on it. But there are case the first `BExp` looks for a match in the startup output.

tests/storage_test.go
AlonaKaplan: Batches that are not using the newly introduced method `ExpectBatchWithSendValidation` can keep using the `\n`.
However, `ExpectBatchWithSendValidation` is not working with `\n`, since `\n` is translated to `\r\n` which causes the send validation to fail.

I agree it is confusing. I will add a commit to the PR introducing an LD (line delimiter).

oshoval: can you enforce that there isnt just \n alone , or even replace it with \r ?

tests/vmi_lifecycle_test.go
AlonaKaplan: My mistake, the \n may be needed, depends on timing (the required `consoleText` is printed during the startup and also after the startup finished if `\n` is sent. But since the epxpecter may be created before/during or after the startup we cannot rely on text printed during the startup).
I also updated `ExpectBatchWithSendValidation` to return an error if the batch doesn't start with a send.

tests/utils.go
AlonaKaplan: Done, added it to the method.

AlonaKaplan:
> I think it's a good idea overall to have a function that sanitizes batches before running them.
> Did you consider running `stty -echo` (or equivalent) after logging in? That would ensure BExp never sees what was sent.
IIUC,  `stty -echo` causes the send command not to be printed. Since that's the way BExp validates the expected value is the result of the send command, I'm not sure how it helps us. Or you suggesting it as another approach?

If you meant as another approach, than yes, it may reduce the flakiness since a lot of the problematic cases were caused by the `BExp` finding a match in the `BSnd` command instead of the actual output.
But validating the expected value is the result of the sent command covers more flaky cases (when the `BExp` finds a match which is left over of a previous command result).

Looking at the man page, maybe `stty -ocrnl` can help us with the `\r` issue since it translates carriage return to newline. I will try it. Thanks for the hint.

> For all the prompt `BExp`s, I'd recommend using raw strings (backticks) instead of interpreted strings (double-quotes), which would allow us to avoid escaping the escape character (`\\`)...

It seems that in most of the tests the fact that the value passes to BExp is regexp is ignored. So it is a good idea to go over all the tests and change most of the BExp to use backticks.
But I think it worth a separate PR.

jean-edouard:
@AlonaKaplan switching to backticks in a separate PR seems fair, I've been meaning to give it a shot, I'll try to remember it after this is merged.

Sorry for not being more explicit, but I suggested `stty -echo` because I'm not a huge fan of messing with the batch to "consume" every send on receive. (I also think it's ok to do multiple sends in a row)
As you touched on in a comment, serial consoles don't know how wide the terminal is and default to 80 (another thing that can be fixed with stty!).
In your tests, you encountered wraps after ~70 chars, meaning the prompt must be ~10 chars.
However, some distros show the hostname (VMI name) in the prompt, which is pretty long in the functests. In that case, the wrap can happen after ~20 chars instead of 70...

AlonaKaplan:
> @AlonaKaplan switching to backticks in a separate PR seems fair, I've been meaning to give it a shot, I'll try to remember it after this is merged.
> 
> Sorry for not being more explicit, but I suggested `stty -echo` because I'm not a huge fan of messing with the batch to "consume" every send on receive. (I also think it's ok to do multiple sends in a row).

When doing multiple sends in a raw, we cannot be sure the expect found a match only after all the sends were executed. It is safer to have one send. Multiple.commands in the send can be split with ';'.
If for some reason we need to have multiple sends in a raw, we can use the regular `expect.expectBatch` for this use case.

> As you touched on in a comment, serial consoles don't know how wide the terminal is and default to 80 (another thing that can be fixed with stty!).
> In your tests, you encountered wraps after ~70 chars, meaning the prompt must be ~10 chars.
> However, some distros show the hostname (VMI name) in the prompt, which is pretty long in the functests. In that case, the wrap can happen after ~20 chars instead of 70...

Why does it mean the prompt must be ~10 chars? It only means the the BExp will try to find a match only to the first 70 chars of the sent command. The length of the prompt or any other value in the expect is not limited.

tests/vmi_configuration_test.go
```diff
@@ -1959,11 +1960,13 @@ var _ = Describe("Configurations", func() {
...
 				defer expecter.Close()
 
-				res, err := expecter.ExpectBatch([]expect.Batcher{
-					&expect.BSnd{S: "[ $(free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2) -lt 80 ] && echo 'pass' || echo 'fail'\n"},
-					&expect.BExp{R: "pass"},
-					&expect.BSnd{S: "swapoff -a && dd if=/dev/zero of=/dev/shm/test bs=1k count=118k; echo $?\n"},
-					&expect.BExp{R: "0"},
+				res, err := tests.ExpectBatchWithValidatedSend(expecter, []expect.Batcher{
+					&expect.BSnd{S: "[ $(free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2) -lt 80 ] && echo 'pass' || echo 'fail'\r"},
+					&expect.BExp{R: "\npass.*\\$"},
```
oshoval, line 1965: can tests.Retcode be used ? or because the .* it cant ?

AlonaKaplan: Yes, I think it can be used. The purpose of the the retcode was to use it for `echo $?`. But you're right, it is a good idea to use it for any value we search for.
Once this PR is merged we should introduce followup PRs using the Retcode for expected values and  tests.ExpectBatchWithValidatedSend instead of `expect.ExpectBatch`.

We should have constants for the prompts or specifying the same prompt for all the vm types on login.

oshoval:
Would be great please if at the end, if the commits can be squashed to several commits with topic on each,
for example 
1. create the mechanism.
2. use the mechanism.
far more re-viewable in the future this way imo.

tests/utils.go
```diff
@@ -2801,21 +2844,36 @@ func CheckForTextExpecter(vmi *v1.VirtualMachineInstance, expected []expect.Batc
...
 	}
 	return err
 }
 
+func configureConsole(expecter expect.Expecter, prompt string) error {
+	batch := append([]expect.Batcher{
+		&expect.BSnd{S: "stty -inlcr\n"},
+		&expect.BExp{R: prompt},
+		&expect.BSnd{S: "sudo dmesg -n 1\n"},
+		&expect.BExp{R: prompt},
```
oshoval, line 2859: wont you get a prompt even if the command failed ? 
so maybe worth to make sure all commands didnt fail ?

tests/vmi_multus_test.go
```diff
@@ -213,7 +213,7 @@ var _ = Describe("Multus", func() {
 					&expect.BExp{R: "\\$ "},
 					&expect.BSnd{S: cmdCheck},
 					&expect.BExp{R: "\\$ "},
-					&expect.BSnd{S: "ip addr show eth1 | grep 10.1.1 | wc -l"},
+					&expect.BSnd{S: "ip addr show eth1 | grep 10.1.1 | wc -l\n"},
 					&expect.BExp{R: "1"},
```
oshoval, line 217: 1 here appear in the above Snd, is it dangerous, isnt it ?

tests/utils.go
```diff
@@ -2792,6 +2792,46 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
...
+			if expectFlag == 1 {
+				return nil, fmt.Errorf("Two sequential expect.BExp are not allowed")
+			}
+			expectFlag = 1
+			sendFlag = 0
+			bExp, _ := batch[i].(*expect.BExp)
+			previousSend := regexp.QuoteMeta(previousSend)
+
+			// Remove the newline character since it is by the console to carriage return.
+			previousSend = previousSend[0 : len(previousSend)-1]
```
oshoval, line 2814: output = strings.TrimSuffix(output, "\n")

AlonaKaplan: done

AlonaKaplan:
> Would be great please if at the end, the commits will be squashed to several commits with topic on each,
> for example
> 
>     1. create the mechanism.
> 
>     2. config consoles.
> 
>     3. use the mechanism in the tests.
> 
> 
> far more re-viewable in the future this way imo.

I've re-arranged the commits.

tests/pausing_test.go
```diff
@@ -363,30 +364,32 @@ var _ = Describe("[rfe_id:3064][crit:medium][vendor:cnv-qe@redhat.com][level:com
 
 	Context("A long running process", func() {
 
+		grepSleepPid := func(expecter expect.Expecter) string {
+			res, err := tests.ExpectBatchWithValidatedSend(expecter, []expect.Batcher{
+				&expect.BSnd{S: `pgrep -f "sleep 5"` + "\n"},
```
oshoval, line 369: shouldnt it start with Snd \n ?

tests/config_test.go
```diff
@@ -498,16 +498,16 @@ var _ = Describe("[rfe_id:899][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 				res, err := expecter.ExpectBatch([]expect.Batcher{
 					// mount iso Secret image
 					&expect.BSnd{S: "sudo su -\n"},
-					&expect.BExp{R: "#"},
+					&expect.BExp{R: "\\#"},
```
oshoval, line 501: some of the prompts have `\\# ` some doesnt have the space, maybe best to keep one method ?
or at least in other PR where there will be a define for the prompt as we spoke please

AlonaKaplan: Yes, we defiantly should do a cleanup PR for the prompts issue.

tests/utils.go
```diff
@@ -2795,6 +2795,46 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
...
+				return nil, fmt.Errorf("Two sequential expect.BExp are not allowed")
+			}
+			expectFlag = 1
+			sendFlag = 0
+			bExp, _ := batch[i].(*expect.BExp)
+			previousSend := regexp.QuoteMeta(previousSend)
+
+			// Remove the \n since it is translated by the console to \r\n.
+			previousSend = strings.TrimSuffix(previousSend, "\n")
+			bExp.R = fmt.Sprintf("(?s)%s%s%s", previousSend, ".*", bExp.R)
```
oshoval, line 2818: maybe the (?s) and the reason for it worth comment (let . match \n)

tests/vmi_configuration_test.go
```diff
@@ -561,11 +561,13 @@ var _ = Describe("Configurations", func() {
 				Expect(err).ToNot(HaveOccurred())
 				defer expecter.Close()
 
-				res, err := expecter.ExpectBatch([]expect.Batcher{
+				res, err := tests.ExpectBatchWithValidatedSend(expecter, []expect.Batcher{
 					&expect.BSnd{S: "[ $(free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2) -gt 200 ] && echo 'pa''ss'\n"},
-					&expect.BExp{R: "pass"},
-					&expect.BSnd{S: "swapoff -a && dd if=/dev/zero of=/dev/shm/test bs=1k count=118k; echo $?\n"},
-					&expect.BExp{R: "0"},
+					&expect.BExp{R: tests.Retcode("pass", "\\$ ")},
```
oshoval, line 566: maybe worth to rename Retcode so it will fit to strings and retcodes ?

tests/pausing_test.go
AlonaKaplan: Snd \n + Exp prompt is done to make sure the buffer is clean (and actually a lot of our expecter flaky tests are because this cleanup failed). I think `tests.ExpectBatchWithValidatedSend` is safe enough and the extra cleanup in the beginning of the batch is not needed.

tests/vmi_configuration_test.go
```diff
@@ -1979,11 +1982,13 @@ var _ = Describe("Configurations", func() {
...
 
-				res, err := expecter.ExpectBatch([]expect.Batcher{
-					&expect.BSnd{S: "[ $(free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2) -lt 80 ] && echo 'pa''ss'\n"},
-					&expect.BExp{R: "pass"},
-					&expect.BSnd{S: "swapoff -a && dd if=/dev/zero of=/dev/shm/test bs=1k count=118k; echo $?\n"},
-					&expect.BExp{R: "0"},
+				res, err := tests.ExpectBatchWithValidatedSend(expecter, []expect.Batcher{
+					&expect.BSnd{S: "[ $(free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2) -lt 80 ] && echo 'pass' || echo 'fail'\n"},
+					&expect.BExp{R: tests.Retcode("pass", "\\$ ")},
+					&expect.BSnd{S: "swapoff -a && dd if=/dev/zero of=/dev/shm/test bs=1k count=118k\n"},
```
oshoval, line 1988: aint this block is repeating (even with delicate changes) ? maybe worth on other PR to make a func for it

tests/utils.go
AlonaKaplan: Well, I don't think so. If a developer doesn't understand a regex, he/she can go to the regex [documentation](https://golang.org/pkg/regexp/syntax/). There is no need to duplicate it here.

oshoval: reason was as its a fundamental match of this method, that is eating the \n, but up to you.
btw , maybe it would be stronger to enforce that exactly one \n can come in between, this way no other stderr can occur in the middle ?

tests/pausing_test.go
oshoval: it means that each previous batch must clean its output, unless you implement the snd random + exp random, we might encouter problems in the future in case previous exp spill chars (saw it in the past), but up to you

another corner this is prev batch shouldnt leave bg processes that might write stuff to stderr / stdout, but this is out of this PR anyway.

tests/utils.go
AlonaKaplan: I've update the regex so the `(?s`) will affect only the `.*` and not the whole expression.

tests/vmi_configuration_test.go
AlonaKaplan: What name do you suggest? RetVal?

tests/vmi_configuration_test.go
AlonaKaplan: Maybe, but not in this PR:)

tests/utils.go
oshoval: it still allow whatever test in between the previousSend and the bExp.R isnt it ?
dont we want more strict ?

AlonaKaplan: > reason was as its a fundamental match of this method, that is eating the \n, but up to you.

It is not related the the `\n` we trim. It just allows the output of the sent command to contain and value including `\n`.

> btw , maybe it would be stronger to enforce that exactly one \n can come in between, this way no other stderr can occur in the middle ?
The output of the sent command can be any sequence of characters. Why should we restrict it?

tests/pausing_test.go
AlonaKaplan: Correct written batch is cleaning its output. But even if not, since we make sure the sent command appear before the expected value, even if there is leftover in the buffer, the test will work as expected and match would be found in the leftover.

2b. point in the [document](https://docs.google.com/document/d/16qGpswxJP9nDCprwNckXu5RHSUWF8lDFNmVEZT539aM/edit) suggests cleaning the buffer after each `BSnd`, `BExp` pair. We may have a follow-up PR doing it. Or as an alternative do a cleanup to our batches and make sure buffer is cleaned after each `BSnd` `BExp` pair. This can be achieved by using the new `Retcode` method in the `BExp`s. The method always waiting for the prompt.

tests/vmi_configuration_test.go
oshoval: RetString ?
RetValue

tests/utils.go
oshoval: > > reason was as its a fundamental match of this method, that is eating the \n, but up to you.
> 
> It is not related the the `\n` we trim. It just allows the output of the sent command to contain and value including `\n`.
> 

didnt meant otherwise, i just afraid if anything can be in between the two halfs, and then combined with other changes it can create new problems, but up to you, its a long shot that we can take care of in the future, unless you want already to make it stricter, which is good imo.

> > btw , maybe it would be stronger to enforce that exactly one \n can come in between, this way no other stderr can occur in the middle ?
> > The output of the sent command can be any sequence of characters. Why should we restrict it?

AlonaKaplan: How can I make it stricter?

oshoval: thinking out loud with you,
expect exactly previousSend + \n (or \r\n) + bExp.R 
without allowing wild card .* in between that can match anything (if i am not wrong) ?

AlonaKaplan: That won't work since the expected value in most of the cases doesn't include the whole output of the command.

oshoval: so imo , in the future, bExp.R should be a regex matching the expected result, instead having here a wildcard which can consume all the in between expressions.

tests/vmi_configuration_test.go
```diff
@@ -1941,7 +1943,8 @@ var _ = Describe("Configurations", func() {
 				log.DefaultLogger().Object(cpuVmi).Infof("%v", res)
 				Expect(err).ToNot(HaveOccurred())
 			})
-			It("should be able to start a vm with guest memory different from requested and keep guaranteed qos", func() {
+			It("should be able to start a vm with guest memory different from requested and keed guaranteed qos", func() {
```
jean-edouard, line 1946: Minor: this reintroduces a typo, probably not desired.

tests/vmi_multus_test.go
```diff
@@ -215,7 +215,7 @@ var _ = Describe("Multus", func() {
 					&expect.BExp{R: "\\$ "},
 					&expect.BSnd{S: cmdCheck},
 					&expect.BExp{R: "\\$ "},
-					&expect.BSnd{S: "ip addr show eth1 | grep 10.1.1 | wc -l"},
+					&expect.BSnd{S: "ip addr show eth1 | grep 10.1.1 | wc -l\n"},
```
jean-edouard, line 218: Wow... Proof right there that this PR is desperately needed!

tests/vmi_configuration_test.go
AlonaKaplan: Changed to RetValue

tests/pausing_test.go
```diff
@@ -363,30 +364,32 @@ var _ = Describe("[rfe_id:3064][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
 			By("Checking the long running process")
-			res, err := expecter.ExpectBatch([]expect.Batcher{
-				&expect.BSnd{S: `pgrep -f "sleep 5"` + "\n"},
-				&expect.BExp{R: "sleep 5"},    // command
-				&expect.BExp{R: "[0-9]+\r\n"}, // pid
-			}, 15*time.Second)
-			log.DefaultLogger().Infof("a:%+v\n", res)
-			Expect(err).ToNot(HaveOccurred())
-			return strings.TrimSpace(res[1].Match[0])
+			return grepSleepPid(expecter)
```
EdDev, line 392: I think you should return the output of the command and do the assert and pid extraction here. Otherwise the name of the func does  not represent well what is going on (and it will be nice to separate the operation from the assertion).

tests/pausing_test.go
```diff
@@ -363,30 +364,32 @@ var _ = Describe("[rfe_id:3064][crit:medium][vendor:cnv-qe@redhat.com][level:com
...
 				&expect.BExp{R: "\\# "}, // prompt
-				&expect.BSnd{S: `pgrep -f "sleep 5"` + "\n"},
-				&expect.BExp{R: "sleep 5"},    // command
-				&expect.BExp{R: "[0-9]+\r\n"}, // pid
 			}, 15*time.Second)
 			log.DefaultLogger().Infof("a:%+v\n", res)
 			Expect(err).ToNot(HaveOccurred())
-			return strings.TrimSpace(res[2].Match[0])
+
+			return grepSleepPid(expecter)
```
EdDev, line 387: Makes sense to call `checkProcess`. See next comment below.

tests/config_test.go
oshoval: as Edy also said in this PR, lets please at least line up the changes that this PR did ?

tests/utils.go
```diff
@@ -2795,6 +2795,46 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
 	}, timeout, opts...)
 }
 
+// The method adds the expect.BSnd command to the exect.BExp expression.
```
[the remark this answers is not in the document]
oshoval: i suggested that there will be a short summary of the known culprits and suggested usage. even in a README.md for tests

tests/vmi_multus_test.go
```diff
@@ -215,7 +215,7 @@ var _ = Describe("Multus", func() {
...
 					&expect.BExp{R: "1"},
```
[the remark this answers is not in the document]
oshoval: this is also another reason i dont think we should be depend on correct previous expecters, in one of the threads here, i can find it if needed

tests/config_test.go
EdDev: This is not changing the existing, so I do not think it is a must here. It can come in a following change.

tests/pausing_test.go
oshoval: need to consume the prompt at grepSleepPid as well ?

tests/config_test.go
oshoval: `\\` were added, it doesnt count ?

tests/vmi_multus_test.go
AlonaKaplan: Added the test id

AlonaKaplan:
> This is a great improvement.
> 
> Although it is a too late now, but for the sake of future cases like this, it would be nice to separate the main enhancement in the tooling and the various fixes.
> The enhancement may come with a few fixes to show it works, but the rest will fit nicely in following PR/s.

We still need a follow-up to use the `ExpectBatchWithValidatedSend` instead of `expecter.ExpectBatch` where possible. This PR doesn't cover all the places.
Also we need follow-up to use `RetValue` where possible in `BExp`, the PR cover just the `echo $?` results.

tests/utils.go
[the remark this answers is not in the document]
AlonaKaplan: @EdDev done

tests/utils.go
```diff
@@ -2795,6 +2795,46 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
...
+// It is done to make sure the match was found in the result of the expect.BSnd
+// command and not in a leftover the wasn't removed from the buffer.
```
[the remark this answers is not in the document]
AlonaKaplan: Done, thanks!

tests/utils.go
```diff
@@ -2795,6 +2795,46 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
...
+			if expectFlag == 1 {
+				return nil, fmt.Errorf("Two sequential expect.BExp are not allowed")
+			}
+			expectFlag = 1
+			sendFlag = 0
+			bExp, _ := batch[i].(*expect.BExp)
+			previousSend := regexp.QuoteMeta(previousSend)
+
+			// Remove the \n since it is translated by the console to \r\n.
+			previousSend = strings.TrimSuffix(previousSend, "\n")
```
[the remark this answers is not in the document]
AlonaKaplan: The method doesn't support multiline commands. In multiline commands after `\n` a `>` is added, which makes it hard to find a match for the sent command. I will add it to the description of the method.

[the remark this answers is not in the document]
oshoval: maybe worth to add a detection and error in case \n is in the middle ?

tests/utils.go
```diff
@@ -2795,6 +2795,46 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
 	}, timeout, opts...)
 }
 
+// The method adds the expect.BSnd command to the exect.BExp expression.
+// It is done to make sure the match was found in the result of the expect.BSnd
+// command and not in a leftover the wasn't removed from the buffer.
+func ExpectBatchWithValidatedSend(expecter expect.Expecter, batch []expect.Batcher, timeout time.Duration) ([]expect.BatchRes, error) {
+	sendFlag := 0
+	expectFlag := 0
```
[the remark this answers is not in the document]
AlonaKaplan: Changed to boolean

tests/utils.go
```diff
@@ -2795,6 +2795,46 @@ func NewConsoleExpecter(virtCli kubecli.KubevirtClient, vmi *v1.VirtualMachineIn
...
+			previousSend = strings.TrimSuffix(previousSend, "\n")
+			bExp.R = fmt.Sprintf("%s%s%s", previousSend, "((?s).*)", bExp.R)
+			previousSend = ""
+		case expect.BatchSend:
+			if sendFlag == 1 {
+				return nil, fmt.Errorf("Two sequential expect.BSend are not allowed")
+			}
+			sendFlag = 1
+			expectFlag = 0
+			previousSend = batcher.Arg()
```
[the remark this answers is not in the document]
AlonaKaplan: Yes, we can use `previousSend` (unless we send empty string, but it is a bug by its own). But I think it is more readable this way.

tests/vmi_configuration_test.go
```diff
@@ -1979,10 +1982,12 @@ var _ = Describe("Configurations", func() {
 				Expect(err).ToNot(HaveOccurred())
 				defer expecter.Close()
 
-				res, err := expecter.ExpectBatch([]expect.Batcher{
-					&expect.BSnd{S: "[ $(free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2) -lt 80 ] && echo 'pa''ss'\n"},
-					&expect.BExp{R: "pass"},
-					&expect.BSnd{S: "swapoff -a && dd if=/dev/zero of=/dev/shm/test bs=1k count=118k; echo $?\n"},
+				res, err := tests.ExpectBatchWithValidatedSend(expecter, []expect.Batcher{
+					&expect.BSnd{S: "[ $(free -m | grep Mem: | tr -s ' ' | cut -d' ' -f2) -lt 80 ] && echo 'pass' || echo 'fail'\n"},
```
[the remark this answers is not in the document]
AlonaKaplan: Rebase error. Thanks was the catch!

tests/vmi_configuration_test.go
```diff
@@ -2326,9 +2331,9 @@ var _ = Describe("Configurations", func() {
 
 			By("Check value in VM with dmidecode")
 			// Check on the VM, if expected values are there with dmidecode
-			res, err := expecter.ExpectBatch([]expect.Batcher{
-				&expect.BSnd{S: "[ $(sudo dmidecode -s chassis-asset-tag | tr -s ' ') = Test-123 ] && echo 'pa''ss'\n"},
-				&expect.BExp{R: "pass"},
+			res, err := tests.ExpectBatchWithValidatedSend(expecter, []expect.Batcher{
+				&expect.BSnd{S: "[ $(sudo dmidecode -s chassis-asset-tag | tr -s ' ') = Test-123 ] && echo 'pass'\n"},
+				&expect.BExp{R: tests.Retcode("pass", "\\#")},
```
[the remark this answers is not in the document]
AlonaKaplan: done

tests/pausing_test.go
AlonaKaplan: @oshoval this is an internal method for a very specific purpose. The prompt is known, there is no need to pass it as a parameter.

oshoval: i meant to consume it, not the pass it

AlonaKaplan: @EdDev the pupose of this internal method was reuse. This method as its name specifies does grep to get the sleep pid. The method is internal and not a tool that can be resued in other places that may want to have different assersion. I think it is fine and convenient to have the assertion inside it.

tests/pausing_test.go
AlonaKaplan: I think that `grepSleepPid` explains better what this method is doing. It doesn't check a generic process, it greps the sleep (not any sleep, the `sleep 5` that we just sent).

AlonaKaplan:
@oshoval - couldn't quote reply. In this [comment](https://github.com/kubevirt/kubevirt/pull/3882#discussion_r464811556), do you mean add space after the `#`?

oshoval:
> @oshoval - couldn't quote reply. In this [comment](https://github.com/kubevirt/kubevirt/pull/3882#discussion_r464811556), do you mean adding space after the `#`?

yep , just on the lines you added `\\`

tests/utils.go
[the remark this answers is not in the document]
AlonaKaplan: If I will add detection I can add support:) The detection is not trivial, it has to differentiate between a `\n`s which are acceptable (for example writting to a file, part of a grep, etc) to `\n`s which cause the command to be multiline. Currently, in the scope of this PR it is not supported, I think the comment is enough. We can always add support or automatic error in the future.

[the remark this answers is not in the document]
oshoval: sgtm

tests/pausing_test.go
AlonaKaplan: > i meant to consume it, not the pass it

Sorry, I don't understand the comment.
`grepSleepPid` waits for the prompt. What else do you think it should do?

oshoval: if it Exp for the prompt at the end of it, then all good

oshoval:
Thanks
/lgtm

oshoval:
/hold
@EdDev please remove the hold if you are fine with it as well, (since there is an approve i added the hold)

tests/utils.go
```diff
@@ -2851,6 +2851,27 @@ func CheckForTextExpecter(vmi *v1.VirtualMachineInstance, expected []expect.Batc
...
+func configureConsole(expecter expect.Expecter, prompt string, shouldSudo bool) error {
+	sudoString := ""
+	if shouldSudo {
+		sudoString = "sudo "
+	}
+	batch := append([]expect.Batcher{
+		&expect.BSnd{S: "stty columns 500\n"},
+		&expect.BExp{R: prompt},
+		&expect.BSnd{S: "echo $?\n"},
+		&expect.BExp{R: Retcode("0", prompt)},
```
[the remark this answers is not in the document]
EdDev: Did you missed this one?

tests/utils.go
```diff
@@ -2851,6 +2851,27 @@ func CheckForTextExpecter(vmi *v1.VirtualMachineInstance, expected []expect.Batc
...
+		&expect.BSnd{S: fmt.Sprintf("%sdmesg -n 1\n", sudoString)},
+		&expect.BExp{R: prompt},
+		&expect.BSnd{S: "echo $?\n"},
+		&expect.BExp{R: Retcode("0", prompt)}})
+	resp, err := expecter.ExpectBatch(batch, 30*time.Second)
+	if err != nil {
+		log.DefaultLogger().Infof("%v", resp)
```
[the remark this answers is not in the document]
EdDev: Did you missed this one?

tests/utils.go
```diff
@@ -2844,7 +2844,7 @@ func CheckForTextExpecter(vmi *v1.VirtualMachineInstance, expected []expect.Batc
 	}
 	defer expecter.Close()
 
-	resp, err := expecter.ExpectBatch(expected, time.Second*time.Duration(wait))
+	resp, err := ExpectBatchWithValidatedSend(expecter, expected, time.Second*time.Duration(wait))
```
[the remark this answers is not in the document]
EdDev: Did you missed this one?

tests/utils.go
```diff
@@ -5031,6 +5031,6 @@ func IsLauncherCapabilityValid(capability k8sv1.Capability) bool {
 	return false
 }
 
-func retcode(retcode string) string {
-	return "\n" + retcode
+func Retcode(retcode string, prompt string) string {
```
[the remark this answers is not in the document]
EdDev: Did you missed this one?

tests/vmi_lifecycle_test.go
```diff
@@ -298,7 +298,6 @@ var _ = Describe("[rfe_id:273][crit:high][vendor:cnv-qe@redhat.com][level:compon
 
 				By("Checking console text")
 				err = tests.CheckForTextExpecter(vmi, []expect.Batcher{
-					&expect.BSnd{S: "\n"},
```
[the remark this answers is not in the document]
EdDev: Did you missed this one?

tests/utils.go
```diff
@@ -4407,9 +4407,9 @@ func GenerateHelloWorldServer(vmi *v1.VirtualMachineInstance, testPort int, prot
 	serverCommand := fmt.Sprintf("screen -d -m sudo nc -klp %d -e echo -e 'Hello World!'\n", testPort)
 	if protocol == "udp" {
 		// nc has to be in a while loop in case of UDP, since it exists after one message
-		serverCommand = fmt.Sprintf("screen -d -m sh -c \"while true\n do nc -uklp %d -e echo -e 'Hello UDP World!'\ndone\n\"\n", testPort)
+		serverCommand = fmt.Sprintf("screen -d -m sh -c \"while true; do nc -uklp %d -e echo -e 'Hello UDP World!';done\"\n", testPort)
```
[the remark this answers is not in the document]
EdDev: Documentation added, detecting and error not.
As the implementation is not trivial, I think it is good enough for now.

tests/vmi_multus_test.go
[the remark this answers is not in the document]
EdDev: Did you missed this one?

tests/pausing_test.go
EdDev: > I think it is fine and convenient to have the assertion inside it.

I do not like exactly this inner assertion.
It gives bad tracing and usually can be "fxied" by using an offset-assertion which is not well maintained (as it depends the stack through which it is called).
Inner assertions also gives less control.

Anyway, I will not insist on this.

tests/utils.go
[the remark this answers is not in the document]
AlonaKaplan: yes, you are right. That's the price of re-ordering/changing the commits again and again:). Done.

tests/utils.go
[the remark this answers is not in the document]
AlonaKaplan: The extra value of printing the response is printing the current buffer which is very useful for debugging in case of an error (that is the best way to find out what the current buffer contains and why the match didn't work).

tests/utils.go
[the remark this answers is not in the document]
AlonaKaplan: The PR doesn't aim to change all the occurrences, it just gives the infra methods and some examples of how to use it. I will update the PR comment to reflect it.

tests/utils.go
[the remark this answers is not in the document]
AlonaKaplan: I agree, we can. But please let's add this improvement to the followup PR.

Regarding the nit - done.

tests/vmi_lifecycle_test.go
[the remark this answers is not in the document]
AlonaKaplan: done

tests/vmi_networking_test.go
```diff
@@ -652,13 +653,15 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
...
+				&expect.BSnd{S: "\n"},
 				&expect.BExp{R: "\\#"},
 				&expect.BSnd{S: "cat /dhcp-env\n"},
 				&expect.BExp{R: "new_bootfile_name=config"},
+				&expect.BSnd{S: "\n"},
 				&expect.BExp{R: "\\#"},
 				&expect.BSnd{S: "cat /dhcp-env\n"},
-				&expect.BExp{R: "new_ntp_servers=127.0.0.1 127.0.0.2"},
-				&expect.BExp{R: "new_unknown_240=private.options.kubevirt.io"},
+				&expect.BExp{R: regexp.QuoteMeta("new_ntp_servers=127.0.0.1 127.0.0.2") + "((?s).*)" + regexp.QuoteMeta("new_unknown_240=private.options.kubevirt.io")},
```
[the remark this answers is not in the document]
AlonaKaplan: Sequential `BExp` are flaky for sure. I don't think it is a good idea to support it. We should find all the places that have it and have a proper fix to each one, I don't think we need to have a generic solution for it.

tests/vmi_multus_test.go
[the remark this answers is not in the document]
AlonaKaplan: @EdDev yes, we don't wait for the prompt and we should, relevant here and relevant in may other places. The PR doesn't aim to fix all the places.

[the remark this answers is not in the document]
oshoval: aint it wait because it use the new mechanism which does wait ?
this is why i overlooked it

it even doesnt wait for \n1 so it must use the new mech unless it already does. (which what i glanced and believed is does)

tests/vmi_networking_test.go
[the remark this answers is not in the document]
EdDev: I meant that you in the wrapper, to detect sequential BExp and merge them as you you did above.
It is not reasonable (and pretty ugly) to ask testers to do what you just did.

tests/vmi_multus_test.go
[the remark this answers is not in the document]
EdDev: This commit claims to fix this test. As you touched it, I would expect you to leave it in a good shape.

tests/vmi_monitoring_test.go
```diff
@@ -70,9 +70,9 @@ var _ = Describe("Health Monitoring", func() {
 			By("Killing the watchdog device")
 			_, err = expecter.ExpectBatch([]expect.Batcher{
 				&expect.BSnd{S: "watchdog -t 2000ms -T 4000ms /dev/watchdog && sleep 5 && killall -9 watchdog\n"},
-				&expect.BExp{R: "#"},
+				&expect.BExp{R: "\\#"},
```
oshoval, line 73: we want to add Snd \n + Exp prompt before the first Snd ?

tests/vmi_multus_test.go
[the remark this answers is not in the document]
oshoval: btw it made me think, since CheckForTextExpecter uses now ExpectBatchWithValidatedSend
which wait for the prompt on each Exp that use RetValue
we arent allowed to wait for prompt after use of RetValue in case there is (didnt find so far) or will be one in the future.

tests/vmi_monitoring_test.go
AlonaKaplan: Why? And how this is relevant to this PR?

oshoval: it not must be in this PR, 
i just saw it and wrote, as most of the batches start with it , to be independent on the previous as a preliminary act, i think its a good practice, at least in the current state of the code.

tests/vmi_networking_test.go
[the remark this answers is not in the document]
jean-edouard: As ugly as it is, the only viable alternative we've found so far is to `cat` twice, and `BExp` a different line each time. Probably a better approach here, a bit ugly too but in a different way!

tests/vmi_configuration_test.go
```diff
@@ -1942,6 +1944,7 @@ var _ = Describe("Configurations", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 			It("should be able to start a vm with guest memory different from requested and keep guaranteed qos", func() {
+				Skip("Skip test till issue https://github.com/kubevirt/kubevirt/issues/3910 is fixed")
```
oshoval, line 1947: btw new changes in this code were skiped testing from some point, unless you checked them manually

AlonaKaplan: I checked them manually. But the developer who will fix the issue will have to verify.

oshoval:
/lgtm cancel
you said there is one issue found

EdDev:
> /lgtm cancel
> you said there is one issue found

My review stands, this looks fine to me.
The owner can hold it if there is something that needs investigation.

/lgtm

AlonaKaplan:
/hold

https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3882/pull-kubevirt-e2e-k8s-1.16/1291187384536272896

AlonaKaplan:
/hold cancel

The error in test https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_kubevirt/3882/pull-kubevirt-e2e-k8s-1.16/1291187384536272896 doesn't see to be related to the PR. A random ` \r` was added to the command. Although I have no idea why it happened, a follow-up PR will be sent changing the failing test to have shorter line and not using sudo.

tests/utils.go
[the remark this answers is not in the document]
AlonaKaplan: The follow-up PR answering the comment - https://github.com/kubevirt/kubevirt/pull/3956

